PR #33, opened 2024-04-09
feat: use htmltools instead of str

SaaiVenkat:
Maidr object has been modified to reflect HTMLDocument of htmltools. Users can use this to render the html or save as a string to a file.


SaaiVenkat:
> @SaaiVenkat As far as I recall, htmltools object has `.show()` method that opens the current html object in user's default browser.
> 
> 1. Please test whether this method works for maidr object now.
> 2. If that works, please use that method instead of file save in all the examples.

1. I have updated Maidr object to use HTML Tools with show(). However, this might be working only in the interactive session because of the HTTP server being closed when using scripts.

2. Since they are examples, I have kept both save_html() as well as show().

jooyoungseo (approved):
LGTM!

PR #35, opened 2024-05-06
docs(example): add quarto (qmd)

SaaiVenkat:


jooyoungseo:
@SaaiVenkat When you include other examples for qmd, please only include the qmd files without the html/css/js dependencies because they are going to be generated on users' end.

Issue #65, opened 2024-08-09
feat(maidr.show): support py-shiny renderer


jooyoungseo:
Currently, `py_maidr` is not compatible with [`py-shiny`](https://github.com/posit-dev/py-shiny). We need to add a custom renderer to support dynamic shiny dashboard.

PR #67, opened 2024-08-13
feat(maidr.show): support py-shiny renderer

krishnanand5:
This PR adds support for maidr on py-shiny through a custom renderer called `render_maidr`. This decorator internally calls the newly created `maidr.render` function to transform a plot of the matplotlib type to a maidr object and envelops it as an iframe srcdoc for display on a py-shiny web dashboard.

Closes [#65]

### DIFF
--- a/example/py-shiny/example_pyshiny_reactive_scatter.py
+++ b/example/py-shiny/example_pyshiny_reactive_scatter.py
@@ -41,6 +41,7 @@ def server(input, output, session):
         ax.set_xlabel(input.x_var().replace("_", " ").title())
         ax.set_ylabel(input.y_var().replace("_", " ").title())  # Hello
         return s_plot
+
     output.reactivebarplot = render_maidr(create_reactivebarplot)
 
 
--- a/example/py-shiny/example_pyshiny_reactive_scatter.py
+++ b/example/py-shiny/example_pyshiny_reactive_scatter.py
@@ -1,0 +1,52 @@
+import matplotlib.pyplot as plt
+import seaborn as sns
+from shiny import App, ui
+
+from maidr.widget.shiny import render_maidr
+
+# Load the dataset
+iris = sns.load_dataset("iris")
+
+# Define the UI components for the Shiny application
+app_ui = ui.page_fluid(
+    ui.row(
+        ui.column(
+            3,
+            ui.input_select(
+                "x_var",
+                "Select X variable:",
+                choices=iris.select_dtypes(include=["float64"]).columns.tolist(),
+                selected="sepal_length",
+            ),
+            ui.input_select(
+                "y_var",
+                "Select Y variable:",
+                choices=iris.select_dtypes(include=["float64"]).columns.tolist(),
+                selected="sepal_width",
+            ),
+        ),
+        ui.column(9, ui.output_ui("create_reactivebarplot")),
+    )
+)
+
+
+# Define the server
+def server(input, output, session):
+    def create_reactivebarplot():
+        fig, ax = plt.subplots(figsize=(10, 6))
+        s_plot = sns.scatterplot(
+            data=iris, x=input.x_var(), y=input.y_var(), hue="species", ax=ax
+        )
+        ax.set_title(f"Iris {input.y_var()} vs {input.x_var()}")
+        ax.set_xlabel(input.x_var().replace("_", " ").title())
+        ax.set_ylabel(input.y_var().replace("_", " ").title())  # Hello
+        return s_plot
+    output.reactivebarplot = render_maidr(create_reactivebarplot)
+
+
+# Create the app
+app = App(app_ui, server)
+
+# Run the app
+if __name__ == "__main__":
+    app.run()
--- a/example/py-shiny/example_pyshiny_reactive_scatter.py
+++ b/example/py-shiny/example_pyshiny_reactive_scatter.py
@@ -2,7 +2,7 @@ import matplotlib.pyplot as plt
 import seaborn as sns
 from shiny import App, ui
 
-from maidr.widget.shiny import render_maidr
+from maidr.widget.shiny import RenderMaidr
 
 # Load the dataset
 iris = sns.load_dataset("iris")
@@ -32,7 +32,7 @@ app_ui = ui.page_fluid(
 
 # Define the server
 def server(input, output, session):
-    @render_maidr
+    @RenderMaidr
     def create_reactivebarplot():
         fig, ax = plt.subplots(figsize=(10, 6))
         s_plot = sns.scatterplot(
--- a/example/py-shiny/example_pyshiny_reactive_scatter.py
+++ b/example/py-shiny/example_pyshiny_reactive_scatter.py
@@ -32,6 +32,7 @@ app_ui = ui.page_fluid(
 
 # Define the server
 def server(input, output, session):
+    @render_maidr
     def create_reactivebarplot():
         fig, ax = plt.subplots(figsize=(10, 6))
         s_plot = sns.scatterplot(
@@ -39,10 +40,8 @@ def server(input, output, session):
         )
         ax.set_title(f"Iris {input.y_var()} vs {input.x_var()}")
         ax.set_xlabel(input.x_var().replace("_", " ").title())
-        ax.set_ylabel(input.y_var().replace("_", " ").title())  # Hello
+        ax.set_ylabel(input.y_var().replace("_", " ").title())
         return s_plot
-
-    output.reactivebarplot = render_maidr(create_reactivebarplot)
 
 
 # Create the app
--- a/example/py-shiny/example_pyshiny_reactive_scatter.py
+++ b/example/py-shiny/example_pyshiny_reactive_scatter.py
@@ -2,7 +2,7 @@ import matplotlib.pyplot as plt
 import seaborn as sns
 from shiny import App, ui
 
-from maidr.widget.shiny import RenderMaidr
+from maidr.widget.shiny import render_maidr
 
 # Load the dataset
 iris = sns.load_dataset("iris")
@@ -32,7 +32,7 @@ app_ui = ui.page_fluid(
 
 # Define the server
 def server(input, output, session):
-    @RenderMaidr
+    @render_maidr
     def create_reactivebarplot():
         fig, ax = plt.subplots(figsize=(10, 6))
         s_plot = sns.scatterplot(
--- a/maidr/widget/shiny.py
+++ b/maidr/widget/shiny.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import Any, Callable, Coroutine, Optional
-
 from shiny.render import ui
 from shiny.types import Jsonifiable
 
@@ -14,45 +12,18 @@ class render_maidr(ui):
 
     This class extends the Shiny UI rendering functionality to handle Maidr objects.
 
-    Attributes
-    ----------
-    fn : Callable[[], Coroutine[Any, Any, Optional[Any]]]
-        An asynchronous function that returns the value to be rendered.
-
     Methods
     -------
     render()
         Asynchronously renders the Maidr object.
     """
 
-    def __init__(self, fn: Callable[[], Coroutine[Any, Any, Optional[Any]]]):
-        """
-        Initialize the render_maidr instance.
-
-        Parameters
-        ----------
-        fn : Callable[[], Coroutine[Any, Any, Optional[Any]]]
-            An asynchronous function that returns the value to be rendered.
-        """
-        super().__init__(fn)
-
-    async def render(self) -> Optional[Jsonifiable]:
-        """
-        Asynchronously render the Maidr object.
-
-        This method calls the provided function, renders the result using Maidr,
-        and transforms it into a Jsonifiable format.
-
-        Returns
-        -------
-        Optional[Jsonifiable]
-            The rendered and transformed Maidr object, or
-            None if the initial value is None.
-        """
-        initial_value: Optional[Any] = await self.fn()
+    async def render(self) -> Jsonifiable:
+        """Return maidr rendered object for a given plot."""
+        initial_value = await self.fn()
         if initial_value is None:
             return None
 
-        maidr_rendered: Any = maidr.render(initial_value)
-        transformed: Jsonifiable = await self.transform(maidr_rendered)
+        maidr_rendered = maidr.render(initial_value)
+        transformed = await self.transform(maidr_rendered)
         return transformed
--- a/maidr/widget/shiny.py
+++ b/maidr/widget/shiny.py
@@ -6,7 +6,7 @@ from shiny.types import Jsonifiable
 import maidr
 
 
-class render_maidr(ui):
+class RenderMaidr(ui):
     """
     A custom UI rendering class for Maidr objects in Shiny applications.
 
--- a/maidr/widget/shiny.py
+++ b/maidr/widget/shiny.py
@@ -6,7 +6,7 @@ from shiny.types import Jsonifiable
 import maidr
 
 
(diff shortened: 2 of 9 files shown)


maidr/custom_renderers/pyshiny_renderer.py
jooyoungseo, line 1: Add type annotations for all the functions

jooyoungseo: Change the module name to `shiny_renderer`

maidr/custom_renderers/pyshiny_renderer.py
```diff
@@ -0,0 +1,48 @@
...
+            <html>
+            <head>
+                {dependency_tags}
+            </head>
+            <body>
+                {html_content}
+            </body>
+            </html>
+            """
+
```
jooyoungseo, line 37: any reasons why you don't use htmltools package and its tag methods instead of the direct string literals for html tags?

example/py-shiny/example_pyshiny_scatter.py
```diff
@@ -0,0 +1,38 @@
...
+        ax.set_xlabel("Sepal Length")
+        ax.set_ylabel("Sepal Width")
+        return s_plot
+
+    @output
+    @render.ui
+    def scatterplot():
+        return render_maidr(create_scatterplot)()
+
+
```
jooyoungseo, line 32: This is too simple. Demonstrate the example Shiny app where users can choose two different numeric variables (one for x; the other for y) from the iris dataset and update the plot reactively against the variable change.

krishnanand5: Sure, Professor. I will create a more complex example to demonstrate the feature more intuitively.

maidr/custom_renderers/pyshiny_renderer.py
```diff
@@ -0,0 +1,48 @@
+from matplotlib.axes import Axes
+from matplotlib.figure import Figure
+import maidr
+from shiny import ui
+from typing import Any
```
jooyoungseo, line 5: Add the new shiny dependency via `peotry add shiny` to update the `project.toml`

maidr/custom_renderers/pyshiny_renderer.py
jooyoungseo: > During the initial development of the feature, I was able to load the entire maidr object on to the shiny dashboard component but was unable to make use of maidr's interactiveness. Upon analyzing the scripts, I realized that maidr's javascript objects were being loaded on to the webpage but shiny was preventing me from using it.
> 
> In order to have control over the HTML component I am feeding into iframe, I decided to use generic string structures for the HTML tags instead of html tools.
> 
> With the current implementation as a viable fallback, I will begin the migration to htmltools for this section.

Have you taken a look at this method by chance?: https://shiny.posit.co/py/api/core/Htmltools.html

jooyoungseo:
@krishnanand5 I think you have accidentally closed this PR. Please reopen.

krishnanand5:
Greetings Professor,

In an attempt to resolve a merge conflict with poetry.lock, I synced my fork and it caused this unforeseen change. I am rectifying my local repo and will be making a commit shortly with the updated scripts.

Regards,
Krishna Anandan Ganesan.

jooyoungseo:
Aw, I see. Thanks for letting me know!

krishnanand5:
Greetings Professor,

I have addressed all the initial concerns in the respective sections and would like to discuss one change that I have implemented in the `shiny_renderer`. 

As shiny provides complete freedom to the user to define the cosmetics of a component that they are creating, the same applies to plot objects as well. In order to view a maidr plot, we require the component area to be slightly bigger so that the user can experience the interactiveness without having to use scrollbars - in other words, overflow of the web component is a factor to consider/document. 

For now, I have modified the render to expect a width and height parameter during function invocation. If you feel we should handle this in a better way, please do suggest any measures and I will begin working on them. I am investigating `htmltools` to identify any support that could help us achieve this.

Regards,
Krishna Anandan Ganesan.

jooyoungseo:
@krishnanand5 I cannot execute the new example:

image

example/py-shiny/example_pyshiny_reactive_scatter.py
```diff
@@ -0,0 +1,57 @@
+from maidr.custom_renderers.shiny_renderer import render_maidr
+import matplotlib.pyplot as plt
+import pandas as pd
```
jooyoungseo, line 3: Why are you importing `pandas`? Is this used?

maidr/custom_renderers/shiny_renderer.py
```diff
@@ -0,0 +1,63 @@
+from matplotlib.axes import Axes
+from matplotlib.figure import Figure
+import maidr
+from shiny import ui
+from htmltools import tags
+from typing import Any, Optional, Callable, TypeVar, Union, List
+
```
jooyoungseo, line 7: Are `matplotlib` and `htmltools` used after you import?

maidr/custom_renderers/shiny_renderer.py
jooyoungseo, line 1: Some docstrings exceed our line length limit (our default 88 chars). If you use VSCode, you can easily see the linter errors.

jooyoungseo:
@SaaiVenkat and @dakshpokar -- Please review this PR as well.

example/py-shiny/example_pyshiny_reactive_scatter.py
krishnanand5: `pandas` was used with the older example but I did not have that removed for the new scatterplot. It has been removed in the latest commit.

maidr/custom_renderers/shiny_renderer.py
krishnanand5: Yes both imports are not used after they are imported. Remnants from the older version of the `shiny_renderer` and they have been removed in the latest commit.

maidr/custom_renderers/shiny_renderer.py
krishnanand5: I was using `Pylint` on my vscode which was set to a default of `130 characters`.  I realized now that `flake8` is the repository's lint module and hence changed the linter on my local vscode as well. Such issues should not arise in future.

jooyoungseo (approved):
LGTM!

maidr/custom_renderers/shiny_renderer.py
```diff
@@ -0,0 +1,67 @@
...
+    Parameters
+    ----------
+    func: Callable[..., T]
+        The Shiny function which returns an object that can be rendered for
+        displaying the maidr plot.
+    """
+
+    def wrapper(
+        *args: Any,
+        width: Optional[str] = None,
```
dakshpokar, line 23: Hi @krishnanand5,
Can you automatically resize the iframe if height and width are not passed as arguments? Please feel free to refer to #64 for guidance. Considering this library's target audience, we should try to ensure the interface is as user-friendly and accessible as possible.

Apart from that the PR looks clean, good work!

As for the [comment](https://github.com/xability/py_maidr/pull/64/files#r1720523929) that you left on #64 - My observation is that Python notebooks typically have a considerable level of flexibility when it comes to handling cell rendering, which makes suitable style changes possible. Overflow characteristics, however, can be activated if pre-defined styles are used, particularly if the content is larger than the cell's dimensions.

To handle this, you may want to try utilizing Inline-JS/CSS to more accurately regulate the overflow behavior.

example/py-shiny/example_pyshiny_reactive_scatter.py
```diff
@@ -0,0 +1,57 @@
...
+            data=iris, x=input.x_var(), y=input.y_var(), hue="species", ax=ax
+        )
+        ax.set_title(f"Iris {input.y_var()} vs {input.x_var()}")
+        ax.set_xlabel(input.x_var().replace("_", " ").title())
+        ax.set_ylabel(input.y_var().replace("_", " ").title())
+        return s_plot
+
+    @output
+    @render.ui
+    def reactivebarplot():
```
SaaiVenkat, line 48: Could `@render_maidr` be used to render the example? If so, please update this to reflect the same.

example/py-shiny/example_pyshiny_reactive_scatter.py
```diff
@@ -0,0 +1,57 @@
...
+            ),
+        ),
+        ui.column(9, ui.output_ui("reactivebarplot")),
+    )
+)
+
+
+# Define the server
+def server(input, output, session):
+    @reactive.Calc
```
SaaiVenkat, line 35: `reactive.Calc` is mainly used to perform mathematical intensive calculations. May I know why this has been used here?

example/py-shiny/example_pyshiny_reactive_scatter.py
```diff
@@ -0,0 +1,57 @@
...
+        fig, ax = plt.subplots(figsize=(10, 6))
+        s_plot = sns.scatterplot(
+            data=iris, x=input.x_var(), y=input.y_var(), hue="species", ax=ax
+        )
+        ax.set_title(f"Iris {input.y_var()} vs {input.x_var()}")
+        ax.set_xlabel(input.x_var().replace("_", " ").title())
+        ax.set_ylabel(input.y_var().replace("_", " ").title())
+        return s_plot
+
+    @output
```
SaaiVenkat, line 46: This plot has been injected into the shiny through `ui.output_ui`. Why `@output` is required?

maidr/custom_renderers/shiny_renderer.py
SaaiVenkat: - Since iframe seems to addressed here as well, I would suggest to merge this PR after #64 . 
- Also, please create a package called 'widget' under 'maidr' package (maidr/widget/shiny.py), and move the contents there. We'll be extending this further to support GUI widgets in the future.
- We are using the htmltools ui underneath, and shiny uses the same. This annotation duplicates many of the functionalities from shiny as well as from htmltools. This could be simplified by the following

```
from __future__ import annotations

from shiny.render import ui
from shiny.types import Jsonifiable

import maidr


class render_maidr(ui):
    async def render(self) -> Jsonifiable:
        value = await self.fn()
        if value is None:
            return None

        value = maidr.render(value)
        rendered = await self.transform(value)
        return rendered
```

maidr/maidr.py
```diff
@@ -10,6 +11,14 @@
 from maidr.core.figure_manager import FigureManager
 
 
+def render(
+    plot: Any, *, lib_prefix: str | None = "lib", include_version: bool = True
+) -> RenderedHTML:
```
SaaiVenkat, line 16: In order to incorporate the simplified rendering we can return the `Tag` object instead of the `RenderedHTML`.

```
def render(plot: Any) -> Tag:
    ax = FigureManager.get_axes(plot)
    maidr = FigureManager.get_maidr(ax.get_figure())
    return maidr.render()
```

And update core/maidr.py to render the `Tag`
```
    def render(self) -> Tag:
        return self._create_html_tag()
```

example/py-shiny/example_pyshiny_reactive_scatter.py
SaaiVenkat, line 35: `reactive.Calc` is mainly used to perform mathematical intensive calculations. May I know why this has been used here?

example/py-shiny/example_pyshiny_reactive_scatter.py
```diff
@@ -0,0 +1,57 @@
...
+    @render.ui
```
SaaiVenkat, line 47: This plot has been injected into the shiny through `ui.output_ui`. Why `@output` is required?

example/py-shiny/example_pyshiny_reactive_scatter.py
```diff
@@ -0,0 +1,57 @@
...
+    def reactivebarplot():
+        return render_maidr(create_reactivebarplot)(width="100%", height="700px")
```
SaaiVenkat, line 49: Could `@render_maidr` be used to render the example? If so, please update this to reflect the same.

SaaiVenkat (changes requested):

example/py-shiny/example_pyshiny_reactive_scatter.py
krishnanand5: Yes, `reactive.calc` is residual from development. Will remove this from the example

example/py-shiny/example_pyshiny_reactive_scatter.py
krishnanand5: `@output` is for rendering reactive output on the UI but since `@render.ui` is sufficient in our case to render the generated HTML on the shiny app, `@output` is redundant. Will have this removed.

example/py-shiny/example_pyshiny_reactive_scatter.py
krishnanand5: As `@render_maidr` is a decorator, we can invoke this directly.

maidr/maidr.py
krishnanand5: Thank you @SaaiVenkat, for these suggestions. These changes have been incorporated and have made the module more compact

maidr/custom_renderers/shiny_renderer.py
krishnanand5: The requested changes have been addressed in [in this commit](https://github.com/xability/py_maidr/commit/4fdfcd119a90c7470430d0e2f158f431cfec717d).

krishnanand5:
Greetings @SaaiVenkat,

Following are the list of changes I have done in the recent commits:

- Moved `shiny_renderer.py` to `maidr/widget/shiny.py`.
- Updated the example plot to use only `render_maidr` and not use default shiny renderers like `@output` and `@render.ui`.
- `@reactive.calc` served no purpose for the provided example script and has been removed.
- `shiny.py` has been simplified to use the `render` of maidr.
- As discussed, I have removed the conditional check in `_inject_plot` to allow shiny to utilize the same iframe generator as the one @dakshpokar developed for Interactive Python notebooks for Shiny as well.

I did check possible collateral effects of removing the conditional check but turns out only the interactive python module and shiny module are currently leveraging it for iframes. Please let me know if the changes made are as per expectations and if there are any further changes you would like on this module.

Regards,
Krishna Anandan Ganesan.

maidr/widget/shiny.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+from shiny.render import ui
+from shiny.types import Jsonifiable
+
+import maidr
+
+
+class render_maidr(ui):
```
jooyoungseo, line 9: We use `PascalCase` for class name in our package. why do you use `snake_case` here? Is this the Shiny convention?

krishnanand5: I checked Shiny and there are no such conventions. I have renamed the class name to adhere to `PascalCase`.

maidr/widget/shiny.py
```diff
@@ -0,0 +1,29 @@
+from __future__ import annotations
+
+from shiny.render import ui
+from shiny.types import Jsonifiable
+
+import maidr
+
+
+class RenderMaidr(ui):
```
SaaiVenkat, line 9: @krishnanand5 shiny breaks the conventional Pascal case for classes, since this will be used as a decorator, like `@render_maidr`. There have followe this convention for all their renderers. So, we should also follow their convention.

cc: @jooyoungseo

SaaiVenkat (changes requested):

maidr/widget/shiny.py
krishnanand5: Thank you for the confirmation @SaaiVenkat. I thought our repository's naming convention superseded supporting libraries but you are right. Better to follow the naming convention of the supporting libraries. I have reverted the class name to `snake_case`

SaaiVenkat (approved):

Issue #85, opened 2024-08-21
feat: support 2-layered scatter plot


jooyoungseo:
# Feature Request

## Description

Currently, our examples only contain single-layered scatter plots. We should add examples that demonstrate how to create 2-layered scatter plots.

## Motivation

`maidr.js`supports multi-layered plot just like [this 2-layered scatter plot](https://xabilitylab.ischool.illinois.edu/maidr/galleries/scatterplot.html). Users can switch between layers using PageUp/PageDown keys. We need to support this feature in `py_maidr` as well.

## Proposed Solution

Support a 2 layered scatter plot like below:

```python
# !pip install statsmodels
import matplotlib.pyplot as plt
import seaborn as sns

# Load the penguins dataset
penguins = sns.load_dataset("penguins")

# Initialize the plot
plt.figure(figsize=(10, 6))

# Layer 1: Scatter plot
sns.scatterplot(data=penguins, x="flipper_length_mm", y="body_mass_g", s=100)

# Layer 2: Smoothed line (LOWESS)
sns.regplot(
    data=penguins,
    x="flipper_length_mm",
    y="body_mass_g",
    scatter=False,
    lowess=True,
    line_kws={"color": "black"},
)

# Add labels and title
plt.title("Penguin Flipper Length vs Body Mass (Smoothed)", fontsize=16)
plt.xlabel("Flipper Length (mm)", fontsize=14)
plt.ylabel("Body Mass (g)", fontsize=14)
plt.show()
```

## Additional Context

N/A

## Checklist

- [x] I have searched for similar feature requests and confirmed that this is a new request.
- [x] I have provided a clear and concise description of the feature.
- [x] I have explained the motivation behind this feature request.
- [x] I have outlined a proposed solution or ideas for implementing this feature.
- [x] I have provided any additional context or screenshots if applicable.

PR #100, opened 2024-08-27
feat: support 2-layered scatter plot

krishnanand5:
## Description

This PR aims to add demonstrations of how users can leverage Maidr to create double-layered scatterplots.

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Documentation update

## Checklist

- [X] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

# Pull Request

## Description
This PR aims to include examples of how double-layered scatter plots can be created using maidr

## Related Issues
#85 

## Changes Made
This is a draft PR that has the new 2-layered scatterplot example. As the example is still in development phase, I have included the html generation as part of the example. Additionally, the `_inject_plot` method has been modified to accept a flag called `html`. If set to `True`, the `_inject_plot` method will return the bare html for the plot. The default value is `False` and if the flag is set to `False`, the iframe element containing the plot will be returned.

## Checklist

- [X] I have read the [Contributor Guidelines](../CONTRIBUTING.md).
- [ ] I have performed a self-review of my own code and ensured it follows the project's coding standards.
- [X] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass.
- [ ] I have added appropriate unit tests, if applicable.

## Additional Notes

I am encountering the following issues by leveraging the current py-binder and upstream dependancies.

- In the maidr [example](https://xabilitylab.ischool.illinois.edu/maidr/galleries/scatterplot.html) shared on the issue, I am unable to use the &#8593; / &#8595; to switch between the line plot and scatterplot. 
- In the example I developed, the text related to the active component gets updated everytime I move from point to point on the scatterplot but the active components are not highlighted. 

I have this error showing up on DevTools when the html is validated:
scatter_2Layer

Whilst the secondary plot toggle seems to be an issue on the upstream repository, I would like to know if any additional changes should be done on the py-binder to activate highlighting of scatterplot components.

### DIFF
--- a/example/scatter/example_scatter_2_layer.py
+++ b/example/scatter/example_scatter_2_layer.py
@@ -1,0 +1,36 @@
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+import maidr
+
+
+penguins = sns.load_dataset("penguins")
+
+# Initialize the plot
+fig, ax = plt.subplots(figsize=(10, 6))
+
+# Prepare Scatter plot layer
+sns.scatterplot(data=penguins, x="flipper_length_mm", y="body_mass_g", s=100, ax=ax)
+
+# Prepare Line plot layer
+sns.regplot(
+    data=penguins,
+    x="flipper_length_mm",
+    y="body_mass_g",
+    scatter=False,
+    lowess=True,
+    line_kws={"color": "black"},
+    ax=ax
+)
+
+ax.set_title("Penguin Flipper Length vs Body Mass (Smoothed)", fontsize=16)
+ax.set_xlabel("Flipper Length (mm)", fontsize=14)
+ax.set_ylabel("Body Mass (g)", fontsize=14)
+
+# Create a Maidr instance
+maidr_plot = maidr.Maidr(fig)
+
+html_string = maidr_plot.render(html=True).get_html_string()
+
+with open("example_scatter_2_layered.html", "w") as f:
+    f.write(html_string)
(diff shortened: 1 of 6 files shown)


jooyoungseo:
@krishnanand5 It's PageUp or PageDown key to switch between layers in the [example](https://xabilitylab.ischool.illinois.edu/maidr/galleries/scatterplot.html). In the smoothed line layer, you need to use the autoplay key because it is a continuous line.

jooyoungseo:
@krishnanand5 -- Actually, the second layer is different from the typical "line" graph where we can use LeftArrow or RightArrow keys. I think we will have to revisit the property value name here if we use "line" for both "line graph" and "smoothed line" which needs to be distinguished from one another.

jooyoungseo:
@krishnanand5 Take a look at [this line](https://github.com/xability/maidr/blob/1ef9a00f7b37a9648b45454773c8c3b535598ed7/galleries/scatterplot.html#L4594-L4595).

You need to do:

1. Inject `x` and `y` values for the second "smooth" layer from `matplotlib`: currently, our py binder only picks up the first "point" layer.

1. Include the second layer data structure within the JSON schema just like [our example](https://github.com/xability/maidr/blob/1ef9a00f7b37a9648b45454773c8c3b535598ed7/galleries/scatterplot.html#L4594-L5878).

PR #102, opened 2024-08-29
feat: fetch LLM API keys from user env variables

### LINKED ISSUE #76: fix(llm): address an issue where saved AI API keys are not carried over to the next sessions
## Reproducible Steps

1. Create any plots and display via `maidr.show()`

1. Hit H key and provide API keys for both OpenAI and Google Gemini

1. Click Save and Close

1. If prompted, save the password in your browser

1. Open LLM from the interactive plot area via Ctrl+Shift+/ (on Windows) or Alt+Shift+/ (on Mac)

1. Make sure the AI responses are working

1. Close the maidr browser and exit out of the current Python repl

1. Repeat the steps above and see if the AI API keys are preserved to the next sessions

## Current Behavior

API keys are not carried over to the next sessions even when it's saved in the browser

## Suggested Solution

1. When `maidr.show` is executed search for the following env variables via `os.getenv()`:

* `OPENAI_API_KEY`

* `GOOGLE_API_KEY`

1. Use these keys as fallback

dakshpokar:
## Description
This pull request fixes the handling of API keys for LLMs in the code. It adds a JavaScript script to handle the API keys for LLMs and initializes the LLM secrets in the MAIDR instance. The script injects the LLM API keys into the MAIDR instance and sets the appropriate settings based on the presence of the Gemini and OpenAI API keys. This ensures that the LLM functionality works correctly with the updated API key handling.

closes #76 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Checklist

# Pull Request

## Description
1. Added a new method called `initialize_llm_secrets()` in environment.py which fetches the keys from the environment variable.
2. Injected the script when the maidr iframe loads initially.

## Checklist

### DIFF
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -1,18 +1,17 @@
 from __future__ import annotations
 
-from typing import Literal
-
 import io
 import json
 import uuid
+from typing import Literal
 
 from htmltools import HTML, HTMLDocument, Tag, tags
 from lxml import etree
-
 from matplotlib.figure import Figure
 
 from maidr.core.context_manager import HighlightContextManager
 from maidr.core.plot import MaidrPlot
+from maidr.util.environment import Environment
 
 
 class Maidr:
@@ -172,6 +171,7 @@ class Maidr:
 
         # Embed the rendering into an iFrame for proper working of JS library.
         base_html = tags.iframe(
+            id="maidr-iframe",
             srcdoc=str(base_html.get_html_string()),
             width="100%",
             height="100%",
@@ -181,7 +181,8 @@ class Maidr:
             onload="""
                 this.style.height = this.contentWindow.document.body.scrollHeight +
                 100 + 'px';
-            """,
+            """
+            + Environment.initialize_llm_secrets(),
         )
 
         return base_html
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -1,10 +1,9 @@
 from __future__ import annotations
 
-from typing import Literal
-
 import io
 import json
 import uuid
+from typing import Literal
 
 from htmltools import HTML, HTMLDocument, Tag, tags
 from lxml import etree
@@ -170,9 +169,11 @@ class Maidr:
             tags.script(maidr),
         )
 
+        unique_id = Maidr._unique_id()
+
         # Embed the rendering into an iFrame for proper working of JS library.
         base_html = tags.iframe(
-            id="maidr-iframe",
+            id=unique_id,
             srcdoc=str(base_html.get_html_string()),
             width="100%",
             height="100%",
@@ -183,7 +184,7 @@ class Maidr:
                 this.style.height = this.contentWindow.document.body.scrollHeight +
                 100 + 'px';
             """
-            + Environment.initialize_llm_secrets(),
+            + Environment.initialize_llm_secrets(unique_id),
         )
 
         return base_html
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -169,7 +169,7 @@ class Maidr:
             tags.script(maidr),
         )
 
-        unique_id = Maidr._unique_id()
+        unique_id = "iframe_" + Maidr._unique_id()
 
         # Embed the rendering into an iFrame for proper working of JS library.
         base_html = tags.iframe(
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
+from typing import Literal
+
 import io
 import json
 import uuid
-from typing import Literal
 
 from htmltools import HTML, HTMLDocument, Tag, tags
 from lxml import etree
--- a/maidr/util/environment.py
+++ b/maidr/util/environment.py
@@ -17,7 +17,7 @@ class Environment:
             return False
 
     @staticmethod
-    def initialize_llm_secrets() -> str:
+    def initialize_llm_secrets(unique_id: str) -> str:
         """Inject the LLM API keys into the MAIDR instance."""
 
         gemini_api_key = os.getenv("GEMINI_API_KEY")
@@ -53,9 +53,11 @@ class Environment:
         elif openai_api_key is not None:
             settings["LLMOpenAiMulti"] = True
             settings["openAIAuthKey"] = openai_api_key
+            settings["LLMModel"] = "openai"
         elif gemini_api_key is not None:
             settings["LLMGeminiMulti"] = True
             settings["geminiAuthKey"] = gemini_api_key
+            settings["LLMModel"] = "gemini"
 
         settings_data = json.dumps(settings)
 
@@ -73,7 +75,7 @@ class Environment:
                 }}
             }}
             addKeyValueLocalStorage(
-                'maidr-iframe', 'settings_data', JSON.stringify({settings_data})
+                '{unique_id}', 'settings_data', JSON.stringify({settings_data})
             );
         """
 
--- a/maidr/util/environment.py
+++ b/maidr/util/environment.py
@@ -1,3 +1,7 @@
+import json
+import os
+
+
 class Environment:
     @staticmethod
     def is_interactive_shell() -> bool:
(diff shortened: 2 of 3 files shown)


dakshpokar:
Greetings Professor @jooyoungseo,

While this PR addresses the issue of LLM keys being lost due to changes in the port number each time a Maidr instance is run, I believe we should consider a more scalable solution, especially in light of potential future changes. I discussed this matter with @SaaiVenkat, and he shares the concern that these fixes do not address the core problem but rather serve as a workaround for a dependency issue with the library. This issue could have been more easily resolved if `py-htmltools` provided a way to fix the port.

In my opinion, a more sustainable solution would be to implement a custom server to serve our HTML files, rather than relying on py-htmltools for this functionality. However, this is just my suggestion, and I would appreciate your thoughts on the matter.

Best Regards,
Daksh Pokar

jooyoungseo:
I had a Slack communication with @dakshpokar  and we agreed that we need to use a custom local server and pin down a port number for an interactive console sessions (e.g., vanilla Python repl; ipython repl) while keeping the current behavior of `htmltools.show()` for shiny and any other web widgets. This way, we can preserve the same API keys across different sessions. Of course, this may be a stopgap solution and there might be a better way to handle this case, but we want to patch this first for the sake of the importance of this issue against our end-users.

maidr/core/maidr.py
```diff
@@ -1,18 +1,17 @@
 from __future__ import annotations
 
-from typing import Literal
```
SaaiVenkat, line 3: Why is this changed?

maidr/core/maidr.py
```diff
@@ -1,18 +1,17 @@
...
-
 import io
 import json
 import uuid
+from typing import Literal
 
 from htmltools import HTML, HTMLDocument, Tag, tags
 from lxml import etree
-
```
SaaiVenkat, line 11: Why is this line removed?

maidr/core/maidr.py
```diff
@@ -1,18 +1,17 @@
 from __future__ import annotations
 
-from typing import Literal
-
 import io
 import json
 import uuid
+from typing import Literal
```
SaaiVenkat, line 6: Please revert this

SaaiVenkat:
> I had a Slack communication with @dakshpokar and we agreed that we need to use a custom local server and pin down a port number for an interactive console sessions (e.g., vanilla Python repl; ipython repl) while keeping the current behavior of `htmltools.show()` for shiny and any other web widgets. This way, we can preserve the same API keys across different sessions. Of course, this may be a stopgap solution and there might be a better way to handle this case, but we want to patch this first for the sake of the importance of this issue against our end-users.

Professor @jooyoungseo , since this is stopgap solution, I would suggest to create a backlog ticket right away to address this issue and link it with this issue.

cc: @dakshpokar

jooyoungseo:
Update: @dakshpokar and I decided to make a separate PR addressing the custom local server implementation. However, I vote for the current approach where users save their API keys in their env variables (e.g., `OPENAI_API_KEY`; `GEMINI_API_KEY`) for the recurrent and stable use across different sessions. This approach also aligns with other major libs (e.g., openai; langchain). I will merge this PR today and will create another issue for the custom server.

jooyoungseo:
@dakshpokar `GEMINI_API_KEY` seems to be an officially recommended key name: https://github.com/google-gemini/generative-ai-python

maidr/util/environment.py
```diff
@@ -11,3 +15,66 @@ def is_interactive_shell() -> bool:
...
+                        iframe.contentWindow.localStorage.setItem(key, value);
+                    }} catch (error) {{
+                        console.error('Error accessing iframe localStorage:', error);
+                    }}
+                }} else {{
+                    console.error('Iframe not found or inaccessible.');
+                }}
+            }}
+            addKeyValueLocalStorage(
+                'maidr-iframe', 'settings_data', JSON.stringify({settings_data})
```
jooyoungseo, line 76: Do NOT use the same `iframeId` here. Based on my testing result, the user env variable change does not get picked up dynamically once a setting is stored to this iframe. since you have pinned down the iframeId as "maidr-iframe-id,"the same LLM setting is preserved regardless of user's env variable update. Use a random id just like https://github.com/dakshpokar/py_maidr/blob/2e08b9d7803c7d544e897cc4b063a5af0f473b8b/maidr/core/maidr.py#L145-L149

However, I still suggest you bind this local store handler to maidr object instead of iframe.

maidr/util/environment.py
```diff
@@ -11,3 +15,66 @@ def is_interactive_shell() -> bool:
...
+            settings["openAIAuthKey"] = openai_api_key
+            settings["LLMOpenAiMulti"] = True
+            settings["LLMGeminiMulti"] = True
+            settings["LLMModel"] = "multi"
+        elif openai_api_key is not None:
+            settings["LLMOpenAiMulti"] = True
+            settings["openAIAuthKey"] = openai_api_key
+        elif gemini_api_key is not None:
+            settings["LLMGeminiMulti"] = True
+            settings["geminiAuthKey"] = gemini_api_key
```
jooyoungseo, line 58: Only when gemini key is available, you need to set to gemini. If we don't this, this will fall back to the default "openai" model.

```suggestion
        elif gemini_api_key is not None:
            settings["LLMGeminiMulti"] = True
            settings["geminiAuthKey"] = gemini_api_key
            settings["LLMModel"] = "gemini"
```

maidr/util/environment.py
jooyoungseo: I have tested this again and the env variable update gets picked up as expected even with this fixed id. However, please use `class` instead of `id` if you need to go with this tweak. For `id`, assign a random unique iframe id if needed.

maidr/util/environment.py
```diff
@@ -11,3 +15,66 @@ def is_interactive_shell() -> bool:
...
+
+        if gemini_api_key is not None and openai_api_key is not None:
+            settings["geminiAuthKey"] = gemini_api_key
+            settings["openAIAuthKey"] = openai_api_key
+            settings["LLMOpenAiMulti"] = True
+            settings["LLMGeminiMulti"] = True
+            settings["LLMModel"] = "multi"
+        elif openai_api_key is not None:
+            settings["LLMOpenAiMulti"] = True
+            settings["openAIAuthKey"] = openai_api_key
```
jooyoungseo, line 55: Specifying "openai" is safer as the default value might change later.

```suggestion
        elif openai_api_key is not None:
            settings["LLMOpenAiMulti"] = True
            settings["openAIAuthKey"] = openai_api_key
            settings["LLMModel"] = "openai"
```

maidr/util/environment.py
dakshpokar: Sure Professor @jooyoungseo, making this change!

maidr/util/environment.py
dakshpokar: Thank you Professor @jooyoungseo, will check this!

maidr/util/environment.py
dakshpokar: Sure Professor @jooyoungseo, making this change!

maidr/util/environment.py
dakshpokar: Professor, @jooyoungseo. I have assigned a unique id to the maidr iframe. I could've used class but it returns an array of matched elements which I have to then loop over to add the key values.

maidr/core/maidr.py
```diff
@@ -170,8 +169,11 @@ def _inject_plot(plot: HTML, maidr: str) -> Tag:
             tags.script(maidr),
         )
 
+        unique_id = Maidr._unique_id()
+
```
jooyoungseo, line 173: Are you recycling the same unique_id used for maidr object towards iframe?This may cause some issues in the long run because there are two matches per id (one for iframe; the other for maidr). What do you think about adding a prefix (e.g., "iframe_") to distinguish the iframe id from the maidr one?

maidr/util/environment.py
```diff
@@ -11,3 +15,68 @@ def is_interactive_shell() -> bool:
             )
         except ImportError:
             return False
+
+    @staticmethod
+    def initialize_llm_secrets(unique_id: str) -> str:
```
jooyoungseo, line 20: Maybe I could be wrong, but don't we use `_` prefix for static method?

dakshpokar: Professor @jooyoungseo, it is generally used for the methods that are private or internal.

maidr/core/maidr.py
dakshpokar: Professor @jooyoungseo, as the method generates a unique uuid everytime the `Maidr._unique_id()` is called, the conflict might not arise, however, to be on a safer side I will add a prefix iframe_ for the id.

jooyoungseo (approved):
I have tested all the possible edge cases with various env variable combinations. It is working very reliably. LGTM! I am merging it now.

PR #121, opened 2024-10-11
feat(maidr.show): use tempfile for interactive sessions

### LINKED ISSUE #103: refactor(maidr.show): use a tempfile instead of a local server in an interactive session
## Current Issue

The default behavior of `maidr.show()` is serving an HTML content in a local server with a random port number. This depends on `htmltools.show()`.

However, serving our static content in a local server is an overkill and unnecessary. The random port assignment causes some issues like [#76](https://github.com/xability/py_maidr/issues/76)76.

Unless we need to keep watching the figure updates, we need to use a lightweight solution. For example, we can use a tempfile instead of a local server in an interactive session.

Even if we want to watch the figure updates, the current `thmltools.show()` shuts down the server right after displaying the content. Thus, there is no reason for us to use a local server for the current purpose.

Then, why do we need to use `hmlttools`? Quarto, Shiny, and R Markdown, all of which we need to support for reproducibility depends on this package. So, I will not remove the dependency on `htmltools`.

## solutions

1. Check if the current session is interactive.

1. If interactive, `save_html()` as a  tempfile, and open it with the default browser.

1. If not interactive, use `htmltools.show()` as it is.

dakshpokar:
## Description
This PR introduces a more efficient and lightweight solution by serving static HTML content through temporary files in interactive sessions. Instead of relying on a local server and random port assignments, this update will create and open a temporary HTML file using tempfile and webbrowser.open()

closes #103 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Checklist

# Pull Request

## Description

1. Introducing _open_plot_in_browser(), which generates a temporary file for the rendered plot and opens it in the browser.
2. Adding logic in show() to handle browser-based rendering when the environment is not a notebook or interactive session.

## Changes Made

1. The code now avoids serving static HTML through a local server by using temporary files in interactive sessions.
2. Introduction of _open_plot_in_browser(): A new method _open_plot_in_browser() has been added to handle rendering the HTML content into a temporary file and automatically opening it in the browser.
3. Logic Update in show() Method: The show() method now checks if the environment is interactive and, if so, uses the new temporary file approach instead of relying on a local server.
4. get_renderer() Method Update: The get_renderer() method was added to return the appropriate renderer (ipython or browser), ensuring compatibility with different environments.


dakshpokar:
Greetings Professor @jooyoungseo,

This PR introduces a key optimization by replacing the existing behavior of serving HTML content via a local server with random port assignment, which was causing issues, with a more lightweight approach for interactive sessions. Instead of using htmltools.show() to serve content on a local server, we now use a temporary file and open it in the browser directly, which avoids unnecessary server overhead and port conflicts. A new method, _open_plot_in_browser(), handles the creation and opening of this temporary file. The show() method has been updated to implement this new logic when the environment is interactive. 

Best Regards,
Daksh Pokar

jooyoungseo (approved):

Issue #119, opened 2024-10-09
fix(save_html): address a critical issue where LLM secret keys are stored


jooyoungseo:
# Bug Report

## Description

I have found this critical issue when working with Quarto document. Saving the LLM API keys within the iframe can cause a critical security breach in the case where users share the rendered HTML document with others.

We need to find a more secure way to store the LLM API keys.

## Steps to Reproduce

Rendre the quarto document in `example/quarto/demo.qmd`

## Actual Behavior

API keys are stored in the rendered HTML document.

## Expected Behavior

API keys should not be stored in the rendered HTML document.

## Screenshots

N/A

## Environment

Universal

## Additional Information

N/A

PR #124, opened 2024-10-17
fix: address iframe resizing issue in jupyter notebooks

dakshpokar:
## Description
This PR includes changes for removing overflow on iframes and resizing them according to content of the plot.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Checklist

# Pull Request

## Description

This PR enhances iframe resizing behavior to dynamically adjust height and width based on content size and user interaction. Specifically, the height is recalculated with additional space (100px) when the braille-input field is focused, and 50px otherwise. This ensures the iframe is not clipped during braille mode. The script also adds event listeners for focusin, focusout, and resize events to trigger appropriate resizing as needed, which were added by @krishnanand5.

### DIFF
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -193,24 +193,36 @@ class Maidr:
             resizing_script = f"""
                 function resizeIframe() {{
                     let iframe = document.getElementById('{unique_id}');
+
                     if (iframe && iframe.contentWindow && iframe.contentWindow.document) {{
                         let iframeDocument = iframe.contentWindow.document;
-                        let brailleContainer = iframeDocument.getElementById('braille-div');
+                        let brailleContainer = iframeDocument.getElementById('braille-input');
+
+                        iframe.style.height = 'auto';
+
                         let height = iframeDocument.body.scrollHeight;
                         if (brailleContainer && brailleContainer === iframeDocument.activeElement) {{
-                            height *= 1.35;  # Increase height by 35% if braille-container is in focus
+                            height += 100;
+                        }}else{{
+                            height += 50
                         }}
-                        iframe.style.height = (height + 150) + 'px';
+
+                        iframe.style.height = (height) + 'px';
                         iframe.style.width = iframeDocument.body.scrollWidth + 'px';
                     }}
                 }}
                 let iframe = document.getElementById('{unique_id}');
+                resizeIframe();
                 iframe.onload = function() {{
                     resizeIframe();
                     iframe.contentWindow.addEventListener('resize', resizeIframe);
-                    iframe.contentWindow.document.addEventListener('focusin', resizeIframe);
-                    iframe.contentWindow.document.addEventListener('focusout', resizeIframe);
                 }};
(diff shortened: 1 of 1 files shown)


maidr/core/maidr.py
```diff
@@ -193,24 +193,36 @@ def generate_iframe_script(unique_id: str) -> str:
             resizing_script = f"""
                 function resizeIframe() {{
                     let iframe = document.getElementById('{unique_id}');
+
                     if (iframe && iframe.contentWindow && iframe.contentWindow.document) {{
                         let iframeDocument = iframe.contentWindow.document;
-                        let brailleContainer = iframeDocument.getElementById('braille-div');
+                        let brailleContainer = iframeDocument.getElementById('braille-input');
+
+                        iframe.style.height = 'auto';
```
krishnanand5, line 201: Greetings @dakshpokar,

I think we can set the height to auto in the `tags.iframe` construction. We set the height to 100% there and we are overriding it here. I removed this assignment and checked on Shiny by setting height to auto in `tags.iframe` construction - seems to work fine. Do check with Jupyter and share your thoughts on this.

maidr/core/maidr.py
```diff
@@ -193,24 +193,36 @@ def generate_iframe_script(unique_id: str) -> str:
...
-                    iframe.contentWindow.document.addEventListener('focusout', resizeIframe);
                 }};
+                iframe.contentWindow.document.addEventListener('focusin', () => {{
+                    resizeIframe();
+                }});
+                iframe.contentWindow.document.addEventListener('focusout', () => {{
+                    resizeIframe();
+                }});
             """
             return resizing_script
```
krishnanand5, line 228: on the `tags.iframe` assignment, I did check your old PR where you have loaded the LLM credentials from environment storage. My PR from yesterday inadvertently removed that assignment. You can refer to your old PR to add that to the `onLoad` argument in `tags.iframe`.

krishnanand5 (commented):
Hey @dakshpokar,

The changes are fine on Shiny. Thank you for taking care of the dynamic height adjustment when braille is active. Let me know if we can address my comments.

Regards,
Krishna Anandan Ganesan.

maidr/core/maidr.py
dakshpokar: @krishnanand5 We had to remove the LLM keys injection script, due a security flaw that we discovered in #119.

maidr/core/maidr.py
dakshpokar: That's true I was going to add in the tags but the problem with that is that everytime the focus in and out script runs I want the original scrollable height of the content inside iframe. If we don't reset the previous height, the script was adding additional (not required) height. Hence, I just reset the height when the script runs so as to get the minimum scrollable height and then on the basis of that do the rest of the calculations.

krishnanand5: Sure, in that case this change is feasible.

krishnanand5 (approved):

jooyoungseo (approved):
LGTM!

PR #129, opened 2025-01-15
ci: update poetry.lock to adhere to v2.0.0

krishnanand5:
## Description

This PR updates poetry.lock to aid in virtualenv version updation.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,7 +47,6 @@ jobs:
       - name: Install dependencies
         if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
         run: |
-          poetry lock --no-update
           poetry install
           python -m pip install pytest-github-actions-annotate-failures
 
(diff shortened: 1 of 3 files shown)


krishnanand5:
Some changes had to be made in accordance with Poetry bump up to 2.0.0 to adhere to new format changes. 
metadata under [tools.poetry] has been moved to [project] section to comply with PEP 621. Authors info was expected to be an object rather than string so the representation has been updated.

Additionally, I investigated the changes that have come in Poetry 2.0.0 and these are the two major changes:
- Poetry 2.0.0 (released January 4, 2025) changed the default behavior of the lock command
- The --no-update flag is now the default behavior in Poetry 2.0.0

Hence from `release.yml`, we need to exclude `poetry lock --no-update`. The reason why the current jobs failed is because the job in main still has the old copy of `release.yml`. 

I tested the changed `release.yml` file using netkos/act and the execution was successful. So we may have to merge this PR and validate if it fixes the broken action.

jooyoungseo (approved):

PR #59, opened 2024-07-23
feat: support hightlighing except for segmented plots and boxplots

SaaiVenkat:

PR #132, opened 2025-02-10
fix: address an issue where stacked bar plot doesn't work

### LINKED ISSUE #82: fix(stackedplot): address an issue where stack plot doesn't work
# Bug Report

## Description

We have two stacked plot examples in: example/stacked/matplotlib/example_mpl_stacked.py and example/stacked/seaborn/example_sns_stacked.py. Both of which are failing. @SaaiVenkat has mentioned that this has happened with the new API design. 

## Steps to Reproduce

1. Run the example/stacked/seaborn/example_sns_stacked.py or example/stacked/matplotlib/example_mpl_stacked.py
2. Check the output, examples fail to run.

## Error/Stack-Trace Information
```
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
  File "<stdin>", line 3, in main
  File "/Users/dakshpokar/RA/py_maidr/maidr/api.py", line 29, in stacked
    return FigureManager.create_maidr(ax, PlotType.STACKED)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dakshpokar/RA/py_maidr/maidr/core/figure_manager.py", line 53, in create_maidr
    raise ValueError("No plot found.")
ValueError: No plot found.
```

## Additional Information

Cannot add an example file for the stacked plot in #64 because of this issue.

dakshpokar:
## Description
This PR fixes the issue with stacked bar plot in pymaidr. Also it solves the problem where user had to create the object of stacked bar plot and then pass it to the function maidr.stacked(). Now with this change user will not need to pass the plot object to maidr.stacked(plot) instead they can directly call maidr.show(plot) like we do it for the other plots.

closes #82

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Checklist


dakshpokar:
Greetings Professor @jooyoungseo & @SaaiVenkat,
This pull request resolves the issue with the stacked bar plot by implementing a minor patch to the bar function. Consequently, the previous requirement to create a separate object for the stacked bar plot and subsequently invoke interactivity via the maidr.stacked method has been eliminated. We can now directly call maidr.show(plot) in the same manner as with other plot types.

Please note that this pull request does not address the highlighting issue. The highlight feature in PyMaidr was introduced in [this pull request](https://github.com/xability/py-maidr/pull/59/). To fully resolve the highlighting functionality for stacked bar plots, we will need to develop a more generic solution. Currently, the desired elements are extracted by adding the attribute maidr="true" in the SVG; however, a more universal approach is necessary to extend this functionality across similar plots.
I will be creating a separate issue for this as we need to support highlight issue for not just segmented plots like stacked bar plot but also for box plot.

Thanks,
Daksh Pokar

maidr/patch/common.py
```diff
@@ -20,6 +21,11 @@ def common(plot_type, wrapped, _, args, kwargs) -> Any:
         # Patch the plotting function.
         plot = wrapped(*args, **kwargs)
 
+    if "bottom" in kwargs:
+        bottom = kwargs.pop("bottom")
```
SaaiVenkat, line 25: Is there any requirement for popping the bottom instead of just fetching it?

maidr/patch/common.py
```diff
@@ -20,6 +21,11 @@ def common(plot_type, wrapped, _, args, kwargs) -> Any:
         # Patch the plotting function.
         plot = wrapped(*args, **kwargs)
 
+    if "bottom" in kwargs:
```
SaaiVenkat, line 24: common patch exists to reduce the duplicate among all the common elements. If we going to introduce plot based condition, I would suggest to create the separate patch for bar and instead of specifying them in the generic patch

SaaiVenkat (changes requested):

maidr/patch/common.py
dakshpokar: @SaaiVenkat that seems fair, let me address this!

maidr/patch/common.py
dakshpokar: @SaaiVenkat let me fix this!

SaaiVenkat (approved):

Issue #185, opened 2025-06-06
feat(histogram): support KDE curve overlay


jooyoungseo:
# Feature Request

## Description

Once #184 is addressed, we need to support histograms having KDE curve overlay.

## Proposed Solution

Check the current histogram has the KDE overlay. If so, add a smooth layer on top of it like [this example](https://github.com/xability/maidr/blob/1bb83c2d4db45dd70657ce36548e0d1cc6bee021/examples/smoothplot.html#L1-L5678).

## Additional Context

Below is a simple histogram having the KDE overlay:

```python
import numpy as np
import seaborn as sns
import matplotlib.pyplot as plt

# 1) Sample data
data = np.random.randn(500)

# 2) Plot histogram + KDE together
sns.histplot(data, bins=30, stat='density', kde=True, alpha=0.3)

plt.xlabel('Value')
plt.ylabel('Density')
plt.title('Seaborn Histogram with KDE Overlay')
plt.tight_layout()
plt.show()
```

Issue #184, opened 2025-06-06
feat: support density plot


jooyoungseo:
# Feature Request

## Description

We need to support density plots (aka, KDE plot).

## Proposed Solution

Use smooth layer for density plots like [this example](https://github.com/xability/maidr/blob/1bb83c2d4db45dd70657ce36548e0d1cc6bee021/examples/smoothplot.html#L1-L5678). We just need a single-smooth layer.

## Additional Context

Below is a simple density plot example:

```python
import numpy as np
import seaborn as sns
import matplotlib.pyplot as plt

# 1) Sample data
data = np.random.randn(500)

# 2) Plot the KDE curve
sns.kdeplot(data)
plt.xlabel('Value')
plt.ylabel('Density')
plt.title('Seaborn KDE Plot')
plt.tight_layout()
plt.show()
```

PR #189, opened 2025-06-13
feat: add density layer support in kde and hist plot

### LINKED ISSUE #183: feat(scatterplot): support an overlaid regression line
# Feature Request

## Description

We need to support two-layered scatterplot like [this example](https://github.com/xability/maidr/blob/1bb83c2d4db45dd70657ce36548e0d1cc6bee021/examples/smoothplot.html#L1-L5678).

## Proposed Solution

When `sns.regplot()` is used, check whether it is set to include an overlaid regression line. If not, create a single point-only layer like what we do for `sns.scatterplot()`. If the plot has a regression line layer, add the smooth layer [like this](https://github.com/xability/maidr/blob/1bb83c2d4db45dd70657ce36548e0d1cc6bee021/examples/smoothplot.html#L1-L5678).

## Additional Context

Below is a simple scatterplot example having an overlaid regression line:

```python
import numpy as np
import seaborn as sns

# 1) Generate sample data
#    Let x be evenly spaced, and y = sin(x) plus noise.
np.random.seed(0)
x = np.linspace(0, 10, 100)
y = np.sin(x) + 0.3 * np.random.randn(100)

# 2) Create scatter + LOESS smoothing
plt.figure(figsize=(6, 4))
sns.regplot(x=x, y=y, lowess=True, 
            scatter_kws={'s': 30, 'alpha': 0.6},  # point size and transparency
            line_kws={'color': 'red', 'lw': 2})   # LOESS curve style

plt.xlabel('x')
plt.ylabel('y')
plt.title('Scatterplot with LOESS (Seaborn)' )
plt.tight_layout()
plt.show()
```

nk46-cloud:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Checklist

- [ ] My changes generate no new warnings

# Pull Request

## Description

1. Added KDE plot functionality handing for matplotlib & seaborn (registered as smooth curves)
2. Added new example scripts to demonstrate histogram plots with KDE overlays for both Matplotlib and Seaborn, showcasing MAIDR integration.
3. Enhanced the backend (maidr.pt) to better distinguish line vs smooth curves(moved from lineplot.py as it broke highlighting for line)
4. Implemented cache busting in maidr.py by appending a timestamp query parameter to the MAIDR JS CDN URL, ensuring the latest frontend code is always loaded.

## Related Issues
Closes  #183 #184 #185 

## Changes Made

1. Updated examples in example/kde, example/histogram
2. Implemented cache busting in maidr.py
3. Added patch files for seaborn, matplotlib kde plots
4. Histogram with kde curve is also now supported


example/kde/seaborn/multiple_kde_seaborn.py
jooyoungseo, line 1: Add a single KDE plot example for sns.

nk46-cloud:
> ` [cab7934](/xability/py-maidr/pull/189/commits/cab793468704f005628a353f7708f6e82e3b2bb5)`

Addressed in commit cab7934

jooyoungseo (approved):

PR #299, opened 2026-08-04
feat(step): add step plot support with ordinal level names

jooyoungseo:
## Description

Adds `PlotType.STEP` so matplotlib step charts are exported as MAIDR step layers instead of being mislabelled as line layers.

The motivating case is a **hypnogram** — sleep stage against time. Its y axis is an ordinal category (Awake / REM / N1 / N2 / N3) and its value is *piecewise constant*: a stage is held for a stretch of the night and then jumps. Exporting that as a line chart tells a blind user the sleeper glided continuously through the intermediate stages, which is not what the data says.

`ax.step(...)` already reached maidr before this change, because matplotlib's `Axes.step` is not its own renderer — it sets `drawstyle="steps-*"` and delegates to `Axes.plot`, which `maidr/patch/lineplot.py` already wraps. So this is a **re-classification**, not a new interception point. That matters: adding a second `wrapt` wrapper on `Axes.step` would have registered two layers (STEP *and* LINE) for a single call. Instead the existing `line()` wrapper now chooses the type by reading the resulting artists' drawstyles, which covers `ax.step`, `plt.step`, `ax.plot(drawstyle="steps-post")` and `sns.lineplot(drawstyle="steps-mid")` under one rule.

**Dependency status:** the JS side (xability/maidr#723) has **merged**, so `TraceType.STEP` is on `maidr` `main`. Online rendering picks it up via `maidr@latest` as soon as the next release is published — the latest tag is still `v3.75.1`, which predates the merge. Offline rendering (`use_cdn=False`) additionally needs the routine bundled-`maidr.js` bump; `maidr/static/VERSION` is currently 3.73.0.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

**Classification**
- `PlotType.STEP = "step"`; `MaidrKey.STEP_DIRECTION = "stepDirection"`.
- `maidr/util/step_utils.py` — one shared rule for recognising step lines and their convention. An axes is a step plot only when *every* data-bearing line on it is a step line; one ordinary line mixed in keeps the conservative `line` classification, since maidr emits a single layer per axes.
- `maidr/patch/lineplot.py` — picks the type from the artists rather than hard-coding `PlotType.LINE`.

**`StepPlot`** (`maidr/core/plot/stepplot.py`) subclasses `MultiLinePlot`, so coordinate extraction, per-series nesting, GID assignment, selector generation and legend/`z` handling are reused rather than copy-pasted. It adds:
- `stepDirection`, mapped from the drawstyle — `steps-post → hv`, `steps-pre`/`steps → vh`, `steps-mid → mid`. Omitted when the axes does not author one unambiguously, so the description never names a direction the data did not report.
- a per-point `label` carrying the ordinal level *name*, resolved from the y tick labels at **render** time (the idiomatic way to build a hypnogram is to plot numeric codes and name them afterwards with `set_yticks(codes, labels=names)`). `y` stays numeric — it is what drives sonification, braille and the min/max range.

Two details worth a reviewer's eye:
- Tick labels that merely spell out their own position (`"3"` at y = 3) are skipped, so an ordinary numeric step plot gets no labels rather than a redundant one on every point.
- Levels are read straight off the axes rather than through `LevelExtractorMixin.extract_level`, which filters against `ax.dataLim` and would silently drop the boundary levels — precisely the `Awake` and `N3` ends of a hypnogram.

The canonical `axes` contract is untouched: `_extract_axes_data` is not overridden, so keys stay a subset of `{x, y, z}` and level names live in `data`, never in `axes`.

## Testing

`uv run pytest` — **212 passed**. `ruff check` clean on every changed file. CI green across the 3.9 / 3.10 / 3.11 / 3.12 matrix.

`tests/core/plot/test_stepplot.py` asserts that `ax.step`, `ax.plot(drawstyle=...)` and `sns.lineplot(drawstyle=...)` each register **exactly one** layer typed `step`; that all three `where` values map to the right direction; that stage names land on the points; and — the regression that matters most — that a plain `ax.plot()` still classifies as `line`.

It also pins **both directions of the mixed-axes rule**, after review pointed out my original description stated only the conservative half. Classification runs once, on the first plotting call for an axes (`_maidr_plot_created`), so the rule is asymmetric: `plot` then `step` stays LINE, but `step` then `plot` stays STEP with the later line riding inside it. The generous direction is deliberate — downgrading at render time would strip the ordinal level names, and "3" instead of "REM" for a whole hypnogram is a worse accessibility outcome than a slightly generous type. Either way `stepDirection` is dropped, since `resolve_step_direction` re-reads the artists at render time and finds them disagreeing, so no convention is ever claimed that the data cannot support.

Also verified by hand against a real hypnogram figure: one series, one selector, `stepDirection: "hv"`, `[{"x": "0.0", "y": 4.0, "label": "Awake"}, ...]`, and a plain line plot still producing a single `line` layer.

## Checklist

- [x] Any dependent changes have been merged and published in downstream modules — xability/maidr#723 is merged; a published release is still pending (see Dependency status above)

## Additional Notes

`x` is emitted as a string (`"0.0"`) rather than a number. That is pre-existing `MultiLinePlot` behaviour from the categorical tick-label mapping, reproduced identically by a plain line plot on this branch's base, and is left alone here.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,7 +41,7 @@ The `maidr/patch/` modules use `wrapt` to intercept matplotlib/seaborn plot call
 
 ### Supported Plot Types
 
-Defined in `maidr/core/enum/plot_type.py`: BAR, BOX, COUNT, DODGED, HEAT, HIST, LINE, SCATTER, STACKED, SMOOTH, CANDLESTICK.
+Defined in `maidr/core/enum/plot_type.py`: BAR, BOX, COUNT, DODGED, HEAT, HIST, LINE, SCATTER, STACKED, STEP, SMOOTH, CANDLESTICK.
 
 ### Canonical `axes` Payload
 
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -411,9 +411,9 @@ and maidr reports it so the description can say what the chart actually means:
 ::: {.callout-note}
 Level names are read from the y tick labels at render time, so calling
 `set_yticks()` or `set_yticklabels()` *after* plotting works as expected. A tick
-label that merely spells out its own position (`"3"` at y = 3) carries no
-ordinal information and is not emitted, so a purely numeric step plot is
-announced by its numbers as usual.
+label that is itself a number (`"3"`, `"1,000"`, or the rescaled `"0.0"` a
+default axis prints for y = 1000000) carries no ordinal information and is not
+emitted, so a purely numeric step plot is announced by its numbers as usual.
 :::
 
 ### Heat Map
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -354,6 +354,68 @@ plt.show() #<<
 
 ```
 
+### Step Plot
+
+A step plot is the right chart when the value is *piecewise constant*: it is
+held across an interval and then jumps, rather than sliding continuously
+between samples. `ax.step()` produces one, as does any `drawstyle="steps-*"`
+passed to `ax.plot()` or `sns.lineplot()`.
+
+The classic case is a **hypnogram** — an ordinal sleep stage against time. Plot
+the stages as numeric codes so they keep driving sonification, braille and the
+min/max bounds, then name the codes with `set_yticks(..., labels=...)`. maidr
+attaches the names to each point and announces "REM" instead of "3".
+
+```{python}
+#| fig-alt: Hypnogram of sleep stage across one night
+
+import matplotlib.pyplot as plt
+
+import maidr #<<
+
+# Ordinal sleep stages, deepest first, so "up" means lighter sleep.
+stage_codes = [0, 1, 2, 3, 4]
+stage_names = ["N3", "N2", "N1", "REM", "Awake"]
+
+# One reading every half hour across a night's sleep.
+hours = [i * 0.5 for i in range(17)]
+stages = [4, 3, 2, 1, 0, 0, 1, 3, 2, 1, 0, 1, 3, 2, 1, 3, 4]
+
+fig, ax = plt.subplots(figsize=(10, 5))
+
+# where="post": the stage holds until the next reading, then jumps.
+ax.step(hours, stages, where="post") #<<
+
+# Name the ordinal levels; the underlying y values stay numeric.
+ax.set_yticks(stage_codes, labels=stage_names) #<<
+
+ax.set_title("Hypnogram\nSleep stage across one night")
+ax.set_xlabel("Time asleep (hours)")
+ax.set_ylabel("Sleep stage")
+
+# Add a number formatter for better screen reader output
+ax.xaxis.set_major_formatter("{x:.1f}")
+
+plt.show() #<<
+```
+
+The `where` argument decides which side of each sample the value is held on,
+and maidr reports it so the description can say what the chart actually means:
+
+| matplotlib | maidr `stepDirection` | Meaning |
+|------------|-----------------------|---------|
+| `where="post"` (`drawstyle="steps-post"`) | `hv` | Value holds until the next x value, then jumps |
+| `where="pre"` (`drawstyle="steps-pre"` or `"steps"`) | `vh` | Value jumps at the current x value, then holds |
+| `where="mid"` (`drawstyle="steps-mid"`) | `mid` | Value jumps midway between x values |
+
+::: {.callout-note}
+Level names are read from the y tick labels at render time, so calling
+`set_yticks()` or `set_yticklabels()` *after* plotting works as expected. A tick
+label that merely spells out its own position (`"3"` at y = 3) carries no
+ordinal information and is not emitted, so a purely numeric step plot is
+announced by its numbers as usual.
+:::
+
 ### Heat Map
 
 ```{python}
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -65,6 +65,8 @@ We currently support the following data visualization libraries in Python:, and 
 
 * Multi-line plot having one x (numerical) and one y (numerical) axis with multiple data series on the same axes.
 
+* Step plot having one x (numerical) and one y (numerical or ordinal) axis, for piecewise-constant values that are held over an interval and then jump — e.g. a hypnogram of sleep stage against time. Named y tick labels are announced as level names in place of the numeric level.
+
 * Vertical box plot having one x (categorical) and one y (numerical) axis.
 
 * Horizontal box plot having one x (numerical) and one y (categorical) axis.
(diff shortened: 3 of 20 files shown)


jooyoungseo:
Thanks — good catch on the key name, that was the right thing to block on. Answering it and the rest below.

**`label` vs `level` — confirmed `label` is correct.** The JS side defines the point shape as:

```ts
export interface StepPoint extends LinePoint {
  /** Ordinal level name announced in place of the raw numeric `y`. */
  label?: string;
}
```

and `StepTrace` reads `point.label` when substituting the name for the number in the announcement. `MaidrKey.LEVEL` is unused here because `"level"` is not part of the step contract at all — and it could not be, since the canonical `axes` test lists `level` among the forbidden top-level keys. So the reserved-for-axis-text concern is real as a naming observation, but the wire key genuinely is `label`; the enum member just serves two roles now.

You're right that nothing would have raised if I'd guessed wrong — the field would have silently gone unused. The JS repo is where that contract lives, so it is worth cross-checking there rather than in this repo alone.

**Fixed in e861354:**

- `backend.py` — you were both right that the list was inaccurate; I'd only added `step` to a literal that had already drifted. It is now built from `PlotType`, so it reads `bar, box, candlestick, count, dodged_bar, heat, hist, line, point, smooth, stacked_bar, step, violin_box, violin_kde` and cannot drift again.
- `_is_numeral_echo` — now compares with a tolerance. This is slightly more than cosmetic: a tick label is a *rounded* rendering of its position, so `"0.3"` at `0.30000000000000004` would have failed exact equality and put a redundant label on every point of an ordinary numeric step plot.
- Added `example/step/seaborn/example_sns_step.py` for parity, since the drawstyle-based classification already covers `sns.lineplot(drawstyle=...)`.
- Fixed the missing "that" in the fixture docstring.

**Left alone:** the shared-y sibling walk in `_resolve_y_levels`. It shares only the `get_siblings` call with `extract_shared_ylabel` — that one returns the first non-blank *label string*, this one needs a `{position: name}` map keyed by tick coordinate. Factoring out three lines of iteration would leave both call sites doing their own real work, so I don't think it earns the indirection.

`uv run pytest` — 204 passed; `ruff check` clean.

---

jooyoungseo:
You're right, and thanks for catching it — I traded a stale message for a technically-correct but unfriendly one. Someone who called `ax.scatter()` being told about `point` is worse than the drift I was fixing.

Fixed in 0e4a284 with the `display_name` property you suggested, so the list stays derived (can't drift) *and* readable:

```
bar, box, candlestick, count, dodged bar, heatmap, histogram, line, scatter,
smooth, stacked bar, step, violin
```

`display_name` falls back to `.value` for the members where the two already agree, so only the seven that diverge need an entry. Both violin layers map to `"violin"` and the caller's `set()` folds them, since they are two layers of one plot rather than two plots.

**`data_bearing_lines` vs the extractor's filter** — agreed, a docstring is not a constraint. Added `test_data_bearing_lines_agrees_with_the_extractor`, which draws two step lines plus an empty one and asserts the helper's count equals the number of series that actually reach the payload. It fails the moment the two predicates diverge, which is the failure you described: an axes classified as a step plot whose step line is then dropped.

**The late-added-line case in `is_step_axes`** — accepting it consciously, and your framing is the right one. Concretely: classification runs once, at the first plotting call, because that is when `_maidr_plot_created` fires and maidr emits one layer per axes. So `ax.step(...)` followed by `ax.plot(...)` keeps `type: "step"` while holding one genuinely interpolated series. Narrowing that would mean changing the one-layer-per-axes model, which is well beyond this PR. `stepDirection` does drop out (`resolve_step_direction` re-reads the artists at render time and finds them disagreeing), so the layer at least stops claiming a convention it no longer has.

**Offline rendering** — confirmed, and it errors rather than mis-rendering: the JS `TraceFactory.create` has a `default` branch that throws `Invalid trace type: step`, so a `step` layer against the currently bundled `maidr.js` is a hard failure, not a silent downgrade to `line`. Nothing in this PR gates on it because the bundle bump is a separate scripted chore, but it does mean this should land after the JS side ships and the bundle is refreshed. I'll note the same ordering on the JS PR.

`uv run pytest` — 207 passed. `ruff check` clean on every file this PR touches (the 7 findings across the repo are pre-existing `__init__.py` re-export warnings, present on `main` too).

---

jooyoungseo:
The asymmetry is real and you're right that my description stated only the conservative half of it. Went with option (a) — pinned in 4d9dae2.

**The rule, stated properly.** Classification runs once, on the first plotting call for an axes, gated by `_maidr_plot_created`. So:

- `ax.plot()` then `ax.step()` → **LINE**, no `stepDirection`. Conservative.
- `ax.step()` then `ax.plot()` → **STEP**, no `stepDirection`. Generous, and the case my description skipped.

Both directions now have a test. The second one carries the reasoning on the test itself, so the next person to look at it doesn't have to reconstruct it:

> Downgrading at render time would be worse: it would strip the ordinal level names, and losing the stage names is a bigger accessibility loss than a slightly generous type.

That's the trade I'd defend. A screen-reader user on a `step`-typed layer holding one interpolated series hears a slightly over-strong claim about interpolation; a user on a downgraded `line` layer hears "3" instead of "REM" for every point of a hypnogram. The first is a nuance, the second makes the chart unreadable. And the layer does give up `stepDirection` — `resolve_step_direction` re-reads the artists at render time and finds them disagreeing — so it never claims a convention the data cannot support.

Worth noting this is not step-specific: the one-layer-per-axes model already means an axes mixing plot types gets exactly one. Step just makes the consequence visible.

**On `label` at that position** — confirmed, and I can be definitive since the JS side is the sibling PR (xability/maidr#723). `src/type/grammar.ts` there defines:

```ts
export interface StepPoint extends LinePoint {
  /** Ordinal level name announced in place of the raw numeric `y`. */
  label?: string;
}
```

and `StepTrace.text` substitutes `point.label` for the numeric `y` in the announcement. You're right that it's a new convention for this repo — the reason it can't reuse `LevelExtractorMixin.extract_level`'s "level names in the coordinate" approach is that `y` has to stay numeric: it drives sonification, braille and the min/max range, and a string there becomes `NaN` in all three. The name has to travel *alongside* the number, not instead of it.

**`ax.get_shared_y_axes()`** — checked rather than assumed. No `DeprecationWarning` on matplotlib 3.10.9 (`pytest -W error::DeprecationWarning` on the step suite passes), and `MaidrPlot.extract_shared_ylabel` already uses the same call, so switching would leave the codebase inconsistent for no gain today. The `try/except` stays as the graceful-degradation path you noted.

**Style nit on `Dict`/`List`/`Optional`** — fair; `regplot.py` and `scatterplot.py` do lean on builtin generics. Leaving it rather than churning two new files in a feature PR, but happy to switch if you'd prefer consistency now.

`uv run pytest` — **212 passed**. `ruff check` clean on every file this PR touches.

---

PR #303, opened 2026-08-05
feat(plotly): bind a plotly step chart as a step trace, not a line

jooyoungseo:
## Now targets `main` directly

#299 **merged** (squashed as `72c1a8a`), so this is no longer stacked — `PlotType.STEP` and `MaidrKey.STEP_DIRECTION` are on `main`.

Because #299 was **squash**-merged, its branch commits are not ancestors of `main`, so GitHub's automatic retarget left this PR re-proposing all of #299's work on top of a `main` that already had it: **1985 additions / 29 files / 8 commits**. Rebased onto `main` (`git rebase --onto origin/main 81e7991`), dropping the redundant commits.

Current scope: **1218 additions / 19 deletions / 14 files / 7 commits** — plotly, tests and docs only.

Upstream ordering is satisfied: xability/maidr#723 (the step trace) and **#746 (the adapters)** are both merged.

## Description

Plotly has no step trace type. A staircase is an ordinary `scatter` trace whose `line.shape` makes plotly.js draw risers between the samples instead of interpolating across them — so the shape is the only thing separating piecewise-constant data from a line.

`PlotlyPlotFactory` never read it:

```python
if "lines" in mode and "markers" not in mode:
    return PlotlyLinePlot(trace, layout, **axis_kwargs)
```

So **every plotly step chart was exported as a line**, telling a blind user the value glided through each intermediate level. For a hypnogram that is the opposite of what the data says. This is the same defect #299 fixed for matplotlib, in the one path that doesn't go through matplotlib at all.

## Why Altair needed no code

`maidr/altair/` doesn't build a schema in Python — it ships the Vega-Lite spec to `maidr@latest/dist/vegalite.js` and lets the upstream adapter do the work. **#746 taught that adapter to read a stepping `interpolate`**, so Altair step charts work with zero changes here. Verified against the real spec rather than assumed:

```
mark: {"interpolate": "step-after", "type": "line"}
```

which is exactly the key `STEP_DIRECTION_BY_INTERPOLATE` maps to `hv`. The Altair example and docs added here record that path; they don't implement it.

## Changes Made

**`maidr/plotly/step_shape.py`** — the shape vocabulary, mirroring `src/adapters/plotly/extractor.ts` so a figure describes itself the same way in the browser and through this package:

| `line.shape` | `stepDirection` | Meaning |
|---|---|---|
| `hv` | `hv` | holds, then jumps at the next x |
| `vh` | `vh` | jumps at the current x, then holds |
| `hvh` | `mid` | jumps midway between x values |
| `vhv` | *(none)* | still a step; no direction claimed |

`vhv` is the one shape whose horizontal segments do not sit at a sample's own value, so it has no MAIDR equivalent. It binds as a step — the data *is* piecewise constant — but withholds the direction rather than guessing, which is what an optional `stepDirection` exists for.

**`maidr/plotly/step.py`** — `PlotlyStepPlot`, covering single- and multi-series in one class like `PlotlyMultiLinePlot`. One point per **data sample**, never one per stairstep vertex: the frontend derives transitions and run lengths from consecutive `y` values and rebuilds the risers itself.

**Steps group by convention, one layer each.** A MAIDR layer carries a single `stepDirection` for all of its series, so merging an `hv` trace with a `vh` one would describe one of them wrongly. Traces whose shape reports no direction group together rather than being scattered into whichever directional group came first.

## Four bugs found during review, each fixed and pinned

These were all caught by review rounds on this PR rather than after merge, and each regression test below was checked against the pre-fix code to confirm it actually fails there.

**1. Selector indices counted within the layer, not the subplot.** `nth-child` counts within the subplot's `scatterlayer`, and splitting steps out means a layer's traces are routinely not the leading ones. With a step declared before two plain lines:

```
PlotType.LINE  -> nth-child(1), nth-child(2)
PlotType.STEP  -> nth-child(1)
```

The LINE layer pointed at the step's own element and line 1 instead of lines 1 and 2, with `nth-child(1)` claimed twice. `PlotlyLinePlot` failed differently for the same reason: its unscoped `.trace.scatter path.js-line` matched the staircase too, since a step also renders as `path.js-line`.

The cause is worth stating: splitting steps into their own layers ended an invariant the line classes silently relied on — *a line layer owns every scatter trace on its subplot*. All three classes now build selectors through one shared `PlotlyPlot._scatter_line_selector(position)`, fed from a `position_of` map built once per subplot. A lone line is constructed in `_extract_plots` rather than by the factory, which cannot know a trace's position.

**2. The classification rule was duplicated and had already drifted.** `_extract_plots` and `PlotlyPlotFactory` each spelled out "is this a lines-only scatter trace", and they disagreed on the default for a missing `type`. Consolidated into `is_connected_line_trace()`.

**3. Consolidating (2) introduced a `KeyError`.** The new predicate defaulted a missing `type` to `"scatter"` while the `scatter_family` filter feeding `position_of` still used no default — so a trace could be classified as a line but be absent from the position map, and the lookup crashed. Fixed by extracting `is_scatter_family_trace()` and building `is_connected_line_trace()` on top of it, so the classifier and the selector index are the same test *by construction* rather than by agreement.

**4. A staircase that never authored a `mode` was dropped to scatter.** `to_dict()` omits `mode` when the author never sets one, and the classifier read absent-as-markers. So `add_scatter(..., line_shape="hv")` produced a **scatter** layer at every point count, while plotly still drew a staircase — the chart looked stepped and announced as loose points. Every example here passes `mode="lines"`, so it was invisible to anyone following the docs and silent for anyone who wasn't. Now a declared stepping shape is rescued; an explicit mode still wins, and a mode-less trace with no `line.shape` keeps its prior classification:

```
mode-less + shape hv     SCATTER -> STEP   (the defect)
mode-less, no shape      SCATTER           unchanged
mode=markers + shape     SCATTER           unchanged
mode=lines+markers       SCATTER           unchanged
mode=lines + shape       STEP              unchanged
mode=lines, no shape     LINE              unchanged
```

## Testing

`uv run pytest` — **273 passed**. `ruff check` clean on every file this PR touches. CI's own matrix runs green on Python **3.9 / 3.10 / 3.11 / 3.12**, plus lint, commit-lint, workflow lint and CodeQL, so these numbers are not local-only.

`tests/plotly/test_plotly_step.py` covers: all four shapes binding as step; `linear`/`spline`/absent staying a line; `lines+markers` staying a scatter; a non-dict `line` not raising; the direction mapping including `vhv` withholding one; grouping by convention with order preserved; subplot-relative selector indexing; the mode-less rescue and its three non-rescued neighbours; the classification/position agreement that prevents the `KeyError`; and the line-side regressions — lines after a step, a lone line beside a step, and the no-step case pinned as unchanged.

One note on why that matters here: the original `test_a_step_beside_lines_leaves_the_multiline_layer_intact` asserted only on layer *data*, which is exactly why a wrong-element selector bug sailed through CI. Every selector fix since is pinned by an assertion on the selector itself, verified to fail against the pre-fix code.

Also verified by execution: the example emits `type: step`, `stepDirection: "hv"`, 17 points, canonical `{x, y}` axes, one selector per series; the Altair example builds a spec carrying `interpolate: "step-after"`; and a bar + line + box figure emits layers in the same order as on `main`, since the step block is a no-op when no staircases are present.

## Known gaps, deliberately left

- A mode-less **plain line** (no `mode`, no `line.shape`) is still classified as `SCATTER`. This is pre-existing and affects charts beyond steps; fixing it would reclassify existing users' scatter traces, which deserves its own change. Pinned by `test_a_modeless_trace_without_a_shape_is_untouched` so the behaviour is intentional rather than incidental.
- `scattergl` is in the scatter family, but it draws through WebGL, so `path.js-line` selectors match nothing and highlighting is inert. Pre-existing for lines; documented at `_CONNECTED_TRACE_TYPES`.
- `PlotlyStepPlot`/`PlotlyMultiLinePlot` default `scatter_positions` to leading order when constructed standalone, where `PlotlyLinePlot` falls back to an unscoped selector. Both are documented and neither is reachable from `PlotlyMaidr`.
- No test covers a step beside bar/box traces, or steps across a `row`/`col` subplot grid.

## Examples and docs

- `example/step/plotly/example_plotly_step.py` and `example/step/altair/example_altair_step.py` — hypnograms, alongside the matplotlib and seaborn ones from #299.
- `docs/examples-plotly.qmd` and `docs/examples-altair.qmd` — Step Plot sections with the full shape/`interpolate` mapping tables and the `vhv` caveat.
- `docs/index.qmd` — the step bullet now names all three detection routes.
- `CLAUDE.md` — records that `maidr/plotly/` needs per-type handling in Python while `maidr/altair/` delegates upstream, so the next person doesn't look for Altair code that isn't there.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -43,6 +43,10 @@ The `maidr/patch/` modules use `wrapt` to intercept matplotlib/seaborn plot call
 
 Defined in `maidr/core/enum/plot_type.py`: BAR, BOX, COUNT, DODGED, HEAT, HIST, LINE, SCATTER, STACKED, STEP, SMOOTH, CANDLESTICK.
 
+Note that `maidr/plotly/` builds its MAIDR schema in Python and therefore needs its own
+handling per plot type, whereas `maidr/altair/` delegates entirely to the upstream
+Vega-Lite JS adapter — an Altair-only plot type is implemented there, not here.
+
 ### Canonical `axes` Payload
 
 Every emitted schema's `axes` object follows the canonical per-axis form:
--- a/docs/examples-altair.qmd
+++ b/docs/examples-altair.qmd
@@ -466,6 +466,69 @@ multiline_plot = (
 maidr.show(multiline_plot)  #<<
 ```
 
+### Step Plot
+
+Altair expresses a staircase through the mark's `interpolate` property, which
+Vega-Lite hands to the matching d3 curve. maidr renders Altair charts through the
+upstream Vega-Lite adapter, so a stepping `interpolate` is recognised there and
+the chart is bound as a step layer rather than a line.
+
+A **hypnogram** is the canonical case: the sleep stage is held for a stretch of
+the night and then jumps.
+
+```{python}
+#| warning: false
+#| fig-alt: Hypnogram showing sleep stage as a step chart across one night
+
+import altair as alt
+import pandas as pd
+
+import maidr  #<<
+
+# Ordinal sleep stages, deepest first, so that "up" means lighter sleep.
+STAGE_NAMES = ["N3", "N2", "N1", "REM", "Awake"]
+
+hours = [i * 0.5 for i in range(17)]
+stages = [4, 3, 2, 1, 0, 0, 1, 3, 2, 1, 0, 1, 3, 2, 1, 3, 4]
+
+df = pd.DataFrame({"hours": hours, "stage": stages})
+
+chart = (
+    alt.Chart(df)
+    .mark_line(interpolate="step-after")  #<<
+    .encode(
+        x=alt.X("hours:Q", title="Time asleep (hours)"),
+        y=alt.Y(
+            "stage:Q",
+            title="Sleep stage",
+            axis=alt.Axis(
+                values=list(range(len(STAGE_NAMES))),
+                labelExpr=(
+                    "datum.value == 0 ? 'N3' : datum.value == 1 ? 'N2' : "
+                    "datum.value == 2 ? 'N1' : datum.value == 3 ? 'REM' : 'Awake'"
+                ),
+            ),
+        ),
+    )
+    .properties(title="Hypnogram: sleep stage across one night")
+)
+
+maidr.show(chart)
+```
+
+`interpolate` decides which side of each sample the value is held on:
+
+| Altair `interpolate` | maidr `stepDirection` | Meaning |
+|----------------------|-----------------------|---------|
+| `"step-after"` | `hv` | Value holds until the next x value, then jumps |
+| `"step-before"` | `vh` | Value jumps at the current x value, then holds |
+| `"step"` | `mid` | Value jumps midway between x values |
+
+::: {.callout-note}
+`mark_area(interpolate="step-after")` binds as a step too — the fill is lost in
+the conversion, as it is for a plain area mark.
+:::
+
 ### Scatter Plot
 
 ```{python}
--- a/docs/examples-plotly.qmd
+++ b/docs/examples-plotly.qmd
@@ -399,6 +399,76 @@ fig = go.Figure(
 maidr.show(fig)  #<<
 ```
 
+### Step Plot
+
+Plotly has no step trace type. A staircase is an ordinary `Scatter` trace whose
+`line_shape` tells plotly.js to draw risers between the samples rather than
+interpolating across them, so `line_shape` is the only thing that marks the data
+as *piecewise constant*. maidr reads it and exports a step layer.
+
+A **hypnogram** is the canonical case: the sleep stage is held for a stretch of
+the night and then jumps. Drawn as a line it would claim the sleeper glided
+through every intermediate stage.
+
+```{python}
+#| warning: false
+#| fig-alt: Hypnogram showing sleep stage as a step chart across one night
+
+import plotly.graph_objects as go
+
+import maidr  #<<
+
+# Ordinal sleep stages, deepest first, so that "up" means lighter sleep.
+STAGE_NAMES = ["N3", "N2", "N1", "REM", "Awake"]
+
+hours = [i * 0.5 for i in range(17)]
+stages = [4, 3, 2, 1, 0, 0, 1, 3, 2, 1, 0, 1, 3, 2, 1, 3, 4]
+
+fig = go.Figure(
+    go.Scatter(
+        x=hours,
+        y=stages,
+        mode="lines",
+        line_shape="hv",  #<<
+        name="Night 1",
+    )
+)
+
+fig.update_layout(
+    title="Hypnogram<br>Sleep stage across one night",
+    xaxis_title="Time asleep (hours)",
+    yaxis=dict(
+        title="Sleep stage",
+        tickmode="array",
+        tickvals=list(range(len(STAGE_NAMES))),
+        ticktext=STAGE_NAMES,
+    ),
+)
+
+maidr.show(fig)
+```
+
+`line_shape` decides which side of each sample the value is held on, and maidr
+reports it so the description can say what the chart actually means:
+
+| Plotly `line_shape` | maidr `stepDirection` | Meaning |
+|---------------------|-----------------------|---------|
+| `"hv"` | `hv` | Value holds until the next x value, then jumps |
+| `"vh"` | `vh` | Value jumps at the current x value, then holds |
+| `"hvh"` | `mid` | Value jumps midway between x values |
+| `"vhv"` | *(none)* | Still a step, but no direction is claimed — see below |
+
+::: {.callout-note}
+`"vhv"` is the one shape whose horizontal segments do not sit at a sample's own
+value, so it has no maidr equivalent. The trace still binds as a step, because
+the data is piecewise constant, but no `stepDirection` is emitted rather than
+one being guessed.
+
+A maidr layer carries a single `stepDirection` for all of its series, so traces
+drawn with different shapes are exported as **separate layers**. Merging an
+`"hv"` trace with a `"vh"` one would describe one of them wrongly.
+:::
+
 ### Scatter Plot
 
 ```{python}
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -65,7 +65,7 @@ We currently support the following data visualization libraries in Python:, and 
 
 * Multi-line plot having one x (numerical) and one y (numerical) axis with multiple data series on the same axes.
 
-* Step plot having one x (numerical) and one y (numerical or ordinal) axis, for piecewise-constant values that are held over an interval and then jump — e.g. a hypnogram of sleep stage against time. Named y tick labels are announced as level names in place of the numeric level.
+* Step plot having one x (numerical) and one y (numerical or ordinal) axis, for piecewise-constant values that are held over an interval and then jump — e.g. a hypnogram of sleep stage against time. Named y tick labels are announced as level names in place of the numeric level. Recognised from matplotlib's `drawstyle="steps-*"` (including `ax.step()`), Plotly's `line_shape`, and Altair's `interpolate`.
 
 * Vertical box plot having one x (categorical) and one y (numerical) axis.
 
--- a/example/step/altair/example_altair_step.py
+++ b/example/step/altair/example_altair_step.py
@@ -1,0 +1,59 @@
+"""Hypnogram in Altair: a sleep stage held over an interval, then jumping.
+
+Altair expresses a staircase through the mark's ``interpolate`` property, which
+Vega-Lite hands to the matching d3 curve. maidr renders Altair charts through
+the upstream Vega-Lite adapter, so the step handling lives there and this file
+needs nothing maidr-specific beyond the import.
+
+``interpolate`` maps onto MAIDR's step conventions as:
+
+    "step-after"  -> hold the value, then jump at the next reading  (used here)
+    "step-before" -> jump at the reading, then hold
+    "step"        -> jump midway between readings
+"""
+
+import altair as alt
+import pandas as pd
+
+import maidr
+
+# Ordinal sleep stages, deepest first, so that "up" means lighter sleep.
+STAGE_NAMES = ["N3", "N2", "N1", "REM", "Awake"]
+
+# One reading every half hour across a night's sleep.
+hours = [i * 0.5 for i in range(17)]
+stages = [4, 3, 2, 1, 0, 0, 1, 3, 2, 1, 0, 1, 3, 2, 1, 3, 4]
+
+df = pd.DataFrame(
+    {
+        "hours": hours,
+        "stage": stages,
+        "stage_name": [STAGE_NAMES[s] for s in stages],
+    }
+)
+
+chart = (
+    alt.Chart(df)
+    # "step-after" is the hypnogram reading: the stage holds until the next
+    # reading, then jumps.
+    .mark_line(interpolate="step-after")
+    .encode(
+        x=alt.X("hours:Q", title="Time asleep (hours)"),
+        # The y values stay numeric — they drive sonification, braille and the
+        # min/max range — while the axis labels name the ordinal levels.
+        y=alt.Y(
+            "stage:Q",
+            title="Sleep stage",
+            axis=alt.Axis(
+                values=list(range(len(STAGE_NAMES))),
+                labelExpr=(
+                    "datum.value == 0 ? 'N3' : datum.value == 1 ? 'N2' : "
+                    "datum.value == 2 ? 'N1' : datum.value == 3 ? 'REM' : 'Awake'"
+                ),
+            ),
+        ),
+    )
+    .properties(title="Hypnogram: sleep stage across one night")
+)
+
+maidr.show(chart)
--- a/example/step/altair/example_altair_step.py
+++ b/example/step/altair/example_altair_step.py
@@ -28,7 +28,6 @@ df = pd.DataFrame(
     {
         "hours": hours,
         "stage": stages,
-        "stage_name": [STAGE_NAMES[s] for s in stages],
     }
 )
 
--- a/example/step/plotly/example_plotly_step.py
+++ b/example/step/plotly/example_plotly_step.py
@@ -1,0 +1,53 @@
+"""Hypnogram in Plotly: a sleep stage held over an interval, then jumping.
+
+Plotly has no step trace type. A staircase is an ordinary ``Scatter`` trace
+whose ``line_shape`` tells plotly.js to draw risers between the samples rather
+than interpolating across them, so ``line_shape`` is the only thing that marks
+the data as piecewise constant. maidr reads it and exports a step layer, which
+announces the stage names and lets a user jump transition to transition.
+
+``line_shape`` maps onto MAIDR's step conventions as:
+
+    "hv"  -> hold the value, then jump at the next reading   (used here)
+    "vh"  -> jump at the reading, then hold
+    "hvh" -> jump midway between readings
+    "vhv" -> still a step, but no MAIDR equivalent, so no direction is claimed
+"""
+
+import plotly.graph_objects as go
+
+import maidr
+
+# Ordinal sleep stages, deepest first, so that "up" means lighter sleep.
+STAGE_NAMES = ["N3", "N2", "N1", "REM", "Awake"]
+
+# One reading every half hour across a night's sleep.
+hours = [i * 0.5 for i in range(17)]
+stages = [4, 3, 2, 1, 0, 0, 1, 3, 2, 1, 0, 1, 3, 2, 1, 3, 4]
+
+fig = go.Figure(
+    go.Scatter(
+        x=hours,
+        y=stages,
+        mode="lines",
+        # "hv" is the hypnogram reading: the stage holds until the next
+        # reading, then jumps. maidr exports this as stepDirection "hv".
+        line_shape="hv",
+        name="Night 1",
+    )
+)
+
+fig.update_layout(
+    title="Hypnogram<br>Sleep stage across one night",
+    xaxis_title="Time asleep (hours)",
+    # Name the ordinal levels. The underlying y values stay numeric, which is
+    # what drives sonification, braille and the min/max range.
+    yaxis=dict(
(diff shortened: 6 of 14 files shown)


jooyoungseo:
## `claude-review` is failing on auth, not on this diff

Every functional gate is green — **Python 3.9 / 3.10 / 3.11 / 3.12 tests, `lint`, `Lint workflows and shell scripts`, `commit-lint`**. The only red check is `claude-review`, and it never reaches the code:

```
Requesting OIDC token...
OIDC token successfully obtained
Exchanging OIDC token for app token...
App token exchange failed: 401 Unauthorized - Invalid OIDC token
Retrying in 5 seconds...   → Attempt 2 → Attempt 3
Operation failed after 3 attempts
##[error]Action failed with error: Invalid OIDC token
```

It dies ~16s in, during the app-token exchange, before any review work starts. `CLAUDE_CODE_OAUTH_TOKEN` is present in the job env; `ANTHROPIC_API_KEY` is empty, so the OAuth path is the one being used and it is the OIDC exchange that 401s.

**I re-ran the failed job to rule out a transient outage — it reproduced identically**, same three attempts, same error. So it is systematic rather than flaky, and it is not something a change to this branch can fix.

### One hypothesis, untested

This is the first of the step-plot PRs with a **non-`main` base** (it stacks on #299). `claude-review` passed on #299 and on the upstream maidr PRs, all of which target `main`. An OIDC subject encodes the workflow ref, so a `pull_request` event against a non-default base could plausibly produce a token the app's federation config rejects.

That is a correlation, not a diagnosis — I have not tested it. If it would help, I can retarget this PR's base to `main` temporarily to confirm or kill the theory; the cost is that the diff then shows #299's commits until #299 merges, which is exactly what stacking was avoiding.

### What I did not do

I did not touch `.github/workflows/`. The fix belongs to repo CI auth config, I would be guessing at the cause, and bundling it into a feature PR would be scope creep. Upstream maidr has a related guard (xability/maidr#743, "skip Claude review where secrets are absent") — but that is not quite this case, since the token here is present and the *exchange* is what fails.

Happy to open a separate issue for it if you'd like it tracked.

---

jooyoungseo:
Both comments make the same point and it's correct — fixed in `9bf5006`.

I checked the claim rather than taking it on trust, and it holds exactly: across `maidr/plotly/`, the only `from typing import` lines are `Any`, `Any, Literal, cast` — things with no builtin equivalent — plus my two new files. Every other module annotates with `list[...]` / `dict[...]`, and several already use PEP 604 (`plotly_maidr.py` has `str | None`, `bool | Literal["auto"]`). So the convention here is unanimous and my files were the sole exception.

**Safe on Python 3.9**, which the CI matrix covers: `from __future__ import annotations` leaves annotations unevaluated, nothing in the package calls `get_type_hints`, and `plotly_maidr.py` already ships `str | None` under the same matrix. I confirmed `get_type_hints` resolves both modules without error rather than assuming it.

---

Worth noting for anyone comparing: I **declined** this same nit on #299, and that wasn't inconsistency — the two packages differ.

| package | modules importing `List`/`Dict`/`Optional` | verdict |
|---|---|---|
| `maidr/plotly/` | 0 (before this PR) | unanimous convention → align |
| `maidr/core/plot/` | 7, alongside builtin generics | already mixed → no standard to violate |

`maidr/core/plot/` has `from typing import Dict, List, Optional`, `List, Union`, `Any, Dict, List, Tuple` and more in existing files, so converting #299's two files would have made them consistent with *some* neighbours and inconsistent with others. Here there was one answer, so it was worth taking.

`uv run pytest` — **235 passed**. `ruff check` clean on both files.

On the red check: `claude-review` is still failing on `401 Unauthorized - Invalid OIDC token` during the app-token exchange, which is the auth problem described [above](https://github.com/xability/py-maidr/pull/303#issuecomment-5196582406) and is unrelated to the diff. Every functional gate is green.

---

jooyoungseo:
The selector bug is real and I've fixed it in `75d4996`. Thanks — this was a regression I introduced, and my own test was too shallow to catch it.

Reproduced it before changing anything, with a step declared ahead of two plain lines on one subplot:

```
PlotType.LINE  -> nth-child(1), nth-child(2)
PlotType.STEP  -> nth-child(1)
```

So the LINE layer pointed at the step's own element and line 1, rather than lines 1 and 2 — every line shifted onto its predecessor, and `nth-child(1)` claimed by two layers at once. The single-line case failed differently, exactly as you described: `PlotlyLinePlot`'s unscoped `.trace.scatter path.js-line` matched both the step and the line, since a staircase renders as `path.js-line` too.

Your framing of the cause is the right one. Splitting steps out ended the invariant both classes were relying on — *this layer owns every scatter trace on this subplot* — and I only carried the subplot-relative reasoning into the new class, not into the two whose assumption I'd just invalidated.

**The fix.** All three now build the selector through one shared `PlotlyPlot._scatter_line_selector(position)`, so the rule lives in a single place rather than being re-derived per class. `PlotlyMultiLinePlot` takes `scatter_positions` and `PlotlyLinePlot` takes `scatter_position`, both supplied by `_extract_plots` from the same `position_of` map the step branch already used (now hoisted so all three share it).

A lone line is constructed in `_extract_plots` rather than left to the factory, since the factory has no way to know the trace's position within its subplot. `PlotlyLinePlot` keeps the unscoped form as a fallback for direct construction, documented as such.

After the fix:

```
step + 2 lines:   STEP -> 1        LINE -> 2, 3
step + 1 line:    STEP -> 1        LINE -> 2
lines only:       LINE -> 1, 2     (unchanged)
```

**On the missing assertion.** You're right that `test_a_step_beside_lines_leaves_the_multiline_layer_intact` only checked data, which is why CI stayed green through a wrong-element bug. Three regression tests added: lines after a step, a lone line beside a step, and the no-step case pinned as unchanged. I confirmed the first two actually fail against the previous code rather than assuming they would — a test that passes either way would have been worse than none here.

Suite is 238 passing, `ruff` clean.

**On the two minor points.** The factory's `is_step_trace` branch is indeed unreachable via `PlotlyMaidr` now that `_extract_plots` consumes every step trace — it's exercised only by direct `PlotlyPlotFactory.create` calls in the unit tests. I've left it as the defensive/standalone path rather than deleting it, since the factory is a public dispatch point and the same is now true of the `PlotlyLinePlot` branch for identical reasons. I did not hoist the mid-loop `is_step_trace` import; it matches the deferred-import style used throughout that function, and moving one while leaving the rest would be inconsistent. Happy to do either if you'd prefer.

---

jooyoungseo:
All three verified and addressed in `ceb9a26`.

**1. The factory branch is unreachable — and more so than you noted.** Confirmed by reading `_extract_plots`: `line_traces` (both the `> 1` and the `== 1` branch) and `step_traces` are all added to `merged`, so the fallback loop never sees a lines-mode trace. Worth adding that the `PlotlyLinePlot` branch only *became* dead in this PR — the `elif len(line_traces) == 1` I added to give a lone line a position-scoped selector is what took it out of the fallback path. So it isn't just the new step branch that's unreachable; the whole lines-mode branch is.

I've kept it as the standalone-construction path rather than deleting it, for the reason you identified as the alternative, and it now says so explicitly — including the part that actually matters, that its classification rule has a second copy in `_extract_plots` and nothing forces the two to agree.

**2. `stage_name` was dead.** Confirmed: built in the DataFrame, referenced by no encoding channel. Dropped rather than wired into a tooltip — the y axis already names the levels through `labelExpr`, so the column was redundant as well as unused, and the docs version never had it. Verified afterwards that the emitted spec still carries `mark: {"type": "line", "interpolate": "step-after"}` and that the encoding only ever used `x` and `y`, which is what makes the column provably dead rather than merely unreferenced-looking.

**3. Docstring was stale.** Also correct, and it was my change that made it so. It now covers the step/line split and why it exists, the per-convention step grouping, the lone-line case, and the reason every scatter-family trace is constructed there rather than by the factory — only `_extract_plots` knows a trace's position within its subplot, which the selector depends on. That last point is the thing a future reader most needs, since it's what makes the factory route structurally unable to produce a correct selector.

Suite 267 passed (up from 238 — the branch is now rebased onto `main`, which brought #301's tests), `ruff` clean on every touched file. The seven `F401`s ruff reports repo-wide are pre-existing barrel-`__init__.py` re-exports; I confirmed they reproduce identically with my changes stashed.

Thanks — that's two rounds now where the review caught something real.

---

jooyoungseo:
Points 1 and 2 done in `39cda37`. Point 3 I'm deliberately leaving, with reasoning below.

**1. Import placement — you're right and my earlier answer was wrong.** A previous review raised this and I brushed it off as matching the file's deferred-import style. Your framing showed why that was wrong: the convention in `_extract_plots` isn't "import lazily", it's **import inside the branch that uses it** — which every other deferred import there does. Mine was unconditional in the per-subplot loop body, so it followed neither rule. And as you note, `step_shape.py` imports nothing from `maidr`, so there's no circular risk. It's a normal top-level import in both files now.

**2. Extracted the shared predicate.** `is_connected_line_trace()` now lives in `step_shape.py` as the single definition of "plotly is joining this trace's samples up", used by both `_extract_plots` and the factory. You were right that documenting an unenforced invariant isn't the same as removing it — I'd only left a comment asking the two to be kept in sync, which is the kind of thing that holds until it doesn't.

Worth flagging what consolidating turned up: **the two copies didn't actually agree.** `_extract_plots` tested `t.get("type")` with no default; the factory used `.get("type", "scatter")`. So a hand-built trace dict omitting `"type"` was a line to one and not-a-line to the other. Merging them forces a choice, and the shared predicate takes plotly's own documented default of `scatter`, matching the factory. Only reachable with hand-built dicts — `fig.to_dict()` always emits `type` — and a bare `{}` is still excluded because `mode` defaults to `markers`. Small, but it's exactly the drift you predicted, already present before anyone noticed.

**3. `scattergl` — confirmed, pre-existing, and left alone on purpose.** You're right that WebGL traces render to canvas, so the `path.js-line` selectors match nothing and highlighting is silently inert. But the fix isn't step-shaped: `("scatter", "scattergl")` has always been the line/multiline filter, so excluding `scattergl` would change behaviour for existing **line** charts under a step PR — a real regression risk for anyone currently getting a (broken but harmless) line layer. Changing it needs its own change with its own reasoning about what should happen instead: no layer at all, or a layer with different selectors.

What I have done is write it down at the definition site, on `_CONNECTED_TRACE_TYPES`, so the next person doesn't have to rediscover it from a silent non-highlight. Happy to file the follow-up issue for it and for the multi-subplot step+line test gap you flagged — say the word and I'll open them.

Suite 267 passed, `ruff` clean on every touched file.

---

jooyoungseo:
Confirmed and fixed in `339ea7e`. This one is worth being blunt about: **I introduced it in the commit immediately before, whose entire purpose was to prevent this exact failure mode.**

Reproduced before touching anything:

```python
class F:
    def to_dict(self):
        return {"layout": {}, "data": [{"mode": "lines", "x": [0,1,2], "y": [1,2,3]}]}

PlotlyMaidr(F())   # KeyError: 140708306605120
```

And I checked it was mine rather than pre-existing: at `ceb9a26` both filters used `t.get("type")` with no default, so they agreed and there was no gap. `39cda37` defaulted one side to `"scatter"` and left `scatter_family` alone — I closed a two-copy divergence between `_extract_plots` and the factory, and opened a new one *one filter down in the same function*. Consolidating a rule while leaving a second consumer of that rule spelled out by hand is not consolidation; it just moves the seam.

**Fix.** Extracted `is_scatter_family_trace()`, and `is_connected_line_trace()` is now built on top of it rather than repeating the type test. The classifier and the selector index are now the same test *by construction* instead of by agreement — which is what your suggestion of a shared "is this a scatter-family trace" predicate was pointing at, and is stronger than just aligning the two defaults.

**Two regression tests added**, and I verified both fail against the divergent version rather than assuming they would:

```
FAILED ...::test_a_typeless_line_trace_still_resolves_its_position
FAILED ...::test_a_typeless_step_trace_still_resolves_its_position
2 failed, 26 passed
```

You're right that it's narrow — `Figure.to_dict()` always emits `type`, so it needs a hand-built dict. But that is exactly the path these classifiers are reached on directly, and a `KeyError` is a hard crash rather than a degraded result, so it seemed worth pinning rather than reasoning away.

Suite 269 passed, `ruff` clean.

That's three rounds where review caught something real, and this round caught a regression in the fix from the previous one. Thanks — the `scattergl` note and the multi-subplot test gap from the last round are still open as follow-ups if you'd like them filed.

---

jooyoungseo:
Point 2 turned out to be a real defect in this feature, not a theoretical one — fixed in `85336ab`. Points 1 and 3 I'm leaving, with reasons.

**2. The `mode` default — worse than "in principle".** You said a step chart without explicit `mode` "could in principle be misclassified". I tested it and it wasn't conditional at all:

```
n=5:  mode in to_dict()? False  ->  PlotType.SCATTER
n=25: mode in to_dict()? False  ->  PlotType.SCATTER
```

`to_dict()` omits `mode` entirely when the author never sets one, and the classifier read absent-as-markers. So `add_scatter(..., line_shape="hv")` produced a **scatter** layer at every point count — while plotly still drew a staircase, since both its defaults (`lines+markers` under 20 points, `lines` at or above) draw lines. The chart looked stepped and announced as loose points, losing exactly the piecewise-constant reading this PR exists to convey.

What made it worth fixing rather than documenting: **every example in this PR passes `mode="lines"`**, so anyone following them never hits it, and anyone who doesn't gets silent degradation. That's the worst failure shape for an accessibility feature.

The fix is confined to one cell — only a *declared stepping shape* is rescued, and an explicit mode always wins:

```
mode-less + shape hv     SCATTER -> STEP   (the defect)
mode-less, no shape      SCATTER           unchanged
mode=markers + shape     SCATTER           unchanged
mode=lines+markers       SCATTER           unchanged
mode=lines + shape       STEP              unchanged
mode=lines, no shape     LINE              unchanged
```

So plain mode-less scatters keep the classification they had before steps existed — I deliberately didn't "fix" the broader line case you noted as pre-existing, since that would reclassify existing users' scatter traces. Four tests added; the one pinning the fix was verified to fail against the markers-default version.

**1. Fallback asymmetry — leaving it, deliberately.** You're right that "unscoped" and "assume-leading-position" are different failure modes, and the latter is silently wrong rather than merely unsafe. Both are documented in their docstrings and neither is reachable from `PlotlyMaidr`. I'm leaving it because making `scatter_positions` required is the only fix that actually removes the footgun, and that means changing constructor signatures and every standalone call site — real churn, late in a PR that has already had one fix introduce a `KeyError` into the fix before it. This round is the second time restraint would have been the better call, so I'd rather flag it than churn it.

**3. Test gap — agreed, and it's the one I'd file.** Step alongside bar/box on a subplot, and steps across a `row`/`col` grid. I expect both work, since `position_of` and the merge logic are per-subplot-group, but expectation isn't coverage.

Happy to open follow-up issues for **1**, **3**, and the `scattergl` selector caveat from the previous round — just say the word.

Suite 273 passed, `ruff` clean.

---

jooyoungseo:
No code changes this round — the one open question resolves to "no risk", and I'd rather say that than churn.

**2. Emission order — verified, not just reasoned about.** You asked for a sanity check that nothing downstream assumed line-then-box adjacency. I ran a bar + line + box figure (no steps) against both this branch and `main`:

```
this branch:  ['PlotType.DODGED', 'PlotType.LINE', 'PlotType.BOX']
main:         ['PlotType.DODGED', 'PlotType.LINE', 'PlotType.BOX']
```

Identical. The reason it can't differ: `if step_traces:` is a no-op when the list is empty, so a figure with no staircases takes exactly the path it took before. Only figures that contain steps see the new block, and those are new functionality with no prior order to preserve. So there's nothing for a frontend tab order or a snapshot to have depended on.

Worth noting my first attempt at that comparison was wrong — I ran `git stash` on an already-clean tree, so it stashed nothing and I measured this branch twice. Caught it because the "before" and "after" were suspiciously identical for the wrong reason. The numbers above are from actually checking out `main`'s `maidr/plotly/`, and I've since confirmed the tree is byte-identical to the pushed branch again with the suite green at 273.

**1. Mode-less plain line — agreed, and deliberately left.** That's the scoping call I made when fixing the mode-less *staircase*: rescuing only a declared stepping shape, so plain scatter traces keep the classification they've always had. Fixing the general case means reclassifying existing users' mode-less scatters as lines, which is a real behaviour change that deserves its own PR and its own argument — not a rider on a step PR. It's pinned by `test_a_modeless_trace_without_a_shape_is_untouched` so the current behaviour is at least intentional and visible rather than incidental.

**3. Boilerplate — leaving it.** You're right it's five near-identical blocks now, and right that it matches the file's existing style. I'm not extracting a helper at this point: this PR has already had one review-driven fix introduce a `KeyError` into the fix before it, and a pure-restructuring change to the branch that assigns every layer's `row_index`/`col_index` is exactly where a silent off-by-one would hide. Better as its own change with its own tests.

That's three follow-ups now worth filing — the mode-less plain line, the `scattergl` selector caveat, and the multi-subplot/step-beside-bar-box test gap — plus the fallback asymmetry from two rounds ago. Happy to open them; say the word and I will.

---

jooyoungseo:
Filed the four follow-ups raised across these review rounds, so nothing depends on this thread staying readable:

- **#308** — `fix(plotly)`: a line chart with no explicit `mode` exports as `SCATTER`. Pre-existing; #303 fixed only the staircase case deliberately. Needs a decision on what a mode-less scatter should become, since all three options reclassify something.
- **#309** — `fix(plotly)`: `scattergl` highlighting is silently inert, because WebGL renders to canvas and the `path.js-line` selectors match nothing. Pre-existing for lines. Worth checking with the JS side whether a canvas highlight is feasible before picking a fix.
- **#310** — `test(plotly)`: no coverage for a step beside `bar`/`box` traces, or steps across a `row`/`col` subplot grid. Both expected to pass; the point is that a wrong `nth-child` doesn't raise, it highlights the wrong element — which is how two bugs in this PR stayed invisible until a selector assertion existed.
- **#311** — `refactor(plotly)`: the position fallbacks are inconsistent, and the leading-order one is silently wrong rather than merely unsafe. Suggests removing both and making positions required, since `_extract_plots` already always supplies them.

Each records why it was left out of this PR rather than just what it is, so whoever picks one up doesn't have to reconstruct the reasoning from here.

---

Issue #311, opened 2026-08-07
refactor(plotly): make the selector-position fallbacks consistent, or require positions


jooyoungseo:
# Bug Report

## Description

The three Plotly line-family classes disagree on what to do when nobody tells them where their traces sit in the subplot, and the two behaviours fail in different ways.

`PlotlyLinePlot._get_selector()` falls back to an **unscoped** selector:

```python
if self._scatter_position is None:
    return [f"{self._subplot_css_prefix()}.trace.scatter path.js-line"]
```

`PlotlyStepPlot` and `PlotlyMultiLinePlot` instead default `scatter_positions` to **leading order**:

```python
self._scatter_positions = (
    list(range(len(traces))) if scatter_positions is None else scatter_positions
)
```

## Actual Behavior

The unscoped fallback over-matches — it can select more elements than it should, which is at least visible. The leading-order fallback is **silently wrong**: constructed standalone for traces that are not the subplot's first scatter children, it emits `nth-child(1), nth-child(2), …` pointing at whichever elements happen to occupy those positions, with no error and nothing obviously amiss in the output.

## Expected Behavior

One consistent behaviour when position is unknown, and ideally no way to construct these classes without it.

## Additional Information

**Not currently reachable.** `PlotlyMaidr._extract_plots` always supplies positions to all three classes — it is the only thing that knows them — so no production path hits either fallback today. Both are documented in their docstrings. This is a latent footgun rather than a live bug, which is why it was left out of #303.

**The cleanest fix is probably to remove the fallbacks entirely** and make `scatter_position` / `scatter_positions` required. `_extract_plots` already passes them unconditionally, so the only affected callers are tests that construct these classes directly. That converts a silently-wrong default into a `TypeError` at the call site, which is the right trade for something whose failure mode is "highlights the wrong element".

It was deliberately not done inside #303: changing constructor signatures on the three classes that assign every layer's selectors, late in a PR that had already had one review-driven fix introduce a `KeyError` into the fix before it, is exactly where a silent off-by-one would hide. It wants its own change and its own test run.

Related: #303.

Issue #309, opened 2026-08-07
fix(plotly): highlighting is silently inert for `scattergl` traces


jooyoungseo:
# Bug Report

## Description

`scattergl` traces are classified as lines/steps and given `path.js-line` CSS selectors, but `scattergl` renders through WebGL onto a `<canvas>` rather than as SVG. Those selectors match nothing, so visual highlighting does nothing — with no error and no warning.

The trace-type filter lives in `maidr/plotly/step_shape.py`:

```python
_CONNECTED_TRACE_TYPES = ("scatter", "scattergl")
```

and the selectors built for those traces are SVG paths, e.g. from `PlotlyPlot._scatter_line_selector`:

```
.subplot.xy .scatterlayer > .trace.scatter:nth-child(1) path.js-line
```

## Actual Behavior

A `scattergl` line or step chart produces a MAIDR layer with data, braille and sonification that all work, but whose highlight selectors resolve to zero elements. Navigating the chart gives correct audio and text with no visible highlight.

## Expected Behavior

Either the highlight resolves to the rendered geometry, or the layer declares that it has no highlightable elements — rather than emitting selectors that silently match nothing.

## Additional Information

**Pre-existing, not a regression.** `("scatter", "scattergl")` has been the line/multiline filter for as long as the Plotly path has existed; #303 only inherited it when adding steps. It is documented in place at `_CONNECTED_TRACE_TYPES` so it is not rediscovered from a silent non-highlight, but documenting is not fixing.

This matters most for large data, which is exactly when an author reaches for `scattergl` — so the charts affected are the ones where a sighted user most benefits from a highlight tracking their position.

**The decision this issue needs** is what should happen instead. Options:

1. Find the WebGL-rendered geometry and build a selector that matches it, if the frontend can highlight a canvas region at all.
2. Emit the layer without selectors, so the other three modalities still work and nothing claims a highlight that will not appear.
3. Convert or warn at export time.

Worth checking against the JS side (`xability/maidr`) whether a canvas-backed highlight is feasible before choosing, since option 1 depends entirely on that.

Related: #303.

PR #313, opened 2026-08-07
fix(plotly): resolve an absent mode the way plotly does before classifying

### LINKED ISSUE #308: fix(plotly): a line chart that never authored a `mode` is exported as scatter
# Bug Report

## Description

A Plotly line chart built without an explicit `mode="lines"` is classified as `SCATTER`, so it is announced as loose points rather than a connected line.

`Figure.to_dict()` omits `mode` entirely when the author never sets one, and `is_connected_line_trace()` in `maidr/plotly/step_shape.py` reads an absent `mode` as markers-only:

```python
mode = trace.get("mode")
if mode is None:
    return is_step_trace(trace)   # only a declared stepping shape is rescued
```

Plotly's own default draws lines either way — `"lines+markers"` below 20 points, `"lines"` at or above — so the chart on screen is a line while MAIDR describes scattered points. The connectivity of the data is lost for a screen-reader user.

## Steps to Reproduce

```python
import plotly.graph_objects as go

fig = go.Figure()
fig.add_scatter(x=list(range(25)), y=list(range(25)))   # no mode=

print([p.schema[MaidrKey.TYPE] for p in PlotlyMaidr(fig)._plots])
```

## Actual Behavior

`[<PlotType.SCATTER: 'scatter'>]` — at any point count.

## Expected Behavior

`[<PlotType.LINE: 'line'>]`, matching what Plotly actually renders.

## Additional Information

**This is pre-existing and predates step support** — the same `mode` default has governed line classification for as long as the Plotly path has existed. It is filed now because #303 hit the same defect for *staircases* and fixed only that case.

#303 deliberately scoped its fix to traces that declare a stepping `line.shape`, on the grounds that widening it would reclassify existing users' mode-less scatter traces — a real behaviour change that deserves its own change and its own argument rather than riding along on a step PR. The current behaviour is pinned by `test_a_modeless_trace_without_a_shape_is_untouched` in `tests/plotly/test_plotly_step.py`, so it is intentional rather than incidental.

**The decision this issue needs** is what a mode-less scatter trace should become. Options:

1. Mirror Plotly's own rule — `lines+markers` under 20 points, `lines` at or above — which is faithful but makes classification depend on data length.
2. Treat an absent `mode` as `lines` unconditionally, which is simpler and matches the larger-data case, but reclassifies small mode-less traces that today announce as scatter.
3. Leave it, and document the need for an explicit `mode`.

Whichever is chosen, `test_a_modeless_trace_without_a_shape_is_untouched` encodes the current behaviour and will need updating alongside it.

Related: #303.

jooyoungseo:
## Description

A Plotly line chart written without an explicit `mode="lines"` was exported as `SCATTER`, so it announced as loose points while plotly drew a connected line. The connectivity of the data — the thing a line chart exists to show — was lost for a screen-reader user.

The cause is that `Figure.to_dict()` **omits** `mode` unless the author set one, and the classifier read that absence as markers-only:

```python
mode = trace.get("mode")
if mode is None:
    return is_step_trace(trace)   # only a declared stepping shape was rescued
```

An absent `mode` is not "no drawing mode". It is "whatever plotly's default resolves to" — and plotly's default always draws lines.

## Related Issues

Closes #308. Pre-existing behaviour, surfaced by #303.

## Changes Made

**`maidr/plotly/step_shape.py`** — a new `default_mode(trace)` reproduces the rule plotly documents on its own `scatter.mode` attribute, quoted verbatim from the installed plotly 6.7.0:

> If there are less than 20 points and the trace is not stacked then the default is "lines+markers". Otherwise, "lines".

Both halves are implemented: the `< 20` boundary, and the stacked-trace exclusion (`stackgroup` defaults to `"lines"` at any size). `_trace_point_count` counts the points actually drawn — the shorter of `x`/`y`, since they are zipped — and tolerates a `y`-only trace or a trace carrying no sequence at all.

`is_connected_line_trace` then stands that default in before applying the existing rule, instead of short-circuiting:

```python
mode = trace.get("mode")
if mode is None:
    if is_step_trace(trace):
        return True          # #303's rescue, unchanged
    mode = default_mode(trace)

return "lines" in mode and "markers" not in mode
```

## Which option this takes, and why

#308 offered three. This is **option 1 — mirror plotly's own rule** — and the reason is consistency rather than fidelity for its own sake.

Option 2 (treat every absent `mode` as `lines`) would make a mode-less six-point trace a `LINE` while an explicit `mode="lines+markers"` six-point trace stays a `SCATTER` — even though plotly draws those two **identically**. That is two classification rules for one rendered picture. Option 1 keeps one rule: resolve what plotly would draw, then classify that.

Resulting behaviour, with only the first row changed:

| trace | before | after |
|---|---|---|
| mode-less, no shape, ≥ 20 points | `SCATTER` | **`LINE`** |
| mode-less, no shape, < 20 points | `SCATTER` | `SCATTER` |
| mode-less, `stackgroup` set, any size | `SCATTER` | **`LINE`** |
| mode-less + stepping `line.shape` | `STEP` | `STEP` |
| explicit `mode="markers"` | `SCATTER` | `SCATTER` |
| explicit `mode="lines+markers"` | `SCATTER` | `SCATTER` |
| explicit `mode="lines"` | `LINE` | `LINE` |

**What this deliberately does not fix.** A short mode-less line chart is still `SCATTER`. That is not an oversight — it is the same answer plotly's own default gives, and the same answer an explicit `lines+markers` gets today. The adjacent question is whether `lines+markers` should be a `LINE` at all, since plotly does connect those points. Changing *that* reclassifies every explicit `lines+markers` trace users have today, which is a much larger behaviour change than #308 describes and wants its own argument. Happy to file it separately if you want it looked at.

The `#303` step rescue is untouched and now runs **before** the default is resolved, so a staircase short enough to default to markers still binds as a `STEP`.

## Verification

New `tests/plotly/test_plotly_mode_default.py` — 24 tests covering the default rule, the classifier, and the exported layer end to end.

Passing is not evidence on its own, so each group was checked against a deliberately broken build:

| Mutation | Tests killed |
|---|---|
| restore the pre-fix `return False` for an absent mode | 5 — the classifier, the exported line, the stacked case, both selector tests |
| `< 20` → `<= 20` | 2 — the boundary tests |
| drop the `stackgroup` exclusion | 4 — all three stacked parametrisations plus the exported layer |

Two tests go past classification to the emitted selector, because reclassifying is only half the job: a rescued line now competes for `scatterlayer` positions with a step, which is exactly where a wrong `nth-child` highlights the wrong element. Both assert on the selector string.

`tests/plotly/test_plotly_step.py` — `test_a_modeless_trace_without_a_shape_is_untouched` still passes unchanged (its trace is 6 points), but it was renamed to `..._is_not_a_step` and recommented, because the *reason* it is a `SCATTER` changed: it is now plotly's marker default, not an absent mode reading as markers. Its class docstring claimed "everything else keeps its previous classification", which is no longer true; corrected.

Full suite: **297 passed** (273 before, +24). `ruff check` clean on `maidr/plotly/` and `tests/plotly/`.

## Type of Change

- [x] Bug fix
- [x] Behaviour change — mode-less traces of ≥ 20 points, and stacked mode-less traces, now export as `LINE` instead of `SCATTER`

## Checklist

- [x] I have added appropriate unit tests.
- [ ] I have updated the documentation — no user-facing doc describes `mode` handling; the rule is documented at `default_mode`.

## Additional Notes

**Merge-order note.** This is one of four sibling PRs closing the #303 follow-ups. It touches `maidr/plotly/step_shape.py`, which **#309 also touches** (a different region of the same file — the trace-type constants). #312 (tests only) and #311 don't conflict with it. Whichever of this and #309 merges first, the other needs a rebase; the overlap is small and adjacent rather than interleaved.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DZocjnGTiRvLDCbcvqYqa3)_

### DIFF
--- a/maidr/plotly/plotly_plot_factory.py
+++ b/maidr/plotly/plotly_plot_factory.py
@@ -8,8 +8,11 @@ class PlotlyPlotFactory:
     """
     Factory that maps Plotly trace types to PlotlyPlot subclasses.
 
-    For scatter traces, uses the ``mode`` attribute to disambiguate
-    between scatter (markers) and line (lines) plots.
+    For scatter traces, the drawing mode disambiguates loose markers from a
+    connected line. That is resolved through
+    :func:`maidr.plotly.step_shape.is_connected_line_trace` rather than read
+    off the trace, because ``Figure.to_dict()`` omits ``mode`` when the author
+    never set one — and an absent ``mode`` still draws a line.
     """
 
     @staticmethod
--- a/maidr/plotly/step_shape.py
+++ b/maidr/plotly/step_shape.py
@@ -46,6 +46,72 @@ STEP_SHAPE_DIRECTION: dict[str, str] = {
 _CONNECTED_TRACE_TYPES = ("scatter", "scattergl")
 
 
+#: Point count at which plotly's default ``mode`` drops its markers.
+#:
+#: From the plotly.js ``scatter.mode`` attribute: "If there are less than 20
+#: points and the trace is not stacked, then the default is 'lines+markers'.
+#: Otherwise, 'lines'." Both halves matter — a mode-less trace is never
+#: markers-only, so reading an absent ``mode`` as markers describes a chart
+#: plotly draws as a connected line.
+_MARKER_DEFAULT_MAX_POINTS = 20
+
+
+def _trace_point_count(trace: dict) -> int:
+    """
+    Count the points a trace will actually draw.
+
+    ``x`` and ``y`` are zipped when the data is extracted, so the shorter of
+    the two is the number of points that reach the chart. A trace supplying
+    only ``y`` counts that, matching plotly generating ``x`` as ``0..n-1``.
+
+    Parameters
+    ----------
+    trace : dict
+        The plotly trace dictionary.
+
+    Returns
+    -------
+    int
+        The number of drawn points; 0 when neither axis carries a sequence.
+    """
+    lengths = [
+        len(values)
+        for values in (trace.get("x"), trace.get("y"))
+        if values is not None and hasattr(values, "__len__")
+    ]
+    return min(lengths) if lengths else 0
+
+
+def default_mode(trace: dict) -> str:
+    """
+    Resolve the ``mode`` plotly itself applies when the author sets none.
+
+    ``Figure.to_dict()`` omits ``mode`` entirely unless it was set, so the
+    exported dict cannot be read literally — an absent ``mode`` is not "no
+    drawing mode", it is "whatever plotly's default resolves to". Reproducing
+    that default here keeps the classification tied to what is drawn rather
+    than to what happens to be spelled out in the dict.
+
+    Parameters
+    ----------
+    trace : dict
+        The plotly trace dictionary.
+
+    Returns
+    -------
+    str
+        ``"lines+markers"`` for a short unstacked trace, ``"lines"`` otherwise.
+    """
+    # A stacked trace defaults to "lines" at any size -- plotly excludes it
+    # from the marker default explicitly.
+    if trace.get("stackgroup"):
+        return "lines"
+
+    if _trace_point_count(trace) < _MARKER_DEFAULT_MAX_POINTS:
+        return "lines+markers"
+    return "lines"
+
+
 def is_scatter_family_trace(trace: dict) -> bool:
     """
     Report whether a trace belongs to the subplot's scatter layer.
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Follow-up on the review point about `scattergl` — it was worth chasing, and the answer took three steps because the obvious source turned out to be useless.

**`Scattergl`'s Python docstring states no default at all.** Not a different one — none. `Scatter.mode.__doc__` carries the rule sentence; `Scattergl.mode.__doc__` stops after the flaglist. So it can neither confirm nor contradict, and a docstring-guard test modelled on the existing one would simply fail against a docstring that never made the claim.

**plotly.js says they share it.** The threshold is a single `PTS_LINESONLY: 20` constant, and the trace modules coerce with the same `n < PTS_LINESONLY ? "lines+markers" : "lines"` expression.

**Confirmed in a browser**, by reading back plotly's own resolved `gd._fullData[i].mode` — which is the ground truth, since it is what plotly actually decided to draw:

| points | `scattergl` | `scatter` |
|---|---|---|
| 5 | `lines+markers` | `lines+markers` |
| 19 | `lines+markers` | `lines+markers` |
| 20 | `lines` | `lines` |
| 25 | `lines` | `lines` |

Identical at every count, including across the boundary. So applying `default_mode` uniformly to the scatter family is correct, and neither option in your comment (a Scattergl doc-guard, or scoping the threshold to `"scatter"` only) is needed. Pinned by parametrised tests asserting the two resolve identically, with the derivation recorded in the test class docstring so the next reader doesn't repeat it.

Two smaller ones also applied: the drift-guard tests now carry a note that a failure should be checked for wording-only change before any code moves, and `PlotlyPlotFactory`'s docstring no longer claims the `mode` attribute is read directly.

Corrected count for the record: the suite is **306 passing**, not the 297 stated in the PR description above, which predates these additions.

---

PR #315, opened 2026-08-07
refactor(plotly): require a selector position instead of guessing one

### LINKED ISSUE #311: refactor(plotly): make the selector-position fallbacks consistent, or require positions

jooyoungseo:
## Description

The three Plotly line-family classes disagreed about what to do when nobody told them where their traces sit in the subplot, and the two fallbacks failed in different ways.

`PlotlyLinePlot` fell back to an **unscoped** selector:

```python
if self._scatter_position is None:
    return [f"{self._subplot_css_prefix()}.trace.scatter path.js-line"]
```

That over-matches — a step trace renders as `path.js-line` too, so the fallback selected it as well. At least over-matching is visible.

`PlotlyStepPlot` and `PlotlyMultiLinePlot` fell back to **leading order**:

```python
self._scatter_positions = (
    list(range(len(traces))) if scatter_positions is None else scatter_positions
)
```

That one is silent. Constructed for traces that are not the subplot's first scatter children, it emits `nth-child(1), nth-child(2), …` pointing at whichever elements happen to occupy those positions — no error, nothing visibly wrong in the output, and the wrong element highlighted.

## Related Issues

Closes #311. Follow-up to #303. Surfaced #316 (see below).

## Changes Made

Both fallbacks are removed; `scatter_position` / `scatter_positions` are now **required**. A missing position is a `TypeError` at the call site instead of a wrong element at runtime — the right trade for something whose failure mode is "highlights someone else's geometry".

Each class carries a `Notes` section recording *which* wrong default it used to have — these differ per class (unscoped for `PlotlyLinePlot`, leading-order for the other two), so the reasoning survives the diff rather than only living here.

**`PlotlyPlotFactory` is the one caller that genuinely cannot know a position.** It sees a single trace with no idea what else is on its subplot, so `0` is the only assumption available, and it now passes that explicitly rather than inheriting a hidden default. The comment there also records what the assumption costs when wrong — see "Residual risk" below.

**Validation** (`PlotlyPlot._validate_scatter_positions`), added during review: requiring a position closes the hole where a caller supplies *none*, but leaves the one where a caller supplies a **wrong** list, which fails identically silently. The validator checks, in order: container type → length → entry type → non-negativity → uniqueness. `TypeError` for shape problems, `ValueError` for content. Its docstring also states what it *cannot* check (see below).

**Lists are stored by value**, not aliased — a caller mutating its list after construction would otherwise silently change the layer's selectors on the next `render()`: validated input slipping past the guard by being edited after it ran.

## Verification

New `tests/plotly/test_plotly_selector_positions.py` — **22 test functions, 24 collected items**. Full suite **273 → 297 passed**. `ruff check --diff` clean under the version CI pins (0.3.4).

Covers: omission raises `TypeError` on all three classes; malformed lists (wrong length, negative, duplicate, empty, non-int, explicit `None`) raise with a message naming the argument; neither old fallback is reachable (a line built at position 3 emits `nth-child(4)` and is explicitly *not* the old unscoped string; a step layer given `[2, 4]` emits `nth-child(3)`/`(5)` where leading order would have emitted `(1)`/`(2)`); mutating the caller's list after construction changes nothing; the factory's stated position-0 assumption; and the production path still assigns a step and two lines `nth-child(1)`, `(2)`, `(3)`.

The copy and `None` tests were both checked against the un-fixed code and fail there, so they pin the behaviour rather than merely exercising it.

**`PlotlyMaidr._extract_plots` already supplied positions unconditionally**, so no production path changes behaviour — confirmed by the existing 273 tests passing unmodified except at the five direct-construction sites #311 predicted (`test_plotly_plots.py` ×4, `test_plotly_step.py` ×1).

## What this does *not* guard, deliberately

- **No upper bound.** A position beyond the subplot's actual scatter-trace count is well-formed by every rule here, so `scatter_position=99` on a two-trace subplot constructs happily and matches nothing at render. Only `_extract_plots` knows that total, so an upper bound would have to live there. Stated in the validator's docstring so the guard isn't read as exhaustive.
- **No cross-layer duplicate detection.** Two independently-constructed layers can still claim the same position. That needs a shared registry, which is a lot of machinery for a hazard the production path already precludes by centralising assignment.
- **Residual risk in the factory.** A direct caller handing it one trace out of a multi-trace figure used to get an unscoped selector that over-matched — wrong, but visibly. It now gets a confidently wrong one bound to position 0. That is the failure mode this parameter removes elsewhere, surviving in the single place that cannot know better. Documented at the call site.
- **The scalar error message stays list-shaped.** `PlotlyLinePlot(scatter_position=-1)` reports `got [-1]`. Threading a noun through the validator to fix the wording would put back the second code path that routing the scalar case through the list validator exists to avoid, and drift between two copies of the rule is a worse problem than an awkward message on an internal API.

## Type of Change

- [x] Refactor — no behaviour change on any path reachable from `PlotlyMaidr`
- [x] Breaking change for direct construction of `PlotlyLinePlot` / `PlotlyStepPlot` / `PlotlyMultiLinePlot`

On that second box: these are internal classes under `maidr/plotly/`, not exported from `maidr/__init__.py`, and the only in-repo callers are `_extract_plots`, the factory, and tests. Flagged rather than waved away, since anyone constructing them directly will get a `TypeError` — the intended outcome, but it shouldn't be a surprise.

## Checklist

- [x] I have added appropriate unit tests.
- [ ] I have updated the documentation — not applicable; internal classes with no user-facing docs.

## Additional Notes

**Spun out as an issue rather than fixed here: #316.** Reviewing this PR surfaced that `_extract_plot_data` drops a trace whose series comes out empty while `_get_selector` always emits one selector per trace, so an empty series desynchronises the two lists and every later series inherits its predecessor's selector. Same wrong-highlight failure, reached by a different route. Fixing it changes what `_extract_plot_data` emits, which would cost this PR its "no production path changes behaviour" property, so it is filed separately with the options laid out.

**Merge-order note.** The last of four sibling PRs closing the #303 follow-ups (#312, #313, #314). This one and **#314** both touch `line.py`, `multiline.py` and `step.py` — #314 changes what `_get_selector` returns, this one changes the constructors, so the overlap is real but in different methods. #313 and #312 don't conflict.

Suggested order to minimise rebasing: **#312 → #313 → this → #314**, since #314's `_get_selector` bodies are the most rewritten and are easiest to rebase last.

### DIFF
--- a/maidr/plotly/line.py
+++ b/maidr/plotly/line.py
@@ -42,7 +42,7 @@ class PlotlyLinePlot(PlotlyPlot):
     ) -> None:
         # Routed through the list validator so the non-negative rule has
         # one home rather than a scalar copy that can drift from it.
-        self._validate_scatter_positions([scatter_position], 1)
+        PlotlyPlot._validate_scatter_positions([scatter_position], 1)
 
         super().__init__(trace, layout, PlotType.LINE, **kwargs)
         self._scatter_position = scatter_position
--- a/maidr/plotly/line.py
+++ b/maidr/plotly/line.py
@@ -15,19 +15,29 @@ class PlotlyLinePlot(PlotlyPlot):
         The scatter/lines trace dict.
     layout : dict
         The Plotly figure layout.
-    scatter_position : int, optional
+    scatter_position : int
         The trace's zero-based position among the subplot's scatter-family
-        traces. Pass this whenever the subplot holds more than this one
-        scatter trace — a step trace beside it makes that the normal case.
+        traces. Required: it is the only thing that makes this layer's
+        selector address *this* trace rather than whichever one happens to
+        sit first. See the class note below on why it has no default.
     **kwargs : str
         Axis names forwarded to the parent class.
+
+    Notes
+    -----
+    ``scatter_position`` deliberately has no default. It previously fell back
+    to an unscoped ``.trace.scatter path.js-line``, which over-matched — a
+    step trace renders as ``path.js-line`` too, so the fallback selected it
+    as well. A caller that cannot supply a real position has to say so at the
+    call site now, rather than silently getting a selector that is wrong in a
+    way nothing reports.
     """
 
     def __init__(
         self,
         trace: dict,
         layout: dict,
-        scatter_position: int | None = None,
+        scatter_position: int,
         **kwargs: str,
     ) -> None:
         super().__init__(trace, layout, PlotType.LINE, **kwargs)
@@ -37,21 +47,12 @@ class PlotlyLinePlot(PlotlyPlot):
         """
         Return the selector for this line's rendered path.
 
-        With a known position the selector is scoped to that one trace.
-        Without one it falls back to the unscoped subplot-wide form, which
-        assumes this is the only ``path.js-line`` on the subplot. That
-        assumption held while a line layer owned every scatter trace, but a
-        step trace renders as ``path.js-line`` too, so the unscoped form
-        would match both. ``PlotlyMaidr`` therefore always supplies a
-        position; the fallback is for direct/standalone construction.
-
         Returns
         -------
         list of str
-            A single CSS selector.
+            A single CSS selector, scoped to this trace's position among the
+            subplot's scatter traces.
         """
-        if self._scatter_position is None:
-            return [f"{self._subplot_css_prefix()}.trace.scatter path.js-line"]
         return [self._scatter_line_selector(self._scatter_position)]
 
     def _extract_plot_data(self) -> list[list[dict]]:
--- a/maidr/plotly/line.py
+++ b/maidr/plotly/line.py
@@ -40,6 +40,10 @@ class PlotlyLinePlot(PlotlyPlot):
         scatter_position: int,
         **kwargs: str,
     ) -> None:
+        # Routed through the list validator so the non-negative rule has
+        # one home rather than a scalar copy that can drift from it.
+        self._validate_scatter_positions([scatter_position], 1)
+
         super().__init__(trace, layout, PlotType.LINE, **kwargs)
         self._scatter_position = scatter_position
 
--- a/maidr/plotly/multiline.py
+++ b/maidr/plotly/multiline.py
@@ -21,6 +21,8 @@ class PlotlyMultiLinePlot(PlotlyPlot):
     scatter_positions : list of int
         Each trace's zero-based position among the subplot's scatter-family
         traces, in trace order. Required: see the note below.
+    **kwargs : str
+        Axis names forwarded to the parent class.
 
     Notes
     -----
@@ -41,6 +43,10 @@ class PlotlyMultiLinePlot(PlotlyPlot):
         scatter_positions: list[int],
         **kwargs: str,
     ) -> None:
+        if not traces:
+            raise ValueError("a multi-line layer needs at least one trace")
+        self._validate_scatter_positions(scatter_positions, len(traces))
+
         super().__init__(traces[0], layout, PlotType.LINE, **kwargs)
         self._traces = traces
         self._scatter_positions = scatter_positions
--- a/maidr/plotly/multiline.py
+++ b/maidr/plotly/multiline.py
@@ -45,7 +45,7 @@ class PlotlyMultiLinePlot(PlotlyPlot):
     ) -> None:
         if not traces:
             raise ValueError("a multi-line layer needs at least one trace")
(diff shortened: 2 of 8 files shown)


jooyoungseo:
Applied the review point about validating the list itself — it's the right catch, and it exposes that this PR as originally written only did half the job.

Requiring a position closes the hole where a caller supplies *none*. It leaves the one where a caller supplies a **wrong** list, which fails exactly the same silent way:

- **wrong length** — slides every later series onto another element;
- **negative** — builds `nth-child(0)` or lower, matching nothing;
- **duplicate** — points two series at one element.

None of those raise on their own, which is the failure mode the whole parameter exists to prevent. A shared `PlotlyPlot._validate_scatter_positions` now checks all three, with the lone-line case routed through it as `[position]` so the non-negative rule has one home rather than a scalar copy that can drift from it. An empty `traces` list is rejected explicitly instead of dying on `traces[0]` inside the parent constructor.

**The useful signal: the full suite passed untouched after adding the validation.** The production path has never violated these invariants — the guard is there for callers that don't exist yet, not for a bug being fixed.

Also folded in the two nits: `**kwargs` is now documented on `PlotlyMultiLinePlot`, and the over-wrapped constructs that fit on one line are collapsed.

On the boundaries you flagged as out of scope — no upper-bound check, and no cross-layer duplicate detection — I agree and left both. The first is genuinely unknowable here: this method sees the *layer's* trace count, never the subplot's total. The second would need a shared registry, which is a lot of machinery for a hazard the production path already precludes by centralising assignment in `_extract_plots`. Worth stating rather than silently not doing.

Corrected count for the record: the suite is **291 passing**, not the 283 stated in the PR description above, which predates the validation tests.

---

jooyoungseo:
On the empty-series point you flagged as out of scope — I reproduced it, and it's worse than a length mismatch. Filed as #316.

```
series in data : 2 ['a', 'c']
selectors      : 3 [(1), (2), (3)]
  series 'a' -> nth-child(1)
  series 'c' -> nth-child(2)     <- 'c' is the third trace, renders at (3)
```

The lists don't just differ in length; because the frontend pairs positionally, every series after the empty one inherits its predecessor's selector. So it's the same silent-wrong-highlight failure this PR is about, reached by a different route.

I left it out of this PR deliberately. Fixing it means changing what `_extract_plot_data` emits, and this PR's whole claim is that it's a no-op on every production path — folding a real behaviour change in would cost that. #316 lays out the three options and argues for filtering `scatter_positions` by the same predicate that filters the data, so both lists come from one pass instead of diverging again.

Also fixed the pluralisation nit — `position(s)`/`trace(s)` is now proper singular/plural.

---

PR #319, opened 2026-08-08
fix(smooth): keep a regression line navigable point by point

### LINKED ISSUE #318: fix(smooth): regression line collapses to 2 navigable points
## Summary

On the [Regression Plots example](https://py.maidr.ai/examples.html#regression-plots), pressing <kbd>Page Up</kbd> to reach the second layer (the `smooth` trace, i.e. the fitted line) leaves only **two** navigable data points — the start and the end of the line. Auto-play is a single jump, and one <kbd>→</kbd> press already reports "No more data to display".

The KDE curve overlaid on the [Histogram example](https://py.maidr.ai/examples.html#histogram) is the same `smooth` trace type and navigates normally with ~30 points, so the two layers behave inconsistently for the same kind of overlay.

## Reproduction

Browser-driven against the live payload from `py.maidr.ai`, activating the figure and pressing <kbd>Page Up</kbd> then <kbd>→</kbd> repeatedly:

**Regression plot — layer 2 (`smooth`)**

```
Layer 2 of 2: smooth plot at X values is 0.0, Y values is 1.13
X values is 10.0, Y values is 19.97
No more data to display
```

**Histogram — layer 2 (`smooth`, the KDE curve)**

```
Layer 2 of 2: smooth plot at Petal Length (cm) is 1.2, Frequency is 13
Petal Length (cm) is 1.3, Frequency is 14
Petal Length (cm) is 1.4, Frequency is 15
… (30 points total)
```

The emitted `smooth` layer data confirms it:

| example | points in `smooth` layer | source `Line2D` vertices |
|---|---|---|
| regression plot | **2** | 100 |
| histogram + KDE | 30 | 200 |

## Root cause

`SmoothPlot._extract_plot_data()` in `maidr/core/plot/regplot.py` thins the fitted line with Ramer–Douglas–Peucker via `simplify_curve()`:

```python
if len(x_data) > _DEFAULT_MAX_SMOOTH_POINTS:
    mask = simplify_curve(
        np.column_stack([x_data, y_data]),
        target=_DEFAULT_MAX_SMOOTH_POINTS,
    )
```

RDP preserves *shape*, dropping vertices that lie close to the chord between their neighbours. `sns.regplot`'s default fit is a straight line evaluated on a 100-point grid, so every interior vertex is exactly collinear, every perpendicular distance is `0`, and the `dists[max_rel] > epsilon` test never fires — even at `epsilon = 0`. The 100 vertices collapse to the 2 endpoints. RDP is doing precisely what it is designed to do; it is the wrong objective for this trace.

## Why spacing matters more than shape here

A `smooth` trace is navigated and auto-played one vertex at a time at a fixed rate — `SmoothTrace` in the JS side maps vertex *index* to time and `y` to frequency. So the vertices need to be evenly spaced along the line for the sonified sweep to match the visual trend. Shape-preserving thinning warps that time axis.

This shows up on the KDE curve too, just less obviously. Its 30 surviving x values are clustered around the bends:

```
1.00 1.18 1.33 1.44 1.50 1.59 1.71 1.83 2.27 2.48 2.69 2.81 2.87 2.93 3.02 3.13
3.37 4.11 4.38 4.53 4.65 4.77 4.82 4.91 5.03 5.15 5.42 6.40 6.66 6.90
```

Steps range from `0.059` to `0.978` — a **16.6×** spread, including a visible skip from 1.8 straight to 2.3. Auto-play therefore races through the flat stretches and crawls through the bends.

## Scope

- **py-maidr only.** `maidr/core/plot/regplot.py` is the only smooth-layer emitter that applies RDP.
- `maidr` (JS) consumes the points as given; `SmoothTrace` needs no change.
- r-maidr is unaffected — `Ggplot2SmoothLayerProcessor` emits every built vertex, and its only `simplify_curve()` caller is the violin processor.
- `ViolinKdePlot`'s use of `simplify_curve()` is a different problem (picking the levels that outline a violin) and is correct as is.

## Expected behaviour

The `smooth` layer keeps up to `_DEFAULT_MAX_SMOOTH_POINTS` (30) vertices spread evenly along the fitted line — including for a perfectly straight fit — so the regression line navigates and auto-plays exactly like the KDE curve.

## Environment

- py-maidr `main` @ `13002c3`
- maidr JS `@latest` from jsDelivr, Chromium 1194
- seaborn regplot / histplot(kde=True), matplotlib Agg

jooyoungseo:
## Description

The `smooth` layer of a `sns.regplot` collapsed to **two** navigable points. Pressing <kbd>Page Up</kbd> onto the fitted-line layer on the [Regression Plots example](https://py.maidr.ai/examples.html#regression-plots) landed on the start of the line, and a single <kbd>→</kbd> jumped to the end and reported "No more data to display". Auto-play was one jump. The KDE curve overlaid on a histogram is the same trace type and navigates with 30 points, so the two overlays behaved inconsistently.

`SmoothPlot._extract_plot_data()` thinned the fitted line with Ramer–Douglas–Peucker via `simplify_curve()`. RDP preserves *shape*, dropping vertices that lie on the chord between their neighbours. A seaborn regplot fit is a straight line sampled on a 100-point grid, so every interior vertex is exactly collinear, every perpendicular distance is `0`, and `dists[max_rel] > epsilon` never fires — not even at `epsilon = 0`. The line reduced to its two endpoints.

Shape preservation is the wrong objective for this trace. A `smooth` layer is navigated and auto-played one vertex at a time at a fixed rate (`SmoothTrace` maps vertex index to time and `y` to frequency), so the vertices need to be *evenly spaced along x* for the sonified sweep to track the visual trend. The same thinning warped the KDE curve as well, clustering its 30 vertices around the bends until the step along x varied by 16.6× (a visible skip from 1.8 straight to 2.3), racing through flat stretches and crawling through curves.

## Related Issues

Fixes #318

## Scope

This PR is deliberately limited to the reported bug. It contains two things:

1. **RDP → even resampling** — the fix for #318.
2. **Even by vertex index → even along x** — because step 1 alone made `sns.regplot(..., lowess=True)` on a skewed sample *worse* than what it replaced (**319×** step spread versus RDP's 87×). A regression this PR introduced, so it is fixed here rather than deferred.

Screen-space pacing — which matters once an axis is log-scaled, where even-along-x stops being even-on-screen — was split out to **#320** so this PR stays close to the bug it fixes. `resample_curve`'s docstring records that limitation until #320 lands.

## Changes Made

- `maidr/util/rdp_utils.py` — added `resample_curve()`, which thins a polyline to points evenly spaced along x, always keeping the first and last. It places samples on an even x grid and interpolates y whenever x increases monotonically, and falls back to even-by-vertex-index for a curve that decreases or doubles back. Module docstring explains why two thinning strategies now coexist.
- `maidr/core/plot/regplot.py` — `SmoothPlot._extract_plot_data()` calls `resample_curve()` instead of `simplify_curve()`, with a comment on why spacing beats shape here.
- `tests/core/plot/test_smooth_sampling.py` — new. End-to-end fixtures for a linear `regplot`, `histplot(kde=True)`, and a clustered-x `lowess` fit, plus unit coverage of `resample_curve`.

`ViolinKdePlot`'s use of `simplify_curve()` is untouched — outlining a violin genuinely wants shape preservation, and it remains that function's only caller.

## Verification

Browser-driven with Playwright, activating the figure and pressing <kbd>Page Up</kbd> then <kbd>→</kbd> repeatedly.

**Regression plot, layer 2 — before**

```
Layer 2 of 2: smooth plot at X values is 0.0, Y values is 1.13
X values is 10.0, Y values is 19.97
No more data to display
```

**Regression plot, layer 2 — after**

```
Layer 2 of 2: smooth plot at X values is 0.0, Y values is 1.10
X values is 0.3, Y values is 1.68
X values is 0.7, Y values is 2.45
X values is 1.0, Y values is 3.03
… (30 points, evenly spaced)
```

**Histogram + KDE, layer 2 — before** (note the skip from 1.8 to 2.3)

```
1.2 → 1.3 → 1.4 → 1.5 → 1.6 → 1.7 → 1.8 → 2.3 → 2.5 → 2.7 → 2.8 → 2.9
```

**Histogram + KDE, layer 2 — after**

```
1.2 → 1.4 → 1.6 → 1.8 → 2.0 → 2.2 → 2.4 → 2.6 → 2.8 → 3.0 → 3.2 → 3.4
```

Spacing spread (max x-step ÷ min x-step) across the smooth layer:

| fit | before (RDP) | after |
|---|---|---|
| `regplot` (order=1) | 2 points only | 1.00× |
| `regplot(order=2)` | — | 1.00× |
| `regplot(lowess=True)`, skewed x | 86.7× | 1.00× |
| `histplot(kde=True)` | 16.6× | 1.00× |

### Nothing changes for sighted users

The emitted `svg_x`/`svg_y` feed `SmoothTraceSvgXY`'s highlight overlay, so this was checked rather than assumed. Comparing old and new extraction with `ci=None` — seaborn's bootstrap CI band otherwise makes any two runs differ by ~1 px regardless of this change, a confound that has to be removed first:

- The fitted line's SVG `d` attribute is **byte-identical**, 100 vertices in both. `_extract_plot_data` only reads `Line2D.get_xydata()`; hashed the vertex array before and after extraction to confirm.
- Rendered screenshots of the un-activated chart: **0 differing pixels** out of 490,000.
- Mid-navigation screenshots differ by 323 pixels — the highlight dot at a different point along the line, which is the fix.
- The overlay circles stay hidden: 31 exist after activation (30 points + the highlight clone) versus 3 before, but exactly **1 is visible** in both.

Interpolated points land on the drawn line to **7e-12 px** on a linear axis. A log axis leaves ~0.107 px, sub-pixel and addressed by #320.

### Tests are not vacuous

Each behaviour was re-run against the code it replaced: 3 tests fail against the pre-fix RDP extractor, and the `lowess` end-to-end case plus the clustered-curve unit test fail against index sampling.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Checklist

- [x] I have tested the changes locally, and all tests related to this pull request pass — `uv run pytest` green at **404 passed**, and `ruff check` clean on the changed files. The pre-existing `ruff` errors elsewhere in `maidr/`, and the one 98-char signature in `rdp_utils.py` that this branch never touches, are left alone.
- [ ] I have updated the documentation, if applicable — no docs change needed; the example pages regenerate from the same source.

## Additional Notes

Scope is py-maidr only. The `maidr` JS side consumes the points as given and needs no change, and r-maidr is unaffected — `Ggplot2SmoothLayerProcessor` emits every built vertex and its only `simplify_curve()` caller is the violin processor.

Earlier revisions of this branch also carried the screen-pacing work; that history now lives on `claude/smooth-screen-pacing-vqse7v` and is reviewed in #320, so review comments on this PR referring to `to_scaled_coords` or `_thin_to_even_steps` belong there.

---

### DIFF
--- a/maidr/core/plot/regplot.py
+++ b/maidr/core/plot/regplot.py
@@ -6,7 +6,7 @@ from maidr.exception.extraction_error import ExtractionError
 import numpy as np
 from maidr.core.enum.plot_type import PlotType
 from maidr.core.enum.maidr_key import MaidrKey
-from maidr.util.rdp_utils import simplify_curve
+from maidr.util.rdp_utils import resample_curve
 from maidr.util.regression_line_utils import find_regression_line
 from maidr.util.svg_utils import data_to_svg_coords
 
@@ -70,14 +70,16 @@ class SmoothPlot(MaidrPlot):
         xydata = np.asarray(regression_line.get_xydata())
         x_data, y_data = xydata[:, 0], xydata[:, 1]
 
-        # Simplify the curve with RDP before computing SVG coordinates.
-        if len(x_data) > _DEFAULT_MAX_SMOOTH_POINTS:
-            mask = simplify_curve(
-                np.column_stack([x_data, y_data]),
-                target=_DEFAULT_MAX_SMOOTH_POINTS,
-            )
-            x_data = x_data[mask]
-            y_data = y_data[mask]
+        # Thin the curve before computing SVG coordinates.  The points are
+        # navigated and auto-played one at a time at a fixed rate, so they have
+        # to stay evenly spaced along the line: shape-based simplification
+        # would collapse a straight fit to its two endpoints and stretch the
+        # flat stretches of a curved one over fewer steps than the bends.
+        xy = resample_curve(
+            np.column_stack([x_data, y_data]),
+            target=_DEFAULT_MAX_SMOOTH_POINTS,
+        )
+        x_data, y_data = xy[:, 0], xy[:, 1]
 
         x_svg, y_svg = data_to_svg_coords(self.ax, x_data, y_data)
         return [
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -155,8 +155,9 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     covers every curve seaborn fits.  Sampling by vertex index instead would
     only be even in x when the source grid already was — true of a plain
     ``regplot`` fit or a KDE, but not of a ``lowess=True`` fit, which lands on
-    the observed x values and inherits their clustering.  Curves that double
-    back carry no usable x ordering, so those fall back to index sampling.
+    the observed x values and inherits their clustering.  Anything else — a
+    curve that doubles back, or one running right to left — falls back to
+    index sampling.
 
     Interpolated points sit exactly on the drawn line: matplotlib renders the
     curve as straight segments between its vertices, which is the same path
@@ -187,7 +188,9 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
         grid = np.linspace(x[0], x[-1], count)
         return np.column_stack([grid, np.interp(grid, x, y)])
 
-    # No usable x ordering, so fall back to even spacing by vertex index.
+    # Left-to-right is the only order worth special-casing, since it is the one
+    # every fit produces; a decreasing or self-crossing curve falls through to
+    # even spacing by vertex index rather than growing a case for each shape.
     # ``n > count >= 2`` puts the step ``(n - 1) / (count - 1)`` strictly above
     # 1, so rounding can never land two samples on the same vertex: the result
     # always holds exactly ``count`` distinct points.
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -159,9 +159,13 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     curve that doubles back, or one running right to left — falls back to
     index sampling.
 
-    Interpolated points sit exactly on the drawn line: matplotlib renders the
-    curve as straight segments between its vertices, which is the same path
-    the interpolation walks.
+    Interpolated points sit on the drawn line, which matters because they are
+    what the highlight ring lands on.  Matplotlib renders the curve as straight
+    segments between its vertices, so on a linear axis — where data and display
+    space differ by an affine map — the interpolation walks exactly that path.
+    A log axis bends each segment between the vertices, leaving a gap of about
+    a tenth of a pixel at this point count; still far under the ring's radius,
+    just not the exact landing a linear axis gives.
 
     Parameters
     ----------
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -167,7 +167,9 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     Returns
     -------
     np.ndarray, shape (M, 2)
-        The retained points, where ``M == min(N, max(target, 2))``.
+        The retained points, where ``M == min(N, max(target, 2))``.  A curve
+        that already fits the budget comes back as *points* itself, not a
+        copy, so callers that intend to mutate the result should copy it.
     """
     n = len(points)
     count = max(int(target), 2)
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -151,10 +151,16 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     collapsing to its two endpoints, and a curved line keeps a steady step
     size.  The first and last vertices are always kept.
 
-    Sampling is by vertex index, so the result is evenly spaced *in x* only
-    when the input is too — which holds for the curves matplotlib and seaborn
-    evaluate on a uniform grid.  A curve sampled unevenly would need
-    arc-length interpolation instead.
+    Spacing is measured along x whenever x increases monotonically, which
+    covers every curve seaborn fits.  Sampling by vertex index instead would
+    only be even in x when the source grid already was — true of a plain
+    ``regplot`` fit or a KDE, but not of a ``lowess=True`` fit, which lands on
+    the observed x values and inherits their clustering.  Curves that double
+    back carry no usable x ordering, so those fall back to index sampling.
+
+    Interpolated points sit exactly on the drawn line: matplotlib renders the
+    curve as straight segments between its vertices, which is the same path
+    the interpolation walks.
 
     Parameters
     ----------
@@ -167,7 +173,7 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     Returns
     -------
     np.ndarray, shape (M, 2)
-        The retained points, where ``M == min(N, max(target, 2))``.  A curve
+        The resampled points, where ``M == min(N, max(target, 2))``.  A curve
         that already fits the budget comes back as *points* itself, not a
         copy, so callers that intend to mutate the result should copy it.
     """
@@ -176,6 +182,12 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     if n <= count:
         return points
 
+    x, y = points[:, 0], points[:, 1]
+    if np.all(np.diff(x) > 0):
+        grid = np.linspace(x[0], x[-1], count)
+        return np.column_stack([grid, np.interp(grid, x, y)])
+
+    # No usable x ordering, so fall back to even spacing by vertex index.
     # ``n > count >= 2`` puts the step ``(n - 1) / (count - 1)`` strictly above
     # 1, so rounding can never land two samples on the same vertex: the result
     # always holds exactly ``count`` distinct points.
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -1,9 +1,15 @@
-"""Ramer-Douglas-Peucker curve simplification utilities.
+"""Curve thinning utilities.
 
-These helpers reduce the number of points on a curve while preserving its
-shape.  They are used by :class:`~maidr.core.plot.violin_kde_plot.ViolinKdePlot`
-and :class:`~maidr.core.plot.regplot.SmoothPlot` to keep the MAIDR JSON
-payload compact.
+These helpers reduce the number of points on a curve to keep the MAIDR JSON
+payload compact.  Two strategies live here because the two callers want
+different things:
+
+- :func:`simplify_curve` runs Ramer-Douglas-Peucker, which preserves *shape*
+  and is used by :class:`~maidr.core.plot.violin_kde_plot.ViolinKdePlot` to
+  pick the levels that outline a violin.
+- :func:`resample_curve` keeps the vertices evenly spaced and is used by
+  :class:`~maidr.core.plot.regplot.SmoothPlot`, whose points are navigated and
+  sonified one at a time, so *spacing* matters more than shape.
 """
 
 from __future__ import annotations
@@ -133,3 +139,34 @@ def simplify_curve(
             break
 
     return best_mask
+
+
+def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
+    """
+    Thin a 2-D curve down to *target* evenly spaced vertices.
+
+    Unlike :func:`simplify_curve`, this keeps the retained vertices spread
+    evenly along the curve rather than clustering them where the curve bends.
+    A straight line therefore survives as *target* points instead of
+    collapsing to its two endpoints, and a curved line keeps a steady step
+    size.  The first and last vertices are always kept.
+
+    Parameters
+    ----------
+    points : np.ndarray, shape (N, 2)
+        Ordered (x, y) points.
+    target : int
+        Desired maximum number of retained points.
+
+    Returns
+    -------
+    np.ndarray, shape (M, 2)
+        The retained points, where ``M <= max(target, 2)``.
+    """
+    n = len(points)
+    if n <= target:
+        return points
+
+    count = max(int(target), 2)
+    idx = np.unique(np.rint(np.linspace(0, n - 1, count)).astype(int))
+    return points[idx]
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -151,22 +151,31 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     collapsing to its two endpoints, and a curved line keeps a steady step
     size.  The first and last vertices are always kept.
 
+    Sampling is by vertex index, so the result is evenly spaced *in x* only
+    when the input is too — which holds for the curves matplotlib and seaborn
+    evaluate on a uniform grid.  A curve sampled unevenly would need
+    arc-length interpolation instead.
+
     Parameters
     ----------
     points : np.ndarray, shape (N, 2)
         Ordered (x, y) points.
     target : int
-        Desired maximum number of retained points.
+        Desired number of retained points.  Values below 2 are raised to 2,
+        since the endpoints alone already describe a segment.
 
     Returns
     -------
     np.ndarray, shape (M, 2)
-        The retained points, where ``M <= max(target, 2)``.
+        The retained points, where ``M == min(N, max(target, 2))``.
     """
     n = len(points)
-    if n <= target:
+    count = max(int(target), 2)
+    if n <= count:
         return points
 
-    count = max(int(target), 2)
-    idx = np.unique(np.rint(np.linspace(0, n - 1, count)).astype(int))
+    # ``n > count >= 2`` puts the step ``(n - 1) / (count - 1)`` strictly above
+    # 1, so rounding can never land two samples on the same vertex: the result
+    # always holds exactly ``count`` distinct points.
+    idx = np.rint(np.linspace(0, n - 1, count)).astype(int)
     return points[idx]
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -167,6 +167,15 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     a tenth of a pixel at this point count; still far under the ring's radius,
     just not the exact landing a linear axis gives.
 
+    Known limitation: the grid is even in *data* x, which is even on screen
+    only while the axis is linear.  A log x-axis stretches the same points
+    into a roughly 100:1 spread across the plot, so auto-play sweeps the
+    picture unevenly even though the announced x values step uniformly.
+    Nothing here sets a nonlinear scale — it takes an explicit
+    ``ax.set_xscale`` — and which of the two should stay uniform is a call
+    about what the sweep represents, so this stays as-is until a caller needs
+    otherwise.
+
     Parameters
     ----------
     points : np.ndarray, shape (N, 2)
--- a/tests/core/plot/test_smooth_sampling.py
+++ b/tests/core/plot/test_smooth_sampling.py
@@ -166,12 +166,17 @@ def test_resample_curve_keeps_a_straight_line_at_full_budget():
 
 
 def test_resample_curve_returns_short_curves_untouched():
-    """Nothing to thin when the curve already fits in the budget."""
+    """Nothing to thin when the curve already fits in the budget.
+
+    Identity, not just equality: the docstring promises the input array back
+    rather than a copy, so a refactor that quietly started copying would be a
+    contract change even though every value still matched.
+    """
     points = np.column_stack([np.arange(5.0), np.arange(5.0) ** 2])
 
     kept = resample_curve(points, target=30)
 
-    assert np.array_equal(kept, points)
+    assert kept is points
 
 
 @pytest.mark.parametrize("n,target", [(100, 30), (31, 30), (61, 30), (7, 5)])
--- a/tests/core/plot/test_smooth_sampling.py
+++ b/tests/core/plot/test_smooth_sampling.py
@@ -158,6 +158,35 @@ def test_resample_curve_returns_exactly_the_requested_count(n, target):
     assert len(kept) == target
 
 
+def test_resample_curve_evens_out_a_clustered_source_curve():
+    """A lowess fit lands on the observed x values and inherits their gaps.
+
+    Sampling such a curve by vertex index would carry the clustering straight
+    through into the emitted points, which is what auto-play pacing feels.
+    """
+    x = np.concatenate([np.linspace(0, 1, 90), np.linspace(2, 10, 10)])
+    points = np.column_stack([x, x**2])
+    source_gaps = np.diff(x)
+    assert source_gaps.max() / source_gaps.min() > 50, "fixture must be clustered"
+
+    kept = resample_curve(points, target=30)
+
+    gaps = np.diff(kept[:, 0])
+    assert gaps.max() / gaps.min() < 2.0, f"clustering survived: {gaps}"
+
+
+def test_resample_curve_falls_back_to_index_sampling_when_x_doubles_back():
+    """A curve with no usable x ordering still gets thinned, just by index."""
+    t = np.linspace(0, 2 * np.pi, 100)
+    points = np.column_stack([np.cos(t), np.sin(t)])
+
+    kept = resample_curve(points, target=30)
+
+    assert len(kept) == 30
+    assert np.array_equal(kept[0], points[0])
+    assert np.array_equal(kept[-1], points[-1])
+
+
 def test_resample_curve_keeps_both_endpoints():
     """The first and last vertices anchor the navigable range."""
     points = np.column_stack([np.linspace(0, 7, 61), np.sin(np.linspace(0, 7, 61))])
--- a/tests/core/plot/test_smooth_sampling.py
+++ b/tests/core/plot/test_smooth_sampling.py
@@ -23,6 +23,7 @@ from maidr.core.enum.plot_type import PlotType  # noqa: E402
 from maidr.core.figure_manager import FigureManager  # noqa: E402
 from maidr.core.plot.regplot import _DEFAULT_MAX_SMOOTH_POINTS  # noqa: E402
 from maidr.util.rdp_utils import resample_curve  # noqa: E402
+from maidr.util.regression_line_utils import find_regression_line  # noqa: E402
 
 
 def _smooth_points(fig) -> list[dict]:
@@ -31,6 +32,18 @@ def _smooth_points(fig) -> list[dict]:
     smooth = [p for p in plots if p.type is PlotType.SMOOTH]
     assert len(smooth) == 1, f"expected one smooth layer, got {len(smooth)}"
     return smooth[0].schema["data"][0]
+
+
+def _source_line(fig) -> np.ndarray:
+    """Return the vertices of the line the extractor actually reads.
+
+    Resolved through ``find_regression_line`` — the same lookup
+    ``SmoothPlot`` uses — so a test's idea of the source curve cannot drift
+    from the extractor's if seaborn ever adds another ``Line2D`` to the axes.
+    """
+    line = find_regression_line(FigureManager.get_axes(fig)[0])
+    assert line is not None, "no regression line found on the axes"
+    return np.asarray(line.get_xydata())
 
 
 def _x_values(points: list[dict]) -> np.ndarray:
@@ -77,8 +90,7 @@ def test_straight_regression_line_keeps_the_full_point_budget(regplot_figure):
     so the assertion tracks ``resample_curve``'s contract instead of whichever
     grid size seaborn happens to evaluate the fit on.
     """
-    source = FigureManager.get_axes(regplot_figure)[0].get_lines()[-1]
-    expected = min(_DEFAULT_MAX_SMOOTH_POINTS, len(source.get_xydata()))
+    expected = min(_DEFAULT_MAX_SMOOTH_POINTS, len(_source_line(regplot_figure)))
 
     points = _smooth_points(regplot_figure)
 
@@ -104,8 +116,7 @@ def test_smooth_points_are_evenly_spaced(figure_fixture, request):
 def test_smooth_points_span_the_whole_line(figure_fixture, request):
     """Thinning keeps both endpoints, so the trace covers the fitted range."""
     fig = request.getfixturevalue(figure_fixture)
-    line = FigureManager.get_axes(fig)[0].get_lines()[-1]
-    source_x = np.asarray(line.get_xydata())[:, 0]
+    source_x = _source_line(fig)[:, 0]
     x = _x_values(_smooth_points(fig))
 
     assert x[0] == pytest.approx(source_x[0])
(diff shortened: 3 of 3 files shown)


jooyoungseo:
Addressed two of the three observations in 817455d.

**The exact-count assertion was coupled to seaborn's grid — fixed, by strengthening the contract rather than weakening the test.**

The review is right that `len(points) == 30` was testing seaborn's internal grid size. But checking the collision case showed `np.unique` was never doing anything: thinning only runs when the curve is longer than the target, which puts the step `(n - 1) / (count - 1)` strictly above 1. Two consecutive samples therefore differ by more than 1 before rounding, and rounding moves each by at most 0.5, so they can never land on the same vertex. Swept `count` in 2..200 against `n` in `count+1 .. count+400` — no collisions.

So `resample_curve()` already returned exactly `min(N, max(target, 2))` points and the docstring's `M <= max(target, 2)` hedge understated it. Dropped the dead `np.unique`, moved the short-circuit to compare against the clamped `count` so the degenerate inputs (`n=1`, `target<2`) still return early, and promised the exact count. The budget test now derives its expectation from the fitted line and `_DEFAULT_MAX_SMOOTH_POINTS` instead of hard-coding 30, plus a parametrized unit test pins the exact count at awkward ratios (`31→30`, `61→30`, `7→5`).

**Arc-length caveat — added.** `resample_curve()`'s docstring now states that index sampling is even in x only because the callers evaluate on a uniform grid, and that an unevenly sampled curve would need arc-length interpolation.

**`np.column_stack` on the no-op path — skipped**, agreeing with the review that it's negligible. Re-adding the length guard at the call site to save one stack of a ≤100×2 array would trade the simpler code for nothing measurable.

Behaviour is unchanged — re-ran the browser check after the refactor and the regression layer still steps through 30 evenly spaced points. `uv run pytest` is green at 399 passed; `ruff check` clean on the three changed files. CI was green across Python 3.9–3.12 on 3f8e506.

---

jooyoungseo:
Line-length nit fixed in 1248520 — `_smooth_points` now pulls the plot list out to its own local, and no line in the file exceeds 88.

One correction on the framing, since it affects how the config reads rather than just this line: `ruff check` was genuinely clean, and still is. `E501` lives in ruff's `E5` group, which isn't part of the default enabled set (`E4`, `E7`, `E9`, `F`), so `line-length = 88` in `pyproject.toml` only drives the formatter — `ruff check --diff`, which is exactly what the `lint` job runs via `astral-sh/ruff-action@v3`, never flagged it. That job passed on both prior commits. Confirmed by running `ruff check --select E501` explicitly, which does report the line, versus the bare `ruff check`, which does not.

So the 89-char line was over the configured width but not a CI failure, and it was in company — `rdp_utils.py` and `regplot.py` already carry pre-existing lines at 94–105 chars. Wrapped it anyway since it costs nothing.

`uv run pytest` remains green at 399 passed.

On the `resample_curve` naming note: agreed it's index-based subsampling rather than interpolating resampling. Leaving the name as is — it pairs with the existing `simplify_curve`, and the docstring states the index-sampling behaviour and the uniform-grid precondition directly.

---

jooyoungseo:
Nitpick addressed in 4babbce, and I verified the visual-rendering question this raised.

**The `get_lines()[-1]` vs `find_regression_line()` coupling.** Fixed by removing the indirection rather than documenting it: a `_source_line()` helper now resolves the curve through `find_regression_line()` — the same lookup `SmoothPlot` uses — so the tests and the extractor can never mean different curves. That's stronger than asserting a single `Line2D`, since it stays correct even if a future `regplot` does add one.

**Return-by-reference on the no-op path** — leaving as is. `_extract_plot_data` immediately slices the result into `x_data`/`y_data` and never writes through it, and a defensive copy on a path that by definition returns the caller's own array would be cost without a caller to protect.

**Rendering impact — confirmed nil for the static chart.** Since this changes `svg_x`/`svg_y`, and `SmoothTraceSvgXY.mapToSvgElements` builds overlay circles from them, it was worth confirming what sighted users see. Comparing old and new extraction with `ci=None` (seaborn's bootstrap CI band otherwise makes runs differ by ~1 px regardless of this change — the confound that has to be removed first):

- The fitted line's SVG `d` attribute is **byte-identical**, 100 vertices in both. `_extract_plot_data` reads `Line2D.get_xydata()` and never writes back — hashed the vertex array before and after extraction to confirm.
- The whole SVG is identical once matplotlib's timestamp and random element ids are normalised.
- Rendered screenshots of the un-activated chart: **0 differing pixels** out of 490,000.
- Mid-navigation screenshots differ by 323 pixels — the highlight dot sitting at a different point along the line, which is the intended fix.

The extra overlay circles stay hidden: 31 circles exist after activation (30 points + the highlight clone) versus 3 before, but exactly **1 is visible** in both cases. If the hidden ones painted, the 323-pixel nav diff would be far larger.

So the fix is invisible to anyone who never activates MAIDR, and for those who do the highlight now walks the line instead of jumping between its two ends.

`uv run pytest` green at 399 passed; `ruff check` clean.

---

jooyoungseo:
Both points handled in 88e3a5b.

**Aliasing on the early return — documented.** Taking the docstring option over `.copy()`: the `Returns` section now states that a curve already within the budget comes back as the input array itself, not a copy, and that a caller intending to mutate should copy. That gives a future second caller the warning without charging today's caller for a copy it doesn't need. The contrast with `simplify_curve`'s boolean-mask path is a fair catch — that one never aliased.

**Dropping the `len(x_data) > _DEFAULT_MAX_SMOOTH_POINTS` guard was intentional**, not an oversight. The guard duplicated the `n <= count` check now inside `resample_curve`, and keeping both would have meant two places to update if the budget rule ever changes. `test_resample_curve_returns_short_curves_untouched` pins the no-op path.

CI is green on 4babbce — all 16 checks, including `claude-review`, `Python 3.9/3.10/3.11/3.12`, `lint`, `commit-lint` and CodeQL. `uv run pytest` green at 399 passed; `ruff check` clean on the changed files. The one remaining over-length line in `rdp_utils.py` is the pre-existing `_perpendicular_distance` signature at 98 chars, untouched by this branch.

---

jooyoungseo:
The `lowess=True` gap is real and I've fixed it in 9e5cf15 rather than deferring it — because measuring it turned up one correction to the review's framing that changes the call.

**It was not merely "RDP would've mishandled it too."** On a skewed sample (`x = sorted(exponential(2.0, 80))`), the max-over-min x-step across the emitted points:

| extraction | x-gap ratio |
|---|---|
| RDP (before this PR) | 86.7× |
| index resampling (this PR, 88e3a5b) | **318.8×** |
| even-x interpolation (9e5cf15) | **1.00×** |

So index sampling made that path measurably *worse* than what it replaced — a regression this PR introduced, not a pre-existing gap it left alone. That puts it in scope.

**The fix.** `resample_curve` now places its samples on an even x grid and interpolates y whenever x increases monotonically, which every seaborn fit does. Index sampling stays as the fallback for a curve that doubles back, where there is no usable x ordering.

Interpolating is safe here, and exactly so rather than approximately: matplotlib renders the curve as straight segments between its vertices, which is the same path the interpolation walks. Verified against the rendered SVG — the emitted `svg_x`/`svg_y` sit on the drawn `<path>` to within **1e-6 px** (max over all 30 points).

**Every fit type now paces evenly**, where before only the two uniform-grid cases did:

| fit | x-gap ratio |
|---|---|
| `regplot` (order=1) | 1.00× |
| `regplot(order=2)` | 1.00× |
| `regplot(lowess=True)` | 1.00× |
| `histplot(kde=True)` | 1.00× |

The docstring's uniform-grid caveat is gone, since the guarantee is now unconditional for monotonic x.

**Re-verified the rendering claims after the change**, since interpolation moves `svg_x`/`svg_y`:

- Static chart, MAIDR never activated: **0 differing pixels** out of 490,000 versus the pre-PR RDP output.
- Index sampling vs interpolation, static: **0 differing pixels**.
- Mid-navigation: 323 pixels — the highlight dot, as before.

Two new tests cover it: a clustered source curve (90 dense vertices then 10 sparse ones, source ratio >50×) must come out under 2×, and a circle — where x doubles back — must still thin to exactly 30 points via the index path.

`uv run pytest` green at 401 passed; `ruff check` clean on the changed files. Browser navigation re-checked and unchanged for the default regplot.

---

jooyoungseo:
The coverage gap was real and is closed in 58fcbf8.

**Confirmed it first.** Reverting the interpolation to index sampling and re-running the suite left every extractor test green — `test_resample_curve_evens_out_a_clustered_source_curve` caught it, but only by handing a synthetic curve straight to `resample_curve`. Both end-to-end fixtures fit on a `linspace` grid, so nothing exercised `SmoothPlot` on a clustered source. Exactly as described.

**Added a `lowess=True` fixture** over an exponential sample (`sorted(rng.exponential(2.0, 80))`), whose fitted vertices bunch at one end, and extended the spacing and span tests to run across it — so the `reg` / `kde` / `lowess` triple now covers both grid shapes end to end. A guard test asserts the fixture's own source gaps exceed 50×, so it can't quietly stop being clustered and start proving nothing.

Re-ran the revert experiment against the new tests:

```
FAILED test_smooth_points_are_evenly_spaced[lowess]
FAILED test_resample_curve_evens_out_a_clustered_source_curve
2 failed, 16 passed
```

The end-to-end case now fails without the fix, which it did not before.

`statsmodels>=0.14.4` is in the `visualization` extra and CI installs with `--all-extras`, so the fixture resolves in all four Python jobs.

**Strictly-increasing vs monotonic** — fair, and the fallback comment was the misleading part: it said a curve with "no usable x ordering" falls through, which isn't true of a right-to-left curve. Both the comment and the docstring now say plainly that a decreasing or self-crossing curve takes the index path, and why left-to-right is the only order worth special-casing: it is the one every fit produces, so handling the others would mean a case per shape for no caller.

**Two minor points skipped**, with reasoning:

- *Test module location* — `resample_curve`'s unit tests sit beside the extractor tests they motivate, and there is no `tests/util/test_rdp_utils.py` to join. Creating one to hold half of a single feature's tests would split the story for no gain; worth reconciling if a dedicated module ever appears for `simplify_curve`.
- *`CLAUDE.md` mention* — its architecture notes describe packages and the schema contract, not individual utility functions, and `simplify_curve` isn't listed either. Adding one would set an inconsistent precedent; the module docstring is the right home and already covers the choice.

`uv run pytest` green at 404 passed; `ruff check` clean on the changed files.

---

jooyoungseo:
The log-axis point is right and is fixed in b2636ff. The rounding nit I checked and it doesn't hold — details below.

**Log axis — my "exactly" was overstated.** Measured the real quantity: transform both the source vertices and the interpolated points into display space, then take each emitted point's distance from the chord matplotlib actually draws.

| axis | max deviation from the drawn chords |
|---|---|
| linear | 7.3e-12 px (floating-point noise) |
| log | **0.107 px** |

So exact under an affine map, and about a tenth of a pixel once a log scale bends each segment. The docstring now says which axis earns the exact landing and quotes the log figure, since the number is what shows it stays well under the highlight ring's radius. Nothing to fix in the code — sub-pixel and nothing in the codebase log-scales these axes, as you note.

**The round-half-to-even nit doesn't apply**, and I'd rather not leave it recorded as a known soft spot. It is a universal guarantee, not just a practical one, because the step is *strictly* greater than 1:

`|rint(x) - x| <= 0.5` holds under half-to-even as much as any other tie rule — at an exact `.5` the result is one of the two neighbours, so the deviation is exactly `0.5`, never more. For consecutive samples `a < b`:

```
rint(b) - rint(a)  >=  (b - a) - |rint(b) - b| - |rint(a) - a|  >=  step - 1  >  0
```

Two integers differing by more than 0 differ by at least 1. Ties are the worst case the bound already covers, so they cannot produce a collision. And `step > 1` is strict, not borderline: `n > count` gives `n - 1 >= count`, so `step >= count/(count - 1)`, which exceeds 1 for every `count >= 2`.

Verified alongside the algebra:

- `np.rint([0.5, 1.5, 2.5, 3.5])` → `[0, 2, 2, 4]`, max deviation exactly `0.5` — the bound survives half-to-even.
- Adversarial sweep over `count` in 2..400 plus 1000 and 4096, each against the tightest `n` values (`count+1`, `count+2`, `2*count`, …): **0 collisions**. Tightest margin exercised was `step - 1 = 2.4e-4`, still nine orders of magnitude above `linspace`'s floating-point error.
- Filtered to only those cases where a sample lands within 1e-12 of an exact `.5` boundary — the scenario the nit describes: **0 collisions**.

`uv run pytest` green at 404 passed; `ruff check` clean.

---

jooyoungseo:
Both handled in 2440bf8 — the second one deliberately as a documented limitation rather than a behaviour change, for the reason you asked about.

**Aliasing contract now pinned.** `test_resample_curve_returns_short_curves_untouched` asserted `np.array_equal`, which would have stayed green through a refactor that started copying. It now asserts `kept is points`, so the no-copy promise is actually tested rather than merely documented.

**Log-axis pacing — measured, and it's as you describe.** A `regplot` over `geomspace(1, 1000, 60)` with `ax.set_xscale("log")`:

| axis | data-space x-gap ratio | svg/display-space gap ratio |
|---|---|---|
| linear | 1.00× | 1.00× |
| log | 1.00× | **101.78×** |

So the pacing problem this PR fixes does reappear on screen for a log x-axis, even though the announced x values step uniformly.

**Left as-is, documented.** This isn't a case where one answer is plainly right, which is why it wants a maintainer call rather than my picking quietly:

- *Even in data x* (current): the announced values step uniformly, so a screen reader reads a regular numeric progression. But the sweep does not track the picture.
- *Even in display x*: the sweep matches what a sighted user sees advancing left to right — which is this PR's stated rationale — but the announced numbers then step irregularly (1, 1.3, 1.7, 2.2, …).

The second is arguably truer to MAIDR's purpose, but it trades away the uniform numeric readout, and it would change behaviour for a configuration nothing reaches by default: seaborn's `regplot`/`histplot` never set a nonlinear scale, so it takes an explicit `ax.set_xscale("log")` on the caller's part. Changing it on my own judgement inside a fix scoped to the collapse bug seemed like the wrong place to decide it.

`resample_curve`'s docstring now records the limitation, the measured spread, and that the choice is about what the sweep is meant to represent. Happy to switch to display-space spacing in this PR or a follow-up if you'd prefer it — it's a small change, `np.linspace` over the transformed x then inverse-transform.

**The inline comment length** in `regplot.py` — leaving as you suggest. It's four lines and all of it is the non-obvious "why", which is what CLAUDE.md asks comments to carry.

`uv run pytest` green at 404 passed; `ruff check` clean.

---

jooyoungseo:
Independent CI confirmation, since that was the one thing the sandbox couldn't reach — and it turns up a nuance worth stating precisely rather than just echoing "404 passed".

**All 16 checks are green on 2440bf8**, including `claude-review`. Pulling the pytest summary line straight from the job logs:

| job | summary |
|---|---|
| Python 3.12 | `404 passed, 24 warnings in 13.13s` |
| Python 3.9 | `400 passed, 4 skipped, 21 warnings in 12.94s` |

So 404 collected in both — but on 3.9 four of them **skip**, which the PR description's flat "404 passed" glossed over. The four are pre-existing matplotlib-version gates, not anything from this branch:

```python
_needs_orientation_kwarg = pytest.mark.skipif(
    "orientation" not in inspect.signature(Axes.violinplot).parameters, ...
)
```

`Axes.boxplot`/`Axes.violinplot` only grew the `orientation` parameter in matplotlib 3.10, and the resolver picks an older matplotlib for Python 3.9. `test_boxplot_orientation.py` and `test_violinplot_orientation.py` own those gates and neither is touched here. None of the new smooth-layer tests are gated — all 18 run on every version in the matrix, including the `lowess` fixture, since `statsmodels` comes in via `--all-extras`.

No code change from this round. The other two points I'm taking as written:

- **Always-on `np.column_stack`** — agreed, not worth reintroducing the guard; that would put the budget condition in two places to keep in sync.
- **Docstring length** — leaving it, per your read and the previous round's. It's all *why*-content, and the log-axis limitation in particular is the kind of thing that costs someone an afternoon if it isn't written down. Easy to trim later if a maintainer wants it tighter.

---

jooyoungseo:
Maintainer call came back: **pace by the screen**. Implemented in 0d54330, so the log-axis limitation is now a fixed behaviour rather than a documented caveat.

**What changed.** Thinning now happens in *scale space* — the axis's own scale transform, log10 for a log axis and the identity for a linear one — and maps back to data coordinates afterwards. Scale space is what the display differs from by an affine map, so even steps there are even steps on screen.

Scale space rather than reading `transData` on purpose: `transData` folds in the figure layout, which isn't settled during extraction (`data_to_svg_coords` calls `tight_layout` itself). The scale transform is layout-independent and gives the same answer.

| axis | on-screen gap ratio before | after |
|---|---|---|
| linear | 1.00× | 1.00× |
| log | **101.78×** | **1.00×** |

**The linear path is provably untouched.** A linear axis's scale transform is the exact identity, verified by round-trip, so this is a no-op there. Hashing the full emitted point list for the default `regplot` and the `histplot(kde=True)` fixture before and after gives the same digest — `87bc3065…` both ways.

**It also fixed the other log-axis defect for free.** Interpolating in scale space lands on the chord matplotlib actually renders, so the highlight deviation I measured two rounds ago is gone:

| axis | on-the-line deviation before | after |
|---|---|---|
| linear | 1.1e-13 px | 1.1e-13 px |
| log | 0.107 px | **1.1e-13 px** |

**The trade, stated plainly:** on a log axis the announced x values now step unevenly (data-space ratio 788×) in exchange for an even sweep. That is the intended direction — the sonified sweep tracking the picture is the point — and it is what the new test asserts in both directions, so it cannot pass by the two spacings happening to coincide.

**Guard for degenerate input.** `to_scaled_coords` returns `None` when a scale can't represent the data faithfully — matplotlib clips a non-positive value on a log axis to a sentinel rather than to infinity, so the check is a round-trip, not a finiteness test. The caller then thins in data coordinates as before.

New coverage: a `log_x_regplot_figure` fixture, `test_log_axis_paces_by_the_screen_not_the_data`, and the log axis added to the parametrized spacing and span tests — which now assert on `svg_x` rather than data x, since screen distance is what the contract is about.

`uv run pytest` green at **407 passed**. `ruff check` clean on all four changed files; the 7 pre-existing errors elsewhere in `maidr/` are untouched, confirmed by re-running against a stash of this branch. Browser navigation re-checked on the default regplot and unchanged.

---

jooyoungseo:
All three landed in eb2ddf6. The first two were real, and the third was a fair correction to my own reasoning.

**1. Unconditional round-trip — the drift is real, guard added.** Measured it rather than assuming: a log axis round-trips `x = uniform(0.001, 5000)` with a max relative drift of **4.99e-16**, not bit-exact. So a below-budget curve was coming out fractionally off the line matplotlib drew, where it used to pass through untouched. `_thin_to_even_steps` now returns early when the curve already fits the budget.

Worth noting this guard is *not* the one earlier rounds agreed to drop. That one was redundant with `resample_curve`'s internal short-circuit and cost nothing. This one prevents a transform round trip that didn't exist until the scale-space change, so the trade genuinely reversed.

Pinned by `test_curve_within_budget_is_passed_through_untouched`, which asserts `np.array_equal` against the source vertices. Removing the guard fails it:

```
FAILED test_curve_within_budget_is_passed_through_untouched
1 failed, 23 passed
```

It uses a 12-point `lowess` fit, since `lowess` returns y-hat at the observed x values and is the only way to get a seaborn fit under the 30-point budget.

**2. Fallback path now covered.** Reachable exactly as you suggested — a log y-axis whose fitted line dips below zero. Two tests: one guards that `to_scaled_coords` really does return `None` there (otherwise the second proves nothing), and one asserts the fallback degrades gracefully — a full 30 points, spanning the fit, evenly spaced in data space rather than collapsing or raising.

**3. The docstring reason was weak — corrected.** The layout claim was literally true at that point (`tight_layout` lives inside `data_to_svg_coords`, which runs after `_thin_to_even_steps`, and `grep` confirms that's its only call site), but you're right that it isn't the reason and would send a reader chasing a call-ordering explanation — or "fixing" it by reordering. The docstring now leads with the actual argument: the affine remainder cannot change distance ratios, so ratios in scale space are ratios on screen, and the result therefore holds under any layout rather than depending on when layout settles.

**On running the suite** — noted, and the earlier CI-log check applies: `Python 3.12` reported `404 passed` and `Python 3.9` `400 passed, 4 skipped` on 2440bf8, the 4 skips being pre-existing matplotlib-version gates on the boxplot/violinplot orientation tests. Locally this branch is now at **410 passed**; CI on eb2ddf6 will confirm independently and I'm watching it.

`ruff check` clean on all four changed files.

---

jooyoungseo:
Three of the five acted on in 7ba2e87. The PolyCollection question was the valuable one — that path is real and was untested.

**PolyCollection KDE — reachable, and it was the one blind spot.** `maidr/patch/kdeplot.py` registers the boundary of each `PolyCollection` as its own smooth layer, so `sns.kdeplot(fill=True)` goes straight through `SmoothPlot` with `is_polycollection=True`. Checked what it actually does:

```
smooth layers registered: 1
  layer[0] polycollection=True  points=30  x strictly increasing=False
```

The outline is a **closed loop** — out along the curve, back along the baseline — so x doubles back and the even-x path can't apply. It falls through to index sampling, which is the right handling for a loop, and still yields the full 30 points. So the behaviour was already correct, but nothing proved it: this is the only shipped caller that exercises the doubles-back branch, and `test_resample_curve_falls_back_to_index_sampling_when_x_doubles_back` only covered it with a synthetic circle. Now pinned end to end, with fixture guards asserting the poly path is genuinely hit and the outline genuinely doubles back.

Not reachable from `histplot_kde_figure`, to answer the "or is this moot" — that one has no fill, so it registers a `Line2D` with monotonic x.

**`atol` — you were right that it's what gates the fallback, and the default was too loose.** numpy's default `atol=1e-8` swamps `rtol` for data near zero, so a value mangled from `1e-12` to `5e-12` — five times wrong — still compared as faithful:

```
allclose(rtol=1e-9), default atol : True
allclose(rtol=1e-9, atol=0)       : False
```

Now `atol=0`, so the guard is purely relative and matches what the docstring claims. Safe for the cases in play: a linear axis round-trips exactly, and a log axis with positive data lands within ~5e-16 relative, nine orders inside `rtol`.

**`ruff format`** — applied to the two files this branch rewrote plus the test module; `ruff format --diff` now reports all three clean. I left the one remaining hit alone, the pre-existing `_perpendicular_distance` signature in `rdp_utils.py`, since reformatting a function this branch never touched would put unrelated churn in the diff.

**Two skipped:**

- *Transforms fetched twice* — agreed, negligible at 30–100 points, and threading a transform pair through the call would cost more clarity than it saves.
- *A "this is the degraded path" comment* — the branch already reads `# The scale cannot represent this data, leaving data coordinates as the only ordering to thin along`, which says both that it's constrained and why. A second line would restate it.

`uv run pytest` green at **411 passed**; `ruff check` and `ruff format --diff` clean on the changed files.

---

jooyoungseo:
Item 2 was a genuine error on my part and is fixed; item 1 deserves a straight answer about how the scope grew.

**The description contradicted the code — corrected.** It claimed the `len(x_data) > _DEFAULT_MAX_SMOOTH_POINTS` guard was dropped. That was true when written, but eb2ddf6 re-added it to stop the scale round trip perturbing curves that need no thinning, and I updated the commit message without updating the body. The description now describes the guard that actually ships, along with the log-axis behaviour and the current test count.

**On scope.** Fair to raise, and the description now carries the chain explicitly so it isn't archaeology. The short version is that each step past the minimal fix repaired a defect the previous step introduced, rather than being generality bolted on:

1. RDP → even resampling — the reported bug.
2. Even by index → even along x — because step 1 made `lowess=True` on skewed x **worse than what it replaced** (319× vs RDP's 87×). A regression this PR introduced, so not optional.
3. Even along x → even as drawn — because data-space evenness stops being on-screen evenness on a log axis (102× spread). **This one was a maintainer decision**, not my own call: I measured it, laid out the trade against uniform announced x values, said explicitly I didn't think it was mine to pick, and the answer came back to pace by the screen. That's why it's here rather than in a follow-up.
4. Fallbacks and the passthrough guard — step 3 has to answer for a scale that can't represent the line, a curve with no x ordering, and a curve short enough that the round trip would only perturb it.

Steps 3 and 4 are the surface worth the sanity check, and splitting them out remains viable — 1 and 2 stand alone and fix #318 — at the cost of shipping a known-uneven sweep on log axes in the meantime. I've flagged the option to the maintainer; happy to carve it up if that's the preference.

**Item 3, `rdp_utils.py` hosting `resample_curve`** — agreed the name has stopped matching the contents, and `curve_thinning.py` would be more discoverable than a docstring caveat. Not doing it here, partly because a rename touches `violin_kde_plot.py` and has a mirror in r-maidr's `R/rdp_utils.R` that would drift, and partly because adding a module rename to a PR already being questioned for scope would be the wrong instinct. Good follow-up.

**Item 4** — agreed on where the line sits. `symlog` and a custom `FuncScale` go through the same round-trip check and would be caught or passed on their own merits, but nothing in the codebase reaches them, so I'd rather not add tests asserting behaviour no caller depends on.

Thanks for tracing the `LogTransform` clip semantics independently — that check is exactly the one I'd want a second pair of eyes on, since a finiteness test looks right and silently isn't.

---

jooyoungseo:
Maintainer call on the scope question: **split it.** Done — this PR is now scoped to the reported bug, and the screen-pacing work moved to #320.

**What stayed here** (9 commits, `3f8e506`…`2440bf8`):

1. RDP → even resampling — the fix for #318.
2. Even by vertex index → even along x — because step 1 alone made `lowess=True` on skewed x worse than what it replaced (319× vs 87×). A regression this PR introduced, so it isn't separable from the fix.

**What moved to #320** (3 commits, `0d54330`…`7ba2e87`): scale-space thinning so a log axis paces by drawn distance, plus the fallbacks and budget guard that decision requires — `to_scaled_coords` / `from_scaled_coords`, `_thin_to_even_steps`, the unmappable-scale and closed-curve paths. #320 is stacked on this branch, so its diff shows only that work; it retargets to `main` once this merges.

The branch was force-pushed back to `2440bf8`. Review comments above that discuss `to_scaled_coords` or `_thin_to_even_steps` now apply to #320 rather than here — the analysis in them still stands, and every point raised was addressed on those commits before the split.

`resample_curve`'s docstring on this branch keeps the log-axis limitation recorded, so the gap is written down rather than silently shipped while #320 is pending.

Re-verified after the split: `uv run pytest` green at **404 passed** on this branch, `ruff check` clean, and the scale-space symbols are entirely absent from it (`grep` for `to_scaled_coords` / `_thin_to_even_steps` under `maidr/` returns nothing). #320 is green at 411.

---

jooyoungseo:
The mismatch was real when this review ran, and is already resolved — it read the pre-split description against the post-split code.

The timing: this `claude-review` job started at **17:47:19Z**, and the description was rewritten for the reduced scope at **17:51:05Z** (the PR's `updated_at`). The force-push that removed steps 3–4 landed just before the job started, so it saw new code against an old body — exactly the contradiction it describes.

The current body no longer contains any of it. There is no `svg_utils.py` entry in *Changes Made*, no `to_scaled_coords` / `from_scaled_coords` / `_thin_to_even_steps`, no log-axis row in the spacing table, and no "pace by the screen" framing. What it says instead is that screen-space pacing was split out to #320, and that `resample_curve`'s docstring records the log-axis limitation until that lands. So the description and the docstring now agree rather than contradict.

Everything the review verified against the shipped code stands, and matches what this branch is meant to be:

- `resample_curve` grids on data-space x only — correct here, and the docstring's "known limitation" is the honest statement of it, not an oversight.
- No test exercises a log-scaled axis — correct, and deliberate: that test lives in #320 alongside the code it covers.
- The unconditional `resample_curve(...)` call is right for this branch. #320 re-adds a length guard, but only because thinning in scale space introduces a round trip that isn't bit-exact; there's no such round trip here, so no guard is warranted.
- `ViolinKdePlot` untouched, `simplify_curve`'s only remaining caller — confirmed.

On the return-by-reference edge: agreed it's sharp, and it's why `test_resample_curve_returns_short_curves_untouched` asserts identity rather than equality, so a future `.copy()` "for safety" would fail loudly instead of silently changing the contract.

No code or description changes from this round. If a fresh review pass is wanted against the current body, a re-run would now see them consistent.

---

PR #320, opened 2026-08-08
fix(smooth): pace a smooth layer by drawn distance, not data distance

jooyoungseo:
## Description

Follow-up to #319, split out of it so that PR stayed scoped to the reported bug. #319 has since merged, and this is rebased onto `main`.

#319 made a `smooth` layer's points evenly spaced **along x**. That is evenly spaced **on screen** only while the axis is linear. A log x-axis stretches the same points into a roughly **102:1** spread across the plot, so auto-play sweeps the picture unevenly even though the announced x values step uniformly — the pacing problem #319 fixes, reappearing for a non-default axis scale.

A `smooth` trace is meant to let a blind reader follow the same trend a sighted reader sees moving left to right, so drawn distance is what has to stay even. This PR makes that the behaviour.

## Related Issues

Follows #318 / #319.

## Changes Made

- `maidr/util/svg_utils.py` — added `to_scaled_coords()` / `from_scaled_coords()`, mapping between data coordinates and the axes' scale space. What remains between scale space and the display is an affine map, so ratios of distances there are the ratios on screen; because an affine map cannot change those ratios, the answer holds under any figure layout. `to_scaled_coords` returns `None` when a scale cannot represent the data, guarded three ways — see below.
- `maidr/core/plot/regplot.py` — `SmoothPlot._thin_to_even_steps()` thins in scale space and maps back. It returns the vertices untouched when the curve already fits the budget, since the scale round trip is not bit-exact and would otherwise shift a short fit off the drawn line. It thins in data space when the scale cannot represent the line.
- `maidr/util/rdp_utils.py` — `resample_curve`'s docstring drops the log-axis limitation #319 documented, now that the caller resolves it, and no longer reaches into a specific caller's internals.
- `tests/core/plot/test_smooth_sampling.py` — grown from 15 tests to **37**.

### Three guards, one failure mode each

A scale can refuse data in more than one way, and no single check catches them all:

| guard | catches | pinned by |
|---|---|---|
| finiteness | a scale answering with `inf`/`NaN` — `nonpositive='mask'`, and `logit` at its upper bound | `test_a_masking_scale_is_caught_by_the_finiteness_guard`, `test_a_scale_rejecting_on_an_upper_bound_is_caught` |
| `_clip_sentinel` probe | a clipping scale's sentinel, which reads as an ordinary coordinate | `test_a_fit_touching_zero_on_a_log_axis_is_not_mapped` |
| round trip (`rtol=1e-9, atol=0`) | a value that maps somewhere it cannot come back from | `test_unmappable_scale_is_detected` |

None of the three is redundant, and two of them catch cases the round trip actively reports as fine: a log scale's sentinel inverts back to zero, and `logit`'s infinity inverts back to `1.0`.

`atol=0` is deliberate: numpy's default `atol=1e-8` swamps the relative tolerance near zero, and a value mangled from `1e-12` to `5e-12` compared as faithful under it. It does not cost a legitimate zero, which round-trips exactly on every scale that represents one (`test_a_representable_zero_is_not_rejected`).

The sentinel probe is scale-agnostic — it looks for the coordinate several unrepresentable inputs collapse onto, rather than naming log or its constant, so `symlog` and `asinh` (where zero is representable) are left alone.

## Verification

Spacing spread (max step ÷ min step) across the smooth layer:

| axis | on-screen, before | on-screen, after |
|---|---|---|
| linear | 1.00× | 1.00× |
| log | **101.78×** | **1.00×** |
| `symlog` | — | **1.00×** |
| `asinh` | — | **1.00×** |

**The linear path is provably untouched.** A linear axis's scale transform is the exact identity, verified by round trip. Hashing the full emitted point list for the default `regplot` and the `histplot(kde=True)` fixture before and after gives the same digest, `87bc3065…` both ways.

**It also removes a second log-axis defect.** Interpolating in scale space lands on the chord matplotlib actually renders:

| axis | on-the-line deviation, before | after |
|---|---|---|
| linear | 1.1e-13 px | 1.1e-13 px |
| log | 0.107 px | **1.1e-13 px** |

**A fit touching exactly zero was dragging the trace off the canvas.** Raised in review as a low-severity edge case; measuring it showed otherwise. `x = np.linspace(0, 10, 50)` is an ordinary fixture, and seaborn evaluates the fit from `x.min()`, so `grid[0]` is `0.0` exactly. A log scale parks it on the sentinel and inverting underflows back to zero, so it round-tripped intact and passed as mappable:

```
to_scaled_coords -> MAPPED (slipped through)
  scale-space x[0] = -1000.0   vs x[1] = -0.996
  points off-canvas (svg_x < 0): 29 of 30
```

**29 of 30 navigable points landed outside the plot.** With the sentinel probe, that drops to 1 — the zero itself, which no log axis can place.

**The trade, stated plainly.** On a log axis the announced x values now step unevenly (788× spread) in exchange for an even sweep. That is the intended direction, decided by the maintainer, and `test_log_axis_paces_by_the_screen_not_the_data` asserts both directions so it cannot pass by the two spacings coinciding.

**Round-trip drift is why the budget guard exists.** A log axis round-trips `uniform(0.001, 5000)` with up to **4.99e-16** relative drift, so a below-budget curve would come out fractionally off the line it used to pass through untouched.

**Edge paths covered.** A log y-axis whose fit dips below zero exercises the unmappable-scale fallback, on both axes, with guard tests confirming `to_scaled_coords` really does return `None` there. `sns.kdeplot(fill=True)` registers a `PolyCollection` outline — a closed loop out along the curve and back along the baseline — the only shipped caller whose x doubles back. On a log axis it reaches **both** fallbacks at once, since a KDE's support runs past its data into values a log cannot map; data far enough from zero keeps the outline mappable, so the index fallback also runs inside scale space.

### Tests are not vacuous

Each behaviour was re-run against the code it replaced: the `lowess` case fails against index sampling, the passthrough test fails without the budget guard, and the zero-touching test fails without the sentinel probe.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Checklist

- [x] I have tested the changes locally, and all tests related to this pull request pass — `uv run pytest` green at **423 passed**. `ruff check` and `ruff format --diff` clean on the changed files.
- [ ] I have updated the documentation, if applicable — no docs change needed; the example pages regenerate from the same source.

## Additional Notes

Scope is py-maidr only, as with #319.

The follow-up this PR's notes used to describe — `rdp_utils.py` hosting `resample_curve`, which has nothing to do with Ramer–Douglas–Peucker — is now open as **#321**, stacked on this branch. It moves the function out rather than renaming the module, because r-maidr's `R/rdp_utils.R` holds the RDP trio and nothing else, so renaming would have widened the gap between the two ports instead of closing it.

---

### DIFF
--- a/maidr/core/plot/regplot.py
+++ b/maidr/core/plot/regplot.py
@@ -8,7 +8,11 @@ from maidr.core.enum.plot_type import PlotType
 from maidr.core.enum.maidr_key import MaidrKey
 from maidr.util.rdp_utils import resample_curve
 from maidr.util.regression_line_utils import find_regression_line
-from maidr.util.svg_utils import data_to_svg_coords
+from maidr.util.svg_utils import (
+    data_to_svg_coords,
+    from_scaled_coords,
+    to_scaled_coords,
+)
 
 #: Default maximum number of output points per smooth curve.
 _DEFAULT_MAX_SMOOTH_POINTS = 30
@@ -49,6 +53,47 @@ class SmoothPlot(MaidrPlot):
             return [f"g[id='{self._smooth_gid}'] path"]
         return ["g[id^='maidr-'] path"]
 
+    def _thin_to_even_steps(
+        self, x_data: np.ndarray, y_data: np.ndarray
+    ) -> tuple[np.ndarray, np.ndarray]:
+        """
+        Thin the curve to points that step evenly across the plot.
+
+        The points are navigated and auto-played one at a time at a fixed rate,
+        so the sweep only tracks the drawn line while the steps look even on
+        screen.  Shape-based simplification gives the opposite: it collapses a
+        straight fit to its two endpoints and spends fewer steps on the flat
+        stretches of a curved one than on the bends.
+
+        Thinning happens in scale space, so a log axis paces by the distance it
+        draws rather than by raw data distance — on a linear axis the two are
+        the same thing and this costs nothing.
+
+        Parameters
+        ----------
+        x_data, y_data : np.ndarray
+            Vertices of the fitted line, in data coordinates.
+
+        Returns
+        -------
+        tuple of np.ndarray
+            The retained vertices, in data coordinates.
+        """
+        scaled = to_scaled_coords(self.ax, x_data, y_data)
+        if scaled is None:
+            # The scale cannot represent this data, leaving data coordinates as
+            # the only ordering to thin along.
+            xy = resample_curve(
+                np.column_stack([x_data, y_data]),
+                target=_DEFAULT_MAX_SMOOTH_POINTS,
+            )
+            return xy[:, 0], xy[:, 1]
+
+        xy = resample_curve(
+            np.column_stack(scaled), target=_DEFAULT_MAX_SMOOTH_POINTS
+        )
+        return from_scaled_coords(self.ax, xy[:, 0], xy[:, 1])
+
     def _extract_plot_data(self) -> list:
         """
         Extract XY data from the regression line for serialization, including SVG coordinates.
@@ -70,16 +115,7 @@ class SmoothPlot(MaidrPlot):
         xydata = np.asarray(regression_line.get_xydata())
         x_data, y_data = xydata[:, 0], xydata[:, 1]
 
-        # Thin the curve before computing SVG coordinates.  The points are
-        # navigated and auto-played one at a time at a fixed rate, so they have
-        # to stay evenly spaced along the line: shape-based simplification
-        # would collapse a straight fit to its two endpoints and stretch the
-        # flat stretches of a curved one over fewer steps than the bends.
-        xy = resample_curve(
-            np.column_stack([x_data, y_data]),
-            target=_DEFAULT_MAX_SMOOTH_POINTS,
-        )
-        x_data, y_data = xy[:, 0], xy[:, 1]
+        x_data, y_data = self._thin_to_even_steps(x_data, y_data)
 
         x_svg, y_svg = data_to_svg_coords(self.ax, x_data, y_data)
         return [
--- a/maidr/core/plot/regplot.py
+++ b/maidr/core/plot/regplot.py
@@ -95,9 +95,7 @@ class SmoothPlot(MaidrPlot):
             )
             return xy[:, 0], xy[:, 1]
 
-        xy = resample_curve(
-            np.column_stack(scaled), target=_DEFAULT_MAX_SMOOTH_POINTS
-        )
+        xy = resample_curve(np.column_stack(scaled), target=_DEFAULT_MAX_SMOOTH_POINTS)
         return from_scaled_coords(self.ax, xy[:, 0], xy[:, 1])
 
     def _extract_plot_data(self) -> list:
--- a/maidr/core/plot/regplot.py
+++ b/maidr/core/plot/regplot.py
@@ -79,6 +79,12 @@ class SmoothPlot(MaidrPlot):
         tuple of np.ndarray
             The retained vertices, in data coordinates.
         """
+        if len(x_data) <= _DEFAULT_MAX_SMOOTH_POINTS:
+            # Nothing to thin, so hand the vertices straight back.  Mapping
+            # them into scale space and out again would return them shifted by
+            # the round trip's last bit or two, for no gain.
+            return x_data, y_data
+
         scaled = to_scaled_coords(self.ax, x_data, y_data)
         if scaled is None:
             # The scale cannot represent this data, leaving data coordinates as
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -159,18 +159,17 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     curve that doubles back, or one running right to left — falls back to
     index sampling.
 
-    Spacing is even in whatever space the caller hands over.  ``SmoothPlot``
-    passes scale space rather than data space, so the evenness lands on screen
-    where the sweep is felt; see
-    :meth:`~maidr.core.plot.regplot.SmoothPlot._thin_to_even_steps`.
+    Spacing is even in whatever space the caller hands over, so a caller that
+    wants evenness on screen rather than in raw data has to convert before
+    calling and back afterwards.
 
     Interpolated points sit on the drawn line, which matters because they are
     what the highlight ring lands on.  Matplotlib renders the curve as straight
     segments between its vertices, so interpolating in scale space — an affine
     map away from the display — walks exactly that path.  That exactness is
-    the caller's to earn, though: interpolating a nonlinear axis in data space
-    instead leaves the points a fraction of a pixel off the segment, which is
-    what ``SmoothPlot`` settles for when a scale cannot represent the curve.
+    the caller's to earn, though: handing over data coordinates for a
+    nonlinear axis instead leaves the points a fraction of a pixel off the
+    segment.
 
     Parameters
     ----------
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -166,8 +166,11 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
 
     Interpolated points sit on the drawn line, which matters because they are
     what the highlight ring lands on.  Matplotlib renders the curve as straight
-    segments between its vertices, and scale space differs from the display by
-    an affine map, so the interpolation walks exactly that path.
+    segments between its vertices, so interpolating in scale space — an affine
+    map away from the display — walks exactly that path.  That exactness is
+    the caller's to earn, though: interpolating a nonlinear axis in data space
+    instead leaves the points a fraction of a pixel off the segment, which is
+    what ``SmoothPlot`` settles for when a scale cannot represent the curve.
 
     Parameters
     ----------
--- a/maidr/util/rdp_utils.py
+++ b/maidr/util/rdp_utils.py
@@ -159,22 +159,15 @@ def resample_curve(points: np.ndarray, target: int) -> np.ndarray:
     curve that doubles back, or one running right to left — falls back to
     index sampling.
 
+    Spacing is even in whatever space the caller hands over.  ``SmoothPlot``
+    passes scale space rather than data space, so the evenness lands on screen
+    where the sweep is felt; see
+    :meth:`~maidr.core.plot.regplot.SmoothPlot._thin_to_even_steps`.
+
     Interpolated points sit on the drawn line, which matters because they are
     what the highlight ring lands on.  Matplotlib renders the curve as straight
-    segments between its vertices, so on a linear axis — where data and display
-    space differ by an affine map — the interpolation walks exactly that path.
-    A log axis bends each segment between the vertices, leaving a gap of about
-    a tenth of a pixel at this point count; still far under the ring's radius,
-    just not the exact landing a linear axis gives.
-
-    Known limitation: the grid is even in *data* x, which is even on screen
-    only while the axis is linear.  A log x-axis stretches the same points
-    into a roughly 100:1 spread across the plot, so auto-play sweeps the
-    picture unevenly even though the announced x values step uniformly.
-    Nothing here sets a nonlinear scale — it takes an explicit
-    ``ax.set_xscale`` — and which of the two should stay uniform is a call
-    about what the sweep represents, so this stays as-is until a caller needs
-    otherwise.
+    segments between its vertices, and scale space differs from the display by
+    an affine map, so the interpolation walks exactly that path.
 
     Parameters
     ----------
--- a/maidr/util/svg_utils.py
+++ b/maidr/util/svg_utils.py
@@ -67,6 +67,11 @@ def to_scaled_coords(
     x_data, y_data : np.ndarray
         Data coordinates to map.
 
+    One unmappable point rejects the whole batch rather than being dropped:
+    a curve mapped point by point would mix two coordinate spaces, leaving
+    the spacing it is thinned by — and any interpolation across it —
+    meaningless.
+
     Returns
     -------
     tuple of np.ndarray, or None
--- a/maidr/util/svg_utils.py
+++ b/maidr/util/svg_utils.py
@@ -46,6 +46,32 @@ def unique_lines_by_xy(lines: List[Line2D]) -> List[Line2D]:
     return unique_lines
 
 
+def _clip_sentinel(transform) -> Optional[float]:
+    """
+    Return the coordinate a scale parks unrepresentable values on, if it clips.
+
+    A clipping scale sends every value it cannot represent to one sentinel, so
+    it stops being injective there.  Probing with several values that only such
+    a scale would reject finds that sentinel without naming a scale or its
+    constant: a log axis collapses them onto a single coordinate, while linear,
+    symlog and asinh keep them apart and report nothing to watch for.
+
+    Parameters
+    ----------
+    transform : matplotlib.transforms.Transform
+        A scale transform, as returned by ``Axis.get_transform()``.
+
+    Returns
+    -------
+    float or None
+        The sentinel coordinate, or ``None`` for a scale that never clips.
+    """
+    probe = transform.transform(np.array([-1.0, -2.0, -4.0]))
+    if probe[0] == probe[1] == probe[2]:
+        return float(probe[0])
+    return None
+
+
 def to_scaled_coords(
     ax: Axes, x_data: np.ndarray, y_data: np.ndarray
 ) -> Optional[Tuple[np.ndarray, np.ndarray]]:
@@ -87,10 +113,19 @@ def to_scaled_coords(
     if not (np.all(np.isfinite(x_scaled)) and np.all(np.isfinite(y_scaled))):
         return None
 
-    # The check above catches only a scale that answers with infinity or NaN.
-    # A log scale clips instead, mapping an unrepresentable value to a sentinel
-    # that reads as a perfectly ordinary coordinate, so the round trip is the
-    # only thing standing between that value and a curve thinned against it.
+    # Zero is the one clipped value the round trip below cannot see: a log
+    # scale parks it on the sentinel, and inverting that underflows back to
+    # zero, so the value appears to have survived the journey intact.  Ask the
+    # scale where it parks what it rejects, and refuse anything sitting there.
+    for transform, scaled in ((x_transform, x_scaled), (y_transform, y_scaled)):
+        sentinel = _clip_sentinel(transform)
+        if sentinel is not None and np.any(scaled == sentinel):
+            return None
+
+    # The finiteness check above catches only a scale that answers with
+    # infinity or NaN.  A log scale clips instead, mapping an unrepresentable
+    # value to a coordinate that reads as perfectly ordinary, so the round trip
+    # is what stands between that value and a curve thinned against it.
     if not (
         np.allclose(
             x_transform.inverted().transform(x_scaled), x_data, rtol=1e-9, atol=0
--- a/maidr/util/svg_utils.py
+++ b/maidr/util/svg_utils.py
@@ -62,6 +62,13 @@ def _clip_sentinel(transform) -> Optional[float]:
     finiteness check in :func:`to_scaled_coords` instead.  The two guards cover
     one mode each; neither is redundant.
 
+    This reads matplotlib's behaviour rather than a promise it makes, so a
+    future release that stops collapsing rejected values would go undetected
+    here and the sentinel would simply never be found again.  Nothing raises
+    in that case — a curve touching zero on a log axis would quietly go back
+    to being thinned against the sentinel — so a caller left wondering why
+    such a curve drifts off the canvas should suspect this probe first.
+
     Parameters
     ----------
     transform : matplotlib.transforms.Transform
--- a/maidr/util/svg_utils.py
+++ b/maidr/util/svg_utils.py
@@ -60,17 +60,17 @@ def to_scaled_coords(
     axes' position and limits to settle.  A linear axis passes through
     unchanged; a log axis becomes its logarithm.
 
+    One unmappable point rejects the whole batch rather than being dropped:
+    a curve mapped point by point would mix two coordinate spaces, leaving
+    the spacing it is thinned by — and any interpolation across it —
+    meaningless.
+
     Parameters
     ----------
     ax : Axes
         The axes whose x and y scales apply.
     x_data, y_data : np.ndarray
         Data coordinates to map.
-
-    One unmappable point rejects the whole batch rather than being dropped:
-    a curve mapped point by point would mix two coordinate spaces, leaving
-    the spacing it is thinned by — and any interpolation across it —
-    meaningless.
 
     Returns
     -------
--- a/maidr/util/svg_utils.py
+++ b/maidr/util/svg_utils.py
@@ -52,11 +52,13 @@ def to_scaled_coords(
     """
     Map data coordinates into the axes' scale space.
 
-    Distances there are proportional to distances on screen, because what
-    remains between scale space and the display is an affine map.  A linear
-    axis passes through unchanged; a log axis becomes its logarithm.  Working
-    in scale space rather than reading ``transData`` keeps the result
-    independent of figure layout, which is not settled during extraction.
+    What remains between scale space and the display is an affine map, so
+    ratios of distances there are the ratios on screen — which is what makes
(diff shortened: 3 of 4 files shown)


jooyoungseo:
One fix in f875d68; the rest I checked and they don't need code.

**The exactness claim was overstated — corrected.** You're right that "the interpolation walks exactly that path" reads as unconditional when it's contingent on the caller passing scale space. The fallback drops back to data space and the sub-pixel deviation returns there. The docstring now says the exactness is the caller's to earn and names what `SmoothPlot` settles for when a scale can't represent the curve.

**Masked arrays — checked, and the concern doesn't materialize.** Rather than guard speculatively I ran it: with `nonpositive="mask"` matplotlib returns a plain `ndarray` carrying `nan`, not a `MaskedArray`.

```
type: ndarray   is_masked: False
value: [nan  0.  1.  2.]
np.all(np.isfinite(out)) -> False   (plain bool)
```

So the existing finiteness check catches it and degrades to the data-space fallback exactly as intended — no special handling needed, and no silent-wrong-answer path.

**Log-x + linear-y is already covered**, so that gap isn't open: `log_x_regplot_figure` is precisely that combination — `geomspace(1, 1000, 60)` on `set_xscale("log")` with an untouched linear y — and it drives `test_log_axis_paces_by_the_screen_not_the_data` plus both parametrized spacing and span tests. The log-y-crosses-zero fixture is the separate unmappable case.

**Two flagged and deliberately not changed:**

- *All-or-nothing fallback across both axes.* Per-axis handling is possible, but the case it would rescue is a plot that's already degenerate — a log axis holding values matplotlib can't draw. Adding a branch so that a half-broken plot keeps screen-accurate pacing on its good axis buys precision in a situation where the picture itself is wrong. The all-or-nothing rule is easier to reason about, and the fallback is still a correct thinning, just paced by data distance.
- *Custom scales without a working `inverted()`.* Agreed this would raise rather than degrade, and agreed it isn't worth a speculative guard. Every built-in scale implements it, and nothing that reaches `SmoothPlot` goes near a custom one.

`uv run pytest` green at **411 passed**; `ruff check` clean.

On retargeting: this stays based on `claude/smooth-line-navigation-issue-vqse7v` until #319 merges, at which point I'll retarget it to `main`.

---

jooyoungseo:
Both actionable items in 4ce7079.

**The log-x gap is closed, and I checked it rather than trusting the symmetry.** You were right to flag it as an assumption: `to_scaled_coords` tests the two axes in separate expressions, so the x branch wasn't actually exercised by anything. Verified a `regplot` over `linspace(-3, 12, 60)` with `set_xscale("log")`:

```
fitted x range: -3.0 -> 12.0   vertices: 100
to_scaled_coords -> None
emitted 30 points; spans -3.000..12.000; data-gap ratio=1.000
```

Same detection, same graceful degrade. Both fallback tests are now parametrized over `log_y` and `log_x` rather than a second copy, and the fixture guard asserts the fit really does reach below zero on whichever axis is under test.

**All-or-nothing is now documented, with your reasoning rather than mine.** I'd justified it in a comment as "the plot is already degenerate", which is true but beside the point. Yours is the real one, and it's what the docstring now says: mapping point by point would mix two coordinate spaces, leaving the spacing the curve is thinned by — and any interpolation across it — meaningless. That's a property of the algorithm rather than a judgement about how rare the input is, so it belongs in the docstring.

**Two left alone**, both consistent with earlier rounds: re-resolving the transforms in `from_scaled_coords` (cheap, and threading a transform pair through would complicate the pair's API for nothing), and the docstring density (dense but carrying non-obvious invariants).

Thanks for working the `transData` decomposition through independently — `transScale → transLimits → transAxes` with only the first nonlinear is exactly the argument the docstring rests on, and it's worth having a second derivation of it on the record.

`uv run pytest` green at **413 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
The docstring nit was a real convention break of mine — fixed in ae527ca.

I'd added that paragraph in the previous round and dropped it straight after the parameter entries, which puts free text between `Parameters` and `Returns` with no header. NumPy style has nowhere for prose there, so numpydoc would fold it into the parameter list or lose it. It now sits in the extended summary alongside the affine-map rationale, which is where the rest of the reasoning already lived.

Checked the other three docstrings this PR adds or rewrites rather than assuming the mistake was isolated — `from_scaled_coords`, `resample_curve` and `_thin_to_even_steps` all read extended summary → `Parameters` → `Returns`, with no free text after a section header. So this was the only one.

**Tolerance doing double duty** — agreed, and worth having on the record. `rtol=1e-9, atol=0` is currently detecting two different things: a scale that clipped, and ordinary float drift. They're separable in principle, but only log and linear reach this code today and their round-trip error is many orders inside the threshold, so splitting the check now would be inventing structure for a scale that doesn't exist yet. Noted as an assumption rather than changed.

**`_is_polycollection` in the test** — leaving it. There's no public accessor, and the alternative is asserting on the emitted selector string, which would swap a private attribute for a coupling to the selector's exact format. As a fixture guard rather than the assertion under test, reaching for the attribute is the more honest of the two.

**On CI confirming the numbers** — it has. All four Python jobs (3.9–3.12) plus lint and commit-lint are green on this branch, and the earlier log check on #319 showed the matrix reporting the counts directly rather than my repeating them. Locally this branch is now at **413 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
Both notes acted on in 10ddc6e — and the composition gap turned out more interesting than "should compose fine".

**Filled KDE on a log axis hits *both* fallbacks, stacked.** I built the case expecting it to be a formality and it isn't: a KDE's support runs past its data, so even strongly positive input (`lognormal(3.0, 0.45, 500)`) produces an outline reaching **x = -2.52**. A log x-axis can't map that, so `to_scaled_coords` returns `None` and thinning drops to data space — *and then* the outline is a closed loop, so it drops again to vertex index.

```
polycollection: True   source vertices: 403   source x min: -2.52334
to_scaled_coords -> None (fallback)
emitted: 30   x strictly increasing: False
```

Still a full 30-point trace. Each fallback was covered alone; nothing showed they compose, and this is the configuration that reaches both at once. Now pinned, with guards on all three preconditions so it can't quietly stop testing what it claims.

**The `isfinite` comment now says why it can't stand alone.** You're right that the old wording explained the round trip without warning off the "simplification" — someone could read `isfinite` as already covering it. It now states the split directly: the finiteness check catches only a scale answering with infinity or NaN, while a log scale *clips*, returning a sentinel that reads as an ordinary coordinate, so the round trip is the only thing between that value and a curve thinned against it.

**Two acknowledged, not changed:**

- *`rtol=1e-9, atol=0` doing double duty* — same as the last round: separable in principle, but only log and linear reach this code and their round-trip error sits many orders inside the threshold.
- *The 87-char line* — `ruff format` chose that layout and `ruff format --diff` is clean, so reflowing it by hand would just be undone. Worth noting it's in `regplot.py`'s `_thin_to_even_steps` rather than `svg_utils.py`. If the constant or an import ever grows, the formatter will rewrap it.

**On the base branch** — deliberate and flagged in the description. This must not land on `main` before #319; I'll retarget it once #319 merges.

`uv run pytest` green at **414 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
The layering smell was real — fixed in cb9b642. The other two I checked and they don't need changes.

**A shared utility shouldn't document one caller's internals.** `resample_curve` named `SmoothPlot` and linked its *private* `_thin_to_even_steps`, which points the dependency the wrong way and would rot the moment that method is renamed. It now states the requirement rather than the caller: spacing is even in whatever space it's handed, so a caller wanting evenness on screen converts on the way in and back out. Same for the second reference further down. The module docstring still names both callers, which is the file's existing style and useful as a map — the difference is that's a class-level pointer, not a reach into a private method.

**The `Tuple` vs `tuple` split isn't something this PR introduced.** Both files were already committed to their own style, and I followed each:

- `svg_utils.py:9` — pre-existing `data_to_svg_coords` returns `Tuple[...]`. My `to_scaled_coords` / `from_scaled_coords` match it.
- `regplot.py:101` — pre-existing `_extract_plot_data(self) -> list:` uses the builtin. My `_thin_to_even_steps` matches it.

So each file is internally consistent, and switching either would make the new code clash with its own neighbours. A PEP 585 sweep would be the right way to reconcile them, repo-wide rather than in the two files a bugfix happens to touch.

**`get_transform()` being lower-level** — agreed, and it's a deliberate trade. `transData` is the friendlier API but the wrong one here: it folds in the affine part, which is exactly what has to be excluded for the ratio argument to hold, and it depends on layout that isn't settled during extraction. The scale-only transform is the piece the reasoning needs.

One note on process: running `ruff format` over `rdp_utils.py` also rewrapped the pre-existing `_perpendicular_distance` signature, which I'd said in an earlier round I'd leave alone. I reverted that hunk — the diff here stays docstring-only (6 insertions, 7 deletions) rather than picking up churn in a function this branch never touches.

`uv run pytest` green at **414 passed**; `ruff check` clean.

---

jooyoungseo:
The `symlog` gap is now closed in 9e2c0fa — and I'm reversing an earlier answer of mine to do it.

An earlier round raised `symlog`/`FuncScale` and I declined, on the grounds that I'd rather not pin behaviour no caller depends on. That reasoning was weaker than it looked: the generality isn't incidental here, it's a property the code deliberately claims by reading `axis.get_transform()` instead of testing for a log axis. Your framing — confidence in the generality the fix asserts — is the better one, and the difference now is that I measured it rather than assumed it:

| x scale | mappable | on-screen gap ratio | data-space gap ratio |
|---|---|---|---|
| linear | yes | 1.00× | 1.00× |
| `symlog` | yes | **1.00×** | 959.77× |
| `asinh` | yes | **1.00×** | 769.43× |

Both hold. The test drives data spanning several decades either side of zero, which is the shape these scales exist for, and asserts the trade in both directions so it can't pass by the two spacings coinciding.

Worth noting these exercise a path the log tests don't: `symlog` and `asinh` *can* map zero and negatives, so unlike the log fixtures they stay on the scale-space route rather than falling back. That means the suite now covers scale-space thinning on a nonlinear axis without relying on log, alongside the fallback when a scale can't cope.

Left `FuncScale` out. An arbitrary user-supplied transform has no behaviour to pin beyond "whatever the user's function does", so a test would assert my fixture rather than the code.

**On the stacked base** — confirmed, and it's the intended arrangement. CI here validates against `claude/smooth-line-navigation-issue-vqse7v`, so this isn't independently mergeable; I'll retarget to `main` once #319 lands.

`uv run pytest` green at **416 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
#319 has merged, so this is rebased onto `main` and retargeted. And the exact-zero nit turned out to be **considerably worse than low severity** — thank you for catching it.

**Rebased.** #319 was squash-merged, so `main` gained one commit sharing no history with the nine on this branch; GitHub showed 18 commits and `mergeable_state: dirty`. Rebased the ten follow-up commits onto the squashed `main` with `--onto`, applied cleanly, base is now `main`. The diff is back to 4 files.

**The exact-zero case is reachable and severe, not exotic.** You framed it as needing a fit to land on exactly `0.0`, which sounds unlikely — but `x = np.linspace(0, 10, 50)` is an entirely ordinary fixture, and seaborn evaluates the fit on a grid starting at `x.min()`, so `grid[0]` is `0.0` exactly. Measured on that:

```
fit x[0] exactly zero: True
to_scaled_coords -> MAPPED (slipped through)
  scale-space x[0] = -1000.0   vs x[1] = -0.996
  points landing off-canvas (svg_x < 0): 29 of 30
```

Thinning built its grid from the sentinel to the visible range, so **29 of 30 navigable points ended up outside the plot** — for a blind user, a trace that is almost entirely nowhere. Fixed rather than commented.

**Why not the `!= 0` guard you suggested.** That would have broken the `symlog`/`asinh` support added a commit earlier, where zero *is* representable and legitimately maps. The fix instead asks the scale where it parks what it rejects: probe with several values only a clipping scale would refuse, and if they collapse onto one coordinate, that's the sentinel — then refuse any point sitting on it.

```
linear   sentinel=      None    exact-zero caught: False
log      sentinel=   -1000.0    exact-zero caught: True
symlog   sentinel=      None    exact-zero caught: False
asinh    sentinel=      None    exact-zero caught: False
```

No scale named, no constant hardcoded, and the scales where zero is fine are left alone. After the fix, off-canvas points drop from 29 to **1** — the zero itself, which no log axis can place.

Pinned by `test_a_fit_touching_zero_on_a_log_axis_is_not_mapped`, which fails without the guard.

**The two test-coverage suggestions** — a direct `from_scaled_coords(ax, *to_scaled_coords(ax, x, y))` round-trip test is a fair ask and I'll add it if you'd like, though the pair is now exercised through five scales and both fallbacks. The "linear output byte-identical to the pre-change path" check can't become a checked-in test: it compared *old code against new*, and there's no old code left to compare against once this merges. The durable version of that guarantee is the identity round trip, which `test_curve_within_budget_is_passed_through_untouched` already asserts with `np.array_equal`.

`uv run pytest` green at **417 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
Checked the `nonpositive='mask'` case rather than guarding speculatively, and it already fails safe — so 77e2c9f documents why instead of adding a check.

The premise doesn't hold: matplotlib returns a **plain ndarray carrying NaN**, not a masked array.

```
nonpositive='clip'
  probe -> [-1000. -1000. -1000.]   type=ndarray  masked=False
  _clip_sentinel -> -1000.0
  to_scaled_coords(curve touching 0) -> None (fails safe)
  to_scaled_coords(curve with -5)     -> None (fails safe)

nonpositive='mask'
  probe -> [nan nan nan]            type=ndarray  masked=False
  _clip_sentinel -> None
  to_scaled_coords(curve touching 0) -> None (fails safe)
  to_scaled_coords(curve with -5)     -> None (fails safe)
```

So the chained equality behaves ordinarily: `nan == nan` is `False`, `_clip_sentinel` reports no sentinel, and that's the *correct* answer — a masking scale doesn't have one. Its rejected values are then caught by the finiteness check, which a clipping scale's sentinel slips past. **The two guards cover one mode each, and neither is redundant** — which is the part worth writing down, since the failure mode here would be someone deleting one as duplicated effort. That's now in `_clip_sentinel`'s docstring.

I'd have written the caveat you suggested if the behaviour were undefined, but "unspecified" would have been the wrong thing to record when both paths demonstrably return `None`.

**The `Tuple` / `tuple` nit** — same conclusion as an earlier round, and you reached it too: each file matches its own pre-existing convention, so changing either would make the new code clash with its neighbours. A PEP 585 sweep is the right vehicle, repo-wide rather than in two files a bugfix touched.

Worth noting the sentinel probe you're reviewing landed *because* of the previous round's exact-zero catch, and this round's question tested it from the other side — the clip/mask split is exactly where a probe like that could have been fragile. Good angle to check.

`uv run pytest` green at **417 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
The "best-effort heuristic" note is in 7139023, and your remark about relying on a stated test count turned out to be well aimed — the description was stale.

**The probe's failure mode is now written down**, which was the part worth having. You're right that it reads matplotlib's behaviour rather than a promise, and right that the breakage would be silent: if a release stopped collapsing rejected values, the sentinel would simply never be found again, nothing would raise, and a curve touching zero on a log axis would quietly go back to drifting off the canvas. The docstring now says exactly that, and points whoever meets that symptom at the probe first — that's the debugging thread five years from now.

**The stated count was out of date.** The body said 411 passed and "six added tests"; the branch has since taken the rebase onto the squashed `main` plus the symlog/asinh coverage and the zero-touching fix, and is at **417 passed** with the test module at **31 tests**. Refreshed the description accordingly — it also still claimed the base was #319's branch and said nothing about the sentinel guard. CI on this branch has run the matrix independently rather than my repeating the number.

**The three-guard split is now in the description as a table**, since between your round and the last one it's become the part of this PR most likely to be misread as redundancy:

| guard | catches |
|---|---|
| finiteness | a scale answering `inf`/`NaN` — including `nonpositive='mask'` |
| `_clip_sentinel` probe | a clipping scale's sentinel, which reads as an ordinary coordinate |
| round trip (`atol=0`) | a value that maps somewhere it cannot come back from |

**`_clip_sentinel` recomputing per call** — agreed it's negligible, and left alone. It runs twice per extraction on a 3-element array; caching it would mean holding state keyed on a transform object whose identity isn't guaranteed stable, which trades a real correctness question for an imperceptible saving.

**`rdp_utils.py` naming** — agreed, still future work, and the r-maidr mirror is the reason it isn't bundled here.

`uv run pytest` green at **417 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
Two of the three acted on in 74d4dde. The coverage gap was constructible, and the docstring point has now come up enough times that I took it.

**The missing combination exists and is now pinned.** You were right that it's a distinct path rather than a formality. The reason no test hit it: a filled KDE's support runs past its data, so on a log axis it *usually* trips the unmappable fallback first — which is what the existing test covers. Data far enough from zero keeps the whole outline positive:

```
lognormal(3, .45)     outline x min=     -2.5233   scale-space mappable=False
normal(1000, 5)       outline x min=    982.3451   scale-space mappable=True
```

So `normal(1000, 5)` maps cleanly, thinning stays in scale space, and the closed loop falls back to vertex index *there* — index sampling on scale-space coordinates, which nothing previously exercised.

**Docstring trimmed.** This is the third round to raise the density, and looking back I could see the mechanism: I added a paragraph to `_clip_sentinel` per review — mask handling, then guard complementarity, then the silent-failure warning — each individually justified, cumulatively bloated. Compressed roughly in half, keeping the two facts that are load-bearing rather than explanatory: that the guards cover a mode each (otherwise someone deletes one as redundant, which an earlier round specifically worried about), and that the probe fails quietly if matplotlib ever stops clipping (the debugging thread). The fuller rationale lives in the PR description, as you suggest.

**`ax.transScale` — checked, and it isn't a drop-in.**

```
transScale: TransformWrapper  input_dims: 2
per-axis  : LogTransform      input_dims: 1
```

It's the 2-D blend, while `_clip_sentinel` probes each axis *separately* — with the blended form I'd have to build 2-D probes and then work out which axis clipped, to answer a question that's per-axis by nature. It would also flatten the current behaviour of rejecting the batch when *either* axis fails into something harder to state. The per-axis transforms are what the guard logic actually needs, so this reads as the more idiomatic choice here despite `transScale` being the shorter call.

`uv run pytest` green at **418 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
Your instinct about the trap was right and the mechanism turned out to be different — `logit` is already safe, but for a reason nothing had written down. Pinned in 0c1cca9.

**`logit` doesn't clip to a sentinel at all.** It signals with infinity, at both ends:

```
log      lower-probe sentinel=     -1000.0   upper-probe sentinel=        None
logit    lower-probe sentinel=        None   upper-probe sentinel=        None
symlog   lower-probe sentinel=        None   upper-probe sentinel=        None
```

So an upper-side probe wouldn't have found anything to find. What makes your point worth acting on is the half you got exactly right — the trap class is real and the round trip cannot see it:

```
logit, x touching exactly 1.0:
  forward:    [-0.602  0.  0.954  inf]
  round trip: [0.2 0.5 0.9 1.0]   matches input: True
  _clip_sentinel: None
  to_scaled_coords -> None (caught)
```

Inverting `inf` returns `1.0` unchanged, so the round-trip check reports the value as perfectly intact — precisely the shape of the zero-on-log bug. It's the **finiteness** guard that catches it, not the sentinel probe. That the three guards happened to cover this between them was luck no test was protecting, so there's now one asserting each step: no sentinel, non-finite forward, a round trip that lies, and `to_scaled_coords` returning `None` regardless.

**No upper-side probe added.** It would guard a scale that doesn't exist — every matplotlib scale that rejects on an upper bound uses infinity — and I'd be adding a speculative branch of exactly the kind I've declined elsewhere in this PR.

**One clause added to the docstring**, since you're right that the one-sidedness undercut the "names no scale" framing: it now says the probe looks downward on purpose, because a lower bound is the one a scale parks values *below*, while an upper bound is signalled with infinity. Kept to a sentence — I trimmed this docstring last round precisely because it had been growing one paragraph per review, and I'd rather not restart that.

**The linear-path hash as an automated test** — third time this has come up, so to be concrete about why it can't become one: the comparison was *old code vs new code*, and once this merges there's no old code to compare against. A fixed-seed golden-value test would pin seaborn's fit output rather than this code's behaviour, and would break on any seaborn change. The durable form of that invariant is the identity assertion in `test_curve_within_budget_is_passed_through_untouched`, plus the linear fixtures running through the same parametrized spacing tests.

`uv run pytest` green at **419 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
Both concrete items in 991937a.

**The missing annotation was a real gap against CLAUDE.md** — `_clip_sentinel(transform)` was the one function on this branch without one, and you're right that the rest is fully annotated. Now `transform: Transform`, imported from `matplotlib.transforms` rather than quoted, so it type-checks rather than just reads well.

**The `atol=0` / representable-zero question — checked, and it holds.** This was the one point that could have been a live bug rather than a note, since a false rejection would silently drop a legitimate curve to the fallback:

| scale | zero round-trips to | passes `rtol=1e-9, atol=0` | result |
|---|---|---|---|
| linear | `0.0` exactly | yes | mapped |
| `symlog` | `0.0` exactly | yes | mapped |
| `asinh` | `0.0` exactly | yes | mapped |

Zero survives exactly on all three, so `\|0 - 0\| <= rtol * \|0\| = 0` is satisfied and nothing is falsely rejected. Worth noting `symlog` is *not* bit-exact across the whole array — its largest drift was `1.78e-15` at `x = -10`, a relative `1.78e-16` — so the relative tolerance is doing real work there even though zero itself is untouched. That's now pinned by `test_a_representable_zero_is_not_rejected`, parametrized across the three scales, so the interaction can't regress unnoticed.

**The two you flagged without asking for action, acknowledged rather than changed:**

- *`_clip_sentinel`'s residual risk is the one guard no test can protect* — correct, and unfixable by definition: it would need a matplotlib that doesn't exist yet. That's why the failure mode is spelled out in the docstring instead, pointing at the probe as the first suspect if a curve ever drifts off-canvas again.
- *Tests importing a private helper* — a fair coupling to note. The alternative is asserting only the observable `to_scaled_coords(...) is None`, which is what the older fallback tests do; the two that reach for `_clip_sentinel` do so to record *which* guard is load-bearing for that case, which is the fact worth freezing given how easily the three read as redundant. Worth revisiting if the internal is ever refactored.

`uv run pytest` green at **422 passed**; `ruff check` and `ruff format --diff` clean.

---

jooyoungseo:
You caught something real, though not the bug you suspected — the behaviour is safe, but nothing in the suite was holding it there. Pinned in d2c7409.

**Your reading of the code was right; the premise about matplotlib isn't.** `np.all` on a masked array would indeed treat masked entries as vacuously true and wave the curve through. It never gets one:

```
returned type : ndarray
is masked     : False
values        : [ nan -inf   0.   1.]
isfinite      : [False False  True  True]   type: ndarray
np.all(...)   : False                       type: bool
to_scaled_coords -> None (caught)
```

`nonpositive='mask'` hands back a **plain ndarray** carrying NaN and `-inf` — note it uses both, NaN for the negative and `-inf` for the zero — so `np.isfinite` returns a plain bool array and the reduction a plain `bool`. The guard behaves exactly as the table claims.

**But "verified by hand and written down as settled" was the actual gap.** I'd checked this twice across earlier rounds and reported it in comments both times, which left the claim resting on my say-so while every other guard had a pin test. That asymmetry is what you spotted, and it's the right thing to have flagged even though the code was fine. The test now asserts the load-bearing preconditions rather than just the outcome: that the array is unmasked, that the reduction returns a plain bool, that `_clip_sentinel` finds nothing for this mode, and that `to_scaled_coords` returns `None`. If matplotlib ever switches to returning a genuine masked array, the first two assertions fail loudly instead of the guard silently going vacuous.

**Custom scales that raise instead of returning NaN — declining, consistently with the rest of this PR.** Every built-in degrades gracefully (log clips or masks, logit and symlog/asinh signal with infinity or map cleanly, all verified here). Guarding would mean wrapping the probe in a bare `except`, which trades a loud failure for a silent one and hides genuine errors from a scale that's misbehaving. A custom scale that raises on input outside its domain is violating the contract every built-in keeps; failing loudly at that point seems more useful than quietly thinning in data space and leaving no trace of why.

`uv run pytest` green at **423 passed**; `ruff check` and `ruff format --diff` clean.

---

PR #322, opened 2026-08-09
feat(pie): support matplotlib, seaborn and plotly pie charts

jooyoungseo:
## Description

Adds pie chart support: `PlotType.PIE`, a `PiePlot` extractor, a `wrapt` patch on `Axes.pie`, and plotly `go.Pie`/donut support. `ax.pie(...)` and `go.Pie(...)` now emit a MAIDR pie layer with no user code change.

Companion to xability/maidr#767, which adds `TraceType.PIE` to the renderer. **That PR must ship first** — the JS `TraceFactory` throws on an unknown trace type, so a pie layer is only renderable once the JS release lands. This PR is correct and tested at the schema level today.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Checklist

The last box is deliberately unchecked: it depends on xability/maidr#767 merging and releasing.

# Pull Request

## Description

See above.

## Related Issues

None — implements pie support from scratch. Upstream companion: xability/maidr#767.

## Changes Made

**Core** — `PlotType.PIE`, `maidr/core/plot/pieplot.py`, `maidr/patch/pieplot.py`, registered in `MaidrPlotFactory`, `maidr/patch/__init__.py` and `FigureManager.PLOT_TYPE_PRIORITY`.

**The one subtle bit — where the values come from.** `Axes.pie` normalises its input: when `sum(x) > 1` it plots `x / sum(x)`, and each `Wedge` retains only `theta1`/`theta2`. Recovering magnitudes from the drawn geometry would report `0.3 / 0.5 / 0.2` for a caller who wrote `ax.pie([30, 50, 20])`, and lose float32 precision doing it. So values are taken from the patched call's arguments instead. The angular path survives only as a fallback for a directly-constructed `PiePlot`, where fractions are genuinely all matplotlib kept — documented in-method as a deliberate degradation, not a correct answer.

**Other decisions**

- The patch does not route through `patch/common.py`: `common` feeds the return value to `FigureManager.get_axes`, which handles `list` but not the `tuple` `Axes.pie` returns, and its `kwargs` are simultaneously the wrapped call's and the constructor's, leaving no room for extracted data. It mirrors `common`'s body instead (warning filter, `ContextManager` guard).
- Both `(wedges, texts)` and `(wedges, texts, autotexts)` returns are handled.
- `Axes.pie` sits behind `@_preprocess_data`, so `ax.pie("v", labels="l", data=df)` arrives as bare strings; a `_resolve` helper mirrors matplotlib's `_replacer`. Arguments are read positionally *or* by keyword, since matplotlib ≤3.9 allowed `labels` positionally and 3.10 made it keyword-only.
- Labels resolve as matplotlib does: `labels` argument → `Wedge.get_label()` → slice index. numpy/pandas scalars are unwrapped so a `labels=df["year"]` column does not make `json.dumps` raise.
- Nested pies (two `ax.pie` calls on one axes) use the wedges the patch handed over rather than scanning `ax.patches`, so each layer gets its own ring instead of both seeing all six wedges.
- `axes` emits exactly `{x, y}` per the canonical contract in `CLAUDE.md`, defaulting to `"Category"`/`"Value"` rather than the base class's `"X"`/`"Y"` — "X: Apples" would be announced against every slice.
- Highlighting needed no change to `patch/highlight.py`: `Wedge` subclasses `Patch`, which `tag_elements` already wraps.

**Plotly** — `maidr/plotly/pie.py` handles `go.Pie` and the donut variant (`hole` changes geometry, not data). plotly supplies `labels[]`/`values[]` directly, so there is no normalisation problem.

**Docs** — pie sections in `docs/examples.qmd` and `docs/examples-plotly.qmd`, `docs/llms.txt`, a runnable `example/pie/`, and the stale supported-types list in `CLAUDE.md` brought up to date.

n/a

## Checklist

## Additional Notes

`uv run pytest` — **660 passed**, 0 failed.

`ruff check` reports 52 findings, but that is the count on `main` too — verified by stashing this branch and re-running. No new lint findings were introduced; the existing ones are pre-existing `F401` re-export warnings across `__init__.py` files.

Tests cover the thing most likely to regress: that `ax.pie([30, 50, 20])` yields `30/50/20` and not matplotlib's normalised fractions. Also covered — the label fallback chain, `data=` column resolution, both return shapes, nested-pie layer separation, per-slice selectors in slice order, and `tests/core/test_axes_schema.py` compliance.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,7 +41,7 @@ The `maidr/patch/` modules use `wrapt` to intercept matplotlib/seaborn plot call
 
 ### Supported Plot Types
 
-Defined in `maidr/core/enum/plot_type.py`: BAR, BOX, COUNT, DODGED, HEAT, HIST, LINE, SCATTER, STACKED, STEP, SMOOTH, CANDLESTICK.
+Defined in `maidr/core/enum/plot_type.py`: BAR, BOX, COUNT, DODGED, HEAT, HIST, LINE, PIE, SCATTER, STACKED, STEP, SMOOTH, CANDLESTICK, VIOLIN_KDE, VIOLIN_BOX.
 
 Note that `maidr/plotly/` builds its MAIDR schema in Python and therefore needs its own
 handling per plot type, whereas `maidr/altair/` delegates entirely to the upstream
--- a/docs/examples-plotly.qmd
+++ b/docs/examples-plotly.qmd
@@ -139,6 +139,69 @@ fig = go.Figure(
 maidr.show(fig)  #<<
 ```
 
+### Pie Chart
+
+A pie is one flat row of slices, so navigation runs left and right across them
+and there is no second dimension to move through. Each slice is announced by
+its label, its value, and the share of the whole that value works out to. The
+percentage is derived from the values at render time rather than authored, so
+it can never disagree with the numbers it is derived from.
+
+Plotly draws a pie's wedges largest first — `sort` defaults to true — so the
+order of the `values` array is not the order of the slices. maidr reproduces
+that ordering, because the slice a user hears has to be the slice that
+highlights. Pass `sort=False` to keep the authored order in both.
+
+```{python}
+#| warning: false
+#| fig-alt: Pie chart showing the share of tips recorded on each day of the week
+
+import plotly.graph_objects as go
+import seaborn as sns
+
+import maidr  #<<
+
+# Load the tips dataset
+tips = sns.load_dataset("tips")
+
+# Count the tips recorded on each day of the week
+day_counts = tips["day"].value_counts()
+
+fig = go.Figure(
+    go.Pie( #<<
+        labels=day_counts.index.tolist(),
+        values=day_counts.values.tolist(),
+        # Keep the authored order rather than letting plotly re-sort it.
+        sort=False,
+    )
+)
+
+fig.update_layout(
+    title="Share of Tips by Day",
+    # Name the two dimensions of a slice, so it is read out as
+    # "Day: Sat, Number of tips: 87" rather than "Label: Sat, Value: 87".
+    xaxis=dict(title="Day"), #<<
+    yaxis=dict(title="Number of tips"), #<<
+)
+
+maidr.show(fig)  #<<
+```
+
+A donut is the same chart with `hole` set: it changes the wedge geometry and
+leaves the data behind it untouched, so it is emitted as a pie layer too.
+
+::: {.callout-note}
+Plotly names neither axis of a pie, so the layout's `xaxis`/`yaxis` titles are
+what tell maidr how to announce a slice; without them it falls back to "Label"
+and "Value".
+
+Plotly builds its own slice list before drawing, and maidr follows it exactly
+so that data index *k* and highlighted wedge *k* stay the same slice: entries
+plotly does not read as numbers draw no wedge, repeated labels merge into one
+wedge holding their sum, and a label listed in `layout.hiddenlabels` is
+dropped along with its wedge.
+:::
+
 ### Histogram
 
 ```{python}
--- a/docs/examples-plotly.qmd
+++ b/docs/examples-plotly.qmd
@@ -192,8 +192,14 @@ leaves the data behind it untouched, so it is emitted as a pie layer too.
 
 ::: {.callout-note}
 Plotly names neither axis of a pie, so the layout's `xaxis`/`yaxis` titles are
-what tell maidr how to announce a slice; without them it falls back to "Label"
-and "Value".
+what tell maidr how to announce a slice; without them it falls back to
+"Category" and "Value", the same pair the matplotlib side uses.
+
+Those titles are only borrowed when the figure holds nothing but pies. A
+cartesian trace that declares no axis pair of its own shares the same default
+`xaxis`/`yaxis`, and those titles describe *its* axes — so in a mixed figure
+the pie keeps the generic pair rather than announcing a bar's axis names
+against its slices.
 
 Plotly builds its own slice list before drawing, and maidr follows it exactly
 so that data index *k* and highlighted wedge *k* stay the same slice: entries
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -255,7 +255,9 @@ the slice values measure. Leave them unset and maidr falls back to "Category"
 and "Value", which at least read as English where "X" and "Y" would not.
 
 Negative sizes are rejected — matplotlib refuses to draw a wedge with no area,
-and so does maidr. Drawing options that only move the wedges around
+so `ax.pie()` raises before maidr sees the call at all; maidr repeats the check
+for anyone building a `PiePlot` directly. Drawing options that only move the
+wedges around
 (`startangle`, `counterclock`, `explode`, `shadow`) leave the slice order
 alone, so the data stays index-aligned with what is on screen.
 :::
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -43,7 +43,7 @@ maidr.set_backend(use_maidr=True)    # plt.show() renders accessible maidr outpu
 
 #### Unsupported Plot Types
 
-For plot types not yet supported by maidr (e.g. pie charts, contour plots, 3D plots), `plt.show()` will automatically fall back to displaying a static image and emit a warning. No code changes are needed — your plots will always be visible.
+For plot types not yet supported by maidr (e.g. contour plots, 3D plots), `plt.show()` will automatically fall back to displaying a static image and emit a warning. No code changes are needed — your plots will always be visible.
 
 ### Bar Plots
 
@@ -199,6 +199,66 @@ ax.yaxis.set_major_formatter("{x:.0f}")
 # Display the plot
 plt.show() #<<
 ```
+
+### Pie Chart
+
+A pie is one flat row of slices, so navigation runs left and right across them
+and there is no second dimension to move through. Each slice is announced by
+its label, its value, and the share of the whole that value works out to. The
+percentage is derived from the values at render time rather than authored, so
+it can never disagree with the numbers it is derived from.
+
+`ax.pie()` normalises what it is given — it draws `x / sum(x)`, and each wedge
+keeps only its angles — so the counts below survive as counts because maidr
+reads them off the call rather than off the drawn wedges.
+
+```{python}
+#| warning: false
+#| fig-alt: Pie chart showing the share of tips recorded on each day of the week
+
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+import maidr #<<
+
+# Load the tips dataset
+tips = sns.load_dataset("tips")
+
+# Count the tips recorded on each day of the week
+day_counts = tips["day"].value_counts()
+
+fig, ax = plt.subplots(figsize=(8, 8))
+
+# autopct draws the percentages onto the chart for sighted readers; maidr
+# derives its own from the values, so the two cannot drift apart.
+ax.pie( #<<
+    list(day_counts.values),
+    labels=list(day_counts.index),
+    autopct="%1.1f%%",
+    startangle=90,
+)
+
+ax.set_title("Share of Tips by Day")
+
+# Name the two dimensions of a slice, so it is read out as
+# "Day: Sat, Number of tips: 87" rather than "X: Sat, Y: 87".
+ax.set_xlabel("Day") #<<
+ax.set_ylabel("Number of tips") #<<
+
+plt.show() #<<
+```
+
+::: {.callout-note}
+A pie has no x or y scale, so its two axis labels name the *dimensions* of a
+slice instead: `xlabel` says what the slice labels mean and `ylabel` says what
+the slice values measure. Leave them unset and maidr falls back to "Category"
+and "Value", which at least read as English where "X" and "Y" would not.
+
+Negative sizes are rejected — matplotlib refuses to draw a wedge with no area,
+and so does maidr. Drawing options that only move the wedges around
+(`startangle`, `counterclock`, `explode`, `shadow`) leave the slice order
+alone, so the data stays index-aligned with what is on screen.
+:::
 
 ### Histogram
 
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -2,7 +2,7 @@
 
 > py-maidr is a Python library that makes matplotlib, seaborn, and plotly data visualizations accessible to blind and low-vision users through sonification, braille, text, and AI-powered conversational modalities. Developed by the (x)Ability Design Lab at the University of Illinois Urbana-Champaign.
 
-maidr (pronounced "mader") monkey-patches matplotlib, seaborn, and plotly at import time so users just `import maidr` and call `maidr.show(plot)` to generate interactive HTML with keyboard-navigable sonification, braille, text, and AI chat support. Supported plot types: bar, stacked bar, dodged bar, count, histogram, KDE, line, multi-line, step, box, violin, heatmap, scatter, smooth/regression, candlestick, multi-panel, and facet plots. Works in Jupyter, Colab, VS Code, Streamlit, Shiny, and Quarto.
+maidr (pronounced "mader") monkey-patches matplotlib, seaborn, and plotly at import time so users just `import maidr` and call `maidr.show(plot)` to generate interactive HTML with keyboard-navigable sonification, braille, text, and AI chat support. Supported plot types: bar, stacked bar, dodged bar, count, histogram, KDE, line, multi-line, step, pie, box, violin, heatmap, scatter, smooth/regression, candlestick, multi-panel, and facet plots. Works in Jupyter, Colab, VS Code, Streamlit, Shiny, and Quarto.
 
 ## Docs
 - [Documentation Home](https://py.maidr.ai/): Installation, quick start, and overview
--- a/example/pie/matplotlib/example_mpl_pie_plot.py
+++ b/example/pie/matplotlib/example_mpl_pie_plot.py
@@ -1,0 +1,45 @@
+"""Pie chart: parts of one whole, read out as slices.
+
+A pie is one flat row of slices, so navigation runs left and right across
+them and there is no second dimension to move through. Each slice is
+announced by its label, its value, and the share of the whole that value
+works out to — the percentage is derived from the values at render time
+rather than authored, so it can never disagree with them.
+
+``Axes.pie`` normalises what it is given: it draws ``x / sum(x)`` and each
+wedge keeps only its angles, so the counts below survive as counts only
+because maidr reads them off the call. Set ``xlabel`` and ``ylabel`` to name
+what a slice *is* and what it *measures*; otherwise a slice is announced with
+the generic "Category" and "Value".
+"""
+
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+import maidr  # noqa: F401
+
+# Load dataset
+tips = sns.load_dataset("tips")
+
+# Count the tips recorded on each day of the week.
+day_counts = tips["day"].value_counts()
+
+fig, ax = plt.subplots(figsize=(8, 8))
+
+# autopct draws the percentages onto the chart for sighted readers; maidr
+# derives its own from the values, so the two cannot drift apart.
+ax.pie(
+    list(day_counts.values),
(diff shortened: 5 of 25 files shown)


jooyoungseo:
Thanks — the subplot finding was real, and I reproduced it before touching anything. Pushed in 404f5f5.

**The bug, confirmed.** Two `go.Pie` traces in a 1×2 domain grid:

```
PIE grid:  [0][0] holds BOTH pie layers        ← wrong
BAR control: [0][0] and [0][1], one each       ← correct
```

Your diagnosis was exactly right, including the part about `_subplot_grid_position` having no domain-trace handling.

**The fix.** Row and column now come from the trace's own `domain` rectangle. The one thing worth calling out is that I did *not* give domain traces their own coordinate universe: `_subplot_domain_starts` collects fractional starts from **both** layout axis domains and trace domains into one ordered set. Separate universes would put a pie and a cartesian subplot both in column 0 — which is the mixed-grid case you asked about specifically.

`pie_position` is untouched, for the reason you identified: plotly draws every pie into one figure-level `pielayer` regardless of placement, so that index has to keep counting across the whole figure. Grid position and `pie_position` now derive from two different things, which is correct but non-obvious, so there's a comment at the call site and a test pinning it.

**Tests.** New `TestPlotlyPieSubplotGrid` — 1×2 and 2×2 grids, pie-vs-cartesian ordering in both positions, `pie_position` still numbering the pielayer 1-4 across those subplots, and a lone undomained pie staying at `[0][0]`. I checked the tests actually bite: reverting only `plotly_maidr.py` gives **4 failed, 2 passed** — the four grid tests fail while the two "must not regress" guards still pass.

**Axis-label defaults** — unified on matplotlib's `"Category"`/`"Value"`. Agreed the announcement should follow `PlotType.PIE`, not the library behind it; `"Label: Apples"` also reads as a tautology against a slice.

**Redundant negative check** — tidied. `_extract_values` now early-returns after checking the caller's magnitudes, and the angular fallback returns below with a one-liner noting matplotlib already rejected negatives.

**On the PR title** — you're right, and thank you for catching it. There is no seaborn-specific patch; seaborn has no pie function, so `ax.pie` funnels through the same matplotlib patch. I'll drop "seaborn" from the squash-merge title rather than imply dedicated support.

`uv run pytest -q` → **666 passed**, 0 failed. `ruff check` → 52 findings, identical to the count on `main` (verified by stashing); zero in any file this PR touches.

One case I did **not** handle, documented in `_trace_domain_start` rather than guessed at: a pie positioned via `domain={"row": 0, "column": 1}` against a `layout.grid` still falls back to cell (0,0). `make_subplots` always emits explicit x/y, so this only affects hand-written grid-referencing figures, and resolving `layout.grid` geometry is a separate feature rather than part of this bug.

---

jooyoungseo:
The `hiddenlabels` finding is real. Reproduced it, fixed in 58c6a62.

```
no hiddenlabels        : [('B', 2), ('null', 1)]
hiddenlabels=[None]    : [('B', 2), ('null', 1)]   ← wedge stayed visible
hiddenlabels=['null']  : [('B', 2)]
```

Your reasoning was exactly right: the wedge side normalises before comparing, the lookup side didn't, so `str(None)` = `"None"` never met `"null"`. And `None` is what an author hiding that slice would naturally write, since it's the label they passed.

Both sides now go through one `_wedge_label(label, index)` helper. After the fix, `hiddenlabels=[None]` and `hiddenlabels=["null"]` both hide it.

**On the empty-string half of that finding, I landed somewhere different and want to be explicit about it.** I left `""` unnormalised, deliberately. A `hiddenlabels` entry carries no position to substitute, and the empty label it would name has already become that entry's own index — so `""` names no wedge. Plotly can't match one either, for the same reason: its legend shows the index, so a legend click puts the *index* in `hiddenlabels`, never `""`. Normalising it would mean inventing a position the entry doesn't have. The helper takes `index: int | None` so the two call sites differ in exactly that one way, and there are now tests pinning both directions — `hiddenlabels=[""]` matches nothing, `hiddenlabels=["0"]` hides it.

Four new tests, including the `None` case you identified as the gap.

**Typing style** — fixed. `patch/pieplot.py` now uses plain `tuple`/`dict` like `patch/common.py`.

**`pie/calc.js` drift** — good call, and I've added the version note you suggested (verified against plotly.py 6.7.0), plus a line making the mitigation explicit: the tests pin one rule each, so a drift surfaces as a named failure rather than as silently shifted slices. That mattered more than usual here because the failure mode is positional — the first divergence lands every later slice on the wrong wedge, and the announcement would be confidently wrong rather than obviously broken.

**Negative-value guard reachability** — agreed, and thank you for stating it precisely; it's for the standalone class path, not the patch path. I've opened xability/maidr#771 to settle negative-value policy once across all three bindings, because they currently disagree: Python raises, r-maidr's ggplot2 path silently reports the absolute value via `ymax - ymin`, and the JS renderer accepts whatever arrives.

**Docstring length** — fair, and I'll take the note. The density is concentrated in the two places where the *why* is genuinely non-obvious (normalisation recovery, and the calc.js rules). I'd rather not thin those two specifically, but I agree they're a departure from the surrounding style.

`uv run pytest -q` → **670 passed**, 0 failed. `ruff check` → 52, unchanged from `main`; clean on every file this commit touches.

---

jooyoungseo:
Both selector-alignment questions resolved — and checking #1 turned up that **my previous `hiddenlabels` fix was wrong**. Correcting it here.

## Correction: 58c6a62 diverged from plotly, a7f098c and 076ad0a fix it

I read the actual `pie/calc.js` in the bundled `plotly.min.js` this time instead of reasoning about it. The hidden-marking line is:

```js
p = a[c], (p === void 0 || p === "") && (p = c), p = String(p);
...
k = i.indexOf(p) !== -1
```

`i` is `hiddenlabels` and `p` is the **already-stringified** label. `indexOf` is strict equality, so `[null].indexOf("null")` is `-1` — **plotly keeps drawing that wedge.** My earlier commit made py-maidr drop it, which would have left a drawn wedge with no data point behind it and shifted every later slice onto the wrong element. Precisely the failure this PR is supposed to prevent, introduced by a fix for it.

The rule is now what plotly actually applies: only string entries can match. `[5].indexOf("5")` is `-1` too, so a numeric entry hides nothing either — which the old code got wrong in the opposite direction. Six tests pin both directions.

So my previous reply was wrong on this point. The empty-string half of my earlier answer holds up for a different reason than I gave: `""` matches nothing because the label became `String(c)`, and plotly's `indexOf` sees the same mismatch.

## 1. Zero-valued slice — no bug, now pinned

The filter is `r.filter(function(L){ return L.v >= 0 })`. Plotly keeps zero-valued slices; `> 0` was the concern, `>= 0` is the code. Ours already matched exactly.

Added the regression tests you asked for either way: `values=[0, 5, 3]` keeps all three in order, and two entries cancelling to zero via the merge step (`labels=["A","A","B"], values=[-2,2,5]`) keeps the merged `("A", 0)`.

## 2. Pie inheriting a cartesian trace's axis titles — real, fixed

Reproduced:

```
PIE: {'x': 'Month', 'y': 'Revenue'}    ← the bar's titles
BAR: {'x': 'Month', 'y': 'Revenue'}
```

So a pie in that figure announced "Month is Apples, Revenue is 30". Titles are now borrowed only when every trace in the group is a pie. After the fix a lone pie still takes `xaxis_title="Fruit"`, while the mixed figure gives the pie `Category`/`Value` and leaves the bar's own titles alone. Three tests.

You called it "might be intentional/acceptable" — I don't think it is, for the same reason as the selector cases: it fails *confidently* rather than visibly. A vague label is recoverable; a wrong one isn't.

## 3. Docs sequencing

Agreed and worth flagging on the release, not just here. Tracked in #323, which covers re-vendoring the bundle after xability/maidr#767 ships; I've noted the Quarto docs-site ordering there too.

## Minor

- `_resolve`'s bare `except Exception` — narrowed as you suggested.
- Docstring density — noted; I'd rather not thin the two spots carrying genuinely non-obvious rationale, but I agree it's a departure.

`uv run pytest -q` → **677 passed**, 0 failed. `ruff check` → 52, unchanged from `main`; clean on every file touched.

---

jooyoungseo:
Three reviews this round; everything actionable is in c59af4c and ebe812c.

**The "Label" vs "Category" doc bug is real, and it was mine.** I changed the plotly default to `"Category"` to match matplotlib earlier in this PR and didn't update the callout that named the old value. Fixed, and the page now also states the condition under which the layout titles get borrowed at all — which the doc never mentioned even before the change.

**`_resolve`'s bare `except Exception`** — narrowed to `(KeyError, IndexError, TypeError)` in ebe812c. Two of you raised this independently; you're right that a wide net there hides a broken `data` object rather than a missing name.

**Length-mismatch fallback now warns.** Good catch that this is the dangerous kind of silent: reaching that branch through the patch would mean the wedges and the call's own `x` had come apart, and `0.3` where the caller wrote `30` is plausible enough to sail past review *and* past a user. It says so now.

**`.schema` on a pie with no drawn slice** — added, both all-zero and all-hidden. It doesn't raise; the layer reaches the wire with `data: []` and a well-formed `axes`. Worth having pinned, since a throw there takes down every other layer in the figure, not just the pie.

**Early-exit comment** — added, saying explicitly that it's a short circuit rather than a rule of its own, and why it agrees with the merge/filter path below.

## Not doing, with reasons

**A runtime `plotly.__version__` check.** I considered it and decided against: a warning on major-version drift fires for everyone on every new plotly regardless of whether the calc rules actually changed, and the one time it matters it says nothing about *which* rule moved. The tests already pin one rule each, so a real drift surfaces as a named failure in CI. What would genuinely close this is your other suggestion — cross-checking emitted order against a rendered `pielayer` — and that needs a browser harness this repo doesn't have. Tracked in #323 alongside the post-release verification.

**Trimming the docstrings.** Three reviewers have now flagged the density, so I'll take the signal, but I'd rather not thin the two blocks carrying the normalisation and calc.js rationale — they're the reason the two subtlest bugs in this PR were catchable at all.

## Still open, deliberately

The selector's positional assumptions — `.pielayer` trace order matching enumeration order, and `hole` geometry not adding elements that offset `nth-child` — cannot be exercised until xability/maidr#767 ships. Everything asserted here is schema-level. That is the single thing I'd want checked against a real render before this is called done, and it's the first item in #323.

`uv run pytest -q` → **679 passed**, 0 failed. `ruff check` → 52, unchanged from `main`; clean on every file this PR touches.

---

jooyoungseo:
This is the review that earned holding the merge. All four findings reproduced before being touched; all four were real. Fixed in 7cfd415.

## B1 — holding, and you're right about the mechanism

I checked npm: `maidr@latest` is still **3.75.1**, so the pie release has not published. xability/maidr#767 is now merged and semantic-release should publish shortly; I'm watching for it and will not merge this until it lands.

Your framing is the important part and I want to restate it because it's sharper than the PR description was: this isn't "pie doesn't render yet", it's that tracking the figure *removes* the `_show_fallback` path. A figure mixing a working bar subplot with a pie goes from "bar renders, pie warns" to "the whole figure fails", because `_flatten_maidr` combines all layers into one schema. That's a regression in the exact case the PR's own `TestMixedFigure` covers.

## H2 — confirmed, fixed

Reproduced: bar + `ax.pie([])` → `ExtractionError`, bar layer lost.

Your caution about not silencing a real failure is the crux, and there's a clean line: the patch always hands over `wedges=plot[0]`, this call's own list. **Empty list from the patch** = matplotlib drew nothing = legitimately empty → emit an empty layer. **`wedges is None`** = the layer fell back to scanning the axes and found no pie = real failure → still `ExtractionError`. Both branches tested.

You were right that this contradicted the plotly side's stated principle. It was my own reasoning, applied to one path and not the other.

## M3 — confirmed, fixed

Reproduced: `subplot_titles=("Q1","Q2")` → both pies got `title: ''`, annotations at x=0.225/0.775 against a pie anchor stuck at 0.5.

Rather than override `_get_subplot_annotation_title` and duplicate the matching loop, I extracted the only thing that differs — `_title_anchor() -> (x_mid, y_top)` — and let the pie override that. Matching stays in the base class so a pie and a bar can't drift on what counts as a match.

## M4 — confirmed as a real behaviour change, and scoped

I built the figures and compared against the base commit in a throwaway worktree:

| figure | base | branch (before fix) |
|---|---|---|
| `go.Table` left, bar right | bar col 0 | bar col **1** |
| 2×2 sunburst/bar/treemap/bar | bars col 0 | bars col **1** |
| `go.Indicator` left, bar right | bar col 0 | bar col **1** |

So yes — a `feat(pie)` PR was re-placing figures with no pie in them.

Scoped to renderable trace types via `_PLACED_BY_DOMAIN`. The reasoning: the grid maidr emits is the grid the **user navigates**, and a trace maidr draws no layer for occupies no cell in it. Reserving one adds a dead stop to tab through *and* displaces every renderable subplot beside it — trading an accessibility regression for a fidelity gain no maidr user can perceive. Placement is now byte-identical to base for pie-free figures.

Flagging this as the judgement call worth a second opinion: it means the grid deliberately does not mirror plotly's visual layout when unsupported domain traces are present. If the project's position is that it should, it's a one-line change and the placement shift should be documented as intentional instead.

## Low — the unreachable warning

Confirmed exactly as you described: after any `ax.pie()`, `warnings.filters[0]` is `('ignore', None, Warning, None, 0)`, and `catch_warnings(record=True)` around `plot.schema` catches nothing. Routed through `logging`. Two tests — one asserts the message reaches `caplog`, one pins *why* by asserting a `warnings.warn` wouldn't have been heard.

On the filter's breadth: it's verbatim parity with `patch/common.py`, which does the same on every other patched type. Kept deliberately — a pie that filtered differently would make matplotlib's warnings appear or disappear depending on which plot type the user drew. Narrowing belongs in `common`, for all types at once.

## One repro that doesn't work

`ax.pie([1,2], labels="missing_col", data=df)` never reaches `_resolve`'s fallback — `_preprocess_data` leaves the string unresolved and matplotlib rejects it first with `ValueError: 'labels' must be of length 'x', not 11`. The fallback *is* reachable, so I tested what actually exercises it: `_resolve` directly across all three caught exceptions plus the no-data path, and one end-to-end case matplotlib does accept.

## Your `_as_list` question turned up a live bug — filed as #325

You asked whether the base64 decoding is pie-specific. It isn't, and the other extractors are **broken on `main` without it**:

```python
df = pd.DataFrame({"c": ["a","b","c"], "v": [10, 20, 30]})
px.bar     → [{'x': 'a', 'y': 'dtype'}, {'x': 'b', 'y': 'bdata'}]
px.scatter → [{'x': 'dtype', 'y': 'dtype'}, {'x': 'bdata', 'y': 'bdata'}]
px.pie     → correct
```

Every `px` chart from a DataFrame with numeric columns emits the *dict keys* of plotly's typed-array spec as data — confidently wrong, and truncated to two points. Pre-existing, so out of scope here, but it affects far more users than pie does. Filed as #325 before it could get lost.

Also added: `borrows_axis_titles` documented, matplotlib donut and zero-valued-slice tests (the zero slice matters — matplotlib keeps the zero-span `Wedge`, so dropping it would misalign the selector).

`uv run pytest -q` → **704 passed**, 0 failed. `ruff check` → 52, and I diffed the concise output against the base commit: identical list, same files, same rules.

---

jooyoungseo:
Thanks — both nits handled, and the second one is a better find than "nit" suggests.

**Signature formatting** — collapsed to one line in f33bfd8. You were right that it fits: 79 characters against an 88 limit, so the wrap was noise rather than a limit being respected.

**Accumulating `warnings.filters`** — filed as #326. You're right that parity with `common.py` is the correct call *for this PR*, and also right that duplicating it into a second file is exactly the moment to stop and track it. Worth adding one thing to the issue that this PR ran into concretely: because the filter is process-wide and persistent, it swallows warnings raised **much later**, far from the plot call. A diagnostic added here in `PiePlot._extract_values` could never be heard, because the filter installed during `ax.pie()` was still in effect when `render()` ran. That one was worked around with `logging`, but any future `warnings.warn` in the library has the same problem — which makes this more than a memory-growth nit.

**Runtime plotly version guard** — I considered this after an earlier reviewer suggested the same thing, and landed on not doing it. A warning on version drift fires for everyone on every new plotly regardless of whether the calc rules actually moved, and in the one case it matters it can't say *which* rule moved. What would genuinely close it is cross-checking emitted slice order against a rendered `pielayer`, which needs a browser harness this repo doesn't have. Tracked in #323 for post-release, alongside re-verifying the `nth-child` selector assumptions against a real render.

**Release sequencing** — still holding. `maidr@latest` on npm is **3.75.1**; xability/maidr#767 is merged to `main` (`821de63`) but semantic-release hasn't published yet. I'm watching for it and will not merge until it lands, for the reason an earlier review put well: tracking the figure removes the `_show_fallback` path, so a bar+pie figure would go from "bar renders, pie warns" to "the whole figure fails."

Follow-ups filed from this review round:
- #325 — `px.bar`/`px.line`/`px.scatter` emit `'dtype'`/`'bdata'` as data (pre-existing, affects every plotly.express DataFrame chart)
- #326 — this one
- #323 — post-release verification and bundle re-vendoring

`uv run pytest -q` → **704 passed**, 0 failed. `ruff check` → 52, identical to `main`.

---

jooyoungseo:
Both caught something real. Fixed in ac5743b.

**The mislabelled param is a genuine coverage gap, not just a labelling one.** Verified:

```
list     -> TypeError: list indices must be integers or slices, not str
ndarray  -> IndexError: only integers, slices (`:`), ellipsis (`...`), ...
dict     -> KeyError: 'missing_col'
```

So the case labelled `IndexError` was exercising the same branch as the one below it, and that arm of `_resolve`'s except clause was never run. Swapped in `np.array(...)` — exactly the suggestion — and split the two `TypeError` shapes into distinct ids (`TypeError-list`, `TypeError-unindexable`) so the parametrisation says what it actually covers. Added a comment recording *why* a list can't stand in, so it doesn't get "simplified" back.

**Negative sizes** — you're right that the doc was true but implied the wrong thing about where the check bites. Reworded: matplotlib raises before maidr sees the call, so maidr's own check is a backstop for a directly-built `PiePlot` rather than the thing a user hits. Worth being precise about, since "and so does maidr" reads as if maidr is what protects you.

`uv run pytest -q` → **705 passed**, 0 failed. `ruff check` → 52, identical to `main`.

**Still holding on the release.** `maidr@latest` is **3.75.1**; xability/maidr#767 is on `main` (`821de63`) but semantic-release hasn't published. I'm watching npm and will merge once it lands — nothing else is outstanding on this PR.

---

jooyoungseo:
The `pie_owns_axes` edge case is real — fixed in 17c853d before merging.

You were right that `group_traces` includes unrendered domain traces. The framing that made the fix obvious is yours: those traces land in the default group **for exactly the reason a pie does** — none of them names an axis — so none of them has a claim on `layout.xaxis`/`yaxis`. Only a *cartesian* trace takes those titles away.

Verified all three cases:

```
pie + sunburst : ('Fruit', 'Units')       ← was ('Category', 'Value')
pie + bar      : ('Category', 'Value')    ← unchanged, correct
lone pie       : ('Fruit', 'Units')       ← unchanged, correct
```

I'd push back gently on "the fallback is merely less specific, not wrong": losing a title the author explicitly wrote is a real loss for the person listening, since the generic pair tells them nothing about what the slices measure. Worth fixing rather than noting. Test added alongside `TestPlotlyUnrenderedDomainTraces` as you suggested.

**Negative-value check on the mismatch branch** — agreed, and I'll leave it. It's unreachable through the patch, the angular fallback is non-negative by construction, and the branch now logs when it's taken (added in an earlier commit for exactly this class of "silently plausible" problem). Enforcing it there would add a path with no caller.

**`CLAUDE.md` violin entries** — correct that it's out of scope. I was editing that line to add pie and didn't want to leave a list I knew was wrong; I should have split it. Noting rather than reverting, since removing them would re-break the doc.

**plotly slice-order drift** — this is now the third reviewer to raise it, so it's clearly the right instinct. My position: a version-check warning fires for everyone on every plotly bump and can't say which rule moved, so it trades noise for no real signal. What actually closes it is your suggestion — rendering a pie and checking DOM order against the emitted data, the way the matplotlib side does via `_highlight_groups`. That needs a browser harness this repo doesn't have; it's the first item in #323, tied to the post-release verification.

`uv run pytest -q` → **706 passed**, 0 failed. `ruff check` → 52, identical to `main`.

---

PR #291, opened 2026-07-29
feat(cdn): resolve @latest to a concrete version and report bundle drift

jooyoungseo:
## Description

`save_html(fig, use_cdn=True)` and `use_cdn="auto"` emitted `maidr@latest/dist/maidr.js`. jsDelivr serves that dist-tag with `Cache-Control: public, max-age=604800`, and the URL never changes, so a browser that fetched it once replays week-old JavaScript across every release in that window — silently, with no way for a reader to tell.

This resolves the tag to a concrete published version in Python, at render time, and emits the immutable URL instead. Because the URL now changes on every release, the cache key changes with it, and the seven-day window stops mattering.

Issue #290 also raised a second problem: the bundled offline fallback had drifted eight minor versions behind the release "without user visibility." That is addressed at runtime (`bundle_status()` plus a thresholded one-shot report on the paths where the bundle can actually execute) and in CI (a weekly job that fails when the wheel's bundle falls behind).

Two things were added to the branch after the original scope, both prompted by upstream moving underneath it: the bundled copy is refreshed to maidr 3.75.1, and the KaTeX stylesheet split that release introduced is handled. See **Changes Made** — they are called out separately because they are not part of the CDN-resolution fix.

## Related Issues

Fixes #290.

Follow-ups filed rather than resolved here, each needing a maintainer decision: #292 (`STALE_MINOR_GAP` vs upstream's release cadence), #293 (splitting `dependencies.py`), #294 (`set_cdn_version()` raising eagerly, and renaming `get_cdn_version()`), #295 (making `bundled` the default), #296 (the lookup blocking a Shiny event loop), #297 (no downgrade protection), #298 (the freshness job failing open).

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

### CDN version resolution (the fix for #290)

- `maidr/util/dependencies.py` resolves `@latest` to a concrete version once per process and emits the immutable form — `https://cdn.jsdelivr.net/npm/maidr@3.75.1/dist/maidr.js` — everywhere. `@latest` survives only as the last-resort fallback when no version can be established.
- The lookup is bounded by a shared time budget across both registry endpoints (`deadline - monotonic()` per attempt, not per endpoint), reads are byte-capped, and every failure degrades to `@latest` rather than raising. A failed attempt is cached so an offline session does not retry once per figure.
[...]

PR #324, opened 2026-08-09
fix(cdn): resolve @latest to a concrete version in emitted CDN URLs

### LINKED ISSUE #290: fix(cdn): address an issue where the system doesn't update maidr live
# Bug Report

## Description
`save_html(fig, use_cdn=True)` (and `use_cdn="auto"`) correctly emit a `<script src="https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js">` tag — so the HTML markup itself is right. But because the URL points at the mutable `@latest` npm dist-tag, and jsDelivr serves that URL with `Cache-Control: public, max-age=604800` (7 days) and no cache-busting query string, a browser that has ever loaded `maidr@latest/dist/maidr.js` before will keep serving that exact cached snapshot for up to a week: even though jsDelivr's "latest" resolution has since moved on to a newer published build. In practice this means `use_cdn=True` does not reliably deliver "the current maidr", it delivers "whatever was cached the first time this browser fetched the `@latest` URL", which silently defeats the purpose of opting into the CDN for freshness.

Separately, the `use_cdn=False` / `"auto"`-fallback path references a copy of `maidr.js` bundled with the installed Python package (`lib/maidr-<version>/maidr.js`), and that bundled copy is pinned well behind the current npm `latest`. At the time of filing:

| Source | Version |
|---|---|
| jsDelivr `maidr@latest` (live) | 3.74.0 |
| Bundled copy shipped with `maidr` (PyPI) 1.18.0, used when `use_cdn=False`/on CDN-fetch failure | 3.66.1 |

So if the CDN `<script>` ever fails to load (network policy, ad blocker, sandboxed iframe, offline), `"auto"` silently falls back to a build that is 8 minor versions old, with no visible indication to the user that they're not on the version they expect.

We hit this in a downstream app (`xability/candlestick-dashboard`) that builds an accessible "Download Multimodal Plot in HTML" / "Embed Code" feature via `maidr.save_html(fig, file=temp_path, use_cdn=True)`. After
shipping fixes upstream in `maidr` (JS), our own testing kept showing old behavior in the generated HTML even with `use_cdn=True` explicitly set: traced to the browser cache described above, not to our own code.

We also noticed the documented, "showcased" entry point for previewing a plot is [`Maidr.show()`](https://py.maidr.ai/api/Maidr.show.html), which has its own `use_cdn` handling — but persistence/embedding use cases (saving a standalone HTML file, building an iframe `srcdoc` for an "Embed Code" feature, etc.) go through `save_html()` instead, whose CDN behavior isn't documented alongside `Maidr.show()`'s. Worth
clarifying whether `save_html()` is meant to offer the same freshness guarantees as `Maidr.show()`, since right now the two aren't discoverable from the same docs page and (per the caching issue above).

## Steps to Reproduce
1. `import maidr`, build any matplotlib figure maidr supports (e.g. via `mplfinance`), then call    `maidr.save_html(fig, file="out.html", use_cdn=True)`.
2. Open `out.html` in a browser that has previously loaded `https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js` (e.g. from an earlier test, or from having opened another `maidr` HTML export) within the last 7 days.
3. Inspect the actual `maidr.js` served to that page (e.g. via browser dev tools → Network tab → check the `Age`/`Date` response headers on the `maidr.js` request, or compare behavior against a fix known to be published to npm `maidr@latest` after that browser's cache was populated).
4. Separately: call `maidr.save_html(fig, file="out2.html", use_cdn=False)` and inspect the referenced local script path/version (`lib/maidr-<version>/maidr.js`) against the current npm `latest`

## Actual Behavior

- `use_cdn=True` embeds a `<script src="https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js">` reference, but because jsDelivr returns `Cache-Control: public, max-age=604800` for the mutable `@latest` alias with no cache-busting parameter, any browser that already has that URL cached continues to execute a stale build for up to a week after a new version is published — with no error, warning, or version indicator visible to the user or developer.
- `use_cdn=False` (and the client-side fallback path of `use_cdn="auto"`) loads a copy of `maidr.js` bundled with the installed Python package, which at the time of filing (`maidr` 1.18.0 on PyPI) is pinned to 3.66.1 — multiple minor releases behind the current npm `latest` (3.74.0 as of this report).

## Expected Behavior

- `use_cdn=True` should reliably deliver the current published build e.g. by resolving `@latest` to a concrete version server-side (Python) at render time and emitting a versioned URL (`maidr@<resolved-version>/dist/maidr.js`) instead of the literal string `@latest`, so the browser's cache key changes whenever the published version changes. (A versioned URL is also permanently cacheable, which is strictly better than the current always-revalidate-eventually behavior of an unversioned `@latest` URL.)

## Screenshots

N/A: this is a caching/versioning behavior, not a visual bug.

## Additional Information

```python
import maidr
maidr.save_html(fig, file="test_use_cdn_true.html", use_cdn=True)
maidr.save_html(fig, file="test_use_cdn_false.html", use_cdn=False)
maidr.save_html(fig, file="test_use_cdn_auto.html", use_cdn="auto")
```

- use_cdn=True output contains: <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.css"> and <script src="https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js">.
- use_cdn=False output contains: <script src="lib/maidr-3.66.1/maidr.js"> (relative path to a locally-copied bundle next to the output file).
- use_cdn="auto" output contains both: the CDN <script>/<link> tags and a local lib/maidr-3.66.1/ dependency copy: implying the fallback-to-bundled-copy logic is wired up client-side (via onerror), which is reasonable, but means the 3.66.1 copy is silently reachable any time the CDN request hiccups.
- Confirmed via curl -I https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js: Cache-Control: public, max-age=604800 and X-JSD-Version: 3.74.0 (checked against https://data.jsdelivr.com/v1/packages/npm/maidr, whose tags.latest was also 3.74.0).

Reported from downstream usage in xability/candlestick-dashboard (app.py's save_html_utf8() calls
maidr.save_html(fig, file=temp_path, use_cdn=True) for its "Download Multimodal Plot in HTML" and "Embed Code" features).

jooyoungseo:
Emitted script/link tags pointed at the mutable `maidr@latest` dist-tag,
which jsDelivr serves with `Cache-Control: public, max-age=604800` and no
cache-busting query string. A browser that had ever loaded that URL kept
replaying its cached snapshot for up to a week, so users could keep
executing outdated maidr.js days after a new release.

Resolve `latest` to a concrete version in Python at render time (jsDelivr
data API first, npm registry dist-tags as backup) and emit
`maidr@<version>/dist/maidr.js`. The browser's cache key now changes
whenever the published version changes, which removes the stale window
and lets each build be cached permanently.

- The lookup runs at most once per process and caches both success and
  failure, so an offline session never stalls on a doomed request per
  figure.
- A failed lookup degrades to the previous `@latest` URL rather than
  breaking rendering.
- `use_cdn=False` never touches the network.
- Only well-formed semver is spliced into a URL, so neither a hostile
  `MAIDR_CDN_VERSION` nor a compromised registry response can steer the
  request at an unintended path.
- Escape hatches: `maidr.set_cdn_version()` / `MAIDR_CDN_VERSION`
  (a concrete version, `bundled`, or `latest` to skip resolution) and
  `MAIDR_CDN_TIMEOUT`.

Applies to the matplotlib, Plotly, and Altair renderers plus
`init_notebook()`.

Closes #290

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01StweVDnnVYjC9HZAopSARn


jooyoungseo:
Closing: this work is already on `main`, and merging this PR would revert other work rather than add any.

**It landed as #291**, squash-merged in `d865b4e` about four hours before this PR was opened. Verified against `main` rather than assumed:

| | on `main` |
|---|---|
| `maidr/static/maidr-math.css` | 367,928 bytes |
| bundled `VERSION` | 3.75.1 |
| `_MIN_CDN_TIMEOUT` (the timeout floor) | present |
| `data-maidr-math` (the inlined-KaTeX marker) | present |

**Why this needs closing rather than merging.** The head here, `b4b6d14`, was branched from `c152a9f` — `main` as it stood *before* #291 merged. `main` has moved on since. Comparing the two trees directly (`git diff main b4b6d14`) gives **170 insertions against 6,046 deletions**: this branch is missing step-plot support (`stepplot.py`, `plotly/step.py`, `step_shape.py`, `step_utils.py`), plotly scattergl and selector positions, bar/box/violin orientation handling, the smooth-sampling work, and their test suites.

The dangerous part is that GitHub does not see that as a conflict — `mergeable_state` is `unstable` (checks pending), not `dirty`. So the merge button would have applied cleanly and quietly deleted roughly six thousand lines of merged work.

The branch is being deleted alongside this, so the same stale head cannot be picked up again.

---

Issue #326, opened 2026-08-09
Every patched plot call appends another entry to warnings.filters


jooyoungseo:
Raised during the review of #322. **Pre-existing** — the pie patch mirrors it deliberately for parity, which is what makes it worth fixing in one place rather than per-patch.

## The problem

`maidr/patch/common.py::common` calls:

```python
warnings.filterwarnings("ignore")
```

unconditionally, on every patched plot call. `filterwarnings` **prepends to the global `warnings.filters` list and never clears prior entries**, so a session that draws many plots — a notebook, a loop, a Shiny app re-rendering — accumulates identical filter entries without bound. Every subsequent `warnings.warn` then walks a list that keeps growing.

#322 duplicated the same call into `maidr/patch/pieplot.py`, because that patch cannot route through `common` (`common` feeds the wrapped return value to `FigureManager.get_axes`, which handles `list` but not the `tuple` `Axes.pie` returns). Parity was the right call for that PR — a pie that filtered differently would make matplotlib's warnings appear or disappear depending on which plot type the user happened to draw — but it means the pattern now lives in two files.

## Second-order effect worth knowing

The filter is process-wide and persistent, so it also swallows warnings raised **much later**, far from the plot call. #322 hit this concretely: a diagnostic added in `PiePlot._extract_values` (values and wedges having come apart, which silently rescales every announced magnitude) could never be heard, because the filter installed during `ax.pie()` was still in effect when `render()` ran. That one was worked around by routing through `logging`, but any future `warnings.warn` anywhere in the library has the same problem.

## Suggested direction

Scope the suppression to the wrapped call rather than the process:

```python
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    plot = wrapped(*args, **kwargs)
```

That keeps the intent — matplotlib's own noise should not reach a screen-reader user mid-render — without leaking a global filter or growing the list. It needs doing in `common.py` and `pieplot.py` together so the two do not diverge.

Worth checking whether any current behaviour depends on the filter outliving the call before changing it.

jooyoungseo:
**Correcting this issue: the "unbounded growth" half is wrong.** I filed it and did not verify it.

CPython's `warnings._add_filter` removes an identical entry before inserting:

```python
def _add_filter(*item, append):
    # Remove possible duplicate filters, so new one will be placed
    # in correct place. If append=True and duplicate exists, do nothing.
    if not append:
        try:
            filters.remove(item)
        except ValueError:
            pass
[...]

PR #327, opened 2026-08-09
fix(plotly): decode typed arrays in every extractor, and stop a plot call muting the process

### LINKED ISSUE #325: plotly.express figures emit the literal strings 'dtype' and 'bdata' as data for bar, line and scatter
Found while reviewing #322. **Pre-existing on `main`** — not introduced by that PR, and affecting far more users than pie does.

## The problem

Since plotly 6.x, `Figure.to_dict()` exports a numeric column as a base64 typed-array spec rather than a list:

```python
{"dtype": "i1", "bdata": "ChQe"}
```

The bar, line and scatter extractors iterate that object directly, so they walk its **two dict keys** and emit those key names as the data:

```python
import plotly.express as px, pandas as pd

df = pd.DataFrame({"c": ["a", "b", "c"], "v": [10, 20, 30]})

PlotlyMaidr(px.bar(df, x="c", y="v"))._plots[0].schema["data"]
# [{'x': 'a', 'y': 'dtype'}, {'x': 'b', 'y': 'bdata'}]     ← wrong

PlotlyMaidr(px.scatter(df, x="v", y="v"))._plots[0].schema["data"]
# [{'x': 'dtype', 'y': 'dtype'}, {'x': 'bdata', 'y': 'bdata'}]   ← wrong

PlotlyMaidr(px.pie(df, names="c", values="v"))._plots[0].schema["data"]
# [{'x': 'c', 'y': 30}, {'x': 'b', 'y': 20}, {'x': 'a', 'y': 10}]   ← correct
```

Verified on plotly.py 6.7.0.

## Why it matters

This is every `px` chart built from a DataFrame with numeric columns — the most common way people use plotly. The chart draws correctly, so nothing looks wrong; only the accessible layer is wrong, and it is wrong in the worst way: the announcement is confident, plausible-looking, and bears no relation to the data. A blind user navigating an `px.bar` hears two points named `dtype` and `bdata` instead of their three bars.

It also silently truncates: a three-row DataFrame yields two points, because the spec object has two keys.

## Why pie is unaffected

`maidr/plotly/pie.py::_as_list` decodes the spec (`np.frombuffer(base64.b64decode(...), dtype)`) before reading it. Pie is simply the first extractor to handle the shape.

## Suggested fix

Lift `_as_list` out of `maidr/plotly/pie.py` to somewhere the other extractors can reach — `plotly_plot.py` alongside `domain_interval` is the natural home — and call it wherever a trace array is read. Worth auditing every extractor under `maidr/plotly/` rather than just the three named here.

Tests should cover a `px`-generated figure per plot type, since a hand-built `go.Bar(y=[1,2,3])` passes a plain list and does **not** reproduce this — which is why the existing suite is green.

jooyoungseo:
Closes #325. Refs #326.

## Description

Two follow-ups filed off #322. Both were reproduced before being fixed, and **one of them turned out to be half wrong — see below.**

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

# Pull Request

### #325 — typed arrays, and it is wider than the issue said

Since plotly 6.x, `Figure.to_dict()` exports a numeric column as a base64 typed-array spec rather than a list:

```python
{"dtype": "i1", "bdata": "ChQe"}
```

The issue named bar, line and scatter. Auditing every extractor with `px` figures found **all of them affected except pie**, in three distinct ways:

| Extractor | Baseline symptom |
|---|---|
| bar, line, multiline, scatter, grouped_bar, step | emitted `'dtype'`/`'bdata'` as data, truncated to 2 points |
| box, multibox | **raised** — `TypeError: float() argument must be … not 'dict'` |
| histogram | fell to the categorical path, emitted `[{'x':'dtype','y':1},{'x':'bdata','y':1}]` |
| heatmap | emitted `[['d','t','y','p','e'],['b','d','a','t','a'],['s','h','a','p','e']]` |

The chart draws correctly in every case, so nothing looks wrong — only the accessible layer is, and it is wrong confidently rather than visibly.

Pie's decoder moves to `plotly_plot.as_list` and is called wherever a trace array is read. A spec that will not decode now **logs** and returns empty: silently-empty is better than silently-garbled, but it is still silent, and silence is the fault this decoder exists to fix.

**Two things the issue did not name:**

1. **Heatmap's 2-D spec carries its own extents** (`{'dtype':'i1','bdata':'AQIDBAUG','shape':'2, 3'}`), so the decode reshapes rather than flattening — otherwise `for v in row` would iterate a float.
2. **`_trace_point_count` was silently miscounting.** `len()` of a spec is 2 whatever it holds, so a 30-point numpy-backed scatter counted as 2 and `default_mode` returned `lines+markers` where the same data as a list returned `lines`. That is a plot-type misclassification, not just wrong data. Only reachable via a hand-built `go.Scatter` with no `mode` (px always writes one), which is why nothing caught it.

Also fixed a latent `ValueError: truth value of an array … is ambiguous` in `scatter._get_axis_range`, from px exporting string columns as object ndarrays.

### #326 — the leak is real, the growth claim is not

**I filed #326 and half of it was wrong.** There is no unbounded growth: CPython's `warnings._add_filter` removes an identical entry before inserting, so 1000 identical `filterwarnings("ignore")` calls add one entry, and 500 patched `ax.bar` calls leave exactly one. No code was changed for that half. I've corrected the issue.

The leakage harm is real and reproduced: the filter is process-wide and persistent, so one patched plot call swallowed any `warnings.warn` raised **later, anywhere** — arbitrarily far from the plot. Suppression is now scoped to the wrapped call through a shared `_draw_quietly` helper, which `pieplot` reuses, so its parity with `common` is enforced by construction rather than by a comment.

## Behaviour changes a reviewer should weigh

**1. Render-time warnings are no longer suppressed — deliberately.**

`_extract_plot_data` runs **lazily**, when `.schema` is read at render time, not inside `create_maidr`:

```
after the patched call:  []
after touching .schema:  ['extract']
```

Extraction warnings were previously suppressed not because they sat inside any scope, but because the filter installed during the draw was *still in effect* at render time — which is precisely the leak this PR removes. So a numpy `RuntimeWarning` from a percentile or histogram edge case now reaches the user where it did not before.

That is the point rather than a side effect: the leak's most concrete casualty was MAIDR's own diagnostics — `PiePlot._extract_values`'s "values and wedges have come apart" warning was unhearable for exactly this reason and had to be rerouted through `logging` to be seen at all. Re-suppressing render time to keep numpy quiet would re-break that, and every future diagnostic with it. If extraction warnings prove noisy in practice, the fix belongs at the numpy call sites that raise them, not in re-muting the process.

**2. Nine patch modules lose incidental suppression.**

`boxplot`, `violinplot`, `candlestick`, `heatmap`, `histogram`, `lineplot`, `kdeplot`, `regplot`, `mplfinance` call `wrapped()` directly and were getting suppression only if a `common`-routed call had already run — already order-dependent, so a fresh process whose first plot was `ax.violinplot` was never suppressed. Now consistently off for them. Tracked in #328.

**3. `as_list` rejects a bare string**, returning `[]`. Inherited verbatim from pie's decoder so it is not new for pie, but it now applies everywhere: a trace whose `x` is a bare string previously decomposed into one point per letter and now yields nothing. Not valid plotly input, and the one case where this is not a strict superset of the old behaviour.

## Additional Notes

`uv run pytest -q` → **751 passed** (706 before). `ruff check --output-format=concise` diffed against the pre-edit capture: **identical**, no new findings.

New `tests/plotly/test_plotly_typed_arrays.py` uses `px` figures built from a DataFrame — a hand-built `go.Bar(y=[1,2,3])` passes a plain list and does **not** reproduce this, which is exactly why the existing suite was green. Reverting only the extractors makes **16 of 32 fail**.

`tests/core/test_patch_warning_scope.py` pins the suppression fix on `common`'s path — what every plot type but pie is drawn through. All seven fail against the process-wide filter they replace. The recorder deliberately installs no filter of its own; `simplefilter("always")` would override a leaked "ignore" and the tests would pass either way, which is how the first draft of them did.

Follow-ups filed: #328 (inconsistent suppression across patch modules), #329 (thread safety of the scoped filter).

### DIFF
--- a/maidr/core/plot/pieplot.py
+++ b/maidr/core/plot/pieplot.py
@@ -189,12 +189,13 @@ class PiePlot(MaidrPlot):
         # future bug that decouples the two is visible rather than quietly
         # rescaling every announcement.
         #
-        # This is logged rather than warned because a warning cannot be heard
-        # from here: `maidr.patch.pieplot.pie` calls
-        # `warnings.filterwarnings("ignore")` on every `Axes.pie`, so by the
-        # time `render()` reaches this line the process has a persistent
-        # catch-all "ignore" filter in front of it. `logging` is the channel
-        # that patch does not close, and the one `maidr.backend` already uses.
+        # This is logged rather than warned because it reports a MAIDR bug,
+        # not anything the caller wrote or can act on -- a warning would put
+        # it in front of the reader of an otherwise correct pie. `logging` is
+        # the channel `maidr.backend` already uses for exactly that, and
+        # `maidr.patch.pieplot.pie` suppresses warnings only for the duration
+        # of the `Axes.pie` call, so this line is reached with neither channel
+        # closed.
         if values is not None:
             _logger.warning(
                 "maidr: pie has %d values for %d wedges; reporting each "
--- a/maidr/patch/common.py
+++ b/maidr/patch/common.py
@@ -112,6 +112,14 @@ def _draw_quietly(wrapped: Callable, args: tuple, kwargs: dict) -> Any:
     MAIDR's own diagnostics, which are raised while the schema is built and
     not while the figure is drawn.
 
+    ``catch_warnings`` saves and restores the *global* filter list, so it is
+    not thread safe: two threads drawing at once can restore each other's
+    state, and one may briefly miss a warning the other suppressed. The
+    process-wide filter this replaced shared that flaw and never restored at
+    all, so this is not a regression -- but the window is now per call rather
+    than permanent, which is what makes it reachable in a threaded server.
+    Python 3.14's context-aware filters would close it.
+
     Parameters
     ----------
     wrapped : Callable
--- a/maidr/patch/common.py
+++ b/maidr/patch/common.py
@@ -101,6 +101,36 @@ def resolve_orientation(wrapped: Callable, args: tuple, kwargs: dict) -> str:
     return "horz" if orientation == "horizontal" else "vert"
 
 
+def _draw_quietly(wrapped: Callable, args: tuple, kwargs: dict) -> Any:
+    """
+    Call a patched plotting function with its warnings suppressed.
+
+    The suppression is not to confuse screen-reader users with warnings they
+    did not ask for and cannot act on. It is scoped to the call rather than
+    installed process-wide, because a filter that outlives the call goes on
+    swallowing warnings raised much later and far from any plot -- including
+    MAIDR's own diagnostics, which are raised while the schema is built and
+    not while the figure is drawn.
+
+    Parameters
+    ----------
+    wrapped : Callable
+        The original plotting function.
+    args : tuple
+        Positional arguments the caller passed.
+    kwargs : dict
+        Keyword arguments the caller passed.
+
+    Returns
+    -------
+    Any
+        Whatever the wrapped function returned.
+    """
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        return wrapped(*args, **kwargs)
+
+
 def common(
     plot_type: PlotType | Callable[[Axes], PlotType], wrapped, _, args, kwargs
 ) -> Any:
@@ -130,17 +160,14 @@ def common(
     Any
         Whatever the wrapped function returned.
     """
-    # Suppress warnings not to confuse screen-reader users
-    warnings.filterwarnings("ignore")
-
     # Don't proceed if the call is made internally by the patched function.
     if ContextManager.is_internal_context():
-        return wrapped(*args, **kwargs)
+        return _draw_quietly(wrapped, args, kwargs)
 
     # Set the internal context to avoid cyclic processing.
     with ContextManager.set_internal_context():
         # Patch the plotting function.
-        plot = wrapped(*args, **kwargs)
+        plot = _draw_quietly(wrapped, args, kwargs)
 
     # Extract the data points for MAIDR from the plot.
     ax = FigureManager.get_axes(plot)
--- a/maidr/patch/pieplot.py
+++ b/maidr/patch/pieplot.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import warnings
 from typing import Any, Callable
 
 import wrapt
@@ -9,7 +8,7 @@ from matplotlib.axes import Axes
 from maidr.core.context_manager import ContextManager
 from maidr.core.enum import PlotType
 from maidr.core.figure_manager import FigureManager
-from maidr.patch.common import _argument
+from maidr.patch.common import _argument, _draw_quietly
 
 
 def pie(wrapped: Callable, instance: Axes, args: tuple, kwargs: dict) -> tuple:
@@ -46,23 +45,18 @@ def pie(wrapped: Callable, instance: Axes, args: tuple, kwargs: dict) -> tuple:
         `(wedges, texts)`, or `(wedges, texts, autotexts)` when the caller
         passed `autopct` — whichever the original function returned.
     """
-    # Suppress warnings not to confuse screen-reader users.
+    # The call is drawn through the same helper `maidr.patch.common.common`
+    # uses, so a pie suppresses matplotlib's warnings exactly as every other
+    # patched plot type does -- they neither appear nor disappear depending on
+    # which type the user happened to draw.
     #
-    # This is a process-wide, persistent catch-all, which is broader than this
-    # call needs -- but it is exactly what `maidr.patch.common.common` does on
-    # every other patched plot type, and a pie that filtered differently would
-    # make matplotlib's warnings appear or disappear depending on which plot
-    # type the user happened to draw. Parity is kept deliberately; narrowing
-    # it is a change to make in `common` for all plot types at once.
-    warnings.filterwarnings("ignore")
-
     # Don't proceed if the call is made internally by the patched function.
     if ContextManager.is_internal_context():
-        return wrapped(*args, **kwargs)
+        return _draw_quietly(wrapped, args, kwargs)
 
     # Set the internal context to avoid cyclic processing.
     with ContextManager.set_internal_context():
-        plot = wrapped(*args, **kwargs)
+        plot = _draw_quietly(wrapped, args, kwargs)
 
     data = _argument("data", wrapped, args, kwargs)
     values = _resolve(_argument("x", wrapped, args, kwargs), data)
--- a/maidr/plotly/bar.py
+++ b/maidr/plotly/bar.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
-from maidr.plotly.plotly_plot import PlotlyPlot
+from maidr.plotly.plotly_plot import PlotlyPlot, as_list
 
 
 class PlotlyBarPlot(PlotlyPlot):
@@ -15,8 +15,8 @@ class PlotlyBarPlot(PlotlyPlot):
         return f"{self._subplot_css_prefix()}.trace.bars .point > path"
 
     def _extract_plot_data(self) -> list[dict]:
-        x = self._trace.get("x", [])
-        y = self._trace.get("y", [])
+        x = as_list(self._trace.get("x"))
+        y = as_list(self._trace.get("y"))
 
         return [
             {MaidrKey.X: self._to_native(xv), MaidrKey.Y: self._to_native(yv)}
--- a/maidr/plotly/box.py
+++ b/maidr/plotly/box.py
@@ -121,8 +121,8 @@ class PlotlyBoxPlot(PlotlyPlot):
         q1_vals = as_list(self._trace.get("q1"))
         median_vals = as_list(self._trace.get("median"))
         q3_vals = as_list(self._trace.get("q3"))
-        lowerfence = as_list(self._trace.get("lowerfence", q1_vals))
-        upperfence = as_list(self._trace.get("upperfence", q3_vals))
+        lowerfence = as_list(self._trace.get("lowerfence")) or q1_vals
+        upperfence = as_list(self._trace.get("upperfence")) or q3_vals
 
         results = []
         for i in range(len(median_vals)):
--- a/maidr/plotly/box.py
+++ b/maidr/plotly/box.py
@@ -183,9 +183,8 @@ class PlotlyBoxPlot(PlotlyPlot):
         data = y if y is not None else x
         if data is not None:
             arr = np.array(as_list(data), dtype=float)
-            return [
-                self._compute_stats(arr, label=self._trace.get("name", ""))
-            ]
+            stats = self._compute_stats(arr, label=self._trace.get("name", ""))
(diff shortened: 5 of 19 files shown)


jooyoungseo:
Thanks — both actionable points handled.

**Thread safety** — documented on `_draw_quietly`. You're right that the practical exposure changed even though the flaw didn't: the old process-wide filter had the same underlying problem and never restored at all, but a permanent filter has no window to race against. Scoping it to the call is what makes it reachable from a threaded server. The note says that, and points at Python 3.14's context-aware filters as what would actually close it.

**The nine unsuppressed patch modules** — filed as #328 rather than widened into this PR. Worth recording the framing that made it feel worth filing rather than shrugging at: before this change those modules inherited suppression *by accident*, and only if a `common`-routed call had run first — so a fresh process whose first plot was `ax.violinplot` was never suppressed. Which plot types were quiet was an implementation accident, not a decision. The stated rationale ("don't confuse screen-reader users with warnings they can't act on") doesn't vary by plot type, so neither should the behaviour.

Your independent grep matching mine is useful confirmation, since that list was the thing I was least sure I'd enumerated completely.

**`as_list` on a bare string** — agreed, leaving as-is.

Two things you verified that I want to acknowledge, because they were the parts I'd have most wanted a second pair of eyes on:

- That the `shape` reshape sitting *inside* the `try` means a malformed `shape` raises before any large allocation. I put it there for exception coverage, not for that reason, so that's a better justification than the one I had.
- That the local import in `step_shape` is a genuine cycle break rather than style — `plotly_plot` imports `renders_through_webgl` from `step_shape` at module scope, so the top-level import really would fail.

`uv run pytest -q` → **738 passed**. `ruff check` → 52, identical to `main`.

---

jooyoungseo:
Points 2 and 3 were already addressed in `d0a1e19`, which landed just before this review — the box/multibox re-decode is gone and `_decode_typed_array` now logs. So this reply is mostly about point 1, which is the interesting one.

## Point 1 — the effect is real, the mechanism isn't what you described, and it's intended

You wrote that `_extract_plot_data` "actually runs" inside `FigureManager.create_maidr`, now outside the suppression scope. I checked, and it doesn't — extraction is **lazy**:

```
after the patched call:  []
after touching .schema:  ['extract']
```

`create_maidr` only registers the layer. `_extract_plot_data` runs when `.schema` is read, at render time, long after `common()` has returned. I also confirmed directly that a `RuntimeWarning` raised from inside `BoxPlot._extract_plot_data` does not escape the patched call — it isn't in that scope to begin with, before or after this change.

**But the consequence you identified is real, by a different route.** Extraction warnings *were* suppressed before — not because they were inside the scope, but because the process-wide filter installed during the draw was still in effect at render time. That is precisely the leak this PR removes. So yes: a numpy `RuntimeWarning` from percentile or histogram edge cases now reaches the user where it previously did not.

**That is intended, and it is the point.** The leak's most concrete casualty was MAIDR's own diagnostics — `PiePlot._extract_values`'s "values and wedges have come apart" warning was unhearable for exactly this reason, and had to be rerouted through `logging` to be seen at all. Re-suppressing render time to keep numpy quiet would re-break that, and would do it for every future diagnostic too.

If extraction warnings turn out to be noisy in practice, the right fix is at the numpy call sites that produce them — handle the empty-slice and divide-by-zero cases, or suppress narrowly right there — not to re-mute the whole process from the draw call onward. I'd rather find that out from a real report than pre-emptively paper over it.

I've added this to the PR description rather than leaving it in a thread.

## Confirmations you asked for

- **scatter `_get_axis_range`** — your reading is right, and it is an explicit call-site change rather than a free side effect: `self._trace.get("x", [])` became `as_list(self._trace.get("x"))`. Because `as_list` always returns a plain list, `if data:` never sees an ndarray, so the ambiguous-truth-value `ValueError` cannot arise. Same change, applied at that call site.
- **The 8 patch modules** — filed as #328.
- **Thread safety** — filed as #329, taking your framing from the earlier round: "not a regression" and "safe for our actual concurrent caller" are different claims, and `widget/shiny.py` is the caller that makes the difference matter. The issue lays out four options; my read is that suppressing at the point of *reporting* rather than the point of *drawing* is the one worth exploring, since the goal was never "no warnings exist".

`uv run pytest -q` → **751 passed**. `ruff check` → 52, identical to `main`.

---

jooyoungseo:
**The box crash is real, and it's the best catch on this PR.** Fixed in `2f8e461`.

Reproduced exactly as you described — `median` decodes, `q1` doesn't:

```
maidr: could not decode a typed array (dtype='i1', ...); reporting no data for it.
RAISED IndexError: list index out of range
```

You put your finger on precisely why it matters: this was the one place in the change where a decode failure didn't degrade to empty, it degraded to a crash — and a crash here takes down *every other layer in the figure*, including ones that decoded perfectly. That inverts the whole premise. A decoder whose failure mode is "the pie is empty" is defensible; one whose failure mode is "your dashboard is blank" is not.

The loop is now bounded by the shortest statistic and logs how many boxes it dropped:

```
maidr: box has 3 medians but only 1 complete boxes; describing those and dropping the rest.
```

Applied to `multibox.py` too. Five tests added — one per statistic failing, a partial-length case that keeps the complete boxes, the log assertion, and a silent-when-complete guard.

**On the `or` semantics change you flagged:** you're right that it conflates "absent" with "present but empty", and it's worth saying that this now interacts with the bound rather than standing alone. An empty `lowerfence` falls back to `q1_vals` instead of collapsing `count` to zero — so the two changes together mean a fence that fails to decode costs you the fence's precision, not the box. That's the behaviour I'd want, but it wasn't designed as a pair and I'd rather say so than let it read as intentional from the start.

**Your other confirmations** — scatter's `_get_axis_range` fix is an explicit call-site change (`self._trace.get("x", [])` → `as_list(...)`) rather than a free side effect, though your reading of the mechanism is right: `as_list` always returns a plain list, so `if data:` never sees an ndarray.

The docstring verbosity note is fair and I've taken it; I'd rather not thin the two blocks carrying the normalisation and calc.js rationale, but you're the third reviewer to mention density and that's a signal.

`uv run pytest -q` → **758 passed**. `ruff check` → 52, identical to `main`.

---

jooyoungseo:
You found the sibling path I missed. Fixed in `9c4b83e`.

I hardened the precomputed loop last round and did not go looking for the same shape next door. All three raw-sample paths crashed:

```
raw single    -> RAISED IndexError: index -1 is out of bounds for axis 0 with size 0
raw grouped   -> RAISED IndexError: index -1 is out of bounds for axis 0 with size 0
multibox raw  -> RAISED IndexError: index -1 is out of bounds for axis 0 with size 0
```

`_compute_stats` now answers `None` for an empty array and names the box it dropped; the four call sites skip it. Every quartile of a box with no samples is undefined — `np.percentile` is right to refuse — so there is genuinely nothing to describe. What matters is that the boxes *beside* it still are.

That last part is now pinned directly, since it's the whole argument for dropping rather than raising:

```python
def test_a_readable_box_beside_an_unreadable_one_survives(self):
    ...
    assert len(boxes) == 1
    assert boxes[0]["z"] == "good"
```

Six tests added covering all three raw paths, the log, and that survival case.

**On the fence semantics** — agreed, and worth stating more precisely than I did last round. You're right it's intentional; what I'd add is that it only became *safe* in combination with the bound. On its own, `or q1_vals` conflates "absent" with "explicitly empty". Paired with `min(...)`, the effect is that a fence which fails to decode costs the box its fence precision rather than costing you the box. I like that outcome, but the two changes landed in different commits for different reasons and I'd rather not present it as a single design.

**On #328** — agreed it's user-visible and worth not losing. It's in the PR body and the issue, and I'd flag it explicitly at release-notes time rather than only at review time.

**On the `shape` parsing nit** — correct that the non-string, non-iterable case is caught by accident of the broad except rather than by an explicit check. I'm leaving it: an explicit type check would add a branch for input plotly cannot emit, and the comment above the except now enumerates the three real failure modes so the coverage isn't accidental to a reader even where it is incidental to the code.

`uv run pytest -q` → **763 passed**. `ruff check` → 52, identical to `main`.

---

jooyoungseo:
Both worth acting on. Fixed in `b4e2c07` — with one correction to each diagnosis, because I checked them rather than taking them at face value.

## 1. Test framing — you're right about the substance, the cause isn't pollution

Your point stands: my docstring claimed `common` is "the path every plot type but pie takes", and it isn't. But the mechanism isn't cross-test leakage. I ran each parametrisation **alone** against the old code:

```
OLD code, hist alone:    1 failed
OLD code, plot alone:    1 failed
OLD code, bar alone:     1 failed
OLD code, scatter alone: 1 failed
```

All four fail in isolation, so ordering isn't holding them up. Tracing what actually reaches the helper:

```
hist  -> _draw_quietly called with: ['bar']
plot  -> _draw_quietly called with: ['<lambda>']
```

- **`hist`** does reach the helper — not through its own patch, but because `Axes.hist` draws its bars with the patched `Axes.bar`.
- **`plot`** reaches `common` with the no-op lambda, exactly as you described. The reason it still failed against the old code is that `_draw_quietly`'s `filterwarnings("ignore")` fired **unconditionally on entry**, no-op lambda included — so `common` leaked a filter even when it wrapped nothing.

So the tests aren't order-dependent and won't quietly stop verifying under `-p xdist` or `-k`. But your underlying objection is correct and the fix is the same one you proposed: the docstring now says per kind which cases demonstrate the fix, which reaches it only via an internal draw, and that `plot` demonstrates the absence of the leak rather than its own draw being scoped — pointing at #328 for that.

I'd rather keep all five than drop two: they pin "nothing is left installed", which is true and worth pinning for every kind. What needed fixing was the claim, not the coverage.

## 2. OverflowError — doesn't reproduce here, added anyway

Tested on the pinned numpy:

```
[99999999999999999999, 2]  -> ValueError
[10**30, 1]                -> ValueError
[-1, -1]                   -> ValueError
numpy 2.4.4
```

So on 2.4.4 it's already caught, and `as_list` returns `[]` for all of them. You hedged appropriately about not being able to execute it.

I added `OverflowError` regardless, because your *reasoning* holds even where the specific exception doesn't: `pyproject.toml` accepts `numpy>=1.26`, and the cost of being wrong about one release in that range is an uncaught exception taking down the whole figure — the exact failure this decoder exists to prevent. Four tests pin the impossible-shape cases, and the comment records that it's unreachable on the pinned version so nobody later "cleans up" a clause that looks dead.

## 3. Duplication between box.py and multibox.py

Agreed, and I'm not doing it here. `multibox` already imports `_build_box_selector` from `box`, so the precedent exists and the extraction would be straightforward — but it would touch both extractors' structure in a PR whose subject is typed arrays, and the duplicated logic is currently identical and freshly tested in both. Better as its own change where the diff is legible.

## Minor — the two log sites

Correct, and I'd put it slightly more strongly: when `median` is the array that fails, `count == len(median_vals) == 0` and the truncation warning is skipped entirely, so the *only* signal is `_decode_typed_array`'s own log. That's adequate but not redundant, and it's why that decode-failure log going in earlier this round matters more than it looked at the time.

`uv run pytest -q` → **767 passed**. `ruff check` → 52, identical to `main`.

---

Issue #329, opened 2026-08-09
Warning suppression during a plot call is not thread safe


jooyoungseo:
Raised during the review of #327, where a reviewer made the point that "not a regression" and "safe for our actual concurrent caller" are different claims. They are, and this repo has such a caller.

## The problem

`maidr/patch/common.py::_draw_quietly` suppresses matplotlib's warnings around the wrapped plotting call:

```python
with warnings.catch_warnings():
    warnings.simplefilter("ignore")
    return wrapped(*args, **kwargs)
```

`catch_warnings` saves and restores the **global** `warnings.filters` list. Two threads drawing at once can therefore restore each other's state: thread A's `__exit__` can put back a snapshot taken before thread B's `__enter__`, so B spends the rest of its call with A's filters, or loses its own suppression entirely.

## Why it is worth a ticket rather than a docstring note

The flaw is not new — the process-wide `filterwarnings("ignore")` that #327 replaced mutated the same global state and never restored it at all, so it was arguably worse. But the risk *profile* changed:

- **Before:** permanently wrong, deterministically. Bad, but stable.
- **After:** correct almost always, with a narrow race window that only opens under concurrent draws.

A deterministic bug gets found. A race under load does not.

`maidr/widget/shiny.py` is the one place in this codebase that plausibly serves concurrent requests, which is what turns this from theoretical into worth tracking.

## Options

1. **Leave it and document.** What #327 did. Defensible — the window is small and what gets suppressed today is `PendingDeprecationWarning`/`DeprecationWarning`.
2. **Serialise the draw.** A module-level lock around `_draw_quietly`. Correct, but serialises rendering across threads, which is a real cost for the one caller that motivated the ticket.
3. **Wait for Python 3.14.** It adds context-aware warning filters (`warnings.catch_warnings(context=True)`), which fixes this properly. The project currently supports 3.9+, so this cannot be the only answer for a long while.
4. **Suppress at a different layer.** Capture matplotlib's warnings at the point of *reporting* to the user rather than at the point of drawing, so no global state is touched.

Option 4 is the one worth exploring first — the goal was never "no warnings exist", it was "don't confuse a screen-reader user mid-render", and that is a presentation concern rather than a drawing one.

## Related

- #328 — nine patch modules do not route through this helper at all, so their suppression is inconsistent for a different reason.

PR #330, opened 2026-08-09
fix(patch): draw every patched plot type through the scoped warning filter

### LINKED ISSUE #328: Warning suppression is inconsistent across patch modules, and depends on call order
Follow-up to #327, where a reviewer and I independently landed on the same conclusion.

## The problem

Only `maidr/patch/common.py` and `maidr/patch/pieplot.py` suppress warnings around the wrapped plotting call. Nine other patch modules call `wrapped()` directly:

`boxplot`, `violinplot`, `candlestick`, `heatmap`, `histogram`, `lineplot`, `kdeplot`, `regplot`, `mplfinance`

Before #327, those inherited suppression **by accident**: `common` installed a process-wide `filterwarnings("ignore")`, so once any `common`-routed call had run, every later call in the process was suppressed too. That was already order-dependent — a fresh process whose first plot was `ax.violinplot` was never suppressed at all — so the behaviour a user got depended on which chart they happened to draw first.

#327 scoped the filter to the wrapped call, which fixes a real leak (a plot call was muting `warnings.warn` for the rest of the process, arbitrarily far away). The side effect is that the nine modules above now have suppression consistently **off** rather than inconsistently on.

## Why it is worth fixing

The stated purpose of the suppression is not to confuse screen-reader users with warnings they did not ask for and cannot act on. That rationale does not vary by plot type, so neither should the behaviour. Right now which plot types are quiet is an implementation accident.

## Suggested fix

Route those modules' `wrapped()` calls through the shared `_draw_quietly(wrapped, args, kwargs)` helper that #327 added to `common.py`. Mechanical, one call site each.

Check each module for a second `wrapped()` call on the internal-context early-return path — `pieplot` had one, and it was suppressed before #327 and would need covering too.

## Urgency

Low. The warnings those paths actually raise today are `PendingDeprecationWarning` and `DeprecationWarning`, which Python's default filters ignore anyway, so the user-visible difference is currently near zero. It is worth doing before that stops being true.

jooyoungseo:
Closes #328.

## Description

Only `common` and the pie patch drew through `_draw_quietly`. Nine modules called `wrapped()` directly and had been inheriting the old process-wide filter **by accident** — so whether a violin plot was quiet depended on whether a bar chart had been drawn first in that process. #327 scoped the filter to its call, which removed the accident and left those nine consistently unsuppressed instead.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

# Pull Request

## Description

See above.

## Related Issues

Closes #328. Follows #327, which introduced `_draw_quietly` and disclosed this gap.

## Changes Made

**Reproduced before fixing.** A warning was injected into the draw itself via a data object that warns when numpy converts it. On unmodified `main`, `ax.bar`/`ax.hist`/`ax.pie` were silent (routed through `common`) while all of these leaked:

`ax.plot`, `ax.imshow`, `ax.boxplot`, `ax.violinplot`, `sns.boxplot`, `sns.violinplot`, `sns.kdeplot`, `sns.lineplot`, `candlestick_ohlc`, `mpf.plot`

`sns.boxplot` additionally leaked a real matplotlib 3.10 `PendingDeprecationWarning` ("vert: bool will be deprecated…"), which is now suppressed.

**18 call sites** across the nine modules — several have two, the internal-context early return as well as the main draw:

```
4  boxplot.py     1  candlestick.py   1  heatmap.py
2  histogram.py   1  kdeplot.py       2  lineplot.py
1  mplfinance.py  2  regplot.py       4  violinplot.py
```

Return values and `FigureManager` registration are untouched; only which call is wrapped changed.

Two sites are judgement calls, both wrapped:

- `boxplot.sns_infer_new_orient` wraps seaborn's `_CategoricalPlotter.plot_boxes`. Reached via `seaborn.boxplot` it is already nested inside a suppressed call, so wrapping is a no-op there — but `seaborn.catplot(kind="box")` reaches `plot_boxes` without going through `seaborn.boxplot`, so it genuinely adds suppression for that path. Confirmed by spying on both: a catplot records `plot_boxes` and never `seaborn.boxplot`.
- `regplot.patched_plot` wraps `Axes.plot`, which `lineplot.line` also wraps. Nesting two suppressions is harmless.

`lineplot`'s no-op lambda handed to `common()` was deliberately left alone — the real draw is the direct call above it, and that is what got wrapped.

No module needed to stay unsuppressed; nothing in these nine depends on a warning propagating out of the draw.

n/a

## Checklist

## Additional Notes

`uv run pytest -q` → **805 passed** (baseline 767). `ruff check` → 52, identical to `main`; `ruff check --diff` on the touched paths produced no output.

`tests/core/test_patch_warning_scope.py` grew from 5 kinds to **21**, one or more per patch module, so a module left calling `wrapped()` directly fails on its own row rather than hiding behind another's coverage. There are two parametrised tests: the during-the-draw suppression (what changed here) and the after-the-draw leak guard (what #327 fixed), run for every kind.

The during-the-draw warning is injected by temporarily wrapping `Axes.add_line`/`add_patch`/`add_collection`/`add_image`, so the injection point is strictly inside the wrapped call for every kind and does not depend on a library version choosing to warn about its input. The helper asserts it actually fired, so a kind that stops drawing cannot pass vacuously.

**Verified the tests bite:** with `maidr/` stashed, **13 of the during-the-draw rows fail**, and those 13 cover all nine modules — plot and sns_lineplot (lineplot), hist_step (histogram), imshow and sns_heatmap (heatmap), boxplot and sns_boxplot (boxplot), violinplot and sns_violinplot (violinplot), candlestick, mplfinance, sns_kdeplot (kdeplot), sns_regplot_line_only (regplot).

`hist` and plain `sns_regplot` pass both before and after — those paths were already suppressed through `common`. They are kept as regression guards, with `hist_step` and `sns_regplot_line_only` added alongside because those are the branches the histogram and regplot patches actually had to route themselves.

**The `plot_boxes` wrap needed its own test, and the obvious one is vacuous.** A `sns_catplot_box` row passes with the wrap reverted: the shared injection fires on the first artist added, and for a box draw that is inside `Axes.bxp` — patched and suppressed in its own right — so the warning never reaches the region this wrap covers. What it does cover is what `plot_boxes` raises *around* its call to `bxp`, so `test_catplot_suppresses_what_plot_boxes_raises_around_the_draw` raises from a wrapper installed over maidr's `Axes.bxp`: inside `plot_boxes`, outside `bxp`'s own suppression. That one fails with the wrap reverted. The `sns_catplot_box` row is kept for path coverage, with a comment recording that it does not prove the wrap.

**Worth knowing:** `ax.boxplot` computes its statistics in `Axes.boxplot` and only then calls `Axes.bxp`, which is what maidr patches. Warnings from that statistics pass are outside the patched call and were never suppressible; this does not change that.

This widens the number of call sites subject to the `catch_warnings` thread-safety caveat tracked in #329 — not a regression, since those sites were previously covered by a never-restored process-wide filter, but the window now exists in nine more places.

### DIFF
--- a/maidr/patch/boxplot.py
+++ b/maidr/patch/boxplot.py
@@ -6,21 +6,21 @@ from matplotlib.axes import Axes
 from maidr.core.context_manager import BoxplotContextManager, ContextManager
 from maidr.core.enum import PlotType
 from maidr.core.figure_manager import FigureManager
-from maidr.patch.common import resolve_orientation
+from maidr.patch.common import _draw_quietly, resolve_orientation
 
 
 @wrapt.patch_function_wrapper(Axes, "bxp")
 def mpl_box(wrapped, _, args, kwargs) -> dict:
     # Don't proceed if the call is made internally by the patched function.
     if BoxplotContextManager.is_internal_context():
-        plot = wrapped(*args, **kwargs)
+        plot = _draw_quietly(wrapped, args, kwargs)
         BoxplotContextManager.add_bxp_context(plot)
         return plot
 
     # Set the internal context to avoid cyclic processing.
     with ContextManager.set_internal_context():
         # Patch `ax.boxplot()` and `ax.bxp()`.
-        plot = wrapped(*args, **kwargs)
+        plot = _draw_quietly(wrapped, args, kwargs)
 
     # Set the orientation of the boxplot
     orientation = resolve_orientation(wrapped, args, kwargs)
@@ -40,7 +40,7 @@ def sns_box(wrapped, _, args, kwargs) -> Axes:
     # Set the internal context to avoid cyclic processing.
     with BoxplotContextManager.set_internal_context() as bxp_context:
         # Patch `ax.boxplot()` and `ax.bxp()`.
-        plot = wrapped(*args, **kwargs)
+        plot = _draw_quietly(wrapped, args, kwargs)
         bxp_container = bxp_context
 
     # Set the orientation of the boxplot
@@ -64,7 +64,7 @@ def sns_infer_new_orient(wrapped, instance, args, kwargs) -> str:
         orientation = instance.orient
         BoxplotContextManager.set_bxp_orientation(orientation)
 
-    return wrapped(*args, **kwargs)
+    return _draw_quietly(wrapped, args, kwargs)
 
 
 def patch_seaborn():
--- a/maidr/patch/candlestick.py
+++ b/maidr/patch/candlestick.py
@@ -7,6 +7,7 @@ import numpy as np
 from maidr.core.context_manager import ContextManager
 from maidr.core.enum.plot_type import PlotType
 from maidr.core.figure_manager import FigureManager
+from maidr.patch.common import _draw_quietly
 
 
 def candlestick(
@@ -45,7 +46,7 @@ def candlestick(
     """
     with ContextManager.set_internal_context():
         # Patch the plotting function.
-        plot = wrapped(*args, **kwargs)
+        plot = _draw_quietly(wrapped, args, kwargs)
 
     original_data = None
     date_nums = None
--- a/maidr/patch/heatmap.py
+++ b/maidr/patch/heatmap.py
@@ -7,6 +7,7 @@ from matplotlib.image import AxesImage
 
 from maidr.core.enum import PlotType
 from maidr.core.figure_manager import FigureManager
+from maidr.patch.common import _draw_quietly
 
 
 def heat(wrapped, _, args, kwargs) -> Axes | AxesImage:
@@ -19,7 +20,7 @@ def heat(wrapped, _, args, kwargs) -> Axes | AxesImage:
         optional_params["fmt"] = kwargs["fmt"]
 
     # Patch `ax.imshow()` and `seaborn.heatmap`.
-    plot = wrapped(*args, **kwargs)
+    plot = _draw_quietly(wrapped, args, kwargs)
 
     # Extract the heatmap data points for MAIDR from the plots.
     ax = FigureManager.get_axes(plot)
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -12,7 +12,7 @@ import uuid
 from maidr.core.context_manager import ContextManager
 from maidr.core.enum import PlotType
 from maidr.core.figure_manager import FigureManager
-from maidr.patch.common import common
+from maidr.patch.common import _draw_quietly, common
 
 
 @wrapt.patch_function_wrapper(Axes, "hist")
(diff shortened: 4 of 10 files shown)


jooyoungseo:
Both points were right. Addressed in `2eeacd0` — and chasing the first one turned up something worth knowing.

## The `catplot` gap — real, and the obvious test for it is vacuous

First I confirmed the premise the wrap rests on, by spying on both entry points:

```
sns.catplot(x=[1,2,3,4,5], kind="box")   →   REACHED: ['plot_boxes']
```

`seaborn.boxplot` is never called, so `catplot` does reach `plot_boxes` directly, on seaborn 0.13.2.

Then I added the `sns_catplot_box` row you suggested — and **it passes with the `plot_boxes` wrap reverted.** Tracing where the injected warning actually fires:

```
FIRST ARTIST: ['add_line — INSIDE Axes.bxp']
```

The shared helper fires on the first artist any draw adds, and for a box draw that is inside `Axes.bxp` — which is patched as `mpl_box` and suppressed in its own right. The warning therefore never reaches the region the `plot_boxes` wrap covers, and the assertion holds either way. A row on its own would have looked like coverage while testing nothing.

What the wrap actually covers is what `plot_boxes` raises *around* its call to `bxp`. So the real test raises from a wrapper installed **over** maidr's `Axes.bxp` — inside `plot_boxes`, outside the suppression `bxp` installs for itself:

```python
outer = Axes.bxp

def noisy_bxp(self, *args, **kwargs):
    warnings.warn(raised, UserWarning)
    return outer(self, *args, **kwargs)
```

**That one bites**: `1 passed` with the wrap, `1 failed` with `return wrapped(*args, **kwargs)` restored in `sns_infer_new_orient`.

I kept the `sns_catplot_box` row as well, since it does cover the path for the after-the-draw guard — with a comment saying plainly that its during-the-draw half does not isolate the wrap, and pointing at the dedicated test. A row that passes for the wrong reason is worse than no row if nobody records why.

## The count — you are right, 18

Counted per file against the commit's own parent:

```
4  boxplot.py     1  candlestick.py   1  heatmap.py
2  histogram.py   1  kdeplot.py       2  lineplot.py
1  mplfinance.py  2  regplot.py       4  violinplot.py
--- 18
```

Identical to your breakdown. My "21" came from counting against a stale local `main` that predated #327, which swept in that PR's own `common.py` and `pieplot.py` sites. PR description corrected.

## Thread safety

Already tracked — I filed **#329** for exactly this when #327 landed, and the PR description's closing note links the widened window to it. Nothing further needed here, and agreed it is not a blocker.

`uv run pytest -q` → **805 passed** (was 802). `ruff check` → 52, unchanged from `main`.

---

PR #331, opened 2026-08-09
fix(patch): serialise the warning suppression so concurrent draws cannot leak it

### LINKED ISSUE #329: Warning suppression during a plot call is not thread safe

jooyoungseo:
Closes #329.

## Description

**I filed #329 describing this as a narrow race. Measuring it first showed the description was wrong — it is neither narrow nor transient.**

`catch_warnings` saves the global filter list on entry and restores it on exit. Two threads drawing at once do not merely race on that list; they corrupt it, because the restores nest wrongly:

```
A enters, saving S0          filters = ignore + S0
B enters, saving S1          S1 already contains A's ignore
A exits,  restoring S0       filters = S0            (correct, for now)
B exits,  restoring S1       filters = ignore + S0
```

B puts back a snapshot it took while A was suppressing, so a process-wide `ignore` outlives every draw — **the exact leak #327 removed, reintroduced under concurrency and this time permanently.**

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Related Issues

Closes #329. Follows #327 (which scoped the filter) and #330 (which routed every patch module through the helper).

## Changes Made

**Reproduced before fixing.** Eight threads, sixty draws each, against `main`:

```
filters restored to baseline: False
baseline len: 13   after len: 14
  baseline[0]: ('default', None, DeprecationWarning, re.compile('mplfinance'), 0)
  after[0]   : ('ignore', None, Warning, None, 0)
```

A `('ignore', None, Warning, None, 0)` survives at position 0 and no later draw removes it. Separately, a thread warning during a draw loses its warning **200 times out of 200** — while any draw is in flight the suppression is total, not occasional.

So #329's own wording — a thread "may briefly miss a warning" — understated both halves. The transient half is the *lesser* problem; the leak is the one that matters, because it is permanent and it is what #327 existed to remove.

**The fix** is an `RLock` around the suppression, so the save and restore pair up.

**Reentrant, deliberately:** `regplot.patched_plot` wraps `Axes.plot`, which `lineplot.line` wraps too, so one draw enters the helper twice on the same thread. A plain `Lock` deadlocks there — one of the new tests pins it.

**The cost, stated plainly:** concurrent draws serialise. That is narrow rather than free — Shiny renders on one asyncio loop, so the caller that motivated the ticket sees no contention, and matplotlib's own guidance is that a figure belongs to one thread. It is a real cost for someone drawing separate `Figure` objects on several threads, which is the matplotlib-supported threading pattern.

**What the lock does not fix**, recorded in the docstring rather than left for someone to rediscover: the `ignore` is global while held, so another thread warning at that instant is still silenced. That one ends with the draw rather than surviving it. Closing it would mean not touching the global filters at all — the `showwarning`-dispatch idea from option 4 in the issue — which changes what `catch_warnings(record=True)` callers observe and would need the existing suppression tests' premise revisited. Not folded in here.

Python 3.14's context-aware filters remove the need for the lock entirely; this is the fix for 3.9–3.13.

n/a

## Checklist

## Additional Notes

`uv run pytest -q` → **811 passed** (was 805). `ruff check` → 52, identical to `main`.

`tests/core/test_patch_warning_threads.py` adds six cases. They are about the *filter list*, not about drawing: the work under the lock is a plain `sleep` rather than a plot, so a failure is attributable to the suppression rather than to matplotlib's own thread-safety, and the suite does not depend on concurrent plotting being supported.

**Verified they bite.** With the lock reverted:

```
FAILED test_concurrent_draws_leave_the_filter_list_as_they_found_it
FAILED test_a_warning_after_concurrent_draws_still_reaches_the_caller
2 failed, 4 passed
```

The other four pass either way, and are guards rather than reproductions — that the suppression still works under concurrency, that a nested draw does not deadlock, and that return values are unchanged at two thread counts. Saying so rather than implying all six reproduce the bug.

---

### DIFF
--- a/maidr/patch/common.py
+++ b/maidr/patch/common.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import inspect
+import threading
 import warnings
 from typing import Any, Callable
 
@@ -9,6 +10,12 @@ from matplotlib.axes import Axes
 from maidr.core.context_manager import ContextManager
 from maidr.core.enum import PlotType
 from maidr.core.figure_manager import FigureManager
+
+# Serialises the warning-filter save/restore in `_draw_quietly`; see its
+# docstring for why interleaving those corrupts the global filter list.
+# Reentrant because patches nest: `regplot.patched_plot` wraps `Axes.plot`,
+# which `lineplot.line` wraps as well, so one draw can enter twice.
+_FILTER_LOCK = threading.RLock()
 
 
 def _argument(name: str, wrapped: Callable, args: tuple, kwargs: dict) -> Any:
@@ -112,13 +119,32 @@ def _draw_quietly(wrapped: Callable, args: tuple, kwargs: dict) -> Any:
     MAIDR's own diagnostics, which are raised while the schema is built and
     not while the figure is drawn.
 
-    ``catch_warnings`` saves and restores the *global* filter list, so it is
-    not thread safe: two threads drawing at once can restore each other's
-    state, and one may briefly miss a warning the other suppressed. The
-    process-wide filter this replaced shared that flaw and never restored at
-    all, so this is not a regression -- but the window is now per call rather
-    than permanent, which is what makes it reachable in a threaded server.
-    Python 3.14's context-aware filters would close it.
+    ``catch_warnings`` saves and restores the *global* filter list, which two
+    threads drawing at once will corrupt outright rather than merely race on.
+    Interleave one pair of calls and the restores nest wrongly::
+
+        A enters, saving S0          filters = ignore + S0
+        B enters, saving S1          S1 already contains A's ignore
+        A exits,  restoring S0       filters = S0        (correct, for now)
+        B exits,  restoring S1       filters = ignore + S0
+
+    B puts back a snapshot it took while A was suppressing, so a process-wide
+    ``ignore`` survives every draw -- exactly the leak this helper was written
+    to remove, reintroduced under concurrency and permanently. Measured: eight
+    threads drawing sixty times each leave one ``('ignore', None, Warning,
+    None, 0)`` behind.
+
+    Serialising the draw is what prevents it, since the save and restore have
+    to pair up. The cost is real but narrow: drawing is already effectively
+    single-threaded for the caller that motivated this (Shiny renders on one
+    asyncio loop), and matplotlib's own guidance is that a figure belongs to
+    one thread anyway. Python 3.14's context-aware filters would remove the
+    need for the lock.
+
+    What the lock does *not* fix is that a draw's ``ignore`` is global while
+    it is held, so another thread warning at that moment is still silenced.
+    That one is transient -- it ends with the draw -- rather than surviving
+    it, and closing it would mean not touching the global filters at all.
 
     Parameters
     ----------
@@ -134,7 +160,7 @@ def _draw_quietly(wrapped: Callable, args: tuple, kwargs: dict) -> Any:
     Any
         Whatever the wrapped function returned.
     """
-    with warnings.catch_warnings():
+    with _FILTER_LOCK, warnings.catch_warnings():
         warnings.simplefilter("ignore")
         return wrapped(*args, **kwargs)
 
--- a/tests/core/test_patch_warning_threads.py
+++ b/tests/core/test_patch_warning_threads.py
@@ -1,0 +1,140 @@
+"""Concurrent draws must not leave a filter behind.
+
+``_draw_quietly`` suppresses matplotlib's warnings with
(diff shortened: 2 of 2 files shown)


jooyoungseo:
The flaky-test finding is right, and it is the sharpest thing anyone has caught in this whole series. Fixed in `1181ac3`.

## You were right about the mechanism, and I ran it to check

I ran the file **40 times** before touching it: **0 failures**. So the flakiness was not observable here — but that does not rescue the test, because your argument is structural rather than statistical. The outer `catch_warnings(record=True)` reassigns the same module-global that the code under test manipulates, unsynchronised, on eight threads, outside the lock. The harness reproduced this PR's own bug one layer up. A test that passes 40/40 by timing is still a test that can strip a thread's `ignore` out from under it, and when it passes there is no way to tell which reason it passed for.

Rather than wrap the outer block in `_FILTER_LOCK` — which would serialise the whole loop body and quietly reduce the case to the single-threaded one already covered in `test_patch_warning_scope.py` — it now **reads** the filter list from inside each draw:

```python
def inspect() -> str:
    # `simplefilter("ignore")` prepends ('ignore', None, Warning, None, 0),
    # so the head of the list is this draw's own suppression.
    suppressed.append(bool(warnings.filters) and warnings.filters[0][0] == "ignore")
    time.sleep(0.002)
    return "drawn"
```

Reading races with nothing, it keeps the threads genuinely concurrent, and it asserts the property more directly than recording did: all **320** concurrent draws saw their own suppression installed. 40 further runs after the rewrite: 0 failures.

## Isolation nit — taken

An autouse fixture now restores `warnings.filters` after every test in the file. Your reasoning is exactly why it matters: the failure this file guards against *is* a corrupted global filter list, so a failure here would leak an `ignore` into everything downstream and turn one legible failure into a cascade somewhere unrelated. Verified the corruption tests still bite with the lock reverted — 2 failed, 4 passed, the same two as before.

## Performance trade-off — confirming rather than deferring

Confirmed and accepted, with the reasoning stated where someone will meet it. Two things bound it:

- The caller that motivated the ticket renders on a single asyncio loop, so it never contends.
- matplotlib's own position is that a figure belongs to one thread; the pattern this penalises — separate `Figure` objects per thread — is supported by matplotlib but was already producing a corrupted filter list here, so it is going from *broken and fast* to *correct and serialised*.

Python 3.14's context-aware filters remove the need for the lock entirely, and the docstring says so.

`uv run pytest -q` → **811 passed**. `ruff check` → 52, unchanged from `main`.

---

Issue #333, opened 2026-08-10
Stop bundling the placeholder `maidr.css`, and deprecate the two accessors that point at it


jooyoungseo:
py-maidr ships `maidr/static/maidr.css` and exposes two functions that resolve it. Since maidr 3.75.1 the file is a **406-byte placeholder with no rules in it**, and `maidr/` links no stylesheet at all — the module's own comment says so:

```python
#: maidr 3.75.1 moved it out of :data:`MAIDR_CSS_FILENAME` -- which is now a
#: 406-byte placeholder kept alive only so that existing ``<link>`` tags do
#: not 404 -- and made ``maidr.js`` fetch this file on demand
```

r-maidr, on the same upstream release, simply stopped bundling it. Its `maidr_local_assets()` returns `js` and `math_css` and nothing else. py-maidr kept the file and the accessors.

## Why this is not a delete

`bundled_css_path` is in `maidr/__init__.py`'s `__all__` — a **public API**. `maidr_css_cdn_url` is not exported but is a module-level function in `maidr.util.dependencies` that external code can import. Someone linking the stylesheet themselves in a Shiny or Dash app would call one of them.

Removing the file alone would leave a public function returning a path that does not exist. So this wants a deprecation cycle, not a deletion — which is also why it is filed rather than done.

Neither is called from inside `maidr/`. Verified: the only in-package mention of `maidr_css_cdn_url` is a doc comment naming it, and `bundled_css_path` appears only in the import and `__all__` lists.

## Suggested shape

1. Warn from `bundled_css_path()` and `maidr_css_cdn_url()`, naming `bundled_math_css_path()` as the stylesheet that carries rules.
2. Keep shipping the 406-byte file for now, so nothing breaks mid-cycle.
3. Remove both functions and the file together at the next major.

## Not to be copied from r-maidr: the font stripping

Recording the other half of the comparison so it is not repeatedly re-litigated.

r-maidr strips KaTeX's 20 embedded base64 `@font-face` blocks (~349 kB) out of `maidr-math.css`, leaving 18 kB. **That is a CRAN size-limit compromise, not an improvement**, and its own docstring frames it that way:

> to keep the installed package under CRAN's size limit; KaTeX's layout rules are intact and only the glyphs fall back to system fonts

py-maidr has no equivalent limit, and the arithmetic does not favour copying it:

| | size |
|---|---|
| `maidr.js` | 1.9 MB |
| `maidr-math.css` with fonts | 0.37 MB |
[...]

PR #334, opened 2026-08-10
feat(deps): warn that the placeholder maidr.css accessors are going

jooyoungseo:
## Description

`maidr/static/maidr.css` has been a 406-byte placeholder since maidr 3.75.1 — a comment header and nothing else. Nothing in `maidr/` links a stylesheet, and the module's own comment already said as much. r-maidr, taking the same upstream release, stopped bundling it. py-maidr kept both the file and the two public functions that resolve it.

It cannot simply be deleted. `bundled_css_path` is in `__all__`, so removing the file would leave a public function returning a path that does not exist. So: both accessors warn now, the file keeps shipping, and the two go away together at the next major.

## Related Issues

Refs #333.

## Changes Made

- `maidr/util/dependencies.py`: `_warn_placeholder_css()` emits a `FutureWarning` from `bundled_css_path()` and `maidr_css_cdn_url()`, at `stacklevel=3` so the warning points at the caller rather than at py-maidr's own frame. Both functions gained a `.. deprecated::` directive and a `Warns` section.
- `tests/core/test_offline_bundle.py`: the deliberate calls are wrapped in `pytest.warns(FutureWarning, match=...)`; two new cases — one asserting both accessors warn, one asserting `maidr-math.css` (which does carry rules) stays silent.
- `tests/core/test_cdn_version.py`: a module-level `filterwarnings` for the incidental calls, so the file tests CDN version resolution rather than the deprecation.

### Two deliberate choices worth reviewing

**`FutureWarning`, not `DeprecationWarning`.** Python's guidance splits the two by audience: `DeprecationWarning` for warnings aimed at other library authors, `FutureWarning` for end users of applications written in Python. Someone calling `bundled_css_path()` is assembling a dashboard, not extending py-maidr. The practical half matters more — `DeprecationWarning` is silenced by default outside `__main__`, so a caller inside a Shiny app or any imported module would never see it, and would meet the removal as a breakage instead of a warning. That is the one outcome a deprecation cycle exists to prevent.

**`feat`, not `feat!`.** Nothing breaks here: both functions return exactly what they returned before, and the file they point at still ships. The `!` belongs on the removal — the commit that actually takes something away. I originally typed this `feat(deps)!` and amended it, because `python-semantic-release` honours the marker (verified: `ConventionalCommitParser` returns `bump: major` for that subject), which would have cut **2.0.0 for a change no caller has to react to yet**.

## Checklist

- [ ] I have updated the documentation, if applicable. — no user-facing docs reference these accessors.

## Additional Notes

**Gates.** `uv run pytest` → **813 passed** (from 811). `ruff check` → 52 findings, unchanged from the baseline on `main`. Bite check: deleting the two `_warn_placeholder_css(...)` call sites turns the new assertions red — `3 failed, 41 passed` — so the tests are testing the change and not the scaffolding.

**Not touched, deliberately.** `maidr-math.css` keeps its base64 KaTeX fonts. r-maidr strips them to fit CRAN's package size limit, at the cost of every math glyph falling back to a system font. py-maidr has no such limit, and the saving is under 9% of a package whose JS bundle is five times larger. The two repositories differing here is correct, not drift — #333 records the reasoning so nobody "fixes" the asymmetry later.

**Separate bug found while checking the release wiring, not fixed here.** The `exclude_commit_patterns` allowlist in `pyproject.toml` matches `feat: ` and `feat\(.*\): `, neither of which matches a `!`-marked subject. So every breaking commit in this repository triggers a major version bump and is then **omitted from the changelog** — the worst combination of the two. Filed separately rather than smuggled into this PR.

---

### DIFF
--- a/maidr/util/dependencies.py
+++ b/maidr/util/dependencies.py
@@ -582,7 +582,7 @@ def maidr_js_cdn_url() -> str:
     return cdn_url(MAIDR_JS_FILENAME)
 
 
-def _warn_placeholder_css(name: str) -> None:
+def _warn_placeholder_css(name: str, instead: str) -> None:
     """
     Announce that a ``maidr.css`` accessor is going away.
 
@@ -604,14 +604,19 @@ def _warn_placeholder_css(name: str) -> None:
     name : str
         The function being called, so the message names the caller's own
         call rather than this helper.
+    instead : str
+        What to call in its place, in the same form the caller wanted.  The
+        two accessors return different things -- a local :class:`Path` and a
+        remote URL -- so a single suggestion would hand one of them the
+        wrong type, which is the opposite of what a deprecation message is
+        for.
     """
     warnings.warn(
         f"maidr.{name}() is deprecated and will be removed in the next major "
         "release, along with the bundled maidr.css it resolves. That file has "
         "carried no rules since maidr 3.75.1 and nothing in py-maidr links "
-        "it. Use bundled_math_css_path() for the stylesheet that does carry "
-        "rules, or link no stylesheet at all -- MAIDR styles its interface at "
-        "runtime.",
+        f"it. Use {instead} for the stylesheet that does carry rules, or link "
+        "no stylesheet at all -- MAIDR styles its interface at runtime.",
         FutureWarning,
         stacklevel=3,
     )
@@ -622,9 +627,11 @@ def maidr_css_cdn_url() -> str:
 
     .. deprecated::
         Linking this buys nothing and will be removed, along with
-        ``maidr.css`` itself, in the next major release.  Use
-        :func:`bundled_math_css_path` if what you wanted was the
-        stylesheet that carries rules, or link nothing at all -- MAIDR
+        ``maidr.css`` itself, in the next major release.  If what you
+        wanted was the stylesheet that carries rules, and you want it from
+        the CDN as this function returns it, call
+        ``cdn_url(MAIDR_MATH_CSS_FILENAME)``; :func:`bundled_math_css_path`
+        is the local-file counterpart.  Or link nothing at all -- MAIDR
         styles its interface at runtime.
 
     Nothing in ``maidr/`` emits this any more.  From maidr 3.75.1 the file
@@ -643,7 +650,7 @@ def maidr_css_cdn_url() -> str:
     FutureWarning
         Always.  See :func:`_warn_placeholder_css` for why this category.
     """
-    _warn_placeholder_css("maidr_css_cdn_url")
+    _warn_placeholder_css("maidr_css_cdn_url", "cdn_url(MAIDR_MATH_CSS_FILENAME)")
     return cdn_url(MAIDR_CSS_FILENAME)
 
 
@@ -1433,7 +1440,7 @@ def bundled_css_path() -> Path:
     FutureWarning
         Always.  See :func:`_warn_placeholder_css` for why this category.
     """
-    _warn_placeholder_css("bundled_css_path")
+    _warn_placeholder_css("bundled_css_path", "bundled_math_css_path()")
     return _bundled_asset_path(MAIDR_CSS_FILENAME)
 
 
--- a/maidr/util/dependencies.py
+++ b/maidr/util/dependencies.py
@@ -602,8 +602,12 @@ def _warn_placeholder_css(name: str, instead: str) -> None:
     Parameters
     ----------
     name : str
-        The function being called, so the message names the caller's own
-        call rather than this helper.
+        The function being called, spelled as a path that actually
+        imports, so the message names the caller's own call rather than
+        this helper.  Fully qualified rather than assumed: only one of
+        these two is re-exported from the top-level package, so a fixed
+        ``maidr.`` prefix would name a path that raises
+        :class:`AttributeError` for the other.
     instead : str
         What to call in its place, in the same form the caller wanted.  The
         two accessors return different things -- a local :class:`Path` and a
@@ -612,7 +616,7 @@ def _warn_placeholder_css(name: str, instead: str) -> None:
         for.
     """
     warnings.warn(
-        f"maidr.{name}() is deprecated and will be removed in the next major "
+        f"{name}() is deprecated and will be removed in the next major "
         "release, along with the bundled maidr.css it resolves. That file has "
         "carried no rules since maidr 3.75.1 and nothing in py-maidr links "
         f"it. Use {instead} for the stylesheet that does carry rules, or link "
@@ -630,9 +634,11 @@ def maidr_css_cdn_url() -> str:
         ``maidr.css`` itself, in the next major release.  If what you
         wanted was the stylesheet that carries rules, and you want it from
         the CDN as this function returns it, call
-        ``cdn_url(MAIDR_MATH_CSS_FILENAME)``; :func:`bundled_math_css_path`
-        is the local-file counterpart.  Or link nothing at all -- MAIDR
-        styles its interface at runtime.
+        ``cdn_url(MAIDR_MATH_CSS_FILENAME)`` -- both from this module, which
+        is also the only place this function itself is importable from.
+        :func:`bundled_math_css_path` is the local-file counterpart, and is
+        re-exported as ``maidr.bundled_math_css_path``.  Or link nothing at
+        all -- MAIDR styles its interface at runtime.
 
     Nothing in ``maidr/`` emits this any more.  From maidr 3.75.1 the file
     it points at is a placeholder with no rules in it, published only so
@@ -650,7 +656,10 @@ def maidr_css_cdn_url() -> str:
     FutureWarning
         Always.  See :func:`_warn_placeholder_css` for why this category.
     """
-    _warn_placeholder_css("maidr_css_cdn_url", "cdn_url(MAIDR_MATH_CSS_FILENAME)")
+    _warn_placeholder_css(
+        "maidr.util.dependencies.maidr_css_cdn_url",
+        "cdn_url(MAIDR_MATH_CSS_FILENAME) from maidr.util.dependencies",
+    )
     return cdn_url(MAIDR_CSS_FILENAME)
 
 
@@ -1440,7 +1449,7 @@ def bundled_css_path() -> Path:
     FutureWarning
         Always.  See :func:`_warn_placeholder_css` for why this category.
     """
-    _warn_placeholder_css("bundled_css_path", "bundled_math_css_path()")
+    _warn_placeholder_css("maidr.bundled_css_path", "maidr.bundled_math_css_path()")
     return _bundled_asset_path(MAIDR_CSS_FILENAME)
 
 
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Both cheap notes taken in `eecde6f`. The third I checked and left alone — details below, since the reasoning is the useful part.

**Import order** — fixed, `re` before `warnings`. One correction to the diagnosis, though: it would *not* have registered as a ruff finding. `[tool.ruff]` in this repo sets only `line-length`, so the default `E4,E7,E9,F` selection applies and `I001` is not among the enabled rules — `ruff check tests/core/test_offline_bundle.py` passes on the unfixed file, which is why the "unchanged from the baseline" claim held. Worth knowing for future reviews here: isort is not enforced, so import order is a convention this repo keeps by hand. The ordering was still wrong, so it is fixed.

**The unescaped `.`** — good catch, fixed to `r"ignore:maidr\.maidr_css_cdn_url:FutureWarning"`. Your reasoning was right and I went one step further, because a warning filter that silently stops matching is invisible: there is no `filterwarnings = error` in this repo's pytest config, so a filter that matches nothing changes nothing and no test notices.

So I made it observable rather than trusting it:

```
$ uv run pytest tests/core/test_cdn_version.py -q -W error::FutureWarning
132 passed

# with the mark's message deliberately mismatched:
5 failed, 127 passed
```

The filter is doing real work — five tests depend on it — and the escaped form still matches. That is the check the original note implies but that nothing in the suite was performing.

**The `.. deprecated::` directives, left without a version.** I checked whether anything renders them, and nothing does. This repo builds its API reference with quartodoc, not Sphinx, and `docs/_quarto.yml` lists exactly three symbols under `quartodoc.sections`: `Maidr.show`, `Maidr.save_html`, and `set_backend`. Neither `bundled_css_path` nor `maidr_css_cdn_url` appears in the rendered docs at all, so there is no rendering to verify.

That leaves the directive as a source-level marker for someone reading the docstring, which is what I want it to be, and the missing version argument as the honest state of things — the removal release genuinely is not known yet. Naming a speculative version would be worse than omitting it: a `.. deprecated:: 2.0.0` that turns out to be 3.0.0 is a wrong statement in the source, where an absent argument is merely an incomplete one.

Gates after the change: `uv run pytest` → **813 passed**, `ruff check` → 52, unchanged; the three touched files are clean on their own.

---

jooyoungseo:
The replacement-type point is right and was a real defect in the message. Fixed in `0912283`.

A deprecation warning has one job beyond announcing itself: telling the reader what to do instead. Mine told both callers to use `bundled_math_css_path()`, which is correct for `bundled_css_path` (`Path` → `Path`) and **wrong** for `maidr_css_cdn_url`, whose caller is building a `<link href>` and would get a `Path` object. A message that misdirects is worse than one that says only "deprecated" — the reader who ignores a bare deprecation loses nothing, and the reader who follows a wrong one loses an afternoon.

The message now takes the replacement as an argument, so each accessor names one that returns what that caller was already holding:

```
maidr.bundled_css_path() is deprecated … Use bundled_math_css_path() for the
stylesheet that does carry rules, or link no stylesheet at all …

maidr.maidr_css_cdn_url() is deprecated … Use cdn_url(MAIDR_MATH_CSS_FILENAME)
for the stylesheet that does carry rules, or link no stylesheet at all …
```

And I checked that the URL it now names is real rather than merely plausible, since recommending a 404 would be its own version of the same mistake:

```
$ curl -o /dev/null -w "%{http_code} %{size_download}\n" \
    https://cdn.jsdelivr.net/npm/maidr@4.0.0/dist/maidr-math.css
200 367928
```

Two cases pin it: one asserts each accessor names its own replacement, and one asserts the CDN accessor does not mention the path function at all — that second is the one that would catch a future edit collapsing the two messages back together.

**On the `.. deprecated::` version argument**, raised in both rounds. Nothing renders these docstrings, so there is no rendering to confirm: this repo builds its API reference with quartodoc rather than Sphinx, and `docs/_quarto.yml` lists exactly three symbols — `Maidr.show`, `Maidr.save_html`, `set_backend`. Neither accessor is among them. That leaves the directive as a source-level marker, which is what I intend it to be, and the missing version as the honest state: the removal release is genuinely not decided. A speculative `.. deprecated:: 2.0.0` that turns out to be 3.0.0 is a false statement in the source; an absent argument is only an incomplete one.

| Gate | Result |
|---|---|
| `uv run pytest` | **814 passed** (was 813) |
| `ruff check` | clean on both touched files; 52 repo-wide, unchanged |
| All four Python legs | green on the previous commit, re-running now |

---

jooyoungseo:
Right on every point, including the sting in the tail — this is the same mistake as the `Path`-for-a-URL one, one level up, and I made it in the commit where I was congratulating myself for catching that one. Fixed in `ea25345`.

Verified before fixing:

```
maidr.bundled_css_path                 OK
maidr.maidr_css_cdn_url                AttributeError
maidr.cdn_url                          AttributeError
maidr.MAIDR_MATH_CSS_FILENAME          AttributeError
maidr.bundled_math_css_path            OK
```

So a reader of the CDN accessor's warning got an `AttributeError` on the symbol it named, and a second one on the replacement it suggested — sent looking for the fix and given a path that does not exist, twice.

Each call site now passes the spelling that imports:

```
maidr.bundled_css_path() is deprecated … Use maidr.bundled_math_css_path() …

maidr.util.dependencies.maidr_css_cdn_url() is deprecated … Use
cdn_url(MAIDR_MATH_CSS_FILENAME) from maidr.util.dependencies …
```

**Your closing observation is the important half, and I took it as the instruction.** You noted the existing assertions cannot catch this, and you were exactly right about why: `pytest.warns(match=...)` and `assert instead in message` both pass on a substring, and the wrong path *contains* the right name. Three rounds of tests here have been checking that the message mentions the right words while saying nothing about whether those words denote anything.

So the new test resolves **every dotted path each message contains**, rather than checking for the ones I expect today:

```python
named = set(re.findall(r"\bmaidr(?:\.[A-Za-z_]\w*)+", message))
named -= {"maidr.css", "maidr.js"}          # filenames, not modules
unresolvable = sorted(p for p in named if not _importable(p))
assert not unresolvable
```

That is what makes it catch the *next* wrong path rather than this one. Reverting the prefix produces exactly the diagnosis you wrote by hand:

```
AssertionError: the deprecation for maidr_css_cdn_url names paths that do not
import: ['maidr.maidr_css_cdn_url']
```

Worth recording that the test caught something on its first run — against my own fix. `maidr.css` has the shape of a module path while being a filename, and a message about a stylesheet cannot avoid mentioning it. Excluded by name, with the reason written down, rather than by making the pattern clever enough to tell a module from a file, which it cannot be.

| Gate | Result |
|---|---|
| `uv run pytest` | **815 passed** (was 814) |
| `ruff check` | clean on both touched files; 52 repo-wide, unchanged |

---

jooyoungseo:
Caught a regression I introduced two commits ago, and it is a good one. Fixed in `f38a23d`.

Renaming the symbol to its importable path — the fix for your *previous* round — broke the filter I had proven load-bearing in the round before that. `re.match` anchors at the start, so `maidr\.maidr_css_cdn_url` stops matching the moment `util.dependencies.` appears in between. Confirmed rather than reasoned:

```
$ uv run pytest tests/core/test_cdn_version.py -q -W error::FutureWarning
5 failed, 127 passed          # dead filter
...after the fix:
132 passed
```

Exactly the five tests I had used to demonstrate the filter mattered.

**The literal fix is one line; the interesting part is why I did not just take yours.** You are right that matching the real qualified name works, but that is the third time this string has needed hand-updating to track something that moved, and the failure mode is silent every time — nothing here turns warnings into errors, so a dead filter changes nothing and no test notices. That is the same shape as the `exclude_commit_patterns` bug in #335: a value nothing reads back.

So the filter is now read back:

```python
marks = test_cdn_version.pytestmark
specs = [arg for mark in ... for arg in mark.mark.args]
matched = [spec for spec in specs if re.match(spec.split(":")[1], message)]
assert matched, "no filter in test_cdn_version.py matches the warning it means to silence"
```

It pulls the mark off that module and applies it to the message the accessor really emits, so the two cannot drift apart again. Restoring the dead pattern produces that assertion naming both the filter and the message. And the pattern itself stops caring about the module path, since that is what moved underneath it both times.

**On the version bump** — I think `feat` is right, though your gut check is fair to raise. A deprecation warning is a change users observe: code that was silent now prints, and CI configured with `-W error` will fail on it. That is more than `chore`, and the conventional practice is deprecation-announcement → minor, removal → major. `refactor` would be wrong in the other direction; nothing was restructured.

| Gate | Result |
|---|---|
| `uv run pytest` | **816 passed** (was 815) |
| `uv run pytest tests/core/test_cdn_version.py -W error::FutureWarning` | 132 passed |
| `ruff check` | clean on both touched files; 52 repo-wide, unchanged |

---

jooyoungseo:
All three taken in `2987320`. Two of them were real holes, and one of those found a divergence that already existed.

**Nit 3 — the bare symbols were never checked.** You were right that only the module prefix was validated, so `cdn_url` or `MAIDR_MATH_CSS_FILENAME` could have been misspelled and the test would have passed while claiming to verify exactly that. They are now resolved against every module the message names, plus the top-level package:

```
AssertionError: the deprecation for maidr_css_cdn_url names symbols that do
not resolve: ['MAIDR_MATH_CSS_FILENAMES']
```

One wrinkle worth recording: a constant now has to carry an underscore. The obvious `[A-Z][A-Z0-9_]{2,}` also claims **`MAIDR`**, which these messages say as the product's name — "MAIDR styles its interface at runtime" — not as a symbol. The test told me that on its first run.

**Nit 2 — the docstring seam is now mechanical, and it had already drifted.** Whatever the message says to call, the docstring has to name too. This failed the moment I wrote it, which is the answer to whether the seam was hypothetical:

```
AssertionError: bundled_css_path's docstring no longer names the replacement
its warning does: maidr.bundled_math_css_path()
```

The comparison is on the *symbol*, not the spelling — the message qualifies it so a reader can import it, while the docstring uses Sphinx's `:func:` role, which takes the bare name. Both bite: rename the replacement in one place only and the assertion names the function and the symbol.

Your framing of the risk was the useful part — "changing *next major* to a pinned version number once a release is scheduled" is precisely the edit that would have drifted, and it is the one someone will make.

**Nit 1 — renamed.** You are right that the sentence-style names are a departure; the file says `test_bundled_maidr_css_exists`, not a clause. Five renamed to match, and the docstrings carry the reasoning either way, so nothing is lost:

| was | now |
|---|---|
| `test_the_placeholder_css_accessors_warn_before_they_go` | `test_placeholder_css_accessors_warn` |
| `test_every_path_the_deprecation_names_can_be_imported` | `test_deprecation_names_only_importable_symbols` |
| `test_the_cdn_version_suite_really_silences_the_deprecation` | `test_cdn_version_filter_still_matches_the_warning` |
| `test_the_cdn_accessor_is_not_sent_to_a_local_path` | `test_cdn_accessor_replacement_returns_a_url` |
| `test_the_stylesheet_that_carries_rules_does_not_warn` | `test_math_css_accessors_do_not_warn` |

| Gate | Result |
|---|---|
| `uv run pytest` | **816 passed** |
| `ruff check` | clean on the touched file; 52 repo-wide, unchanged |

---

PR #336, opened 2026-08-10
fix(ci): keep breaking commits in the changelog

### LINKED ISSUE #335: A breaking commit cuts a major release and is then dropped from the changelog
## What happens

`[tool.semantic_release.changelog].exclude_commit_patterns` in `pyproject.toml` is an allowlist expressed as a negative lookahead. Every alternative in it ends in a literal `): ` or `: `:

```
^(?!feat: |feat\(.*\): |fix: |fix\(.*\): |docs: |docs\(.*\): | … ).*
```

A breaking subject is `feat(deps)!: …`. After `feat(` the regex needs to find `): `, and the subject has `)!: ` — a `)` followed by `!`, not by a colon. No alternative matches, the lookahead succeeds, and **the commit is excluded from the changelog**.

Meanwhile the commit parser handles `!` correctly and bumps the major.

So a breaking change produces a major version release whose changelog does not mention it. Measured against the config as it stands on `main`:

```
'feat: plain'                     excluded_from_changelog=False
'feat(deps): scoped'              excluded_from_changelog=False
'feat!: plain breaking'           excluded_from_changelog=True   ← 
'feat(deps)!: scoped breaking'    excluded_from_changelog=True   ← 
'fix(deps)!: scoped breaking fix' excluded_from_changelog=True   ← 
```

and, for the bump, calling the parser directly on a `feat(deps)!:` commit:

```
type: ParsedCommit
bump: major
breaking descriptions: []
```

## Why it is worse than it looks

`breaking_descriptions` is empty when the `!` marker is used without a `BREAKING CHANGE:` footer, which is the normal way to write one. So the release notes have no breaking-changes section to fall back on either. The version number jumps a major and the only artifact explaining why is the commit itself, which the changelog just dropped.

Nothing upstream flags this. Commitlint accepts `!`, the parser accepts `!`, CI is green, the release succeeds. The omission is only visible by reading the published changelog and noticing an entry that should be there is not.

## Fix

Admit an optional `!` before the colon in each alternative — `feat\(.*\)?!?: ` in spirit, though the alternation as written needs each branch updated, or the whole thing collapsed to one pattern:

```toml
exclude_commit_patterns = [
  "^(?!(feat|fix|docs|perf|refactor|test|build|ci|style)(\\(.*\\))?!?: ).*"
]
```

That is equivalent for every subject the current list admits, and additionally admits the `!`-marked ones. Worth confirming against the same subject table above before merging, since a mistake here fails silently in exactly the same way.

A test pinning it would be the real fix. The current pattern has already been wrong once in a different way — the comment above it records that the key lived under the wrong table and was silently dropped by pydantic, so every commit type landed in the changelog for some time. Both failures share a shape: a config value that is never read back, so nothing can notice it stopped working. A test that asserts a table of subjects against the compiled patterns would have caught both.

## Related

Found while typing the commit for #333. The maidr.js repository had the mirror-image bug — there the `!` broke the parse itself and suppressed the release entirely (xability/maidr#778); here the release happens and the changelog entry is what goes missing.

jooyoungseo:
## Description

`exclude_commit_patterns` is an allowlist written out type by type, and every alternative in it ends in a literal `): ` or `: `:

```
^(?!feat: |feat\(.*\): |fix: |fix\(.*\): |docs: |docs\(.*\): | … ).*
```

A breaking subject is `feat(deps)!: …`. After `feat(` the regex looks for `): ` and finds `)!: `. No alternative matches, the negative lookahead succeeds, and the commit is **dropped from the changelog**.

The parser, meanwhile, reads `!` correctly and returns `bump: major`. So a breaking change cut a major release whose changelog did not mention it. And because `breaking_descriptions` is empty when `!` is used without a `BREAKING CHANGE:` footer — which is the normal way to write one — there was no breaking-changes section to fall back on either. The version jumped a major and nothing said why.

Nothing upstream flags this. Commitlint accepts `!`, the parser accepts `!`, CI is green, the release succeeds.

## Related Issues

Closes #335.

## Changes Made

**`pyproject.toml`** — the alternation collapses to one pattern with an optional marker:

```toml
exclude_commit_patterns = [
  "^(?!(feat|fix|docs|perf|refactor|test|build|ci|style)(\\(.*\\))?!?: ).*"
]
```

I compared it against the written-out form over **125 subjects** — every combination of type (including `chore`, `revert`, and near-misses like `feature` and `circle`), scope (none, `(deps)`, `(a)(b)`, `()`), and marker, plus release commits, merge commits, and non-conventional messages. The only subjects whose treatment changes are the `!`-marked ones:

```
'feat!: something'              True -> False
'feat(deps)!: something'        True -> False
…
'style()!: something'           True -> False
'feat(deps)!: multi'            True -> False
```

`chore` and `revert` stay excluded with or without the marker, because the `!` is admitted per listed type rather than globally — that was the thing to get wrong here, and it is the reason `chore!: …` appears nowhere in that diff.

**`tests/core/test_changelog_filter.py`** — 68 cases. They read the pattern out of `pyproject.toml` rather than restating it, and apply it the way `ReleaseHistory.from_git_history` actually does: `pattern.match` against the **whole commit message**, not the subject alone (`semantic_release/changelog/release_history.py:159`).

The test is the part worth having. This value has now been wrong twice — the comment above it records the first time, when the key sat under `changelog.default_templates`, pydantic silently dropped the unknown key, and every commit type landed in the changelog for a while. Both failures have the same shape: **nothing reads the value back**, so a mistake produces a green release and a missing entry rather than an error. A table of subjects checked against the compiled pattern would have caught both.

One case pins something easy to break later: `.` does not cross a newline, so the pattern reads the subject and stops. A commit whose *body* contains the word `chore` must not be evicted, and a `chore:` commit whose body says `feat:` must not be let in.

## Checklist

- [x] I have updated the documentation, if applicable. — the comment above the pattern now explains what `!?` is for.

## Additional Notes

| Gate | Result |
|---|---|
| `uv run pytest` | **879 passed** (from 811 on `main`) |
| `ruff check` | 52 findings, unchanged from `main`; the new file is clean on its own |
| `uv lock --check` | resolved, no drift |

**Verified it bites:** restoring the old pattern turns **28 of the 68** new cases red — the 27 breaking-commit combinations plus the body case.

`fix:` rather than `feat:`: the changelog was wrong, and this makes it right.

---


jooyoungseo:
CI caught a real one on the first run — Python 3.9 and 3.10 failed collection. Fixed in `1902ef0`.

`tomllib` is stdlib only from 3.11, and I wrote the test on 3.11, so it passed locally and broke on the two legs I did not run. Which is the correct place for that to surface, but worth naming plainly: the gate table in the PR body above said "879 passed" and that was one interpreter's answer, not four.

```
ERROR tests/core/test_changelog_filter.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
```

**The fix** falls back to `tomli`, which is the package `tomllib` was adopted from and has the same API. No new dependency: it is already resolved for both interpreters, and — the part that mattered to check — it is present even under the install profile `release.yml` uses, which is `--extra test --extra visualization` **without** the dev group. Had it only come in via `--dev`, this test would have passed every CI run and broken the release job instead, which is a considerably worse place to find out.

Verified on real interpreters rather than reasoned about:

```
$ uv run --python 3.9  --extra test --extra visualization pytest tests/core/test_changelog_filter.py -q
68 passed

$ uv run --python 3.10 --extra test --extra visualization pytest tests/core/test_changelog_filter.py -q
68 passed
```

**Still `importorskip` rather than declaring `tomli`.** If some future resolution drops it, the test skips instead of failing collection. Nothing about a string in `pyproject.toml` varies by interpreter, so 3.11 and 3.12 cover the invariant on their own — and a release must not fall over because a TOML parser is missing. Adding `tomli` to the published `test` extra to run a repo-config check on old interpreters would be the more expensive half of that trade for none of the benefit.

Updated gates, all four legs this time:

| Interpreter | Result |
|---|---|
| 3.9 (release profile) | 68 passed |
| 3.10 (release profile) | 68 passed |
| 3.11 (full suite) | **879 passed** |
| `ruff check` | clean on the new file; 52 repo-wide, unchanged |

---

jooyoungseo:
Both checked. One was right and is fixed in `dc36919`; the other I can confirm from the source.

**`tomli`'s provider — you are right, and for a stronger reason than I gave.** I attributed it to `black`, which does require it below 3.11, but `black` only reaches the dev group transitively through `quartodoc`, and `release.yml` installs the test extra without that group. `pytest` requires it *directly* and is in the `test` extra itself:

```
$ uv run --python 3.9 --extra test --extra visualization python -c "..."
pytest: ['tomli >=1.0.0 ; python_version < "3.11"']
black:  ["tomli>=1.1.0; python_version < '3.11'"]
```

So anything able to run this file at all has the parser — a much better guarantee than the one I wrote down, and it makes the fallback's presence near-unconditional rather than incidental. Comment now points at `pytest`. The `importorskip` stays: a dependency that arrives by implication can stop arriving.

**The `pattern.match(full_message)` premise — confirmed from the installed source**, not inferred from CI. `semantic_release/changelog/release_history.py`, at the version this repo pins:

```python
commit_message = force_str(parsed_result.commit.message)
...
has_exclusion_match = any(
    pattern.match(commit_message) for pattern in exclude_commit_patterns
)
```

`parsed_result.commit.message` is the whole message, subject and body, and it is `match` rather than `search`, so the anchored pattern reads from position 0 and `.` stops at the first newline. `test_the_body_does_not_decide_it`'s premise holds. Worth having asked — you were right that CI passing is weaker evidence than it looks here, since a subject-only application would leave that test passing for the wrong reason.

Gates after the change: `uv run pytest` → **879 passed**, `ruff check` clean on the file.

---

Issue #229, opened 2025-08-08
docs: update user manual to reflect all the supported plot types


jooyoungseo:
Current user manual needs to be updated to reflect all the supported plot types in the library. Take a look at our examples.

Issue #170, opened 2025-04-22
docs: update supported plot types in quartodoc


jooyoungseo:
# Feature Request

## Description

We need to update the documentation to include the supported plot types in `quartodoc`. Multi-line plots, stacked and dodged bar plots, multi-panel, multi-layered, and facet plots need to be added: https://github.com/xability/py-maidr/blob/337a65346aca82f4805bbbac850d1ef9e6146702/docs/index.qmd#L47-L89

PR #346, opened 2026-08-10
feat(heatmap): register pcolormesh and pcolor heatmaps

### LINKED ISSUE #337: feat: support pcolormesh and pcolor as heatmaps
## Description

`maidr/patch/heatmap.py` patches only two entry points:

```python
wrapt.wrap_function_wrapper(Axes, "imshow", heat)
wrapt.wrap_function_wrapper("seaborn", "heatmap", heat)
```

`Axes.pcolormesh` and `Axes.pcolor` are not patched, so a heatmap drawn either of those ways registers no MAIDR plot at all.

## Why this matters

`pcolormesh` is one of the most common ways to draw a heatmap in matplotlib — it is the standard choice whenever the grid is non-uniform or the axes carry real coordinates rather than indices, which covers most scientific use. Users hitting this path get silence, with no indication that anything was missed.

## Why it should be cheap

The highlight side is already wired: `maidr/patch/highlight.py:31` patches `QuadMesh.draw`, and `pcolormesh` returns a `QuadMesh`. What is missing is the registration with `FigureManager` and the data extraction.

## Scope

- [ ] Patch `Axes.pcolormesh` and `Axes.pcolor` in `maidr/patch/heatmap.py`
- [ ] Extract cell values plus the x/y edge coordinates from the `QuadMesh`
- [ ] Emit the same `heat` schema as the `imshow` path
- [ ] Tests parametrized alongside the existing heatmap cases

## Difficulty

Low — likely the cheapest single improvement in the current backlog.

jooyoungseo:
## Description

Only `Axes.imshow` and `seaborn.heatmap` were patched, so a heatmap drawn with `pcolormesh` registered nothing at all — no layer, no warning, just silence with nothing to tell the user a chart had been missed. That is not an obscure path: `pcolormesh` is what you reach for whenever the grid is irregular or the axes carry real coordinates rather than array indices, which covers most scientific use.

Closes #337.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

**The extraction side already worked.** `HeatPlot._extract_scalar_mappable_data` reshapes a flattened `QuadMesh` array against its coordinate mesh and tags the mesh for highlighting — it has always taken that path, because `seaborn.heatmap` draws through `pcolormesh`. Only the patch registration was missing, so this is a three-line change plus the consequence below.

**The consequence is worth reviewing carefully.** `heat` never consulted `ContextManager`. It does not go through `common()` — it has to pop `z_label` before the draw rather than pass it through — and so it had none of the recursion guarding every other patch gets for free. With both `seaborn.heatmap` and `Axes.pcolormesh` patched, the inner call would register a **second, identical HEAT layer** for the same axes, and one `sns.heatmap()` would become two heatmaps for the user to navigate between. Added the guard.

This is pinned by a test rather than left to inspection, since it is exactly the kind of regression that reads as "working" in a screenshot:

```python
def test_seaborn_heatmap_still_registers_one_layer():
```

**`pcolor` is patched too** and reads correctly — matplotlib returns a `PolyQuadMesh` whose value array is already two-dimensional, so it needs no reshape and falls through the existing branch intact. It carries no *visual* highlight, because `patch/highlight.py` tags `QuadMesh` and not `PolyCollection`. I did not widen that: `PolyCollection` backs `fill_between`, violins and others, so tagging it is a change with a blast radius well beyond this issue. The difference is asserted rather than left implicit, so a later change that does tag it has to come back here.

## Checklist

## Additional Notes

Verification actually run:

| Step | Result |
|---|---|
| `uv run pytest` | **819 passed**, 0 failures |
| `ruff check` on the changed files | clean |

`ruff check --diff` across the whole repo reports 37 fixable findings, all of them in `example/**` notebooks and all pre-existing — untouched here.

New tests in `tests/core/plot/test_pcolormesh.py` (8 cases) cover both entry points for registration, grid shape, and highlighting support, plus regression cases for `sns.heatmap` and `imshow` still registering exactly one layer each.

One thing the tests caught in review of my own work: `_support_highlighting` starts `True` and is only cleared *while the data is extracted*, so asserting it on a freshly built plot asserts the default and nothing more. The first version of the `pcolormesh` highlighting test passed for that reason and was vacuous; both highlighting tests now run the extraction first.

**Documentation is unticked deliberately.** The user-facing supported-plots list should gain these two entry points, but that list is tracked separately in #229 / #170 and I did not want to half-update it from here.

---

### DIFF
--- a/maidr/patch/heatmap.py
+++ b/maidr/patch/heatmap.py
@@ -5,12 +5,22 @@ import wrapt
 from matplotlib.axes import Axes
 from matplotlib.image import AxesImage
 
+from maidr.core.context_manager import ContextManager
 from maidr.core.enum import PlotType
 from maidr.core.figure_manager import FigureManager
 from maidr.patch.common import _draw_quietly
 
 
 def heat(wrapped, _, args, kwargs) -> Axes | AxesImage:
+    # `seaborn.heatmap` draws through `Axes.pcolormesh`, and both are patched
+    # here. Without this guard the inner call registers a second HEAT layer for
+    # the same axes, so one `sns.heatmap()` would be announced as two identical
+    # heatmaps the user has to navigate between. Every other patch reaches the
+    # same guard through `common()`; this one does not call it, because it has
+    # to pop `z_label` before the draw rather than pass it through.
+    if ContextManager.is_internal_context():
+        return _draw_quietly(wrapped, args, kwargs)
+
     # Check for additional params used by MAIDR heatmap.
     optional_params = {}
     if "z_label" in kwargs:
@@ -19,8 +29,9 @@ def heat(wrapped, _, args, kwargs) -> Axes | AxesImage:
     if "fmt" in kwargs:
         optional_params["fmt"] = kwargs["fmt"]
 
-    # Patch `ax.imshow()` and `seaborn.heatmap`.
-    plot = _draw_quietly(wrapped, args, kwargs)
+    # Patch `ax.imshow()`, `ax.pcolormesh()`, `ax.pcolor()` and `seaborn.heatmap`.
+    with ContextManager.set_internal_context():
+        plot = _draw_quietly(wrapped, args, kwargs)
 
     # Extract the heatmap data points for MAIDR from the plots.
     ax = FigureManager.get_axes(plot)
@@ -30,8 +41,17 @@ def heat(wrapped, _, args, kwargs) -> Axes | AxesImage:
     return plot
 
 
-# Patch matplotlib function.
+# Patch matplotlib functions.
+#
+# `imshow` is not the only way a matplotlib heatmap gets drawn, and it is not
+# the most common one: `pcolormesh` is what you reach for whenever the grid is
+# irregular or the axes carry real coordinates rather than array indices, which
+# covers most scientific use. Until these two were patched such a figure
+# registered nothing at all, and the user got silence with no indication that
+# anything had been missed.
 wrapt.wrap_function_wrapper(Axes, "imshow", heat)
+wrapt.wrap_function_wrapper(Axes, "pcolormesh", heat)
+wrapt.wrap_function_wrapper(Axes, "pcolor", heat)
 
 # Patch seaborn function.
 wrapt.wrap_function_wrapper("seaborn", "heatmap", heat)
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Both non-blocking notes are addressed in `2cee7d5`, and the `fmt` one turned out to be a real defect rather than a theoretical one.

### `fmt` forwarding — reproduced, then fixed

I checked it before changing anything. Passing `fmt` to a matplotlib entry point does not fail cleanly:

```
ax.pcolormesh(x, y, Z, fmt=".2f")
→ AttributeError: QuadMesh.set() got an unexpected keyword argument 'fmt'
```

`pcolormesh` swallows the kwarg into `**kwargs` and hands it to the artist, so the error surfaces from somewhere the caller has no way to connect back to MAIDR — and it is an `AttributeError`, not the `TypeError` an unexpected kwarg usually raises.

`fmt` is now forwarded only to a function that **declares** it. Testing for a declared parameter rather than for `**kwargs` is the point, and it is the trap I would have walked into: all four patched callables accept `**kwargs`, and only `seaborn.heatmap` can use the value. Covered both ways — the three matplotlib entry points accept `fmt=` without raising and still carry it into the schema, and `sns.heatmap(annot=True, fmt=".1f")` still formats its cell annotations with it.

### Return type annotation

Widened to `Axes | AxesImage | Collection`. `Collection` rather than naming `QuadMesh` and `PolyQuadMesh` directly, since `PolyQuadMesh` is not present in every matplotlib version this package supports.

### `z_label` coverage

You are right that it was the one path in the diff without direct coverage. It goes through the same guard as `fmt`, and the `fmt` parametrisation now exercises that guard across all three new entry points, so the specific risk you named — a param being popped or swallowed differently per entry point — is covered. A dedicated `z_label` test would be closing a gap that predates this PR across every heatmap entry point including `imshow`; I would rather not half-close it from here, and it seems better as its own small change alongside the docs work in #229.

### `heat()` docstring

Left as is, deliberately. It is pre-existing, and the file is one of the denser comment-per-line modules in the package already; adding a NumPy-style docstring on top of the block comments now inside the function would duplicate them rather than replace them. Happy to do it as a focused cleanup if you would rather it land here.

### Test file name

Left as `test_pcolormesh.py`. It matches the `test_stepplot.py` precedent you noted — narrow name, slightly broader scope — and renaming now would detach the file from the issue it closes.

### Verification actually run

| Step | Result |
|---|---|
| `uv run pytest` | **823 passed**, 0 failures |
| `ruff check` on the changed files | clean |

---

PR #347, opened 2026-08-11
feat(heatmap): highlight a pcolor grid, and cover z_label

jooyoungseo:
## Description

`Axes.pcolor` renders a `PolyQuadMesh` where `pcolormesh` renders a `QuadMesh`, and `patch/highlight.py` tagged only the latter. A pcolor heatmap therefore read through audio, text and braille but carried **no visual highlight** — low-vision users lost the cursor on a chart every other reader could follow.

This is the follow-up #346 explicitly deferred, along with the two review notes it left open.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

**The scoping is the whole point of this change, so it is worth reviewing first.** The obvious fix — tagging `PolyCollection` — is the wrong one: `PolyCollection` also backs violin bodies and `fill_between`, and tagging it would hand every one of those a maidr gid and a highlight context they were never extracted for. That blast radius is exactly why #346 left this alone.

`PolyQuadMesh` **inherits** `draw` rather than defining one, so wrapping it installs a subclass-only override and the base class keeps the unwrapped method. Verified rather than assumed:

```
PolyCollection.draw wrapped:  False
PolyQuadMesh owns draw now:   True
```

A test asserts both halves, so a later change that reaches for the base class has to come through here.

**Reshape now driven by the array, not the class.** The two mesh classes disagree about the shape they keep values in — `QuadMesh` flattens them, `PolyQuadMesh` keeps the grid — so reshaping unconditionally would fail on the one that is already two-dimensional.

**`z_label` coverage**, the gap the last review named. Parametrized across all three matplotlib entry points, because the lifting happens once before the wrapped call and a per-entry-point regression would be invisible from any single one of them. Plus a default case.

**`heat()` docstring**, NumPy-style per CLAUDE.md, including why this patch cannot route through `common()`.

## Checklist

## Additional Notes

**One assertion in #346 is deliberately inverted here.** `test_pcolor_reads_without_highlighting` pinned the old behaviour and said in its docstring that a change which starts tagging `PolyCollection` "should have to come back and update this". That is what happened; it is now `test_pcolor_supports_highlighting`.

Verification actually run:

| Step | Result |
|---|---|
| `uv run pytest` | **828 passed** (was 823), 0 failures |
| `ruff check` on the changed files | clean |

The full-suite count matters more than usual here: violin plots and `fill_between` both go through `PolyCollection`, so an over-broad patch would have shown up as failures across those suites rather than in the heatmap tests. It did not.

`ruff check maidr/` reports 7 findings repo-wide, all pre-existing — confirmed by stashing this branch and re-running on the base, which reproduces the same 7. Untouched here.

**Documentation unticked deliberately**, same as #346: the supported-plots list is tracked in #229 / #170 and I would rather not half-update it from here.

---

### DIFF
--- a/maidr/core/plot/heatmap.py
+++ b/maidr/core/plot/heatmap.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import numpy.ma as ma
 from matplotlib.axes import Axes
 from matplotlib.cm import ScalarMappable
-from matplotlib.collections import QuadMesh
+from matplotlib.collections import PolyQuadMesh, QuadMesh
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
@@ -64,11 +64,16 @@ class HeatPlot(
             return None
 
         array = sm.get_array().data
-        if isinstance(sm, QuadMesh):
-            # Data gets flattened in ScalarMappable, when the plot is from QuadMesh.
-            # So, reshaping the data to reflect the original quadrilaterals
-            m, n, _ = ma.shape(sm.get_coordinates())
-            array = array.reshape(m - 1, n - 1)  # Coordinates shape is (M + 1, N + 1)
+        if isinstance(sm, (QuadMesh, PolyQuadMesh)):
+            # The two mesh classes disagree about the shape they keep their
+            # values in: `pcolormesh`'s QuadMesh flattens them, while
+            # `pcolor`'s PolyQuadMesh keeps the grid. Reshaping unconditionally
+            # would fail on the one that is already two-dimensional, so the
+            # recovery is driven by the array rather than by the class.
+            if array.ndim == 1:
+                m, n, _ = ma.shape(sm.get_coordinates())
+                # Coordinates shape is (M + 1, N + 1)
+                array = array.reshape(m - 1, n - 1)
 
             # Tag the elements for highlighting
             self._elements.append(sm)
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Confirming the one point you asked about.

**`imshow` still reads without a visual highlight, and that is unchanged here.** It returns an `AxesImage`, which is not a mesh, so it takes the `else` branch and clears `_support_highlighting` exactly as it did before this PR. Nothing in this change touches that path.

It is a real gap rather than a deliberate exclusion, but a differently shaped one: `AxesImage` is a single raster, not a collection of per-cell artists, so there is no cell element to tag. Highlighting it would mean synthesising an overlay rectangle per cell from the image extent — a different mechanism from the gid-tagging every other path here uses, and well outside a change about `pcolor`. Worth its own issue; happy to file one.

The title says "pcolor grid" specifically for that reason, and `test_pcolormesh.py` now pins all three entry points' behaviour side by side, so the asymmetry is visible in the tests rather than only in the code.

On the end-to-end SVG round-trip: agreed it is the coverage that would actually prove the `maidr` attribute lands on a rendered `pcolor` element, and agreed it is neither present before nor a regression. The `draw`-wrapping mechanism is pinned via `__dict__` membership precisely because that is the part most likely to break silently under a matplotlib upgrade, but that is a proxy for the round-trip rather than the round-trip itself.

---

Issue #341, opened 2026-08-10
feat: support hist2d and hexbin


jooyoungseo:
## Description

Neither `Axes.hist2d` nor `Axes.hexbin` is patched, so 2D density plots produce no MAIDR output. `sns.kdeplot` with both `x` and `y` (bivariate KDE) has the same problem — `maidr/patch/kdeplot.py` handles the univariate case only.

## Why hist2d is nearly free

A rectangular 2D histogram is a heatmap. The core `heat` trace needs no change; the work is extraction — bin edges and counts — and reusing the existing heatmap schema path. This can ship well ahead of the hexbin half.

## hexbin

Needs core support for the offset lattice, tracked at xability/maidr#799. Worth splitting out of this issue if `hist2d` lands first.

## Scope

- [ ] Patch `Axes.hist2d`, emitting `heat` with the bin edges as axis coordinates
- [ ] Handle bivariate `sns.kdeplot` (filled and contour forms)
- [ ] Patch `Axes.hexbin` once xability/maidr#799 lands
- [ ] Announce the count and the bin's coordinate range, not a bin index

## Difficulty

`hist2d` low; `hexbin` medium and gated on the core work.

jooyoungseo:
Went to implement this and found the three halves are in quite different states. Verified each against the current `main` (`cbb13c0`) rather than reasoning from the source.

### `hist2d` — already works, as of #347

`Axes.hist2d` draws through `Axes.pcolormesh`, which #347 patched for its own sake. So a rectangular 2D histogram already registers, with correct bin counts and working highlighting:

```
hist2d -> layers: 1, type: heat
   points: [[12.0, 45.0, 30.0], [22.0, 60.0, 31.0]]   highlight: True
```

Nobody decided that; it fell out of a neighbouring change. Which makes it worth pinning rather than closing quietly — #348 adds the regression tests, including one for the transposition (`hist2d` returns counts indexed `[x, y]` while the mesh it draws is `[y, x]`, so a grid the wrong way round would still navigate and just describe a different chart).

### `hexbin` — blocked on core

```
hexbin -> KeyError: No MAIDR found for figure
```

It renders a `PolyCollection` of hexagons: neither a mesh nor a rectangular grid, so nothing in the heatmap path reaches it. Needs the offset-lattice trace tracked at xability/maidr#799 before there is anything for this binding to emit into.

### Bivariate `kdeplot` — blocked on core, and on a different piece

```
[...]

PR #348, opened 2026-08-11
test(heatmap): pin hist2d, which reads only by accident

jooyoungseo:
## Description

I went to implement `hist2d` support for #341 and found it already works — so this PR is tests, not a feature.

`Axes.hist2d` draws through `Axes.pcolormesh`. Patching the latter in #347 made a rectangular 2D histogram register as a heatmap, with correct bin counts and working highlighting, **without anyone deciding it should**:

```
hist2d -> layers: 1, type: heat
   points: [[12.0, 45.0, 30.0], [22.0, 60.0, 31.0]]   highlight: True
```

That is good news and a liability at the same time. Something that works by accident breaks silently: narrowing which entry points the heatmap patch wraps, or adding a guard that skips nested draws, would take this away with no error and no warning — a chart that used to be navigable simply stops being, and nothing in the suite would say so.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Test coverage for behaviour that currently has none

## Changes Made

`tests/core/plot/test_hist2d.py`, 7 cases.

**What is pinned:**

- The layer registers, once, as `heat`.
- The mesh is tagged, so the cell under the cursor highlights.
- **The transposition.** `hist2d` returns its counts indexed `[x, y]` while the mesh it draws is `[y, x]`, so the test asserts `points == counts.T.tolist()`. A grid that came back the wrong way round would still navigate perfectly well — it would just describe a different chart, which is precisely the kind of bug no smoke test catches.

**What is pinned as *not* working**, so the boundary sits in one place rather than being discovered per bug report:

- `Axes.hexbin` — renders a `PolyCollection` of hexagons, neither a mesh nor a rectangular grid. Blocked on core hexagonal-lattice support (xability/maidr#799).
- `sns.kdeplot(x=, y=)`, filled and unfilled — renders contours. A contour plot is not a heatmap with different colours: the **level** is the navigable object, not the cell, so describing one as a grid would be a different chart rather than an approximate one. Blocked on core contour support (xability/maidr#802).

Those negative assertions are written to fail loudly the day either lands, which is the reminder to rewrite them as positive ones.

## Checklist

## Additional Notes

**#341 should be narrowed rather than closed.** Its `hist2d` half is done — by #347, incidentally — and this PR pins it. Its `hexbin` and bivariate-KDE halves are each blocked on a different piece of core work. I will comment there with the same breakdown so the issue reflects what is actually left.

Verification actually run:

| Step | Result |
|---|---|
| `uv run pytest` | **835 passed** (was 828), 0 failures |
| `ruff check` on the new file | clean |

---

### DIFF
--- a/tests/core/plot/test_hist2d.py
+++ b/tests/core/plot/test_hist2d.py
@@ -1,0 +1,150 @@
+"""Tests for 2D density plots.
+
+``Axes.hist2d`` reads today, and reads by accident: it draws through
+``Axes.pcolormesh``, which was patched for its own sake, so a rectangular 2D
+histogram started registering as a heatmap without anyone deciding it should.
+
+That is worth pinning rather than leaving implicit. A change to the heatmap
+patch -- narrowing which entry points it wraps, say, or adding a guard that
+skips nested draws -- would take this away silently: no error, no warning, just
+a chart that used to be navigable and no longer is. These tests make that a
+failure instead.
+
+The other two 2D density paths do NOT read, and are asserted here too so the
+boundary is visible in one place rather than discovered per bug report:
+
+- ``Axes.hexbin`` renders a ``PolyCollection`` of hexagons, which is neither a
+  mesh nor a rectangular grid.
+- ``sns.kdeplot(x=, y=)`` renders filled or line contours, which need a
+  contour trace to describe levels rather than cells.
+"""
+
+from __future__ import annotations
+
+import matplotlib
+
+matplotlib.use("Agg")
+
+import matplotlib.pyplot as plt  # noqa: E402
+import numpy as np  # noqa: E402
+import pytest  # noqa: E402
+import seaborn as sns  # noqa: E402
+
+import maidr  # noqa: F401,E402  # activates patches
+from maidr.core.enum.plot_type import PlotType  # noqa: E402
+from maidr.core.figure_manager import FigureManager  # noqa: E402
+
+
+#: Two well-separated clusters, so the bin counts are lopsided and a
+#: transposed or mis-shaped extraction cannot pass by symmetry.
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Good catch on `test_hist2d_supports_highlighting` — that assertion was weaker than its own docstring. Fixed in 5a33506.

You traced it correctly: `_support_highlighting` is set `True` in `MaidrPlot.__init__` and only ever cleared, in the `else` of `HeatPlot._extract_scalar_mappable_data`. So the assertion was reporting *which branch the extraction took*, not whether anything got tagged. Drop the `self._elements.append(sm)` inside that branch and the flag stays `True`, the test stays green, and highlighting breaks in silence — which is exactly the failure a test called "supports highlighting" exists to catch. Added the `plot.elements` assertion alongside it, matching the guard `test_pcolormesh.py` already carries, and expanded the docstring to say why both halves are there rather than leaving the second looking redundant.

On the two nits:

- **`test_hist2d_registers_exactly_one_layer` overlapping the type test** — intentional, and I'd like to keep it. The two fail for different reasons: the count in the first test is incidental (I need one plot to index into for the type check), while the second is specifically about the `ContextManager` recursion guard, since `hist2d` draws through `pcolormesh` and both would register if that guard regressed. The docstrings carry that distinction, and a failure naming the recursion guard is worth more than one naming a type mismatch when the guard is what actually broke.
- **Shared fixture/helper duplication** — agreed it wants a `tests/core/plot/conftest.py`, and agreed it shouldn't happen here. There are only two files sharing it so far; I'd rather extract it when a third arrives and the shape is settled than pull test-only infrastructure into a PR whose point is pinning behaviour.

CI is green on the added assertion.

---

PR #349, opened 2026-08-11
feat(errorbar): read an estimate together with its interval

### LINKED ISSUE #340: feat: support error bars (Axes.errorbar)
## Description

`Axes.errorbar` is not patched. Any chart carrying uncertainty — the majority of published statistical figures — loses that information entirely when read through MAIDR.

## Scope

- [ ] New `maidr/patch/errorbar.py` patching `Axes.errorbar`
- [ ] `PlotType.ERRORBAR` in `maidr/core/enum/plot_type.py`
- [ ] Extract centre plus lower/upper bounds; matplotlib returns an `ErrorbarContainer` whose caps and bars are separate artists, so the extraction needs care for asymmetric and one-sided errors
- [ ] Respect the orientation (`yerr` vs `xerr`)
- [ ] Tests covering symmetric, asymmetric and one-sided errors

## Related

- xability/py-maidr#246 (`sns.pointplot`) produces error bars through this same path and should share the extraction.
- Blocked on core support: xability/maidr#789.

## Difficulty

Medium — the matplotlib container layout is the fiddly part, not the schema.

jooyoungseo:
## Description

Closes #340. The Python binding for the `error_bar` trace type, which landed in core as xability/maidr#819.

`Axes.errorbar` was not patched, so any chart carrying uncertainty — the majority of published statistical figures — lost that information entirely. The estimate was announced and the interval dropped, so the comparison the chart was drawn to support could not be made at all.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

`PlotType.ERRORBAR` (wire value `error_bar`, matching core), `maidr/patch/errorbar.py`, `maidr/core/plot/errorbar.py`, factory dispatch, and 18 tests.

### Bounds come from the drawn geometry, not from `yerr`

The issue called the matplotlib container layout "the fiddly part, not the schema", and that is exactly right — so I probed it rather than reasoning about it. The decision that follows from what I found: **read the rendered `LineCollection`, not the caller's `yerr`.**

Those are different quantities. matplotlib takes an *offset*; the MAIDR schema carries an *absolute position*. And the offset arrives in three shapes — scalar, `(N,)`, `(2, N)` — before `uplims`/`lolims` change what the bar means again. The collection matplotlib actually rendered has already resolved every one of those into two endpoints, so reading it is both shorter and correct for cases the arithmetic route would have to enumerate one at a time.

### What the probe turned up, and what each cost

| Case | Finding | Handling |
|---|---|---|
| `fmt="none"` | Container has **no data line** | Centres come from the call arguments. They are genuinely unrecoverable from geometry — an asymmetric bar is *not* centred on its own midpoint |
| `yerr` containing `NaN` | Renders an **empty segment that keeps its position** | Samples stay aligned; that one is emitted with no bounds. Emitting `NaN` would poison the trace's whole pitch range |
| `xerr` **and** `yerr` | Two collections, **x first** | `lines[2][0]` unconditionally would have silently described the x interval on every such chart |
| Two `errorbar` calls | Two containers on one axes | The patch hands over the container **its own call produced** |
| `uplims` / zero on one side | Bar stops at the estimate | Falls out of reading geometry with no special-casing — the reason for reading geometry |
| Categorical x | Labels survive on the data line | Announced by name, not tick position |

The two-container one is worth calling out: a layer that looked up "the" container on the axes would find the first one **both times**, describing the first series twice and dropping the second with no error anywhere. Passing the container from the patch removes the ambiguity rather than papering over it.

### Float noise

A derived bound is not authored. matplotlib draws at `y - err`, and `4.2 - 0.4` is `3.8000000000000003` — which a screen reader spells out in full. Cleaned to 12 significant figures, *not* fixed decimals: rounding to two would report a bound of 0.003 as zero, a worse answer than the noise. The estimate itself is left exact, because nothing subtracted it — only the endpoints arithmetic touched are cleaned.

## A correction to the issue

#340 says #246 (`sns.pointplot`) "produces error bars through this same path and should share the extraction." **That is not true of current seaborn** — `sns.barplot` and `sns.pointplot` render their confidence intervals themselves and never reach `Axes.errorbar`. So this PR does not advance #246, and #246 will need its own extraction.

I've pinned that as a parametrized regression test rather than leaving it as a claim about another library's internals: if seaborn ever does route through `Axes.errorbar`, users would silently get a **second layer** describing the same chart, and that test is what says so.

## Checklist

## Verification actually run

| Step | Result |
|---|---|
| `uv run pytest` | **853 passed** (was 835), 0 failures |
| `uv run pytest tests/core/plot/test_errorbar.py` | **18 passed** |
| `ruff check` on every changed file | clean |

Depends on xability/maidr#819 (merged) for the core trace type.

---

### DIFF
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -617,6 +617,46 @@ ax.yaxis.set_major_formatter("{x:,.0f}")
 plt.show() #<<
 ```
 
+### Error Bar Plot
+
+Uncertainty is usually the finding, not the decoration: whether two group means
+differ is answered by whether their intervals overlap. `ax.errorbar()` draws
+that, and maidr reads the estimate **and** its interval, so the comparison the
+chart was drawn to support can be made by ear.
+
+Navigation is a grid. Left and right walk the samples; up and down at one
+sample walk the lower bound, the estimate and the upper bound — so a reader
+traces one interval rather than rebuilding it from three passes over the chart.
+
+The bounds are read off the drawn bars rather than from the `yerr` you passed,
+so asymmetric errors, one-sided errors, `uplims`/`lolims` and `fmt="none"` all
+read correctly without you having to present them any particular way.
+
+```{python}
+#| fig-alt: Mean response by dose with 95% confidence intervals
+
+import matplotlib.pyplot as plt
+
+import maidr #<<
+
+groups = ["control", "low dose", "high dose"]
+means = [4.2, 5.1, 7.3]
+
+# Asymmetric intervals, as a real confidence interval usually is: the lower
+# and upper distances from the mean are given separately.
+lower = [0.4, 1.1, 0.2]
+upper = [0.4, 1.5, 0.1]
+
+fig, ax = plt.subplots(figsize=(8, 5))
+ax.errorbar(groups, means, yerr=[lower, upper], fmt="o", capsize=5)
+
+ax.set_title("Mean Response by Dose, with 95% Confidence Intervals")
+ax.set_xlabel("Group")
+ax.set_ylabel("Response")
+
+plt.show() #<<
+```
+
 ### Scatter Plot
 
 ```{python}
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -2,7 +2,7 @@
 
 > py-maidr is a Python library that makes matplotlib, seaborn, and plotly data visualizations accessible to blind and low-vision users through sonification, braille, text, and AI-powered conversational modalities. Developed by the (x)Ability Design Lab at the University of Illinois Urbana-Champaign.
 
-maidr (pronounced "mader") monkey-patches matplotlib, seaborn, and plotly at import time so users just `import maidr` and call `maidr.show(plot)` to generate interactive HTML with keyboard-navigable sonification, braille, text, and AI chat support. Supported plot types: bar, stacked bar, dodged bar, count, histogram, KDE, line, multi-line, step, pie, box, violin, heatmap, scatter, smooth/regression, candlestick, multi-panel, and facet plots. Works in Jupyter, Colab, VS Code, Streamlit, Shiny, and Quarto.
+maidr (pronounced "mader") monkey-patches matplotlib, seaborn, and plotly at import time so users just `import maidr` and call `maidr.show(plot)` to generate interactive HTML with keyboard-navigable sonification, braille, text, and AI chat support. Supported plot types: bar, stacked bar, dodged bar, count, histogram, KDE, line, multi-line, step, pie, box, violin, heatmap, scatter, smooth/regression, candlestick, error bar, multi-panel, and facet plots. Works in Jupyter, Colab, VS Code, Streamlit, Shiny, and Quarto.
 
 ## Docs
 - [Documentation Home](https://py.maidr.ai/): Installation, quick start, and overview
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -8,6 +8,7 @@ class PlotType(str, Enum):
     BOX = "box"
     COUNT = "count"
     DODGED = "dodged_bar"
+    ERRORBAR = "error_bar"
     HEAT = "heat"
     HIST = "hist"
     LINE = "line"
@@ -45,6 +46,7 @@ class PlotType(str, Enum):
 #: plot, and callers de-duplicate.
 _DISPLAY_NAMES = {
     PlotType.DODGED: "dodged bar",
+    PlotType.ERRORBAR: "error bar",
     PlotType.HEAT: "heatmap",
     PlotType.HIST: "histogram",
     PlotType.SCATTER: "scatter",
--- a/maidr/core/figure_manager.py
+++ b/maidr/core/figure_manager.py
@@ -58,6 +58,7 @@ class FigureManager:
         PlotType.CANDLESTICK: 1,
         PlotType.VIOLIN_KDE: 1,
         PlotType.VIOLIN_BOX: 1,
+        PlotType.ERRORBAR: 1,
     }
 
     _instance = None
--- a/maidr/core/plot/errorbar.py
+++ b/maidr/core/plot/errorbar.py
@@ -99,9 +99,17 @@ class ErrorBarPlot(MaidrPlot, ContainerExtractorMixin, DictMergerMixin):
             self._elements.append(bars)
 
         # The category runs along the axis the bars do NOT span, and the
-        # magnitude along the one they do. The schema names them x and y in
-        # every orientation and lets `orientation` say which is on screen
-        # where, matching how the box plot already travels.
+        # magnitude along the one they do. The schema names them `x` and `y`
+        # in BOTH orientations, and lets `orientation` say which is on screen
+        # where.
+        #
+        # That differs from how a bar or a histogram travels, and deliberately
+        # so: the shape is set by the consumer. `ErrorBarTrace` reads the
+        # magnitude as `y`/`yMin`/`yMax` with no orientation branch, and
+        # `ErrorBarPoint` declares no `xMin`/`xMax` to put a bound in --
+        # so emitting the screen-aligned form a bar uses would leave a
+        # horizontal chart with no interval at all, which is the one thing the
+        # trace type exists to convey.
         categories, values = (xs, ys) if is_vertical else (ys, xs)
         component = 1 if is_vertical else 0
 
--- a/maidr/core/plot/errorbar.py
+++ b/maidr/core/plot/errorbar.py
@@ -1,0 +1,290 @@
+from __future__ import annotations
+
+from typing import Any, Sequence
+
+import numpy as np
+from matplotlib.axes import Axes
+from matplotlib.collections import LineCollection
+from matplotlib.container import ErrorbarContainer
+
+from maidr.core.enum import MaidrKey, PlotType
+from maidr.core.plot import MaidrPlot
+from maidr.exception import ExtractionError
+from maidr.util.mixin import ContainerExtractorMixin, DictMergerMixin
+
+
+class ErrorBarPlot(MaidrPlot, ContainerExtractorMixin, DictMergerMixin):
+    """
+    A plot that draws an estimate together with the interval around it.
+
+    Uncertainty is not decoration in a statistical graphic; it is frequently
+    the finding. Whether two group means differ is answered by whether their
+    intervals overlap, and before this existed a MAIDR reader got the estimate
+    and nothing else, so the comparison the chart was drawn to support could
+    not be made at all.
+
+    The bounds are read off the **drawn geometry** rather than recomputed from
+    the ``yerr`` the caller passed. Those are not the same quantity:
+    matplotlib takes an *offset* while the MAIDR schema carries an *absolute
+    position*, and the offset form has three shapes -- scalar, ``(N,)``, and
+    ``(2, N)`` -- before ``uplims``/``lolims`` change what the bar means again.
+    The ``LineCollection`` matplotlib actually rendered has already resolved
+    every one of those into two endpoints, so reading it is both shorter and
+    correct for cases this module would otherwise have to enumerate.
+
+    Parameters
+    ----------
+    ax : Axes
+        The axes the error bars were drawn on.
+    **kwargs
+        ``container`` is the ``ErrorbarContainer`` the patched call returned;
+        ``x`` and ``y`` are the centre coordinates the caller passed, used
+        only when the container carries no data line.
+
+    See Also
+    --------
+    MaidrPlot : The base class for MAIDR plot data objects.
+    """
+
+    def __init__(self, ax: Axes, **kwargs) -> None:
+        super().__init__(ax, PlotType.ERRORBAR)
+
+        # The patch hands over the exact container its call produced. Looking
+        # one up on the axes instead would break the moment a figure carries
+        # two `errorbar` calls: both layers would find the first container and
+        # describe the same series twice, silently losing the second.
+        self._container = kwargs.pop("container", None)
+        self._fallback_x = kwargs.pop("x", None)
+        self._fallback_y = kwargs.pop("y", None)
+
+        # Resolved from the drawn container during extraction, which `render`
+        # runs before it reads this.
+        self._orientation = "vert"
+
+    def render(self) -> dict:
+        """
+        Build the layer schema, adding the orientation the bars were drawn in.
+
+        Returns
+        -------
+        dict
+            The MAIDR layer schema.
+        """
+        base_schema = super().render()
+        return self.merge_dict(
+            base_schema, {MaidrKey.ORIENTATION: self._orientation}
+        )
+
+    def _extract_plot_data(self) -> list[dict]:
+        container = self._resolve_container()
+        if container is None:
+            raise ExtractionError(self.type, self.ax)
+
+        # `has_yerr` decides the value axis, and an errorbar carrying neither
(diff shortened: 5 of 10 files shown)


jooyoungseo:
Both minor points were real — one of them more serious than "minor". Fixed in d8ddab4. The main correctness finding I checked carefully and it does **not** hold; details and evidence below.

## The date axis was a genuine bug, and worse than pre-existing

You flagged this as "possibly a pre-existing limitation shared with other plot types". It isn't — I ran it, and error bars were the only patched type that *crashed*:

| Call | Before this fix |
|---|---|
| `ax.errorbar(dates, y, yerr=…)` | **`TypeError: float() argument must be a string or a real number, not 'datetime.date'`** |
| `ax.plot(dates, y)` | reads — `x: '01-01 00'` |
| `ax.scatter(dates, y)` | reads — `x: 20454.0` (the ordinal; degraded, but not fatal) |

So this took out the user's entire figure over an axis matplotlib is perfectly happy with, and error bars on a time series are common. `_scalar` now falls back to `str(value)`, which the schema allows for `x` and which reads better than the bare ordinal a scatter emits. The magnitude and its bounds stay numeric, so sonification is unaffected. Pinned in `test_a_date_axis_does_not_break_the_figure`.

Good catch — thank you for not letting me wave it through as pre-existing.

## `PLOT_TYPE_PRIORITY` — agreed, added explicitly

The implicit `0` made the figure's overall `plot_type` depend on layer *order*: `errorbar` then `bar` promoted to `BAR`, while `bar` then `errorbar` stayed `BAR`. Every other ungrouped type sits at `1`, where ordering doesn't matter. `PlotType.ERRORBAR: 1` now.

## The `x`/`y` orientation finding does not hold

This one I want to be precise about, because the change you propose would **break horizontal error bars rather than fix them**.

The schema's shape is set by its consumer, and the consumer here is `ErrorBarTrace` in maidr core (xability/maidr#819). It reads the magnitude with no orientation branch at all:

```ts
function magnitudeOf(point: ErrorBarPoint, section: Section): number {
  switch (section) {
    case 'lower': return point.yMin ?? Number.NaN;
    case 'upper': return point.yMax ?? Number.NaN;
    default:      return point.y;
  }
}
```

And `ErrorBarPoint` defines exactly four fields — `x`, `y`, `yMin`, `yMax`. There is no `xMin`/`xMax`; the shape you propose isn't expressible in the grammar. Emitting it would leave `magnitudeOf` reading `undefined ?? NaN` for both bounds, so a horizontal chart would lose its interval entirely — the one thing the trace type exists to convey.

That the JS side reads `yMin` as the bound *in horizontal orientation specifically* is pinned by a passing test in that repo (`test/model/errorBar.test.ts`):

```ts
test('reads the same magnitudes as the vertical chart', () => {
  expect(nonEmptyState(horizontal(0, 0)).text.cross.value).toBe(3.8);  // yMin
  expect(nonEmptyState(horizontal(2, 0)).text.section).toBe('upper bound');
});
```

**Why it differs from `BarPlot`/`HistPlot`:** because their JS consumers differ. `AbstractBarPlot` does `toBarValue(orientation === VERTICAL ? point.y : point.x)` — it genuinely swaps which key it reads, so py-maidr's bar emission is screen-aligned and correct. `ErrorBarTrace` does not swap; it keeps `x` = category and the `y` family = magnitude in both orientations and uses `orientation` only to swap the *labels* and the stereo panning. That was deliberate on the core side, for the reason documented on `Candlestick.dimension`: `AutoplayState` is keyed by direction, so transposing the navigation grid mis-paces autoplay and mis-clamps the movement bounds.

You're right that my "matching how the box plot already travels" was a bad citation — `BoxPlot` has no `x`/`y` coordinate fields, so it doesn't establish anything. I've dropped that comparison; the real justification is the consumer, and I've made the code comment say so instead of gesturing at a precedent.

Worth noting the inconsistency you identified is real at the *ecosystem* level — two trace types resolve orientation differently — it just isn't resolvable in this PR, since changing it means changing the JS trace, the grammar, and every binding together. Happy to open an issue on maidr core to unify it if you think that's worth doing.

## Non-blocking, acknowledged

Repeating `ax.hist2d(...)`-style setup across tests: same trade-off as the sibling files, each test reading standalone. Left as is for consistency.

**Verification:** 854 passed (was 853), `ruff check` clean on every changed file.

---

jooyoungseo:
Two reviews landed on this; taking both together. Four changes in e7561ef, and one claim that doesn't hold.

## Fixed: the highlighting gap (both reviews, and the best catch here)

Verified before fixing:

```
with yerr:        _support_highlighting=True  elements=2  selector_in_schema=True
no error at all:  _support_highlighting=True  elements=0  selector_in_schema=True   ← bug
```

A bare `ax.errorbar(x, y)` renders no bar collection, so nothing ever carries the `maidr` attribute — yet the layer still emitted a selector pointing the frontend at paths the document doesn't contain. Now cleared the way `HeatPlot` clears it when the artist isn't a mesh, and pinned in `test_a_call_with_no_error_promises_no_highlight`. You were right that the existing test only exercised the `yerr` path.

## Fixed: `_resolve_container`'s fallback (second review, point 1)

Agreed, and this was the sharpest observation — the fallback *was* the bug the PR argues against, sitting one refactor away from being reachable. Removed rather than commented: it now returns the patch-supplied container or nothing, and the caller raises. Pinned by constructing an `ErrorBarPlot` outside the patch and asserting it refuses to guess rather than describing the first series twice.

## Fixed: the unreachable `get_axes` branch (both reviews)

Removed rather than defended. `instance` is the bound `Axes` by construction for `wrap_function_wrapper(Axes, "errorbar", …)`, and as you both noted `get_axes` has no case for an `ErrorbarContainer` anyway — so the branch could only ever have produced a bare `"No plot found."` The comment now says why it reads `instance` directly.

## Fixed: `capsize` (second review, point 5)

Fair — it's the one permutation that changes how many artists the container holds without changing the interval, and my probing had covered it while my *tests* hadn't. Added.

## Partly right: the line-length flag

The line **was** 89 characters against the configured `line-length = 88`, so I've wrapped it. But `ruff check` does not catch it and CI would not have failed — E501 isn't in this project's enabled rule set, which is why both the local run and the green `lint` job on the previous commit passed. Worth knowing so the next such report isn't trusted to be caught automatically.

## Checked and deliberately not changed

**Element accumulation across renders.** While verifying the highlighting gap I noticed `elements` grows per `render()` call (1 → 2 → 3). I checked whether I'd introduced it — I hadn't: `ScatterPlot` accumulates identically (3 after two renders). It's the established behaviour of the base class, so fixing it here would be an unrelated change riding along in an error-bar PR. Flagging it as a real thing someone may want to look at separately.

**Timestamp variants** (`datetime.datetime`, `pd.Timestamp`, `np.datetime64`): correct that their `str()` differs from `datetime.date`'s. All of them read rather than raising, which was the actual bug; normalising them consistently is a cross-cutting concern — `ScatterPlot` emits raw ordinals, `MultiLinePlot` uses the tick formatter — so it belongs in its own change rather than being solved for one plot type.

**matplotlib's x-before-y collection ordering**: agreed it's an internal detail. Both reviews noted the test pins it so a reordering fails loudly rather than silently swapping axes, which is the mitigation available short of not using it.

**No `example/errorbar/` directory**: happy to add if wanted, though it looked out of scope for this PR.

**Verification:** 857 passed (was 854), `ruff check` clean on every changed file, no line over 88 characters.

---

jooyoungseo:
All four points addressed in 3d91ee0.

## `uplims`/`lolims` — you were right that the narrative outran the test

This was the sharpest catch. My PR body claimed these "fall out of reading geometry with no special-casing", but the closest test exercised a **zero offset**, which is a different thing: `uplims`/`lolims` don't change the offset at all, they change which side of the estimate the bar is drawn on and swap the cap for an arrow. So the claim was unpinned.

Now parametrized over both, asserting the drawn bar:

| kwarg | lower bounds | upper bounds |
|---|---|---|
| `uplims=True` | `[3.7, 4.6, 6.8]` | `[4.2, 5.1, 7.3]` ← coincides with the estimate |
| `lolims=True` | `[4.2, 5.1, 7.3]` | `[4.7, 5.6, 7.8]` |

They passed first run, so the claim was true — it just wasn't being checked.

## Docs — added, both places

`docs/llms.txt` enumerates supported plot types and had genuinely fallen out of date with the package, so that's a correctness fix rather than a nicety. Added a gallery section to `docs/examples.qmd` too, since every other supported type has one. The example uses asymmetric intervals deliberately — that's the case distinguishing "read the drawn bars" from "re-derive from `yerr`", so the example demonstrates the design rather than just producing a picture.

## The comment/code inconsistency — comment corrected, branch kept

Fair hit: the comment asserted `instance` "is the bound Axes, always" while the code still branched. I kept the branch and fixed the comment, rather than the reverse. As you noted it mirrors `lineplot.py`; the comment now says that explicitly and says what it degrades to, instead of claiming unreachability it can't guarantee for a wrapper installed on some other holder of the method.

## Axis-semantics contrast — added to the module docstring

Agreed this is the thing a maintainer skimming `errorbar.py` and `barplot.py` side by side will trip on. The module docstring now states it up front: `x`/`y` mean category/magnitude here in **both** orientations, unlike `BarPlot`/`HistPlot`, because `ErrorBarTrace` reads the magnitude as `y`/`yMin`/`yMax` with no orientation branch and `ErrorBarPoint` declares no `xMin`/`xMax` to hold a bound. Previously that reasoning only lived at the extraction site.

**Verification:** 859 passed (was 857), 24 in the error-bar file (was 22), `ruff check` clean, no line over 88 characters.

---

jooyoungseo:
I ran your exact scenario before answering, because if it held it would transpose every horizontal chart. **It does not hold** — but the coverage gap you identified is real, and is now closed in e60e906.

## What actually happens

Your example, emitted:

```
orientation: horz
axes:    {"x": {"label": "Response"}, "y": {"label": "Group"}}
data[0]: {"x": "control", "y": 4.2, "yMin": 3.9, "yMax": 4.5}
```

You're right that this looks transposed. The resolution is that `ErrorBarTrace` swaps **which label goes with which value**, precisely when the layer is horizontal:

```ts
main:  { label: isHorizontal ? this.yAxis : this.xAxis, value: point.x }
cross: { label: isHorizontal ? this.xAxis : this.yAxis, value: sectionValues[row][col] }
```

So the announcement is:

| | label | value |
|---|---|---|
| `main` | `yAxis` = **Group** | `point.x` = **control** |
| `cross` | `xAxis` = **Response** | magnitude = **4.2** |

"Group: control", "Response: 4.2" — correct. The `"Response: control"` reading assumes a consumer that pairs `axes.x` with `data.x` positionally, and the one consumer this type has does not.

That swap isn't incidental; it's the counterpart of the orientation-invariant data. Screen-aligned labels (as every other plot type emits) plus orientation-invariant data plus the swap in the trace is one coherent design — it's just spread across two repositories, which is exactly why it read as a bug.

## The gap was real

No test asserted both halves *together*: `test_horizontal_error_reports_horz_orientation` checked data and orientation but not labels, so a change to either half alone would have transposed every horizontal announcement with nothing failing. `test_horizontal_axes_labels_pair_with_the_swapped_data` now pins the labels **and** the data in one test, with the trace's pairing written out in the docstring so the next reader finds the answer where the confusion is rather than in review history.

## `np.datetime64`

Fair — `str(np.datetime64(...))` gives `'2026-01-01T00:00:00.000000000'`, which is ugly rather than wrong. It reads instead of crashing, which was the actual bug; normalising timestamp types consistently is cross-cutting (`ScatterPlot` emits raw ordinals, `MultiLinePlot` uses the tick formatter) and belongs in its own change rather than being solved for one plot type.

## Dead-code fallback

Agreed and already adjusted after the earlier review: the branch stays (matching `lineplot.py`) and the comment no longer claims unreachability it can't guarantee.

**Verification:** 25 tests in the error-bar file, full suite green, `ruff check` clean.

---

jooyoungseo:
Thanks — three of the four actioned in de19c56.

## Timestamp coverage (suggestion 4) — the one worth having

You're right that `datetime.date` is not what real code produces. `ax.errorbar(df.index, …)` hands back a `pd.Timestamp` or an `np.datetime64`, so pinning only `date` left the crash this fallback exists to prevent reachable by exactly the two types most likely to appear. Both now parametrized, and both read rather than raising.

Their `str()` forms genuinely differ from each other — one carries a time component — so the assertion is that the label is a non-empty string naming the year, not one exact spelling. Asserting a single literal would have been a test that only passed for whichever type I happened to write it against.

## Redundant `bool()` (3) and the `render()` ordering comment — both taken

`has_xerr`/`has_yerr` are already booleans; dropped. And you're right the ordering was load-bearing but implicit: `super().render()` is what runs `_extract_plot_data`, which is what resolves `self._orientation`, so the read has to come after. Said so at the call site.

## Docstring duplication (1) — fair, and partly deliberate

Genuine trade-off. The repetition isn't accidental — each site answers the question a reader has *at that site* ("why isn't this arithmetic on `yerr`?" reads differently in the patch than in the extractor). But you're right that five copies of one rationale is a sync burden, and I'd rather not defend prose volume as a feature.

I've left it for now rather than doing a broad trim in a PR that's already answered several rounds of review — it's a pure-prose change with no test coverage to catch a mistake, and it'd be better as its own pass. Happy to do that pass if you'd like it before merge.

## `_resolve_container` passthrough (2) — keeping it

Deliberate. It's the seam an alternative construction path would reach for, and a method whose docstring says *why there is no fallback* is more likely to survive a future refactor than a comment on an attribute several lines away. The previous review round caught that the fallback contradicted the design; making its absence explicit and testable (`test_a_layer_built_without_its_container_refuses_to_guess`) is what keeps that from regressing.

**Verification:** 862 passed (was 859), 27 in the error-bar file (was 25), `ruff check` clean.

---

jooyoungseo:
Both concrete nits taken in c11b87c. One of them I checked before writing the test, and the speculation didn't hold — worth recording which way.

## `fmt="none"` + datetime (nit 2) — tested, and the concern doesn't reproduce

You reasoned `np.atleast_1d` might coerce the dates to `datetime64`, giving `fmt="none"` a different label format from the ordinary path. I ran both:

```
normal path: ['2026-01-01', '2026-01-02', '2026-01-03']
fmt=none:    ['2026-01-01', '2026-01-02', '2026-01-03']
```

They agree — `np.atleast_1d` on a list of `datetime.date` keeps object dtype, so `str()` produces the same clean form. But the reasoning behind the concern was sound, and the composition is exactly the kind of thing that regresses quietly: if that coercion ever changed, one chart would read two different ways depending on its marker style. Pinned.

## `test_axes_schema.py` (nit 4) — added, in both orientations

You're right this closes the loop cheaply. I added the horizontal case too, deliberately: this layer's *data* is orientation-invariant while its *axis labels* are not, so the canonical shape needs asserting for the orientation where those two disagree, not just the one where they happen to line up. A single vertical case would have passed without exercising the interesting half.

## matplotlib collection ordering (nit 1)

Agreed on all counts — public type, undocumented ordering, pinned by a test that fails loudly rather than swapping axes silently. Nothing further to do beyond watching it across matplotlib bumps, which is what the test is for.

## Docstring verbosity (nit 3)

Second reviewer to raise it, so I'll stop defending the volume. You're right that the case-by-case table in the class docstring is PR-description material that ended up permanent. I've left it in this PR rather than doing a prose-only sweep at the end of several review rounds — it's the kind of edit with no test coverage to catch a slip, and it'd be cleaner as its own pass. Happy to do that pass now if you'd rather it land before merge than after.

**Verification:** 865 passed (was 862), 30 in the error-bar file, `ruff check` clean.

---

PR #350, opened 2026-08-11
fix(plotly): keep data and selector aligned when a series is empty

### LINKED ISSUE #316: fix(plotly): an empty series desynchronises `data` from `selector`, highlighting the wrong line
# Bug Report

## Description

`PlotlyStepPlot` and `PlotlyMultiLinePlot` build `data` and `selector` from the same trace list, but by **two different rules**:

```python
# _extract_plot_data() -- drops a trace that produced no points
if series:
    all_series.append(series)

# _get_selector() -- always emits one selector per trace
return [self._scatter_line_selector(p) for p in self._scatter_positions]
```

So a trace whose `x`/`y` come out empty is dropped from `data` but still consumes a selector. The two lists then have different lengths, and since the frontend pairs selector *i* with series *i* positionally, **every series after the empty one is paired with its predecessor's selector**.

This is not merely a length mismatch — it is a wrong element highlighted, with no error.

## Steps to Reproduce

```python
from maidr.plotly.step import PlotlyStepPlot

def step(x, y, name):
    return {"type": "scatter", "mode": "lines", "x": x, "y": y,
            "line": {"shape": "hv"}, "name": name}

# the middle trace has no points -- e.g. a series filtered to empty
traces = [step([0, 1], [1, 2], "a"),
          step([], [], "empty"),
          step([0, 1], [2, 1], "c")]

schema = PlotlyStepPlot(traces, {}, scatter_positions=[0, 1, 2]).schema
print(len(schema[MaidrKey.DATA]), len(schema[MaidrKey.SELECTOR]))
```

## Actual Behavior

```
2 3
```

`data` holds `['a', 'c']` while `selector` holds `nth-child(1), nth-child(2), nth-child(3)`. Series `'c'` is the figure's **third** trace and renders at `nth-child(3)`, but is paired with `nth-child(2)` — the empty trace's element. Navigating to `'c'` highlights the wrong line.

## Expected Behavior

`data` and `selector` are the same length and aligned, so series *i* always addresses the element that series *i* is drawn as. Either the empty trace keeps its (empty) series so the lists stay parallel, or its position is dropped alongside its data.

## Additional Information

**Pre-existing, and not a regression.** The old `scatter_positions=range(len(traces))` default had the identical mismatch, so this predates both #303 and #311. Found during review of #315 and confirmed by running the snippet above against that branch.

**Why it matters despite needing an empty trace.** An empty `x`/`y` is not exotic — it is what a series filtered to nothing produces, which is routine in a dashboard or a faceted export where one category has no rows in the current slice. The failure is silent in exactly the way `nth-child` failures always are: the audio, braille and text are all correct, and only the visual highlight is wrong, so a sighted collaborator sees the wrong line highlighted while the blind user has no way to notice.

**Both classes are affected**, and identically — `PlotlyMultiLinePlot._extract_plot_data` has the same `if line_data:` guard. `PlotlyStepPlot` documents the drop as deliberate ("Empty series are dropped, matching `PlotlyMultiLinePlot`"), so whichever way this is resolved, the two should stay consistent with each other.

**The decision this issue needs** is which list gives way:

1. **Keep the empty series** so the lists stay parallel. Simplest, and the selector still points at the right element — but it emits a zero-point series that the frontend then has to tolerate, which may have its own consequences worth checking on the JS side.
2. **Drop the position too**, filtering `scatter_positions` by the same predicate that filters the data. Keeps `data` as it is today and keeps alignment, at the cost of the two lists being built by one shared pass rather than independently.
3. **Reject an empty trace at construction**, consistent with the empty-`traces` guard added in #315 — but that turns a plausible input into a crash, which seems wrong for a series that is legitimately empty.

Option 2 looks right to me: it preserves today's `data` contract and fixes the alignment, and building both lists from one pass is what stops them diverging again. Worth confirming against the JS side whether a zero-point series is acceptable before ruling out option 1.

Related: #311, #315.

jooyoungseo:
## Description

Closes #316.

`data` and `selector` are paired positionally by the frontend, but `PlotlyStepPlot` and `PlotlyMultiLinePlot` built them from the same traces by **two different rules** — the data dropped a trace that produced no points, the selectors did not. Every series after an empty one then addressed its predecessor's element.

Measured on the issue's own snippet, before:

```
step       len(data)=2  len(selector)=3
           nth-child(1), nth-child(2), nth-child(3)
```

Series `'c'` is the figure's **third** trace and renders at `nth-child(3)`, but was paired with `nth-child(2)` — the empty trace's element. After:

```
step       len(data)=2  len(selector)=2
           nth-child(1), nth-child(3)
```

## Related Issues

Closes #316.

## Changes Made

**Option 2 of the three the issue set out**: drop the position alongside the data, from one shared pass — `PlotlyPlot._line_series_with_positions()` returns the drawn series *and* the positions that produced them, so the two cannot be filtered by different predicates.

Why not the other two:

- **Keeping the empty series** so the lists stay parallel would emit a zero-point series for the frontend to tolerate. Dropping the position costs nothing and leaves `data` exactly as it is today — pinned by a test.
- **Rejecting an empty trace** at construction turns a legitimate input into a crash. An empty `x`/`y` is what a series filtered to nothing produces, which is routine in a dashboard or a faceted export where one category has no rows in the current slice.

Building both lists in one pass is the part worth having. `_scatter_line_selectors` already documents this exact hazard from the other direction — *"Dropping one entry from the middle would slide every later series onto the wrong element"* — and the data side was doing precisely that.

**Both classes are fixed together**, as the issue asked, and now share the pass rather than each owning a near-identical copy.

### One behaviour change worth flagging

An all-empty layer now emits **no `selectors` key at all**. `render()` omits it when the list is empty, which is the path a WebGL layer already takes and which the base class documents as the honest answer for a layer with no highlightable geometry. Before, this case was the defect at its purest: zero data, and a selector per trace anyway.

## Checklist

## Additional Notes

| Gate | Result |
|---|---|
| `uv run pytest` | **950 passed** (from 938) |
| `ruff check` | clean on every touched file; 52 repo-wide, unchanged |

12 cases in `tests/plotly/test_plotly_empty_series_alignment.py`, each parametrized across both classes. They assert the selector **strings**, not just that the two lengths match — equal lengths would also be satisfied by pointing every series at the wrong element, which is the failure being fixed.

- an empty trace in the middle does not shift the ones after it;
- nor does a **leading** empty trace;
- `data` is unchanged, so the alternative fix's cost is pinned as not paid;
- a layer with no empty trace keeps every position it was given;
- an all-empty layer emits neither list;
- a trace with **x but no y** counts as empty too — `zip` stops at the shorter side, so it reaches the same guard by a different route.

**Verified they bite:** against the code this replaces, **8 of the 12 fail**.

Worth noting why this stayed hidden: the failure is silent in the way `nth-child` failures always are. Audio, braille and text are all correct, and only the visual highlight is wrong — so a sighted collaborator sees the wrong line highlighted while the person relying on the announcements has no way to notice.

---

### DIFF
--- a/maidr/plotly/multiline.py
+++ b/maidr/plotly/multiline.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
-from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
-from maidr.plotly.plotly_plot import PlotlyPlot, as_list
+from maidr.plotly.plotly_plot import PlotlyPlot
 
 
 class PlotlyMultiLinePlot(PlotlyPlot):
@@ -72,7 +71,13 @@ class PlotlyMultiLinePlot(PlotlyPlot):
             One CSS selector per line, in trace order; empty for a WebGL
             layer.
         """
-        return self._scatter_line_selectors(self._traces, self._scatter_positions)
+        # Positions filtered by the same predicate that filters the data, so
+        # series *i* always addresses the element series *i* is drawn as --
+        # see `_line_series_with_positions`.
+        _, drawn_positions = self._line_series_with_positions(
+            self._traces, self._scatter_positions
+        )
+        return self._scatter_line_selectors(self._traces, drawn_positions)
 
     def _extract_plot_data(self) -> list[list[dict]]:
         """Return multi-line data as a list-of-lists.
@@ -80,24 +85,7 @@ class PlotlyMultiLinePlot(PlotlyPlot):
         Each inner list contains ``{x, y}`` dicts for one line, with an
         optional ``z`` key set to the trace name.
         """
-        all_lines: list[list[dict]] = []
-
-        for trace in self._traces:
-            x = as_list(trace.get("x"))
-            y = as_list(trace.get("y"))
-            name = trace.get("name", "")
-
-            line_data: list[dict] = []
-            for xv, yv in zip(x, y):
-                point: dict = {
-                    MaidrKey.X: self._to_native(xv),
-                    MaidrKey.Y: self._to_native(yv),
-                }
-                if name:
-                    point[MaidrKey.Z] = name
-                line_data.append(point)
-
-            if line_data:
-                all_lines.append(line_data)
-
-        return all_lines
+        lines, _ = self._line_series_with_positions(
+            self._traces, self._scatter_positions
+        )
+        return lines
--- a/maidr/plotly/plotly_plot.py
+++ b/maidr/plotly/plotly_plot.py
@@ -166,6 +166,68 @@ class PlotlyPlot(ABC):
             return []
         return [self._scatter_line_selector(position) for position in positions]
 
+    def _line_series_with_positions(
+        self, traces: list[dict], positions: list[int]
+    ) -> tuple[list[list[dict]], list[int]]:
+        """
+        Build the drawn series and the positions they were drawn at, together.
+
+        ``data`` and ``selector`` are paired positionally by the frontend, so
+        they have to be filtered by the same predicate. They were not: a trace
+        whose ``x``/``y`` came out empty was dropped from the data and still
+        consumed a selector, which slid every later series onto its
+        predecessor's element (#316). The audio, braille and text stayed
+        correct, so only a sighted collaborator could see the wrong line
+        highlighted -- the failure was invisible to the person using it.
+
+        Returning both from one pass is the fix, and is what stops the two
+        drifting apart again. The alternative, keeping the empty series so the
+        lists stay parallel, would emit a zero-point series for the frontend
+        to tolerate; dropping the position costs nothing and leaves ``data``
+        exactly as it is today.
+
+        An empty ``x``/``y`` is not exotic. It is what a series filtered to
+        nothing produces, which is routine in a dashboard or a faceted export
+        where one category has no rows in the current slice.
+
+        Parameters
+        ----------
+        traces : list of dict
+            The traces this layer covers, in series order.
+        positions : list of int
+            Each trace's zero-based position among the subplot's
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Merged, and the double-computation point you raised is real — followed up in #362 (issue #361), because it turned out to be more than a performance cost.

Repeating the pass assumes it can be repeated. `as_list` materialises a trace array with `list(value)`, so a one-shot iterable is spent by the first walk and reads as empty on the second: the layer reports its series and then **no selector at all**. That is the same silent no-highlight this pairing was written to prevent, arrived at from the other side — and it is a regression this PR introduced, not something pre-existing:

```
pre-#350  -> data: 2  selectors: ['… nth-child(1) path.js-line', '… nth-child(2) path.js-line']
post-#350 -> data: 2  selectors: None
```

`Figure.to_dict()` never hands back an iterator, so the export path does not reach it; a caller constructing a layer directly does. Narrow, but invisible when it happens.

On the fix: I went with caching keyed on the identity of the two lists rather than the `PlotlyBoxPlot` idiom you pointed at. You are right that stashing on `self` during `_extract_plot_data` is the established pattern here — but it works because `render()` happens to call data before selector, and a cache keyed on its inputs does not need that to be true. Same single pass, one less thing to preserve.

The performance number, since I had the harness out. 50 series × 10,000 points, `PlotlyMultiLinePlot.render()`, median of five:

| | median render |
| --- | --- |
| before | 1.859 s |
| after | 0.923 s |

Exactly the 2× a doubled walk predicts.

Your second point — the two `_get_selector` bodies now being identical — I left alone and said so on #362. The docstrings differ substantively (the step one explains why a staircase resolves to the same `path.js-line`), so folding the bodies would have to drop or merge that, and it is a three-line call that was duplicated before this PR too.

---

PR #352, opened 2026-08-11
feat(pointplot): read a seaborn point plot's estimates with their intervals

jooyoungseo:
Closes #246.

A point plot estimates a group mean and draws the interval around it. That is the same quantity `Axes.errorbar` carries and the same thing a reader needs from it — **whether two group means differ is answered by whether their intervals overlap** — so it now emits the same `error_bar` layer, and the consumer steps through lower / value / upper at each group with no special case for where the chart came from.

## This is as much a removal as an addition

`sns.pointplot` was already reaching MAIDR, through the generic `Axes.plot` wrapper. What it produced was not "missing intervals" — it was wrong. A three-group chart announced **four series**:

```json
[{"x": "Sat", "y": 22.49}, {"x": NaN, "y": NaN}, {"x": 1.95, "y": 22.49}, {"x": 2.05, "y": 22.49}]
```

That is one of the interval polylines: a NaN separator, and cap offsets like `1.95` sitting among the category names. A reader arrowing through the chart walked cap geometry believing it was data.

| | before | after |
|---|---|---|
| series announced (3 groups) | 4 | 1 |
| interval carried | ✗ | ✓ `yMin`/`yMax` |
| dodged group label | `-0.025` | `Thur` |
| hue series names | off-by-N from `legend_labels[i]` | correct |

## Why patch `pointplot` by name

Only the name says the artists mean something other than what they look like. Seaborn draws no `ErrorbarContainer` — it draws the estimates as one `Line2D` and each interval as another, so nothing downstream can tell an interval from a series.

The split is **verified, not assumed**, since it is a claim about another library's rendering: the estimates have to pair one-to-one with the intervals, group by group, and each interval has to be short enough to be one. When that fails — because a future seaborn renders this differently — the layer falls back to describing the drawn lines the way the generic wrapper did, rather than reading bounds off artists that are not intervals.

## Design notes

**Orientation** is read from the axes, not from the caller's `orient`, which is usually absent — seaborn infers it and does not report what it decided. Two signals do, in order of authority:

1. Seaborn draws its categorical axis through matplotlib's string-category machinery, leaving `UnitData` on that axis and nothing on the other. True whatever the category *values* are, so a numeric grouping column still travels this way.
2. Under `native_scale=True` that machinery is silent, and the intervals answer it themselves: an interval spans the value axis and stands on a single coordinate of the other.

A rule based on "the estimate sits at the centre of its interval" would have looked simpler and been wrong — `errorbar='sd'` draws bounds equidistant from the estimate, so it then sits at the centre along *both* axes. There is a test for exactly that.

**Bounds are the extremes along the value axis**, which is what makes `capsize` irrelevant: a cap is drawn *at* the bound it marks, so a two-vertex line and an eight-vertex NaN-separated polyline give the same answer. Pinned by comparing a capped draw against a bare one.

**`hue` reads as the multi-line chart it is.** `ErrorBarPoint[]` on the consumer side is a single flat series, so a hued point plot cannot carry both its groups and its intervals. The intervals are dropped rather than mis-assigned — one group's bounds under another's estimate would be worse than none — and I will open a follow-up for multi-series intervals, which needs the JS grammar first. The estimates still travel correctly, which they did not before.

**A dodged group is named by the tick it was shifted from.** `dodge` draws the point at `0.975` rather than `1`, the exact-coordinate lookup misses, and the reader heard `-0.025` where the chart says `Thur`. Rounding recovers it, gated on the axis really being categorical — on a numeric axis the same rounding would rename a measurement after whichever tick it fell nearest, and there is a test that fails if that gate is removed.

## Verification actually run

| Step | Result |
|---|---|
| `uv run pytest` | **954 passed** (was 938) |
| `ruff check` on changed files | clean |
| `ruff check` repo-wide | 52 errors, unchanged from `main` (all pre-existing `F401` in `__init__.py`) |

Two of the new tests failed first and found real defects, which is why they are in this shape:

- extraction was **not idempotent** — a layer is rendered more than once, and `_elements` accumulated, so the highlight would land on the wrong group. Now cleared per extraction.
- the numeric-axis guard test initially passed for the wrong reason (a coordinate that happened to sit exactly on a tick). Rewritten so every coordinate sits off one, which is the only way it tests the rounding at all.

`tests/core/plot/test_errorbar.py::test_seaborn_error_bars_do_not_register_a_second_layer[pointplot]` changed from `LINE` to `ERRORBAR`. Its purpose is unchanged — that patching `Axes.errorbar` does not add a *second* layer to seaborn's own charts — and its docstring now says the single layer is registered by the new patch rather than that one.

---

### DIFF
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -657,6 +657,40 @@ ax.set_ylabel("Response")
 plt.show() #<<
 ```
 
+### Point Plot
+
+`sns.pointplot()` is the same reading from the other direction: it estimates a
+group mean from raw observations and draws the interval around it for you. It
+reads as an error bar layer, so navigation is the same grid — left and right
+walk the groups, up and down walk the lower bound, the estimate and the upper
+bound.
+
+`capsize`, `orient` and `errorbar=` change how the chart is drawn, not how it
+is read; the bounds come from the drawn interval either way. A `hue` splits the
+chart into several series, which the error bar layer cannot yet carry
+alongside its intervals, so a hued point plot reads as the multi-line chart it
+is.
+
+```{python}
+#| fig-alt: Mean body mass by penguin species with 95% confidence intervals
+
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+import maidr #<<
+
+fig, ax = plt.subplots(figsize=(8, 5))
+sns.pointplot( #<<
+    penguins, x="species", y="body_mass_g", capsize=0.1, ax=ax
+)
+
+ax.set_title("Mean Body Mass by Species, with 95% Confidence Intervals")
+ax.set_xlabel("Species")
+ax.set_ylabel("Body Mass (g)")
+
+plt.show() #<<
+```
+
 ### Scatter Plot
 
 ```{python}
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -2,7 +2,7 @@
 
 > py-maidr is a Python library that makes matplotlib, seaborn, and plotly data visualizations accessible to blind and low-vision users through sonification, braille, text, and AI-powered conversational modalities. Developed by the (x)Ability Design Lab at the University of Illinois Urbana-Champaign.
 
-maidr (pronounced "mader") monkey-patches matplotlib, seaborn, and plotly at import time so users just `import maidr` and call `maidr.show(plot)` to generate interactive HTML with keyboard-navigable sonification, braille, text, and AI chat support. Supported plot types: bar, stacked bar, dodged bar, count, histogram, KDE, line, multi-line, step, pie, box, violin, heatmap, scatter, smooth/regression, candlestick, error bar, multi-panel, and facet plots. Works in Jupyter, Colab, VS Code, Streamlit, Shiny, and Quarto.
+maidr (pronounced "mader") monkey-patches matplotlib, seaborn, and plotly at import time so users just `import maidr` and call `maidr.show(plot)` to generate interactive HTML with keyboard-navigable sonification, braille, text, and AI chat support. Supported plot types: bar, stacked bar, dodged bar, count, histogram, KDE, line, multi-line, step, pie, box, violin, heatmap, scatter, smooth/regression, candlestick, error bar, point, multi-panel, and facet plots. Works in Jupyter, Colab, VS Code, Streamlit, Shiny, and Quarto.
 
 ## Docs
 - [Documentation Home](https://py.maidr.ai/): Installation, quick start, and overview
--- a/maidr/core/plot/lineplot.py
+++ b/maidr/core/plot/lineplot.py
@@ -1,6 +1,7 @@
-from typing import List, Union
+from typing import List, Optional, Union
 
 from matplotlib.axes import Axes
+from matplotlib.lines import Line2D
 
 from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
@@ -26,6 +27,12 @@ class MultiLinePlot(MaidrPlot, LineExtractorMixin):
         The layer type to emit, by default ``PlotType.LINE``. Subclasses that
         share this extraction logic but describe a different chart — notably
         :class:`maidr.core.plot.stepplot.StepPlot` — override it.
+    lines : list of Line2D, optional
+        The lines to describe, by default every line on the axes. A patch
+        passes this when some of what was drawn is not a series: seaborn's
+        point plot renders each confidence interval as a line of its own, and
+        describing those alongside the estimates announces cap geometry as
+        data.
     **kwargs : dict
         Additional keyword arguments to pass to the parent class.
 
@@ -42,8 +49,30 @@ class MultiLinePlot(MaidrPlot, LineExtractorMixin):
       (fill values) for each point.
     """
 
-    def __init__(self, ax: Axes, plot_type: PlotType = PlotType.LINE, **kwargs):
+    def __init__(
+        self,
+        ax: Axes,
+        plot_type: PlotType = PlotType.LINE,
+        lines: Optional[List[Line2D]] = None,
+        **kwargs,
+    ):
         super().__init__(ax, plot_type)
+        self._lines = lines
+
+    def _series(self) -> List[Line2D]:
+        """
+        Return the lines this layer describes.
+
+        Returns
+        -------
+        list[Line2D]
+            The lines a patch narrowed the layer to, or every line on the axes
+            when it did not.
+        """
+        if self._lines is not None:
+            return self._lines
+
+        return self.ax.get_lines()
 
     def _extract_axes_data(self) -> dict:
         """
@@ -66,7 +95,7 @@ class MultiLinePlot(MaidrPlot, LineExtractorMixin):
 
     def _get_selector(self) -> Union[str, List[str]]:
         # Return selectors for all lines that have data
-        all_lines = self.ax.get_lines()
+        all_lines = self._series()
         if not all_lines:
             return ["g[maidr='true'] > path"]
 
@@ -105,7 +134,7 @@ class MultiLinePlot(MaidrPlot, LineExtractorMixin):
             List of lists, where each inner list contains dictionaries with x,y coordinates
             and line identifiers for one line, or None if the plot data is invalid.
         """
-        all_lines = self.ax.get_lines()
+        all_lines = self._series()
         if not all_lines:
             return None
 
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -11,6 +11,7 @@ from maidr.core.plot.histogram import HistPlot
 from maidr.core.plot.lineplot import MultiLinePlot
 from maidr.core.plot.maidr_plot import MaidrPlot
 from maidr.core.plot.pieplot import PiePlot
+from maidr.core.plot.pointplot import PointPlot
 from maidr.core.plot.scatterplot import ScatterPlot
 from maidr.core.plot.regplot import SmoothPlot
 from maidr.core.plot.stepplot import StepPlot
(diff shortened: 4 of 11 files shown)


jooyoungseo:
Ran all five down. Three were real, one was real for a different reason than described, and two are follow-ups with issues filed. Fixed in 886644d.

## 2 — real, and worse than described. The mechanism is different.

You predicted `native_scale=True` + `capsize>0` + horizontal would fall through to the default. I ran it. The orientation *is* wrong, but not by falling through — and what it produced is worse than a mis-detection:

```json
{"x": 1.0, "y": 1.0, "yMin": 0.95, "yMax": 1.05}
```

Eighteen of those. `0.95`/`1.05` is the **cap's own width**, announced as a confidence interval.

What actually happens: with two numeric columns and no `orient`, seaborn groups by every distinct value, so **every group has one observation**. A single observation has nothing to estimate an interval from, and seaborn draws the polyline anyway with the caps intact and the value coordinates NaN. Then `_extent` answered `0` for an all-NaN axis — a polyline *absent* from an axis is not *flat* along it — so the zero-extent signal read the value axis as the category axis and got the orientation backwards. Reading the caps as bounds followed from that.

Two fixes, at the two different levels:

- A candidate interval now has to be **drawable**: two vertices finite in *both* coordinates. That is what it takes to draw a segment at all, so it needs no view on which axis is which, and it is the exact characterisation of "the chart drew an interval here". Charts whose intervals all fail it read as the line of estimates they are.
- `_extent` answers NaN rather than 0 when nothing is finite, so the orientation signal cannot be fed a fact that isn't one.

`test_a_group_of_one_carries_no_interval` covers it with a plain three-row frame — the failure does not need `native_scale` to reach, which is worth knowing.

## 1 — real, fixed

Bounds now run through `ErrorBarPlot._without_float_noise`. Reusing the base's helper rather than restating the rule, so the two cannot drift to different precisions for the same quantity.

Worth noting the numbers are not quite what you'd expect from an offset subtraction: `mean - sd` for two observations is `0.07928932188134526`, which is a genuine irrational rather than a `3.8000000000000003`-style artifact. It still wants cutting — a screen reader reads all seventeen digits either way — and `test_bounds_are_not_spelled_out_to_seventeen_digits` names it for what it is.

## 4 — real, fixed

`if not estimates: return ax` now falls back to describing the drawn lines. You were right that it contradicted the module's own philosophy: every other verification failure degrades to the generic wrapper's answer, and that answer is poor but silence is worse. Covered.

## 3 — declined here, filed as #354, and it is not the one-liner it looks like

You're right that the pattern is codebase-wide and right that the guard sitting in the subclass is odd. But the central fix you sketched — clearing in `MaidrPlot.render()` before `_extract_plot_data()` — **breaks `ViolinKdePlot`**, which appends to `_elements` in its `__init__`. Clearing at render would discard those and leave the violin bodies untagged, silently.

So it is a fourteen-class audit, not a one-liner, and it wants a test per type asserting a second `render()` leaves the count unchanged. #354 has the class list and that finding written down. Doing it inside this PR would mean changing every plot type on a change about point plots.

## 5 — confirmed and pinned, filed as #353

Confirmed by running it: a horizontal hued dodged chart emits `{"x": 4.33, "y": -0.025}` — the categories as raw dodged offsets. As you say, a pre-existing limit of the shared helper rather than something this PR introduced.

Rather than only note it, `test_a_horizontal_dodge_still_reports_raw_category_offsets` pins the current behaviour and points at #353, so the gap is visible in the suite and the day the helper learns to follow the category axis the test turns red instead of the behaviour changing unnoticed.

## Nit — taken

Factory now dispatches on `isinstance(kwargs.get("estimate"), Line2D)` rather than on the key being present.

**Verification:** `uv run pytest` **958 passed** (was 954); `ruff check` clean on the changed files, and repo-wide unchanged from `main`.

---

jooyoungseo:
Confirmed empirically, as you asked. You traced it exactly right, and it is worse than "a chart that is almost well-formed" — it is any imbalanced frame, which is most real categorical data.

Reproduced with two groups of five observations and one of one:

```
drawn lines: 4 | estimates: [3] | intervals kept: 2 | pairs_up: False
type: LINE | series: 4
[[{"x":"a","y":3.0},{"x":"b","y":12.0},{"x":"c","y":7.0}],
 [{"x":"a","y":1.8},{"x":"a","y":4.2}],
 [{"x":"b","y":10.8},{"x":"b","y":13.2}],
 [{"x":"c","y":NaN},{"x":"c","y":NaN}]]
```

Four series, one of them a pair of NaNs. The exact rendering this PR exists to remove, on a shape more ordinary than either of the two the tests covered.

## The fix, and why it is not option (a)

Relaxing `_pairs_up` to accept a partial match does not work on its own. `PointPlot._extract_plot_data` reads the intervals against the estimates **by position**:

```python
for category, value, interval in zip(categories, values, self._intervals):
```

so a short list does not merely lack one bound — it shifts every later group's interval onto the wrong estimate. Group `b` would have been announced with group `c`'s bounds, which is worse than the fallback.

So the drop moves out of `_split` entirely. The polyline stays in the list holding its group's position, and `_is_drawn` decides what to *do* with it rather than whether it is *there*. Whether the chart is an error bar chart at all is now a separate question — whether **any** group drew an interval — which keeps the all-singles case (`native_scale`) falling back to a line chart as before.

Result on the same frame:

```json
[{"x":"a","y":3.0,"yMin":1.8,"yMax":4.005},
 {"x":"b","y":12.0,"yMin":10.8,"yMax":13.2},
 {"x":"c","y":7.0}]
```

Your read that `_interval_bounds`' existing `None` handling was already the right per-group behaviour and just unreachable was correct — it needed the list to arrive intact, not a new branch.

## One thing the fix surfaced that the review did not

With partial intervals, `_elements` ends up shorter than the point list — 2 elements for 3 points. The consumer resolves the selector to one element per point **in point order and discards the result outright when the count disagrees**, so the layer was emitting a selector that resolved short and silently highlighted nothing, while the schema promised it would.

The selector is now withdrawn unless every point has an element, the same way a bare `ax.errorbar(x, y)` withdraws it. Covered by `test_a_partial_interval_promises_no_highlight`.

## Minor notes

- **Factory dispatch** — agreed it is implicit, and a dedicated `source=` kwarg is the right shape if a third variant appears. Not adding the indirection for two.
- **`_lines_before` touching pyplot** — fair flag. It is the narrowest form I could find: it only reaches for `plt.gca()` when a figure already exists, so it cannot create a figure, and the axes it creates in the no-axes case is the same one the wrapped call is about to create.
- Thanks for reading the follow-up commits as evidence rather than noise — the idempotency and `_extent` bugs both came from tests failing first, and this one now has the same provenance.

**Verification:** `uv run pytest` **960 passed** (was 958); `ruff check` clean on the changed files, repo-wide unchanged from `main`.

---

jooyoungseo:
Correcting something I wrote in the PR description and repeated in my first reply.

I said the hue case "needs the JS grammar first". **That is wrong**, and I checked rather than building on it.

Multiple layers per subplot already work on both sides. `Maidr._merge_plots_by_subplot_position` groups every plot registered at the same `(row, col)` into one subplot's `layers` array, so two `create_maidr` calls on one axes already produce two layers — which is exactly what `test_an_earlier_chart_on_the_same_axes_is_left_alone` in this PR asserts, with `[LINE, ERRORBAR]`. On the JS side `examples/violin.html` puts two layers in one subplot and `TextService` announces the switch between them.

So a hued point plot **could** emit one `error_bar` layer per hue level today, with no grammar change at all.

What stops it is smaller and different: the layer-switch announcement is built from the trace type (`src/service/text.ts:217`), so two `error_bar` layers announce identically —

```
Layer 1 of 2: error_bar plot at Day is Sat, Total bill is 20.4
Layer 2 of 2: error_bar plot at Day is Sat, Total bill is 19.7
```

— and the reader is never told which is Male and which is Female. Emitting per-hue layers before that is fixed would hand a reader two indistinguishable layers, which is worse than the current honest fallback.

`layer.title` looks like the answer and is not: this package puts the *figure* title there for every layer, so preferring it would replace a useful type name with a repeated chart title on every existing multi-layer figure.

**xability/maidr#828 is rewritten** with the corrected framing and three sketches for giving a layer a name of its own. Nothing in this PR changes — the hue fallback is still right, for a better-understood reason — but the follow-up is a much smaller piece of work than I said it was, and on the JS side rather than in the grammar.

---

jooyoungseo:
Both tests added in 195fce2. The blocking question resolves in the direction you flagged as the "moot" branch — but only because of a rendering choice that deserved pinning either way.

## Unbalanced hue × category — checked, and it holds

You asked for it to be double-checked rather than assumed, so I ran it. A hue level absent from one category entirely:

```
half  x  y
g
a     2  2
b     2  2
c     2  0
```

```
estimate point counts: [3, 3]
  [[-0.025, 1.5], [0.975, 10.5], [1.975, 20.5]]
  [[ 0.025, 5.5], [1.025, 14.5], [2.025,  nan]]
intervals: 6 | pairs_up: True
type: LINE | series: 2
```

**Seaborn NaN-pads the missing combination** rather than omitting the point, so both estimate lines carry one vertex per category, the counts still match, and the pairing survives. Two clean series out, no interval polylines among them.

The missing combination arrives as `NaN`, which is a gap rather than a value — the consumer names a non-finite reading "missing" rather than reading it out, pinned on the JS side by `test/service/textGaps.test.ts`.

`test_a_hue_level_missing_a_category_does_not_break_the_pairing` covers it. Written to fail loudly if seaborn ever switches to omitting the point, since that is the version of this where your prediction comes true.

## `markers=""` — added, and it was the right thing to worry about

```
default       markers=['o',  'None', 'None'] -> ERRORBAR, 2 points
marker=''     markers=['',   'None', 'None'] -> ERRORBAR, 2 points
```

The literal comparison is exactly as load-bearing as you say: `if not line.get_marker()` would classify `''` as an interval, every estimate would be misfiled, the counts would stop matching and the chart would drop to the fallback. `test_an_estimate_line_without_a_marker_is_still_an_estimate` is the test that fails when someone makes that simplification — which the comment alone would not have prevented.

## `_lines_before` safety net

Agreed with your read, and it is a better articulation than the comment had. Left as is rather than adding a third paragraph to a function that already carries two — your point stands in this thread.

## Nit: `PLOT_TYPE_PRIORITY`

`PlotType.ERRORBAR: 1` is already there — `maidr/core/figure_manager.py:61`, added with the `Axes.errorbar` patch in #349. No gap.

## Nit: docstring verbosity

Fair. The rationale does appear at module, class and method level. I have left it, because the three answer different questions — *why patch by name*, *why this class exists beside `ErrorBarPlot`*, *why this method reads the extremes* — but the observation is right that they share a premise, and if it grates on a reader it is worth trimming later.

**Verification:** `uv run pytest` **962 passed** (was 960); `ruff check` clean on the changed files, and repo-wide unchanged from `main` (7 pre-existing).

---

PR #355, opened 2026-08-11
ci: refresh the bundled maidr.js when upstream publishes

jooyoungseo:
Receiving half of a two-repo change. The sending half is a PR on xability/maidr, linked below. This mirrors what xability/r-maidr#131 did for the R binding.

## Why

The bundle only ever moved at py-maidr release time, so between releases `main` carried whatever upstream version the last release happened to pin. Checking out `main` to reproduce or test a freshly published upstream fix meant refreshing the bundle by hand first.

## What changes

`update-maidr-js.yml` gains a `repository_dispatch` trigger on `maidr-released`. maidr's release workflow fires it the moment `npm publish` succeeds, carrying the version it published:

```json
{"event_type": "maidr-released", "client_payload": {"version": "4.1.0"}}
```

The job takes the version from `client_payload` when present, falling back to the `workflow_dispatch` input and then to nothing — which makes the script resolve `latest` on its own, exactly as before. Naming the version sidesteps the window right after a publish where the `latest` dist-tag has not propagated.

Everything downstream of that is unchanged: same shared `fetch-maidr-bundle.sh`, same strict version-pattern check, same registry integrity verification, same `chore:`-prefixed commit to `main`.

## What does *not* change, deliberately

**An upstream release still does not reach PyPI on its own.** The commit keeps its `chore:` prefix, so python-semantic-release does not treat a bundle refresh as a py-maidr release; the refreshed bundle ships with the next `feat:` / `fix:` / `perf:` release, whenever that is.

That is the explicit choice here, and it is worth stating plainly because it is the limitation of this approach: it makes `main` current, not PyPI. Making an upstream release cut a py-maidr release would have meant typing the bundle commit as `fix:`, which would let upstream drive py-maidr's version numbers — a bigger change to what a py-maidr version *means*, and not what was asked for.

`release.yml` remains the authority on what users actually receive: it re-resolves the newest upstream when a release is cut, and now simply finds nothing to commit when `main` is already current.

## No scheduled backstop, unlike r-maidr

r-maidr pairs its dispatch with a daily cron. That is right there and wrong here.

r-maidr has no release pipeline — whatever sits on `main` at CRAN submission time is what ships — so a missed refresh there is a missed shipment, and the cron bounds that. Here the release-time refresh in `release.yml` already re-resolves the newest upstream, so a missed dispatch costs only the currency of `main` and never the released artifact. A cron would be redundant with a step that already runs.

## Trust

No new trust is granted. `client_payload` is written by whoever holds the sending token, but that token needs write access to this repository to send a `repository_dispatch` at all, and anyone holding it could push to `main` directly. The version still passes the `^[0-9]+\.[0-9]+\.[0-9]+([-+.][0-9A-Za-z.-]+)*$` check in `fetch-maidr-bundle.sh` before it reaches a URL or a path, and the download is still verified against the registry's published integrity hash.

## Also in this PR

`release.yml` carried a comment stating the bundle "never trickles onto `main` between releases". This change makes that untrue, so it is corrected in the same commit rather than left to contradict the code next to it.

## Verification

`actionlint` passes on both changed workflows. Both files parse, and the trigger sets are `[repository_dispatch, workflow_dispatch]` for the updater and `[schedule, workflow_dispatch]` for the release, as intended.

Note that `repository_dispatch` only ever triggers the workflow file as it exists on the default branch, so nothing can be received until this is on `main`.

CI-only; no Python code, no tests, and nothing in the package itself changes.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DZocjnGTiRvLDCbcvqYqa3)_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -106,9 +106,14 @@ jobs:
         # py-maidr cut its own release.  The refresh is committed locally
         # as ``xabilitylab`` but intentionally *not* pushed here: the
         # semantic-release step below pushes the branch, so this commit
-        # ships only when a release actually happens.  When no release is
-        # pending the commit is discarded with the runner, so the bundle
-        # never trickles onto ``main`` between releases.
+        # ships only when a release actually happens, and when no release is
+        # pending it is discarded with the runner.
+        #
+        # ``main`` may already carry a newer bundle than the last release:
+        # ``update-maidr-js.yml`` refreshes it whenever upstream publishes.
+        # That costs nothing here -- this step re-resolves the newest upstream
+        # either way, and finds nothing to commit when the bundle is already
+        # current -- and this step stays the authority on what users receive.
         run: |
           set -euo pipefail
 
--- a/.github/workflows/update-maidr-js.yml
+++ b/.github/workflows/update-maidr-js.yml
@@ -1,22 +1,40 @@
 name: Update bundled maidr.js
 
-# Manual escape hatch for refreshing the offline-bundled ``maidr.js`` and
-# its assets from the npm registry into ``maidr/static/`` and committing
-# the changes directly to ``main``.
+# Refreshes the offline-bundled ``maidr.js`` and its assets from the npm
+# registry into ``maidr/static/`` and commits the changes directly to
+# ``main``.  Runs when upstream publishes, and on demand.
 #
-# The routine, automatic refresh happens at py-maidr *release* time (see
-# ``release.yml``), so the latest upstream release is bundled *into* each
-# py-maidr release rather than trickling onto ``main`` whenever upstream
-# publishes.  This workflow is kept only for on-demand refreshes between
-# releases and is therefore ``workflow_dispatch``-only -- there is no
-# scheduled trigger.
+# What reaches *users* is still decided at py-maidr release time, not here:
+# ``release.yml`` refreshes the bundle again as part of cutting a release, so
+# a release always ships the upstream version that was current when it was
+# cut.  This workflow moves ``main`` forward in between, which is what makes
+# a freshly published upstream fix testable from a checkout without waiting
+# for the next release.
 #
 # The commit uses the ``chore:`` prefix so python-semantic-release
-# (configured in ``pyproject.toml``) does not treat a bundle refresh
-# as a new py-maidr release.  The next ``feat:`` / ``fix:`` / ``perf:``
-# commit to ``main`` will ship with the refreshed bundle automatically.
+# (configured in ``pyproject.toml``) does not treat a bundle refresh as a new
+# py-maidr release.  That is deliberate, and it is the limit of what this
+# workflow can do: an upstream release alone does not reach PyPI.  It ships
+# with the next ``feat:`` / ``fix:`` / ``perf:`` release, whenever that is.
+#
+# There is no scheduled trigger, and none is needed as a backstop the way
+# r-maidr has one.  r-maidr has no release pipeline -- whatever sits on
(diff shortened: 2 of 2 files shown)


jooyoungseo:
The race is real and it is now handled — thanks, this one earned its keep.

Confirmed the setup you describe: `release.yml:77` declares `concurrency: release` on its own job, `update-maidr-js.yml` declared nothing, and both push to `main` — this one via `git-auto-commit-action`, that one via `python-semantic-release`. Overlapping, the second push is a non-fast-forward against a checkout taken before the first landed, and the loser is a failed release.

I did not take the "rare, self-healing, just comment on it" option, because the fix turns out to be one line. Concurrency groups are scoped to the **repository**, not to a workflow:

> If you have multiple workflows in the same repository, concurrency group names must be unique across workflows to avoid canceling in-progress jobs or runs from other workflows.

— [Workflow syntax → `concurrency`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#concurrency)

So naming the same group in both is exactly the mechanism for serialising them. The updater job now declares `concurrency: release`, sharing `release.yml`'s group deliberately, with a comment recording why the name is shared so nobody "fixes" the apparent collision later.

Queuing rather than cancelling, since waiting is the default. A refresh held behind a release then finds the bundle already current and commits nothing, because the release refreshed it too. The one lossy edge is that a second dispatch arriving while one is pending replaces it — per the same docs, "any existing `pending` job or workflow in the same concurrency group will be canceled" — and the cost there is `main` briefly trailing, never the released artifact.

On the window being narrow: agreed that the schedules are an hour apart on purpose, but that is exactly what stops being load-bearing here. An off-schedule upstream release now fires a dispatch at any hour — one was published manually today at 17:52 UTC — so the separation is no longer something to rely on.

On the live test: agreed, and it has already happened for the sibling repository. The r-maidr half of this mechanism ran end to end on its first real release earlier today — `v4.1.0` published at 17:58:18, dispatch sent at 17:58:19, r-maidr's refresh committed at 17:58:32. Same expression, same fallback chain. This repository's copy will get its own live exercise on the next upstream release.

Taking the comment-length note as read.

---

jooyoungseo:
Thanks — glad the concurrency reasoning reads clearly, since that one came out of your first pass.

On the fork guard, I checked the premise and I'm going to leave it off. The stated rationale is that a fork maintainer firing it by hand would get "a confusing failed run rather than a clean skip", but it would not fail. The job uses no secrets — only the default `GITHUB_TOKEN` — and the only thing it contacts is the public npm registry:

```
REGISTRY="https://registry.npmjs.org/maidr"
```

So on a fork it checks out the fork, downloads from a public registry, and commits to the fork's own `main` with the fork's own token. That succeeds, and what it does is exactly what a fork maintainer pressing that button would want.

The consistency argument doesn't carry either, because the two guards would not be guarding comparable things. `release.yml`'s guard sits directly above `environment: pypi` — it exists because that job publishes to PyPI and cuts releases under this project's identity, which a fork must never do on the project's behalf. A bundle refresh has no effect outside the repository it runs in. Same syntax, different blast radius, and the blast radius is the whole reason the guard is there.

Adding it would remove a harmless and occasionally useful capability from forks to make two jobs look alike. I'd rather the guard keep meaning "this job reaches outside the repository", so that seeing it somewhere is informative.

Everything else in the review matches what I found, including the `""`-is-falsy fallback tracing and the `test_shell_guard_matches_the_python_validator` parity point — worth noting that test keeps the shell regex and the Python validator honest regardless of which trigger supplied the version, which is precisely why this change did not need to touch either.

---

PR #356, opened 2026-08-11
feat(area): read a stackplot as the area chart it is

jooyoungseo:
`Axes.stackplot` registered **nothing at all** — no MAIDR layer, so an area chart carried no data and was unreadable. It now emits an `area` or `stacked_area` layer.

Closes the py-maidr side of xability/maidr#788, whose trace shipped in maidr **4.1.0**.

## Why an area layer rather than a line one

A stacked area draws **two** numbers at each point that a line would conflate:

| | what it is |
|---|---|
| band **height** | the series' own value |
| band **top edge** | the running total |

A line layer announces one number per point with nothing to say which of the two it is, and the reader cannot recover the other. The area trace announces the value and reports the total beside it.

## Reading the arguments, not the geometry — and why that's the right way round here

This is the opposite of how `ErrorBarPlot` and `PointPlot` work, so it's worth being explicit. Those read drawn geometry precisely because *the arguments they were given are not the quantity the schema carries* — matplotlib takes an error **offset** while the schema wants an absolute **bound**.

`stackplot` is the other way round. It is handed each series' own values and does the accumulation itself, so its arguments are exactly the per-series magnitudes the consumer needs. Its polygon, meanwhile, is a closed outline:

```
[[2019, 10], [2019, 0], [2020, 0], [2021, 0], [2022, 0],
 [2022, 30], [2022, 30], [2021, 25], [2020, 20], [2019, 10], [2019, 10]]
```

— forward along the baseline, back along the top, endpoints repeated. Recovering the same numbers from that means undoing both the accumulation and the closure, to arrive back at what the caller passed in.

A test asserts the emitted values are `[5, 8, 12, 14]` and **not** `[15, 28, 37, 44]`, the drawn top edges, so a future change to geometry-reading fails rather than looking plausible.

## Behaviours pinned

- **One band is an `area`, several are `stacked_area`.** A single band has nothing stacked on it; a running total equal to its own value at every point is noise.
- **`stackplot(x, y)` with a 2-D `y`** means the same as `stackplot(x, y1, y2)`, asserted by comparing the two schemas directly.
- **`baseline="wiggle"`** — a streamgraph — reads as the stacked area it is. It floats the stack without changing what any band measures. It looks like it should decide the type and does not.
- **Series names** come from `labels=`; an unlabelled chart gets no invented ones.
- **Re-rendering** does not double the tagged elements.

I also removed a `baseline` parameter I'd initially threaded into `resolve_type` and never used — the reasoning lives in the docstring, where it doesn't need a dead argument to carry it.

## Highlighting

Tags `FillBetweenPolyCollection`, which matplotlib 3.10 split out of `PolyCollection`. That makes it a **subclass-only override** in exactly the way `PolyQuadMesh` already is, so violin bodies and every other `PolyCollection` stay untouched — `highlight.py`'s existing comment explains why blanket-wrapping the base would be wrong.

Guarded, because this package supports matplotlib ≥ 3.8 and the class is 3.10+. On an older matplotlib a band keeps the plain `PolyCollection` it always had and the layer degrades to no highlight rather than failing to import.

The selector is withdrawn unless **every** band was tagged — the consumer resolves it to one element per series and discards the result outright on a count mismatch, so a partial list would promise highlighting and silently deliver none. Same rule I applied in #352.

## What is deliberately not here

**`ax.fill_between` with two curves.** `fill_between(x, lo, hi)` draws a band around a line — an uncertainty ribbon — not an area against a baseline. Registering it as `area` would announce the upper edge as "the value", which is the exact conflation this layer type exists to prevent. It stays unregistered, as today.

## One thing worth flagging

Checking whether the pinned runtime knows `area`, I found **`maidr/static/VERSION` is 4.0.0 while npm publishes 4.1.0**, and none of the recently added trace types — `area`, `error_bar`, `gauge`, `waterfall`, `word_cloud`, `dumbbell` — exist in 4.0.0. The CDN path resolves the published version at render time, so the default is fine. The **bundled** path (`use_cdn=False`, or CDN unreachable) serves 4.0.0, whose `TraceFactory` throws `Invalid trace type` on an unknown type.

`bundle_status()` reports `is_behind=True, is_stale=False` — the warning threshold is `STALE_MINOR_GAP = 5`, so one minor behind is silent. That threshold was calibrated for "old but working" drift; a bundle that predates a trace type we now emit is a different failure. `release.yml` refreshes the bundle at release time so released users are unaffected, but it seemed worth writing down rather than leaving for someone to hit.

## Verification actually run

| Step | Result |
|---|---|
| `uv run pytest` | **974 passed** (was 962) |
| `ruff check` on changed files | clean |
| `ruff check` repo-wide | 7 errors, unchanged from `main` (pre-existing `F401`) |

A probe of every call form found one branch I'd written for a keyword-series call (`stackplot(x, Subs=[...])`) that **matplotlib itself rejects** — its `**kwargs` go to `fill_between`, not to series. Dead code for a call that cannot happen; removed rather than tested.

---

### DIFF
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -657,6 +657,43 @@ ax.set_ylabel("Response")
 plt.show() #<<
 ```
 
+### Area Plot
+
+`ax.stackplot()` reads as an area chart rather than a line one, and the
+distinction matters: a stacked area draws **two** numbers at each point that a
+line would conflate. The band's height is the series' own value; its top edge
+is the running total. maidr announces the value and reports the total beside
+it, so a reader always knows which they heard.
+
+A single band is a plain area — nothing is stacked on it, so a running total
+equal to its own value would be noise. `baseline="wiggle"` (a streamgraph)
+reads the same way as any stacked area: floating the stack moves where the
+bands sit without changing what any band measures.
+
+```{python}
+#| fig-alt: Stacked area chart of revenue by source over four years
+
+import matplotlib.pyplot as plt
+
+import maidr #<<
+
+years = [2019, 2020, 2021, 2022]
+subscriptions = [10, 20, 25, 30]
+services = [5, 8, 12, 14]
+
+fig, ax = plt.subplots(figsize=(8, 5))
+ax.stackplot( #<<
+    years, subscriptions, services, labels=["Subscriptions", "Services"]
+)
+
+ax.set_title("Revenue by Source")
+ax.set_xlabel("Year")
+ax.set_ylabel("Revenue (millions)")
+ax.legend(loc="upper left")
+
+plt.show() #<<
+```
+
 ### Point Plot
 
 `sns.pointplot()` is the same reading from the other direction: it estimates a
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -2,7 +2,7 @@
 
 > py-maidr is a Python library that makes matplotlib, seaborn, and plotly data visualizations accessible to blind and low-vision users through sonification, braille, text, and AI-powered conversational modalities. Developed by the (x)Ability Design Lab at the University of Illinois Urbana-Champaign.
 
-maidr (pronounced "mader") monkey-patches matplotlib, seaborn, and plotly at import time so users just `import maidr` and call `maidr.show(plot)` to generate interactive HTML with keyboard-navigable sonification, braille, text, and AI chat support. Supported plot types: bar, stacked bar, dodged bar, count, histogram, KDE, line, multi-line, step, pie, box, violin, heatmap, scatter, smooth/regression, candlestick, error bar, point, multi-panel, and facet plots. Works in Jupyter, Colab, VS Code, Streamlit, Shiny, and Quarto.
+maidr (pronounced "mader") monkey-patches matplotlib, seaborn, and plotly at import time so users just `import maidr` and call `maidr.show(plot)` to generate interactive HTML with keyboard-navigable sonification, braille, text, and AI chat support. Supported plot types: bar, stacked bar, dodged bar, count, histogram, KDE, line, multi-line, step, pie, box, violin, heatmap, scatter, smooth/regression, candlestick, error bar, point, area, stacked area, multi-panel, and facet plots. Works in Jupyter, Colab, VS Code, Streamlit, Shiny, and Quarto.
 
 ## Docs
 - [Documentation Home](https://py.maidr.ai/): Installation, quick start, and overview
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -12,6 +12,14 @@ class PlotType(str, Enum):
     HEAT = "heat"
     HIST = "hist"
     LINE = "line"
+    #: A filled band between a series and a baseline. Emitted for a single
+    #: `stackplot` band, which has nothing stacked on it.
+    AREA = "area"
+    #: Bands stacked on one another, so a band's height is its own series'
+    #: value while its top edge is the running total. A distinct type because
+    #: a line layer announces one number per point with nothing to say which
+    #: of those two it is.
+    STACKED_AREA = "stacked_area"
     PIE = "pie"
     SCATTER = "point"
     STACKED = "stacked_bar"
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -59,6 +59,7 @@ _DISPLAY_NAMES = {
     PlotType.HIST: "histogram",
     PlotType.SCATTER: "scatter",
     PlotType.STACKED: "stacked bar",
+    PlotType.STACKED_AREA: "stacked area",
     PlotType.VIOLIN_BOX: "violin",
     PlotType.VIOLIN_KDE: "violin",
 }
--- a/maidr/core/figure_manager.py
+++ b/maidr/core/figure_manager.py
@@ -59,6 +59,8 @@ class FigureManager:
         PlotType.VIOLIN_KDE: 1,
         PlotType.VIOLIN_BOX: 1,
         PlotType.ERRORBAR: 1,
+        PlotType.AREA: 1,
+        PlotType.STACKED_AREA: 2,
     }
 
     _instance = None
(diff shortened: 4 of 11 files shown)


jooyoungseo:
The DataFrame one is real, and it is worse than a `KeyError`. Fixed in `29f1685`; the other two I checked and am leaving as they are, with reasons.

## 1 — `_is_two_dimensional` and a DataFrame. Fixed

You predicted a raw `KeyError`. I ran it, and it fails two different ways depending on the column labels, neither of which raises anything a user would see as a bug report:

| `pd.DataFrame([[10, 20, 25, 30], [5, 8, 12, 14]])` | before | after |
|---|---|---|
| named columns (`a, b, c, d`) | `df["a"]`… no, `df[0]` → `KeyError` → **caught** → read as **one** series | 2 series, stacked area |
| positional columns (`0..3`) | `df[0]` → column zero → looks 2-D → `[row for row in df]` → the **labels** `0,1,2,3` → `ExtractionError` | 2 series, stacked area |

The first row is the bad one. The `except KeyError` you can see in `_is_two_dimensional` swallows it, so a two-series chart is described as a one-series area chart: right numbers in the wrong shape, no exception, nothing in the output to say so. The second row at least fails loudly — a chart that draws fine but cannot be described.

The root cause is that both ways of splitting the argument mean *columns* for a DataFrame: `values[0]` is the column labelled zero, and iterating a frame yields the labels themselves. Matplotlib reads it as rows. So the fix is to stop asking the object and ask numpy for the shape:

```python
array = np.asarray(values)
if array.ndim != 2:
    return None
return list(array)
```

`np.asarray` agrees with matplotlib for every form the argument takes — verified across a DataFrame with either label kind, a 2-D ndarray, a list of lists, a 1-D list, and a Series, all six now landing on the same schema as the equivalent `stackplot(x, y1, y2)` call. Ragged input raises `ValueError` from numpy, which is caught and returns `None`; matplotlib has already rejected the call by then, so there is nothing to describe either way.

`test_a_data_frame_of_series_is_read_by_row_like_matplotlib_reads_it` is parameterised over both label kinds, because the two mistakes fail differently and a test for one would have passed on the other. Both parameters fail on the previous commit.

## 2 — the `len(magnitudes) != len(positions)` guard

Your reasoning about the patch's call path is right: `_draw_quietly` runs matplotlib first, matplotlib rejects ragged series, so the guard cannot fire through `Axes.stackplot`.

Leaving it, because that is not the only path in. `AreaPlot` is reachable through `MaidrPlotFactory.create` and by direct construction, neither of which routes through the patch or through matplotlib's validation, and the guard is what stops a mismatched pair being zipped into a schema where the shorter one silently truncates the longer. Dead through one entry point is not dead.

Worth saying plainly, though: `test_a_ragged_call_is_refused_rather_than_described` tests matplotlib's rejection, not this branch, exactly as you say. The test name overstates what it covers and I would rather correct the record than let it read as coverage this branch does not have.

## 3 — `dtype=object` on the positions

Deliberate, and thanks for reading it that way rather than as an oversight. `stackplot`'s x is commonly a `DatetimeIndex` or a list of category strings, and a plain `np.asarray` narrows those in ways `_scalar`'s fallback then has to undo. The allocation is one array per layer at render time, not per point per navigation, so it is not on a path where it costs anything measurable.

## 4 — the bundle gap

Confirmed, and #357 is already open with the refreshed 4.1.0 bundle — `TraceFactory` on 4.0.0 throws `Invalid trace type` for `area`/`stacked_area` under `use_cdn=False`, so that one wants to land before or alongside this.

Full suite: **976 passed**. `ruff check` clean on both changed files.

---

jooyoungseo:
Both actionable ones fixed in `8b12fc2`, and the display-name test is broader than the finding.

## 1 — `_DISPLAY_NAMES` missing `STACKED_AREA`. Real, confirmed

```
AREA           -> 'area'
STACKED_AREA   -> 'stacked_area'      <- the schema token, shown to a user
STACKED        -> 'stacked bar'
DODGED         -> 'dodged bar'
```

Exactly as you described: it lands in `_show_fallback`'s one-sentence list beside "stacked bar" and "dodged bar", so someone who called `ax.stackplot` is shown a wire value. `AREA` needed no override, also as you said.

The test I added asserts **no member's display name contains an underscore**, then pins this one:

```python
for plot_type in PlotType:
    assert "_" not in plot_type.display_name, plot_type.name
```

Written that way because the finding is an instance of a class: `display_name` exists precisely because the wire value is not the user's word, and registering a multi-word type and remembering its override are two separate lines. The narrow test would have passed the next time someone forgets. It fails on the previous commit.

## 2 — highlighting checked counts, not identity. Also real

You put it exactly right: two selectors and two series pass a count check whichever way round they are paired, and the failure is silent. Worse than "a highlight in the wrong place" — everything the layer says about a band (its value, its name, its running total) is wrong for the band actually lit up, with nothing in the output to say so.

`test_the_nth_selector_points_at_the_band_the_nth_series_drew` ties them through **geometry**, deliberately not through the returned list, since that list is the assumption under test. A stacked band is drawn between the running total below it and the running total including it, so band *i*'s extent is a fact about series 0..i that no other band shares:

| band | expected extent | drawn |
|---|---|---|
| 0 (`SUBS`) | baseline → own values | `y=[0, 30]` |
| 1 (`SERVICES`) | band 0's values → the pair's totals | `y=[10, 44]` |

The test computes those bounds from the series arrays, matches each drawn path against them, and then asserts the element carrying that geometry is the one the schema names at the same index. If `stackplot` ever reordered its return, the bounds would stop matching the index and this fails rather than mispairing quietly.

## 3 — the unreachable guard

Agreed and unchanged, for the reason I gave in the previous round: `AreaPlot` is reachable through `MaidrPlotFactory.create` and by direct construction, neither of which routes through matplotlib's validation. Dead through one entry point is not dead. And repeating the correction from that round, since it stands: `test_a_ragged_call_is_refused_rather_than_described` tests matplotlib's rejection, not this branch — the name overstates what it covers.

## 4 — the bundle gap

Closed. #357 refreshed the bundle to 4.1.0 and is **merged**; I verified `maidr.css`/`maidr-math.css` are byte-identical between the two releases by fetching both and comparing, so only `maidr.js` moved. Why the staleness check stayed silent about it — `STALE_MINOR_GAP = 5` measures staleness in minor versions, not in trace types the bundle cannot render — is filed as **#358**.

Full suite: **978 passed**. `ruff check` clean on every file this branch touches (the 7 errors a repo-wide run reports are pre-existing — identical count with these changes stashed).

---

PR #357, opened 2026-08-11
chore: update bundled maidr.js to v4.1.0

jooyoungseo:
Refreshes the bundled `maidr.js` from **4.0.0** to **4.1.0**, using `.github/scripts/fetch-maidr-bundle.sh` — the same script both workflows call, so the SHA-512 verification against the npm tarball is unchanged (`Verified tarball sha512 integrity`).

## Why it did not happen on its own

#355 added the receiving half of the automation, and maidr's `release.yml:118` now sends the `repository_dispatch`. But that sending step landed on maidr's `main` **after v4.1.0 was cut**:

```
$ git show v4.1.0:.github/workflows/release.yml | grep -c repository_dispatch
0
```

So the v4.1.0 release never dispatched, and nothing would have until maidr publishes again. The two halves of the automation landed on either side of the release they were meant to cover.

## Why it matters more than a version number

v4.0.0 predates **every** trace type added since, and `TraceFactory` throws `Invalid trace type` on one it does not recognise — a throw that propagates out of `new Figure(...)`, so the figure renders as *nothing* rather than degrading.

This repository already emits `error_bar` from two patches (#349, #352, both on `main`). Comparing the two bundles on unambiguous tokens:

| token | 4.0.0 | 4.1.0 |
|---|---|---|
| `error_bar` | absent | present |
| `stacked_area` | absent | present |
| `stacked_normalized_area` | absent | present |
| `gauge` | absent | present |
| `waterfall` | absent | present |
| `word_cloud` | absent | present |
| `dumbbell` | absent | present |

(Checked with `grep -F` on those tokens rather than on `area`, which is a substring of ordinary identifiers and gives a false positive — worth noting because my first pass reported it as present in 4.0.0.)

## Scope of the problem this closes

Narrower than the table suggests, and worth stating precisely rather than overselling:

- The **CDN path** — the default — resolves the published version at render time, so it was already serving 4.1.0 and was never affected.
- The **bundled path** (`use_cdn=False`, or CDN unreachable) served 4.0.0.
- `release.yml` refreshes the bundle at release time, so a *released* py-maidr would have carried a matching one regardless. The gap was on `main`, between releases.

What made it worth acting on rather than waiting is that it was **silent**. `bundle_status()` reported:

```
BundleStatus(bundled='4.0.0', published='4.1.0', is_behind=True, is_stale=False)
```

`is_stale` is false because `STALE_MINOR_GAP = 5` and this was one minor behind, so `warn_if_bundle_is_stale()` said nothing. That threshold reads as calibrated for a bundle that is *old but working* — which is a different thing from one that predates a trace type this package now emits, where the failure is not a missing feature but a blank figure.

I have not changed the threshold: what the right rule is — a version floor per emitted type, a check in CI, or simply a smaller gap — is a design question rather than a number to nudge, and this refresh removes the immediate exposure either way.

## Verification actually run

| Step | Result |
|---|---|
| `fetch-maidr-bundle.sh 4.1.0` | `Verified tarball sha512 integrity` |
| `uv run pytest` | **962 passed** |
| `bundle_status()` after | `bundled='4.1.0', published='4.1.0', is_behind=False` |

---


jooyoungseo:
Both notes answered.

## The CSS files — confirmed, and not by inspection

You were right to ask, and right about the mechanism. I checked it directly rather than reasoning from the diff: fetched 4.0.0 and 4.1.0 into separate directories with the same script and compared all three assets byte for byte.

```
maidr.css        byte-identical across 4.0.0 -> 4.1.0
maidr-math.css   byte-identical across 4.0.0 -> 4.1.0
maidr.js         1373982 -> 1409655 bytes
```

So the script did rewrite all three, and git saw a change in one. `maidr-math.css` is KaTeX, which explains it — it moves when KaTeX moves, not when maidr does.

I also re-ran the fetch against the committed tree: all four files, `VERSION` included, come back identical to what is in this PR. Nothing hand-edited, and the diff reproduces from the script.

## Follow-up issue — filed as #358

Filed with the diagnosis rather than just the observation, since the interesting part is *why* it stayed quiet. `STALE_MINOR_GAP = 5` measures staleness as a distance in minor versions, so a bundle one minor behind is never stale — while being unable to render seven of the trace types this package emits. Version distance is a poor proxy for the question a user actually has, which is whether the bundle can draw what it is about to be handed.

#358 sketches three fixes and argues for the one that fails safe: a per-`PlotType` floor checked on the emit path, so a `use_cdn=False` user is told which bundle version they need instead of being shown an empty figure.

All 16 checks green.

---

Issue #353, opened 2026-08-11
fix: category tick labels are recovered on the x axis only


jooyoungseo:
## What

`LineExtractorMixin.extract_line_data_with_categorical_labels` maps a point's coordinate to the name written on the tick beside it. It reads `ax.get_xticks()` / `ax.get_xticklabels()` and nothing else, so the recovery only ever happens for the x axis.

## Why it matters

A horizontal categorical line chart announces its groups as numbers.

`sns.pointplot(frame, y="group", x="value", hue="half", dodge=True)` emits:

```json
[{"x": 4.33, "y": -0.025, "z": "x"}, {"x": 22.0, "y": 0.975, "z": "x"}]
```

`-0.025` and `0.975` are the drawn positions of the categories, dodged aside to make room for the second hue level. The chart writes `a` and `b` on the y ticks; the reader hears the offsets.

The same limit applies to any layer that routes through `MultiLinePlot` with a categorical y.

## Where it stands

`tests/core/plot/test_pointplot.py::test_a_horizontal_dodge_still_reports_raw_category_offsets` pins the current behaviour and points here, so the gap is visible in the suite rather than only in the output. That test should turn red — and be replaced with one asserting the labels — when this is fixed.

## Sketch

The helper needs to follow the *category* axis rather than assume x. `ax.xaxis.units is not None` already identifies a string-category axis (matplotlib leaves `UnitData` there), and the same question can be asked of `ax.yaxis`. The complication is the return contract: it hands back `(x, y)` tuples with only the first labelled, so labelling y means widening what it returns and updating `MultiLinePlot` with it.

Worth doing together with the rounding fallback added in #352, which recovers a *dodged* group's name by rounding to the nearest tick — that has the same x-only limit for the same reason.

jooyoungseo:
This is fixed on `main` and the issue was left open by oversight. Verified by re-running the exact reproduction from the description:

```python
sns.pointplot(frame, y="group", x="value", hue="half", dodge=True)
```

```json
[[{"x": 3.0, "y": "a", "z": "x"}, {"x": 21.0, "y": "b", "z": "x"}],
 [{"x": 4.0, "y": "a", "z": "y"}, {"x": 25.0, "y": "b", "z": "y"}]]
```

The `-0.025` and `0.975` this issue reported are now `"a"` and `"b"` — the names written on the y ticks — and the dodge offsets are recovered along with the plain positions, which was the second half of the ask.

[...]

PR #359, opened 2026-08-12
fix: two silent misreadings — highlight targets and category axis names

### LINKED ISSUE #354: fix: `_elements` accumulates across repeated renders, duplicating highlight targets
## What

`MaidrPlot._elements` is initialised once in `__init__` and appended to during extraction. `render()` can run more than once on the same plot — `set_id` renders again when the schema is not yet cached — and every subclass appends without clearing first, so the list grows by one full set of artists per render.

Reproduced while working on #352:

```python
plot = _plots(fig)[0]
plot.render()
len(plot.elements)   # 6, for a chart with 3 points
```

## Why it matters

`elements` is the ordered list the highlight machinery tags, and the frontend indexes into the resolved selection **by point index**. A doubled list leaves index *n* pointing at the artist for point *n mod count*, so the highlight lands on the wrong element for every point past the first set. Nothing errors.

It is latent rather than active for most types today, because the second render is not guaranteed on every path — which is also why it survived this long.

## Where it stands

`PointPlot._extract_plot_data` clears the list before it appends (added in #352), so that one type is correct. Every other type carries the pattern:

`barplot.py`, `boxplot.py`, `errorbar.py`, `grouped_barplot.py`, `heatmap.py`, `histogram.py`, `lineplot.py`, `mplfinance_barplot.py`, `mplfinance_lineplot.py`, `pieplot.py`, `regplot.py`, `scatterplot.py`, `violin_box_plot.py`, `violin_kde_plot.py`.

## Why it wasn't just fixed in the base

The obvious fix — clear `self._elements` in `MaidrPlot.render()` before calling `_extract_plot_data()` — is **not safe as written**. `ViolinKdePlot` appends to `_elements` in its `__init__`, before any render happens, so a clear in `render()` would discard those and leave the violin bodies untagged.

So the fix is a small audit rather than a one-liner: either move that tagging into extraction and then clear centrally, or clear per class where extraction owns the whole list. Worth doing in one pass with a test per type asserting `render()` twice leaves the element count unchanged, since the failure mode is silent.

jooyoungseo:
## Description

Closes #354 and #353.

Two separate defects, both of which produce a chart that is confidently wrong rather than obviously broken, and neither of which any assertion on the emitted data could see. They are on one branch because the branch is fixed for this work; they are separate commits and can be reviewed as such.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

---

# 1. `_elements` accumulated across renders (#354) — `d5a3fbf`, `0a63f73`

`MaidrPlot._elements` was initialised once in `__init__` and appended to during extraction, so the list grew by a full set of artists per render. It is the ordered list the highlight machinery tags, and the frontend indexes into the resolved selection **by point index** — so a doubled list left point *n* pointing at the artist for point *n mod count*.

Nothing errors. No announcement changes. The data is right, the braille is right, the sonification is right, and the outline lands on a different mark from the one being read. Same shape as the highlight-only bugs on xability/maidr#814.

Rendering more than once is routine: `schema`, `elements` and `set_id` each render when nothing is cached, and a caller may render explicitly besides.

### Changes

**Extraction now owns the whole list** — `render()` clears `_elements` before `_extract_plot_data()`, one mechanism for all fourteen types the issue listed.

Two things made that alone wrong, which is why the issue called for an audit rather than a one-liner:

- **`ViolinKdePlot` registered its bodies in `__init__`.** A central clear would have dropped the `PolyCollection` for each violin and left the bodies untagged — *worse* than the doubling, because the highlight would disappear rather than land on a neighbour. Both its lists come from the constructor, so the bodies are registered in `_extract_plot_data()` now.
- **Three layers re-ran their own extraction inside `render()`, which the clear cannot see.** `CandlestickPlot`, `MplfinanceBarPlot` and `MplfinanceLinePlot` each called `super().render()` — which extracts — and then overwrote `data` and `selector` with a second extraction in order to refresh the axes. Only the axes needed refreshing; `super().render()` had already set both keys to the same values. These three doubled within a **single** render, and the second call comes *after* the clear.

**Two grew the list by rebinding rather than appending** (`self._elements = sorted_patches`, `= [body, wick]`), which would have left `_elements` pointing at a different list from the one cleared. Both `extend` now.

**And one more list with the same problem, found in review** (`0a63f73`). A horizontal violin emits its series in the reverse of the drawn order, to match the box layer's convention. `all_violins` is rebuilt each render so reversing it is idempotent — but `self._poly_gids` comes from the constructor and persists, and it was reversed **in place**. On every second render it was back in drawn order while the data was still reversed, so selector *i* addressed a different violin from point *i*:

```
render 1: gids in drawn order = False    data first-y = 5.025
render 2: gids in drawn order = True     data first-y = 5.025
render 3: gids in drawn order = False    data first-y = 5.025
```

`_get_selector()` now derives the order from a copy and writes nothing back. The first render is unchanged; every even one is corrected.

`AreaPlot` and `PointPlot` already cleared at the top of extraction and keep doing so; their *second* clear — the one that gives up on highlighting when the count disagrees with the data — is a different mechanism and is untouched.

### Tests

New `tests/core/plot/test_element_stability.py`, **33 cases**. The parametrised pair covers fourteen layer types and asserts that rendering twice tags the same artists, and that rendering five times never grows the list — because a fix that cleared on alternate renders would satisfy the first. Then five that check what those cannot:

- **a bar layer tags one artist per bar.** Stable is not the same as right: every assertion above passes on a list consistently twice as long as it should be.
- **the mplfinance layers extract once per render**, with counts checked against the chart.
- **a violin keeps its bodies as well as its curves** — the case that would have caught the naive central clear.
- **a horizontal violin keeps its selectors paired with its data** across four renders, with `_poly_gids` left as it was found.
- **a vertical violin addresses its violins in drawn order** — because a fix that reversed unconditionally would satisfy the case above and pair every vertical violin with the wrong body.

**Verified they bite.** Removing the clear fails **26 of 30**; restoring the redundant re-extraction fails the mplfinance case; restoring the in-place reversal fails the horizontal violin case. None of the three failures overlaps another, so each part of the fix is independently guarded.

---

# 2. Category names were recovered on the x axis only (#353) — `da854ac`

`LineExtractorMixin.extract_line_data_with_categorical_labels` read `ax.get_xticks()` and nothing else, so a **horizontal** categorical chart announced its groups as the positions they were drawn at:

```python
sns.pointplot(frame, y="group", x="value", hue="half", dodge=True)
# before: [{"x": 1.5, "y": -0.025, "z": "x"}, {"x": 20.5, "y": 0.975, "z": "x"}]
# after:  [{"x": 1.5, "y": "a",    "z": "x"}, {"x": 20.5, "y": "b",    "z": "x"}]
```

`-0.025` and `0.975` are where two hue levels were shifted to make room for each other, around the ticks the chart writes `a` and `b` on. The reader was given the offsets and no route to the names.

### Changes

Both axes are asked the same question, and each coordinate is named by its own axis. Which axis is categorical is what matplotlib records in `axis.units`, so a horizontal chart stops being a special case: it is the same lookup on the other axis.

The tick map is now **empty** on a numeric axis rather than merely ungating the rounding, and that fixes a second misreading in the same function. A point sitting exactly on a numeric tick took that tick's *label text*, so:

```python
ax.plot([1.0, 2.0, 3.0], [4.0, 5.0, 6.0])
# before: {"x": "1.00", "y": 4.0}   <- a string, and to two decimal places
# after:  {"x": 1.0,    "y": 4.0}
```

A formatted rendering of a number is not a name for it, and the frontend wants the number. Nothing in the suite depended on the old reading.

### Tests

`tests/core/plot/test_pointplot.py`: the pinned `test_a_horizontal_dodge_still_reports_raw_category_offsets` is **replaced** by one asserting the names, which is what the issue asked for. Four more cover the exact lookup (undodged, on ticks) as distinct from the rounding, the vertical case that already worked, a numeric axis keeping numbers on both axes, and a measurement near a tick not being renamed after it.

**Verified they bite:** reverting the extractor fails 3 of the 5. Two pass either way and are honest about it — the vertical case is a regression guard, and the near-a-tick case guards a future over-correction rather than reproducing the old one.

---

n/a — both changes are in extracted data and tagged artists, pinned by test rather than by eye.

## Checklist

## Additional Notes

| Gate | Result |
|---|---|
| `uv run pytest` | **1015 passed** (was 978) |
| `uv run ruff check` on the changed files | clean |

`uv run ruff check` across the repo reports 7 pre-existing `F401` re-export findings in four `__init__.py` files. None are in files this touches and the count is unchanged by the branch.

### DIFF
--- a/maidr/core/plot/boxplot.py
+++ b/maidr/core/plot/boxplot.py
@@ -250,6 +250,17 @@ class BoxPlot(
         if bxp_stats is None:
             return None
 
+        # Three lists beside `_elements` are filled here and read by
+        # `_get_selector()`, and they carried the same defect: appended to on
+        # every render, so a second one produced twice as many selectors as
+        # there are boxes. Worse than the extra entries, this loop stamps a
+        # fresh gid on each artist every time, so after a second render the
+        # *first* half of the map names gids no artist carries any more and
+        # those selectors resolve to nothing (#354).
+        for gids in self.elements_map.values():
+            gids.clear()
+        self.lower_outliers_count.clear()
+
         whiskers = self._bxp_extractor.extract_whiskers(bxp_stats["whiskers"])
         caps = self._bxp_extractor.extract_caps(bxp_stats["caps"])
         medians = self._bxp_extractor.extract_medians(bxp_stats["medians"])
--- a/maidr/core/plot/candlestick.py
+++ b/maidr/core/plot/candlestick.py
@@ -76,8 +76,10 @@ class CandlestickPlot(MaidrPlot):
             self._maidr_wick_gid = wick_collection.get_gid()
             self._maidr_gid = self._maidr_body_gid or self._maidr_wick_gid
 
-            # Use the original collections for highlighting
-            self._elements = [body_collection, wick_collection]
+            # Use the original collections for highlighting. Extended rather
+            # than rebound, so `self._elements` stays the list `render()`
+            # cleared rather than becoming a different one.
+            self._elements.extend([body_collection, wick_collection])
 
             # Extract data directly from DataFrame
             if self._maidr_original_data is not None and isinstance(
@@ -231,7 +233,8 @@ class CandlestickPlot(MaidrPlot):
                 axis_cfg[MaidrKey.FORMAT] = prev[MaidrKey.FORMAT]
         base_schema[MaidrKey.AXES] = axes_data
 
-        base_schema[MaidrKey.DATA] = self._extract_plot_data()
-        if self._support_highlighting:
-            base_schema[MaidrKey.SELECTOR] = self._get_selector()
+        # Data and selector are left as `super().render()` built them. Running
+        # the extraction a second time here appended a second set of artists
+        # to `self._elements` within one render, which the frontend then
+        # indexes into by point index (#354). Only the axes needed refreshing.
         return base_schema
--- a/maidr/core/plot/maidr_plot.py
+++ b/maidr/core/plot/maidr_plot.py
@@ -77,6 +77,18 @@ class MaidrPlot(ABC, FormatExtractorMixin):
         if format_config:
             self._merge_format_into_axes(axes_data, format_config)
 
+        # Extraction owns the whole element list, so it starts empty every time.
+        #
+        # A layer is rendered more than once -- `schema`, `elements` and
+        # `set_id` each render when nothing is cached, and a caller may render
+        # explicitly besides -- and every subclass appended without clearing,
+        # so the list grew by a full set of artists per render. `elements` is
+        # the ordered list the highlight machinery tags and the frontend
+        # indexes into by point index, so a doubled list leaves point n
+        # pointing at the artist for point n mod count. Nothing errors; the
+        # outline simply lands on the wrong mark (#354).
+        self._elements.clear()
+
         # Generate a unique UUID for this layer to ensure each plot layer can be distinctly identified
         # in the MAIDR frontend. This supports robust layer switching.
         maidr_schema = {
--- a/maidr/core/plot/mplfinance_barplot.py
+++ b/maidr/core/plot/mplfinance_barplot.py
@@ -86,8 +86,11 @@ class MplfinanceBarPlot(
         # Sort patches by x-coordinate to maintain order
         sorted_patches = sorted(volume_patches, key=lambda p: p.get_x())
 
-        # Set elements for highlighting (use the patches directly)
-        self._elements = sorted_patches
+        # Set elements for highlighting (use the patches directly). Extended
+        # rather than rebound: `render()` clears the list it owns, and a
+        # rebinding would leave `self._elements` pointing at a different list
+        # from the one that was cleared.
+        self._elements.extend(sorted_patches)
 
         # Use datetime converter for enhanced data extraction
         if self._maidr_datetime_converter is not None:
@@ -167,7 +170,8 @@ class MplfinanceBarPlot(
                 axis_cfg[MaidrKey.FORMAT] = prev[MaidrKey.FORMAT]
         base_schema[MaidrKey.AXES] = axes_data
 
-        base_schema[MaidrKey.DATA] = self._extract_plot_data()
-        if self._support_highlighting:
-            base_schema[MaidrKey.SELECTOR] = self._get_selector()
+        # Data and selector are left as `super().render()` built them. Running
+        # the extraction a second time here appended a second set of patches
+        # to `self._elements` within one render, which the frontend then
+        # indexes into by point index (#354). Only the axes needed refreshing.
         return base_schema
--- a/maidr/core/plot/mplfinance_lineplot.py
+++ b/maidr/core/plot/mplfinance_lineplot.py
@@ -208,7 +208,8 @@ class MplfinanceLinePlot(MaidrPlot, LineExtractorMixin):
                 axis_cfg[MaidrKey.FORMAT] = prev[MaidrKey.FORMAT]
         base_schema[MaidrKey.AXES] = axes_data
 
-        base_schema[MaidrKey.DATA] = self._extract_plot_data()
-        if self._support_highlighting:
-            base_schema[MaidrKey.SELECTOR] = self._get_selector()
+        # Data and selector are left as `super().render()` built them. Running
+        # the extraction a second time here appended a second set of lines to
+        # `self._elements` within one render, which the frontend then indexes
+        # into by point index (#354). Only the axes needed refreshing.
         return base_schema
--- a/maidr/core/plot/violin_box_plot.py
+++ b/maidr/core/plot/violin_box_plot.py
@@ -164,7 +164,16 @@ class ViolinBoxPlot(MaidrPlot):
             self._elements.append(artist)
 
     def _tag_sns_artists(self) -> None:
-        """Tag seaborn inner-box Line2D artists with GIDs."""
+        """
+        Tag seaborn inner-box Line2D artists with GIDs.
+
+        The gid map is rebuilt rather than appended to. `_build_sns_selectors`
+        emits one selector per entry, so a second render would have doubled
+        the selector list against unchanged data -- the same defect as #354,
+        in a list beside `_elements` rather than in it.
+        """
+        self._sns_gids.clear()
+
         for violin_lines in self._sns_box_lines:
             gid_map: dict[str, str] = {}
             for role, line in violin_lines.items():
--- a/maidr/core/plot/violin_kde_plot.py
+++ b/maidr/core/plot/violin_kde_plot.py
@@ -57,10 +57,30 @@ class ViolinKdePlot(MaidrPlot):
     # Selector
     # ------------------------------------------------------------------
     def _get_selector(self) -> list[str]:
-        """Return one CSS selector per violin using PolyCollection GIDs."""
-        if self._poly_gids:
-            return [f"g[id='{gid}'] path, g[id='{gid}'] use" for gid in self._poly_gids]
-        return ["g[id^='maidr-'] path"]
+        """
+        Return one CSS selector per violin using PolyCollection GIDs.
+
+        In the order the data was emitted, which for a horizontal violin is
+        the reverse of the order the collections were drawn in -- the box
+        layer's convention, which the KDE layer matches.
+
+        The reversal is done here rather than to ``self._poly_gids``, because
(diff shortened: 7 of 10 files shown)


jooyoungseo:
A second issue is now on this branch — #353 as well as #354 — and the PR title and body are updated to say so. They are separate commits (`d5a3fbf` and `da854ac`) touching disjoint files, so they can be reviewed independently; the branch is fixed for this work, which is the only reason they share a pull request.

They belong together in one respect worth naming. Both are readings that are *confidently wrong* rather than broken: #354 puts the outline on a mark the reader is not on, #353 announces a dodge offset as though it were a group name. In each case the data is right, the announcement is fluent, and nothing in the suite could see it — which is the same blind spot xability/maidr#814 ended up recording, and the same one that produced xability/r-maidr#145 and #151.

**One thing found while fixing #353 that the issue did not predict.** The x-side lookup substituted a tick's *label text* whenever a point sat exactly on a tick, without asking whether the axis was categorical at all:

```python
ax.plot([1.0, 2.0, 3.0], [4.0, 5.0, 6.0])
# {"x": "1.00", "y": 4.0}
```

A plain numeric line chart reported its x as the string `"1.00"`. The issue's sketch — gate on `axis.units` — fixes this as a consequence rather than as a separate change, because the gate belongs on building the tick map rather than on the rounding alone. Worth flagging because it is a behaviour change beyond the issue's stated scope: a numeric x is now a number. Nothing in the suite depended on the old reading, and `tests/core/plot/test_pointplot.py::test_a_numeric_axis_keeps_its_numbers` pins it.

## Verification

| Gate | Result |
|---|---|
| `uv run pytest` | **1013 passed** (was 978) |
| `uv run ruff check` on the changed files | clean |

Both fixes verified to bite by reverting them: the element clear fails 26 of 30 of the new stability cases, the redundant re-extraction fails the mplfinance case specifically, and the extractor change fails 3 of the 5 new label cases.

---

jooyoungseo:
The `_poly_gids` finding is real, and it is the best catch on this PR. Fixed in `0a63f73`.

**Reproduced before touching anything**, on a horizontal `Axes.violinplot`:

```
render 1: gids in drawn order = False    data first-y = 5.025
render 2: gids in drawn order = True     data first-y = 5.025
render 3: gids in drawn order = False    data first-y = 5.025
render 4: gids in drawn order = True     data first-y = 5.025
```

Exactly your reading. `all_violins` is rebuilt every call so its reversal is idempotent; `self._poly_gids` is the constructor's list and the in-place reversal toggles it. On every even render the selectors are in drawn order while the data is reversed, so selector *i* addresses a different violin from point *i* — and the highlight lands on a neighbour, with nothing announced differently. It is the same defect as #354 one list further along, which is why it belongs in this PR rather than a follow-up.

**Fixed by deriving the order rather than writing it back.** `_get_selector()` reverses a copy when the layer is horizontal; `_poly_gids` is left as the constructor set it. The first render is unchanged — it produced the reversed order before and still does — and every even one is corrected.

**Two new cases**, and you were right that the existing one could not see this: `test_a_violin_keeps_its_bodies_as_well_as_its_curves` builds a vertical violin, so `is_horz` was never true anywhere in the file.

- `test_a_horizontal_violin_keeps_its_selectors_paired_with_its_data` renders four times and asserts the selector list is identical across all four, that it is the reverse of the drawn gid order — the order the data comes out in — and that `_poly_gids` is left as it was found.
- `test_a_vertical_violin_addresses_its_violins_in_drawn_order` is the counterpart, because a fix that reversed unconditionally would satisfy the stability check above while pairing every vertical violin with the wrong body.

**Verified it bites:** restoring the in-place reversal fails the horizontal case and nothing else.

On your other two points — yes, `AreaPlot`/`PointPlot`'s extraction-time clear is now redundant and left in place deliberately: it also guards a direct `_extract_plot_data()` call, and their *second* clear (giving up on highlighting when the count disagrees with the data) is a different mechanism entirely. And yes on `mplfinance` being a core dependency, which is why the case runs rather than skips — `importorskip` is there as a courtesy to a stripped environment, not because the dependency is optional.

## Verification

| Gate | Result |
|---|---|
| `uv run pytest` | **1015 passed** |
| `uv run ruff check` on the changed files | clean |

---

jooyoungseo:
Two things on the latest review, one of which corrects me.

**The `_poly_gids` finding was already fixed** in `0a63f73`, pushed a few seconds before that review posted — it reviewed the tree without it. Same diagnosis, same fix as suggested. Nothing further needed there.

**But the sweep it reported was wrong, and so was my acceptance of it.** It said `self.<attr>.reverse()/.sort()/.pop()` inside an extractor was the only instance and "the rest of the fix looks complete." I re-ran the sweep for `append` as well as for mutation-in-place, and found the same defect in two more places. `f89af98` fixes them.

**`BoxPlot` keeps four lists beside `_elements`** — `elements_map["min" | "max" | "median" | "boxes" | "outliers"]` and `lower_outliers_count` — all filled in `_extract_bxp_maidr` and read by `_get_selector()`. Reproduced on a two-box chart:

```
render 1: counts=[0, 1]              selectors=2  data=2
render 2: counts=[0, 1, 0, 1]        selectors=4  data=2
render 3: counts=[0, 1, 0, 1, 0, 1]  selectors=6  data=2
```

Three boxes, nine selectors. And it is worse than surplus entries: the same loop stamps a **fresh** gid on each artist every render, so after a second render the first half of the map names gids no artist carries any more — those selectors resolve to nothing at all, and the ones that do resolve are at the wrong indices.

**`ViolinBoxPlot._sns_gids`** is the same shape: appended per violin in `_tag_sns_artists`, and `_build_sns_selectors` emits one selector per entry. Fixed with them, but reported honestly — seaborn 0.13.2 does not reach that path in this environment (its inner box is not classified as box lines, so only a `ViolinKdePlot` is registered), so it is fixed by inspection rather than reproduced.

**Two new cases**, and the first is the general one this file was missing:

- `test_rendering_twice_emits_the_same_selectors` across all fourteen layer types. It asserts the selector **count**, not the payload — several layers mint a fresh gid per artist each render and name the new one, which stays consistent because the artist carries the gid the selector asks for. What must not change is how many, because that is what the frontend pairs with the data.
- `test_a_box_layer_emits_one_selector_per_box` counts against the chart: three distributions, three selectors, three points, on every one of three renders.

**Verified they bite:** removing the clear fails both, and nothing else.

The correction worth stating plainly: my PR description claimed the audit was complete for this defect class, and it was not. `_elements` was the list the issue named, and I fixed the list the issue named. Two of the three follow-on findings — `_poly_gids` and now these — came from looking beside it.

## Verification

| Gate | Result |
|---|---|
| `uv run pytest` | **1030 passed** |
| `uv run ruff check` on the changed files | clean |

---

jooyoungseo:
The one substantive question — *what user-facing path actually reproduces this pre-fix* — deserves a measured answer rather than a confident one, so I measured it. The answer is split.

**#353 needs no repeat render at all.** A horizontal dodged point plot announces `-0.025` and `0.975` on the first and only render. That is the plain `maidr.show()` path.

**One part of #354 is also single-render**, and it is the part I had not separated out. `MplfinanceLinePlot` re-ran its own extraction inside `render()`, so it doubled within one. On the natural lazy path — `_create_html_tag` reading `plot.elements`, one render, nothing called directly:

```
                     before     after
CandlestickPlot         2          2
MplfinanceBarPlot       8          8
MplfinanceLinePlot      2          1     <- one moving average, two tagged artists
```

The other two are unchanged because they *assigned* rather than appended, which is idempotent — which is also why switching them to `extend` was only safe alongside the central clear.

**The rest of #354 needs a second render**, and you are right that the straight-line path does not produce one. `_create_html_tag` reads `plot.elements` first, which renders and caches `_schema`; `_flatten_maidr` then reads `plot.schema`, which is cached. Rendering the same figure again reuses the same `MaidrPlot` objects with the cache intact. So the reachable routes today are a direct `render()` call — which is exactly how the issue was found, while working on #352 — and anything that constructs a layer and renders it more than once.

So the honest framing, which I have not made in the PR body and should have: **one of these was live for users and the rest were latent.** They are worth fixing anyway — `render()` is public, the invariant is the correct one, and a cache that happens to hide a bug is not the same as the bug not being there — but "latent" is the accurate word for most of it, and the PR body implied otherwise.

On the duplicated comment across the three `render()` overrides: agreed it is near-verbatim, and I left it that way deliberately. Each of the three is a separate file a maintainer will land in without the other two open, and the note is what stops someone re-adding the line to refresh the axes. If a fourth appears, it is worth hoisting.

---

PR #360, opened 2026-08-13
feat(bundle): warn when the bundle cannot draw the layer being emitted

### LINKED ISSUE #358: Bundle staleness is measured in version numbers, not in trace types the bundle cannot render
## What happened

`maidr/static/VERSION` sat at `4.0.0` while this package emitted `error_bar`, `waterfall`, `gauge`, `dumbbell`, `word_cloud`, `stacked_area` and `stacked_normalized_area` layers. None of those trace types exist in the 4.0.0 bundle, so under `use_cdn=False` a user got `Invalid trace type` from `TraceFactory` and a blank figure.

Nothing warned. `bundle_status()` reported `is_behind=True, is_stale=False`, and `warn_if_bundle_is_stale()` only speaks on `is_stale`.

## Why it stayed quiet

`maidr/util/dependencies.py`:

```python
STALE_MINOR_GAP = 5
```

Staleness is a distance in minor versions. 4.0.0 against 4.1.0 is a gap of one, so the bundle was nine minors away from qualifying — while being unable to render seven of the layer types this package emits.

That threshold answers "how old is this bundle?" The question a user needs answered is "can this bundle draw what I am about to hand it?", and version distance is a poor proxy: a bundle five minors behind may render everything we emit, and one minor behind may render none of a newly added type.

## The shape of a fix

The information needed is already on both sides. `PlotType` enumerates what this package can emit, and the bundle's own `TraceFactory` enumerates what it can build. Some options, roughly in order of cost:

1. **Declare a floor per plot type.** A mapping from `PlotType` to the bundle version that first rendered it, checked against `maidr/static/VERSION` at render time. Cheap and exact; the cost is one line per new type, and forgetting that line is silent in exactly the way this issue describes.
2. **Ask the bundle.** Grep the bundled `maidr.js` for the trace-type string literals at build or import time. No per-type bookkeeping, but it couples to the bundle's internal spelling.
3. **Warn on any drift once a release adds a type.** Drop `STALE_MINOR_GAP` to something small, or treat `is_behind` as sufficient for a warning. Blunt, and risks warning fatigue on releases that change nothing we depend on.

Option 1 is the one that fails safe: the check is on the emit path, so a user hitting a type the bundle cannot draw is told which version they need rather than seeing an empty figure.

## Not urgent

`release.yml` refreshes the bundle at release time, so a released wheel does not carry a stale bundle for long, and #357 has already closed the current gap. What is worth fixing is the class of failure — a `use_cdn=False` user on an older wheel gets an empty figure and no explanation.

Raised out of the review on #357, which declined to fix it there as out of scope. That was the right call for that PR.

jooyoungseo:
## Description

Closes #358.

`STALE_MINOR_GAP = 5` measures drift in minor versions, which answers *how old is this bundle?* The question a reader needs answered is *can this bundle draw what I am about to hand it?*, and the two come apart in both directions — a bundle five minors behind may render everything this package emits, and one minor behind may render none of a newly added type.

The gap that prompted the issue was a bundle **nine minors short of qualifying as stale** while unable to render seven of the layer types being emitted. `bundle_status()` reported `is_behind=True, is_stale=False`, and nothing spoke.

**It also reaches an audience the existing warning documents itself as unable to reach.** `warn_if_bundle_is_stale` compares against a published version it deliberately will not fetch, so a process rendering `use_cdn=False` offline — the user whose bundle is the only thing that runs — stays silent however old it is. That is written down in its own docstring as a known blind spot. This check reads the installed file and makes no request at all.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

The issue sketched three options. This takes **option 2, "ask the bundle"**, because the information is already there and option 1's per-type table costs one line per new plot type — and forgetting that line is silent in exactly the way the issue describes.

- **`bundle_trace_types()`** collects every quoted lower-snake identifier from the shipped `maidr.js`, once per process. The bundle's own `TraceFactory` switches on these strings, so the file already knows the answer.
- **`schema_trace_types(schema)`** walks an emitted schema for its layer `type` values.
- **`warn_if_bundle_cannot_render(types, *, bundle_is_primary)`** names the types the bundle has no renderer for. Wired into `Maidr._create_html_tag` and the Plotly equivalent, on the two paths where the bundled copy can actually execute — the same `bundle_is_primary` split the staleness warning already uses, so `use_cdn="auto"` reports to the logger rather than raising a warning about code that did not run.

**Two things the first draft got wrong, both caught by rendering a plain bar chart before writing any test:**

1. **The 4.1.0 bundle quotes its identifiers with backticks.** A scan written for `"` and `'` finds nothing, and every layer of every chart is reported missing.
2. **A schema in memory holds a `PlotType`, not a string.** `PlotType` subclasses `str`, so it compares and hashes as its value — while `str()` of it is `"PlotType.BAR"`. Asking the bundle about that name warns about every chart. Both are pinned by tests now.

What it says, on a bundle asked for a type it does not carry:

```
maidr: the bundled copy of maidr.js is 4.1.0, which has no renderer for
sankey, treemap. The bundle is what renders when the CDN is disabled
(use_cdn=False) or unreachable (use_cdn='auto'), and a layer it cannot
build is dropped rather than drawn, so those plots come out blank.
Upgrade py-maidr to pick up a refreshed bundle, or render with
use_cdn=True. Set MAIDR_BUNDLE_STALE_WARNING=0 to silence this warning.
```

**Nothing fires today**, and that is the correct state: the 4.1.0 bundle carries every type this package currently emits — including `error_bar`, `waterfall`, `gauge`, `dumbbell` and `word_cloud`, which #357 refreshed in. It does *not* carry `treemap`, `sankey` or `choropleth`, so the day a binding starts emitting those against this bundle, it says so.

## Checklist

## Additional Notes

**Deliberately biased towards silence.** The scan collects identifiers without regard to where they appear, so a string coinciding with a trace type in unrelated code counts as support. A false *"your bundle is fine"* leaves the reader where they already were; a false *"your bundle cannot draw this"* sends them chasing a version they do not need. The docstring says so rather than implying the check is a proof.

Three refusals guard the loud direction, each with a test: an unreadable bundle says nothing (a missing file is not evidence that nothing renders); `use_cdn=True` says nothing (the page never references the bundle); and a supported chart says nothing, which is the over-correction that would teach everyone to ignore it.

`tests/core/test_bundle_capability.py`, **14 cases**. Beyond the above: each type is named once per process per severity — so a chart in a loop says it once, a *second* unsupported type is not swallowed by the first, and the quiet `auto` report does not consume the loud `use_cdn=False` one, which is the audience the check exists for. Reverting the wiring in `Maidr._create_html_tag` fails the render-path case and nothing else.

One case is worth flagging as a design note rather than a test: `PlotType.COUNT` exists and no schema ever carries `count` — a `seaborn.countplot` is classified `COUNT` and emitted as `bar`. A check driven by the enum would ask the bundle about a string it will never build and warn about a chart that renders perfectly. That is why this reads the schema.

| Gate | Result |
|---|---|
| `uv run pytest` | **1044 passed** (was 1030) |
| `uv run ruff check` on the changed files | clean |

`uv run ruff check` across the repo reports the same 7 pre-existing `F401` re-export findings in four `__init__.py` files; the count is unchanged by this branch.

---


jooyoungseo:
All four taken — `965265d`.

**1. The missing export was the right thing to block on.** `MaidrBundleStaleWarning` is re-exported from the package root and the package docstring shows the filter; the capability warning was reachable only through `maidr.util.dependencies`, which silently breaks the same story for anyone running under `-W error`. And you put your finger on why no test caught it: every case in the file imports from the module, so none of them exercises the package attribute at all. There is one now, asserting both `maidr.MaidrBundleTraceWarning` and its presence in `__all__`, and the class docstring shows the filter the way the sibling's does.

**2. The shared lock.** Agreed and fixed — `_bundle_trace_lock`. They hold no invariant between them, and as you say the staleness warning next door already keeps its two concerns apart, so this was inconsistency rather than economy.

**3. The wiring layer.** I looked at consolidating and did not, for a reason worth stating rather than leaving as a judgement call: `_inject_plot` is where the branching lives, and it receives the serialised `maidr` string rather than the schema — so moving the check down means threading the trace types through a signature that has no other use for them. The duplication stays and both sides now carry a comment naming the other and saying to change them together. That is weaker than structural coupling and I would rather it were structural, but not at the cost of widening `_inject_plot`'s interface for a warning.

**4. The mismatched quote pair.** Not intentional — I wrote the character class twice without thinking about the pairing. It is a backreference now. Worth reporting the size of it: **1167 tokens before, 1166 after**, so exactly one match in the whole 1.4 MB bundle was a mismatched pair. Your reading that it biased safely was correct, and it is still better to mean what it says.

On the minor note about the ~8-line duplication between `maidr/core/maidr.py` and `maidr/plotly/plotly_maidr.py`: also not taken, and for the same reason as (3) — deduping it means a shared helper that both schema shapes can feed, which is a refactor of the sibling warning's wiring too. Flagging it as unaddressed rather than as consistent-so-fine.

## Verification

| Gate | Result |
|---|---|
| `uv run pytest` | **1045 passed** (was 1044) |
| `uv run ruff check` on the changed files | clean |

---

PR #362, opened 2026-08-13
fix(plotly): run the shared line-series pass once per render

### LINKED ISSUE #361: fix(plotly): the shared line-series pass runs twice, and a one-shot trace array loses every selector
# Bug Report

## Description

#350 fixed #316 by giving `PlotlyStepPlot` and `PlotlyMultiLinePlot` one predicate for both `data` and `selector`: `_line_series_with_positions` returns the drawn series **and** the positions they were drawn at, so series *i* always addresses the element series *i* is drawn as.

`render()` then asks for the two halves separately, and each half calls that pass for itself:

```python
# _extract_plot_data()
series, _ = self._line_series_with_positions(self._traces, self._scatter_positions)

# _get_selector()
_, drawn_positions = self._line_series_with_positions(self._traces, self._scatter_positions)
```

So the pass that exists to be shared is not shared — it runs twice per render.

The review on #350 raised this as a performance point and it is one. But repeating the pass also **assumes it can be repeated**, and it cannot. `as_list` materialises a trace array with `list(value)`, so a one-shot iterable is spent by the first walk and comes back empty on the second. The layer reports its series, then reports no selector at all.

That is the same silent no-highlight this pairing was written to prevent, reached from the other side.

## Steps to Reproduce

```python
from maidr.plotly.step import PlotlyStepPlot

def step(x, y, name):
    return {"type": "scatter", "mode": "lines", "x": x, "y": y,
            "line": {"shape": "hv"}, "name": name}

traces = [step(iter([0, 1]), iter([1, 2]), "a"),
          step(iter([0, 1]), iter([2, 1]), "b")]

schema = PlotlyStepPlot(traces, {}, scatter_positions=[0, 1]).schema
print(len(schema[MaidrKey.DATA]), MaidrKey.SELECTOR in schema)
```

## Actual Behavior

```
2 False
```

Two series, and `selector` absent entirely — no highlight for either line, with nothing in the output to say why.

## Expected Behavior

```
2 True
```

The pass runs once, both halves read the same answer, and the selectors are emitted.

## Additional Information

**A regression from #350, not pre-existing.** Before it, `_get_selector` read `self._scatter_positions` directly and never touched the trace arrays, so the same snippet emitted both selectors correctly. Confirmed by running it against `0faa7fc~1`:

```
pre-#350  -> data: 2 selectors: ['... nth-child(1) path.js-line', '... nth-child(2) path.js-line']
post-#350 -> data: 2 selectors: None
```

**How reachable it is.** `Figure.to_dict()` returns lists, numpy arrays and typed-array specs, never an iterator, so the export path does not hit this. A caller constructing a layer directly does, and `as_list` accepts an arbitrary iterable by design (`list(value)`). So this is narrow — but it is the kind of narrow that is invisible when it happens, and the fix that closes it is the same one the performance point asks for.

**The performance point, measured.** 50 series × 10,000 points, `PlotlyMultiLinePlot.render()`, median of five:

| | median render |
| --- | --- |
| today | 1.859 s |
| one pass | 0.923 s |

Exactly the 2× the double walk predicts — the pass allocates a dict per point, and it does it twice.

## Proposed Fix

Cache the pass on the layer, keyed by the identity of the two lists it was built from, so whichever of `_extract_plot_data` / `_get_selector` runs first computes it and the other reads it. The reviewer on #350 suggested the `PlotlyBoxPlot` idiom — stash the positions on `self` during `_extract_plot_data` — which is the established pattern here, but it depends on `render()`'s call order, and a cache keyed on its inputs does not.

jooyoungseo:
# Pull Request

## Description

#350 gave `PlotlyStepPlot` and `PlotlyMultiLinePlot` one predicate for both `data` and `selector`, so series *i* always addresses the element series *i* is drawn as (#316). `render()` then asks for the two halves separately, and each half was calling that pass for itself — so the pass that exists to be shared ran twice.

The review on #350 raised this as a performance point, and it is one. But repeating the pass also **assumes it can be repeated**, and it cannot: `as_list` materialises a trace array with `list(value)`, so a one-shot iterable is spent by the first walk and reads as empty on the second. The layer reports its series, then reports no selector at all — the same silent no-highlight the pairing was written to prevent, reached from the other side.

## Related Issues

Closes #361. Follows up the review on #350; #316 is the bug that pairing exists to fix.

## Changes Made

**`maidr/plotly/plotly_plot.py`** — `_drawn_line_series(traces, positions)` wraps `_line_series_with_positions` and caches its answer on the layer, keyed by the **identity** of the two lists it was built from. Whichever of `_extract_plot_data` / `_get_selector` runs first computes the pass; the other reads it.

The reviewer on #350 pointed at the `PlotlyBoxPlot` idiom — stash the result on `self` during `_extract_plot_data` for `_get_selector` to read afterwards — as the established pattern here, and it is. Keying on the inputs instead does the same work without depending on `render()` calling data before selector, and recomputes rather than answering for lists it was not built from.

**`maidr/plotly/step.py`, `maidr/plotly/multiline.py`** — both call sites go through it.

**`tests/plotly/test_plotly_line_series_pass.py`** — new, both classes parametrized:

| Case | Pins |
| --- | --- |
| `a_one_shot_trace_array_keeps_its_selectors` | the regression: data **and** selectors survive an iterator |
| `the_pass_runs_once_per_render` | one render, one call — a count, not a benchmark |
| `a_second_layer_is_not_served_the_first_one_s_answer` | the cache stays scoped to its layer |
| `different_traces_are_not_answered_from_the_cache` | passing other lists recomputes |
| `rendering_twice_gives_the_same_answer` | a cached pass does not make render 2 differ from render 1 |

## Verification

**The regression is from #350, not pre-existing.** Before it, `_get_selector` read `self._scatter_positions` directly and never touched the trace arrays. Run against `0faa7fc~1` and against `0faa7fc`:

```
pre-#350  -> data: 2  selectors: ['… nth-child(1) path.js-line', '… nth-child(2) path.js-line']
post-#350 -> data: 2  selectors: None
with this -> data: 2  selectors: 2                      (both classes)
```

**The new tests fail without the change** — 6 of 10 on the current `main` source, including both `a_one_shot_trace_array_keeps_its_selectors` cases and both `the_pass_runs_once_per_render` cases.

**Measured, not asserted.** 50 series × 10,000 points, `PlotlyMultiLinePlot.render()`, median of five runs:

| | median render |
| --- | --- |
| `main` | 1.859 s |
| this branch | 0.923 s |

Exactly the 2× a doubled walk predicts — the pass allocates a dict per point.

**Suite:** `uv run pytest -q` → **1067 passed**. `ruff check maidr/plotly tests/plotly` → clean.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Additional Notes

**How reachable the iterator case is, stated plainly.** `Figure.to_dict()` returns lists, numpy arrays and typed-array specs, never an iterator, so the export path does not hit it. A caller constructing a layer directly does, and `as_list` accepts an arbitrary iterable by design. So it is narrow — but it is the kind of narrow that is invisible when it happens, and closing it costs nothing beyond the fix the performance point already asked for.

**Not addressed here:** the reviewer's second point on #350, that `PlotlyStepPlot._get_selector` and `PlotlyMultiLinePlot._get_selector` now have identical bodies. They do, but their docstrings carry different information — the step one explains why a staircase resolves to the same `path.js-line` a plain line does, and why a lone step trace is still `nth-child`-scoped — and folding the bodies into a base method would have to drop or merge that. It is pre-existing duplication of a three-line call, and separable from this.

---

### DIFF
--- a/maidr/plotly/multiline.py
+++ b/maidr/plotly/multiline.py
@@ -73,8 +73,9 @@ class PlotlyMultiLinePlot(PlotlyPlot):
         """
         # Positions filtered by the same predicate that filters the data, so
         # series *i* always addresses the element series *i* is drawn as --
-        # see `_line_series_with_positions`.
-        _, drawn_positions = self._line_series_with_positions(
+        # see `_line_series_with_positions`. The pass is shared with
+        # `_extract_plot_data` rather than repeated: see `_drawn_line_series`.
+        _, drawn_positions = self._drawn_line_series(
             self._traces, self._scatter_positions
         )
         return self._scatter_line_selectors(self._traces, drawn_positions)
@@ -85,7 +86,5 @@ class PlotlyMultiLinePlot(PlotlyPlot):
         Each inner list contains ``{x, y}`` dicts for one line, with an
         optional ``z`` key set to the trace name.
         """
-        lines, _ = self._line_series_with_positions(
-            self._traces, self._scatter_positions
-        )
+        lines, _ = self._drawn_line_series(self._traces, self._scatter_positions)
         return lines
--- a/maidr/plotly/plotly_plot.py
+++ b/maidr/plotly/plotly_plot.py
@@ -253,7 +253,16 @@ class PlotlyPlot(ABC):
 
         Cached against the two lists by identity rather than unconditionally,
         so a caller passing different traces gets an answer for those traces
-        instead of the previous ones.
+        instead of the previous ones. Identity rather than equality because a
+        layer always hands over the same two objects it stored in ``__init__``,
+        so the check is free where it matters and never walks the points to
+        decide whether to walk the points.
+
+        One entry, deliberately — not a memoizer. It exists so that the two
+        halves of a single ``render()`` share one pass, and a layer only ever
+        asks about one pair. Alternating between two pairs would recompute
+        every time, correctly; if that ever becomes a real call pattern, this
+        wants replacing rather than widening.
 
         Parameters
         ----------
--- a/maidr/plotly/plotly_plot.py
+++ b/maidr/plotly/plotly_plot.py
@@ -51,6 +51,10 @@ class PlotlyPlot(ABC):
         self.row_index: int = 0
         self.col_index: int = 0
         self._schema: dict = {}
+        # Set by `_drawn_line_series`; see it for why the pass is cached.
+        self._line_series_cache: (
+            tuple[list[dict], list[int], tuple[list[list[dict]], list[int]]] | None
+        ) = None
 
     @staticmethod
     def _to_native(val: Any) -> Any:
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Observation 1 taken in `f484151`. You are right that "single entry" is a property a reader has to reverse-engineer from the code, and the docstring only said what the key is, not how many of them there are. It now says the cache exists so the two halves of one `render()` share one pass, that a layer only ever asks about one pair, and that alternating pairs would recompute — correctly — and would want this replaced rather than widened.

I also wrote down why it keys on identity rather than equality, since you had to work that out: a layer always hands over the same two objects it stored in `__init__`, so the check is free where it matters and never walks the points to decide whether to walk the points.

**2** — agreed, and that was the reason. The 2× is in the PR body because it is the review point on #350 that started this, but a timing assertion in CI would be flaky and would fail for reasons that have nothing to do with the pass running twice.

**3** — leaving the annotation. A `TypeAlias` reads better in isolation, but it appears once, and the type is the thing a reader needs to see at the point they are asking "what is in this cache". Naming it moves that answer somewhere else for no gain here.

**4** — leaving the docstrings, and worth saying why rather than just declining. Both bugs this file has produced were invisible to every test that looked at the output: #316's desync kept the audio, braille and text correct and moved only the highlight; this one kept the series and dropped the selectors. What makes them findable is a written record of *why* a rule exists — that `data` and `selector` are paired positionally, that `as_list` materialises with `list(value)`. The prose is doing the work a shorter docstring would leave to the next person's memory. I agree it is past NumPy convention; the file made that trade before this PR and I think it was the right one.

Thank you for verifying `as_list` and the `render()` call order against the source rather than the description — the order-independence is the part I would most want a second reader to check, since it is the difference between this fix and the `PlotlyBoxPlot` idiom the #350 review suggested.

---

PR #363, opened 2026-08-13
fix(deps): never resolve the CDN version on an event loop

### LINKED ISSUE #295: Consider falling back to the bundled version, not @latest, when resolution fails
Follow-up from #291.

When the version lookup fails (offline, blocked, or a malformed resolver response), the emitted URL degrades to `https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js`.

## The argument for changing it

`@latest` is the mutable dist-tag whose 7-day `Cache-Control` is the entire bug in #290. So in exactly the failure case, the fix stops applying — a browser can still serve a week-old build.

Falling back to `maidr@<bundled version>` instead would emit an immutable, permanently cacheable URL that matches the copy shipped in the wheel, so the cache-key fix would hold even when the lookup doesn't.

## The argument against

`@latest` was chosen so the failure path is byte-for-byte what users had before this change — nobody ends up worse off than they were, and there is no scenario where the fix makes things newly broken. The bundled fallback trades that for pinning users to a possibly older version whenever the network hiccups, which is quieter but staler.

I think the reviewer's argument is the stronger one, and I said so on #291 — but it changes behaviour on a path that is hard to observe, so it deserves its own change rather than being slipped in at the end of a long review.

## Where it lives

`maidr_js_cdn_url()` / `cdn_url()` in `maidr/util/dependencies.py` fall back via `get_cdn_version()` returning `LATEST_TAG`.

Note the interaction with `MAIDR_CDN_VERSION=bundled`, which already means "serve what is installed" and deliberately stays offline — this would make the failure path converge on the same behaviour, which is arguably a simplification.

Tests that pin the current behaviour: `test_offline_falls_back_to_latest_tag`, `test_malformed_resolver_response_falls_back`, `test_hostile_resolver_version_is_rejected` in `tests/core/test_cdn_version.py`.

### LINKED ISSUE #296: CDN version lookup blocks the event loop in Shiny apps
Follow-up from #291.

`maidr/widget/shiny.py`'s `render_maidr.render()` is `async def` but calls the synchronous `maidr.render(initial_value)`:

```python
async def render(self) -> Jsonifiable:
    ...
    maidr_rendered = maidr.render(initial_value)
```

Under the default `use_cdn="auto"`, the first such render performs a blocking `urlopen` **on the event loop**. Two consequences:

1. The loop stalls for up to `MAIDR_CDN_TIMEOUT` (3s default) — and that budget is approximate, since `urlopen`'s timeout applies per socket operation and does not reliably cover `getaddrinfo`. A broken resolver can exceed it.
2. `_resolve_latest_version()` holds `_resolution_lock` across the network call, so under concurrent sessions every first-render-in-process queues behind the same slow lookup instead of degrading independently.

## Mitigated for now, not fixed

`docs/index.qmd` now recommends `MAIDR_CDN_VERSION=bundled` for Shiny apps, which avoids the lookup entirely while still emitting an immutable URL. That is a workaround, not a fix — the default configuration still has the problem.

## Options

- **Skip resolution when a loop is running.** `asyncio.get_running_loop()` raises `RuntimeError` when there isn't one, so the check is cheap. Emit `bundled_cdn_url()` in that case — no request, still cache-safe. Downside: behaviour depends on execution context, which is its own kind of surprising.
- **Non-blocking lock acquire.** Waiters that lose the race fall back immediately instead of queueing. Removes the pile-up but makes the emitted URL differ between concurrent sessions.
- **Resolve in a background thread**, with the first call emitting `bundled_cdn_url()`. Keeps the loop free, but output then depends on timing.

All three trade determinism for responsiveness in some way, which is why none was folded into #291.

## Related

Interacts with the broader question in #295 — if the default became `bundled`, this problem disappears along with the lookup.

## Where it lives

`maidr/widget/shiny.py`, `_resolve_latest_version()` in `maidr/util/dependencies.py`.

jooyoungseo:
# Pull Request

## Description

`maidr.render()` is synchronous and Shiny calls it from `render_maidr.render()`, which is `async` — so under the default `use_cdn="auto"` the first figure in an app performed a blocking `urlopen` **on the event loop**, and every concurrent session queued behind it while it ran.

`get_cdn_version()` now answers from the bundled version when a lookup would have to be made from a thread running a loop.

## Related Issues

Closes #296. Follows up #291; interacts with #295.

### `maidr/util/dependencies.py`

- **`_resolution_would_block()`** — true only when a lookup would have to be *performed* **and** this thread is running a loop. `asyncio.get_running_loop()` raises `RuntimeError` when there is none, so the check is cheap.
- **`get_cdn_version()`** — a new step 3 between the pin and the lookup.
- **`_offline_version()`** — the pin → cached-lookup → bundled → `latest` order, written once. `bundled_cdn_url()` spelled the same thing out separately; its own comment records what happened when the two disagreed (*"one page loaded two different builds of maidr.js"* — an iframe on the pinned version, its host on the bundled one). Both read it now.

### Which of the issue's three options this is, and why

The issue offered three and folded none in, because "all three trade determinism for responsiveness in some way."

This is **option 1**, and the reason to prefer it is that its context-dependence is bounded in a way the other two are not:

| | never blocks the loop | output independent of timing | output independent of context |
| --- | --- | --- | --- |
| skip when a loop is running | ✅ | ✅ | after the first resolution anywhere |
| non-blocking lock acquire | partly — the winner still blocks | ❌ | ❌ |
| resolve in a background thread | ✅ | ❌ | ❌ |

A completed lookup is *not* blocking, so it is used on the loop like anywhere else. That means the "async is different" window is one process-wide first resolution, not forever — and calling `get_cdn_version()` once from synchronous code at start-up closes it deliberately. Options 2 and 3 both make the emitted URL depend on which session won a race, which is a worse thing to debug than "this app serves its bundled version."

### The trade, stated plainly

An async app no longer picks up a release newer than its wheel automatically. That is the trade `docs/index.qmd` **already recommended making by hand** — `MAIDR_CDN_VERSION=bundled` was the advice for Shiny apps — so this makes the default do it in exactly the context the advice was written for. The docs now say the lookup cannot happen there, and give the two ways to get the newer release back.

One consequence worth knowing rather than discovering: `warn_if_bundle_is_stale()` reads only an already-completed lookup (`bundle_status(resolve=False)`), so a process that never resolves never hears it. Said in the docs and the docstring.

### One correction to the issue

> `_resolve_latest_version()` holds `_resolution_lock` across the network call

It holds **`_fetch_lock`**; `_resolution_lock` is explicitly released before the call, with a comment saying why ("the offline paths read that lock, so holding it here would make a stalled lookup block a render that promised to make no request at all"). The consequence the issue describes — concurrent first-renders queueing — is real and is `_fetch_lock`'s.

**Not fixed, and not claimed to be:** synchronous threads still queue on `_fetch_lock` while the first of them resolves. It is the event loop specifically that must not be made to wait.

## Verification

Measured against a stubbed 3-second resolver:

```
bundled wheel version: 4.1.0
sync                       -> '9.9.9' in 3.00s
async                      -> '4.1.0' in 0.00s
async, second call         -> '4.1.0' in 0.00s
async after a sync resolve -> '9.9.9' in 0.00s
async with an explicit pin -> '1.2.3'
```

`tests/core/test_cdn_event_loop.py` asserts these on the **request count**, not on elapsed time — a stalled loop is what the user experiences, but the round trip is the cause, and it is the thing a test can state without measuring a shared CI runner.

| Case | Pins |
| --- | --- |
| `a_render_on_an_event_loop_makes_no_request` | the bug |
| `a_synchronous_render_still_resolves` | the control |
| `the_version_emitted_on_a_loop_is_concrete` | declining to resolve must not reintroduce `@latest` |
| `an_explicit_pin_still_wins_on_a_loop` | a pin is the caller's own answer |
| `a_completed_lookup_is_used_on_a_loop` | the window is one resolution, not forever |
| `a_cached_failure_is_not_retried_on_a_loop` | the failure path is unchanged |
| `two_concurrent_renders_on_one_loop_both_return` | no pile-up |
| `bundled_cdn_url_still_prefers_a_pin_then_a_lookup` | the shared fallback order after the refactor |

**3 of 8 fail on `main`** without the change.

**Suite:** `uv run pytest -q` → **1075 passed**. `ruff check maidr/util tests/core` → clean.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

## Additional Notes

`asyncio.get_running_loop()` is per-thread, which is what makes this correct rather than merely convenient: a render dispatched to a worker thread by `asyncio.to_thread` has no running loop on that thread and resolves normally, which is right — it is not the loop's thread and cannot stall it.

This does not close #295 ("consider falling back to the bundled version, not `@latest`, when resolution fails"). That is a different question — what to do when a lookup *was* made and failed — and this change deliberately leaves that path exactly as it was.

---

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -297,7 +297,14 @@ One exception is worth knowing about: **Altair** charts always load `maidr.js` f
 ::: {.callout-note}
 ## Shiny and other async servers
 
-`render_maidr` calls into the synchronous render path, so the one-time lookup happens on the event loop and blocks it for up to `MAIDR_CDN_TIMEOUT`. Set `MAIDR_CDN_VERSION=bundled` at startup for Shiny apps to avoid that entirely.
+**No lookup happens on an event loop.** `render_maidr` calls into the synchronous render path, so the one-time lookup used to land on the loop and block it — every concurrent session queued behind one slow request. py-maidr now answers from the bundled version instead whenever a lookup would have to be made from a thread running a loop, which needs no request and is still an immutable URL.
+
+You give up picking up a release newer than your wheel, automatically, in an async app. Two ways to get it back:
+
+- `maidr.set_cdn_version("3.74.0")` at startup, if you know which version you want.
+- Call `maidr.get_cdn_version()` once from synchronous code at startup. That resolves and caches, and every async render then uses the resolved version — the request is paid for once, by your startup, not by a render.
+
+The same caching means this is not "async gets a different answer forever": after any resolution anywhere in the process, the loop and the main thread agree. It also means `warn_if_bundle_is_stale()` stays quiet in an app that never resolves, since it reads only an already-completed lookup.
 :::
 
 `bundle_status()` is the one deliberate exception: it resolves even when pinned, because "how does my bundle compare to what's published?" has no answer otherwise. Pass `resolve=False` if you need it silent.
--- a/maidr/util/dependencies.py
+++ b/maidr/util/dependencies.py
@@ -26,6 +26,7 @@ grows large enough to matter.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import logging
 import math
@@ -357,9 +358,11 @@ def get_cdn_version() -> str:
 
     1. An explicit pin from :func:`set_cdn_version`.
     2. The ``MAIDR_CDN_VERSION`` environment variable.
-    3. The ``latest`` dist-tag resolved over the network (once per
+    3. The version this wheel ships, when resolving would block an event
+       loop — see :func:`_resolution_would_block`, and the note below.
+    4. The ``latest`` dist-tag resolved over the network (once per
        process, then cached).
-    4. The literal string ``"latest"`` when the lookup fails, which
+    5. The literal string ``"latest"`` when the lookup fails, which
        reproduces the library's historical behaviour.
 
     Returns
@@ -369,7 +372,7 @@ def get_cdn_version() -> str:
 
     Notes
     -----
-    Step 4 is what makes this safe to call from a render path: an
+    Step 5 is what makes this safe to call from a render path: an
     unreachable, blocked, or malformed resolver degrades to ``"latest"``
     rather than raising.  That guarantee rests on
     :func:`_fetch_latest_version` honouring its own "never raises"
@@ -384,10 +387,38 @@ def get_cdn_version() -> str:
     A ``KeyboardInterrupt`` is treated differently: it is not cached, so
     interrupting one render does not leave the rest of the process
     emitting ``@latest``.
+
+    Step 3 is the one that depends on where it is called from, and it is
+    a deliberate trade.  ``maidr.render()`` is synchronous and Shiny calls
+    it from ``render_maidr.render()``, which is ``async`` — so under the
+    default the first figure in an app performed a blocking ``urlopen`` on
+    the event loop, stalling every concurrent session behind it (#296).
+    The bundled version is a real published one, so the URL resolves and
+    is immutable; what is given up is picking up a release newer than the
+    wheel, automatically, in an async process.
+
+    That is the trade the docs already recommended making by hand:
+    ``MAIDR_CDN_VERSION=bundled`` was the advice for Shiny apps, and this
+    makes the default do it in exactly the context the advice was for.
+    Two things follow that are worth knowing rather than discovering:
+
+    * :func:`warn_if_bundle_is_stale` reads only an already-completed
+      lookup, so a process that never resolves never hears it.
+    * Calling this once from synchronous code at start-up caches the
+      answer, after which every async render uses the resolved version --
+      which is how an app that wants the newer release gets it without
+      any render paying for the request.
+
+    Synchronous callers are unaffected, including threads: they still
+    resolve once per process and queue on ``_fetch_lock`` while the first
+    of them does.  That queueing is not fixed here; it is the event loop
+    specifically that must not be made to wait.
     """
     pin = _version_pin()
     if pin is not None:
         return pin
+    if _resolution_would_block():
+        return _offline_version()
     return _resolve_latest_version() or LATEST_TAG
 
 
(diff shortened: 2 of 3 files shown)


jooyoungseo:
The test observation was right, and it mattered more than "worth knowing" — the case was named for a property it did not have.

`asyncio.sleep(0, dependencies.get_cdn_version())` evaluates its argument when the coroutine is *created*, so all four calls ran to completion before `gather` scheduled anything. Confirmed rather than taken on trust:

```
eager evaluation order: [0, 1, 2] -- all before gather ran
```

Fixed in `ea23ae8`: each coroutine now `await asyncio.sleep(0)` and *then* calls, so the scheduler interleaves them — which is the shape a Shiny app's sessions actually arrive in, and the shape the pile-up needed. The comment records why it is written that way rather than the shorter form. Renamed to `test_interleaved_renders_on_one_loop_all_return`, since "concurrent" was the word doing the false advertising.

Re-verified against `main`'s `dependencies.py`: **3 of 8 fail**, including the reworked case.

**Item 1 is filed as #364** rather than folded in. You are right that it is pre-existing and right that this PR makes it easier to reach — the tail used to be on `bundled_cdn_url()` only, and it is now also where an event-loop render lands. The issue records something the review pass surfaced and I want written down: there are two roads to `@latest` and they deserve different treatment. A failed lookup is a routine operating condition and should stay quiet; an unreadable bundle is a broken install, and the user is the only one who can fix it. Warning on both would train people to ignore the message.

**Item 2** — agreed, and worth being precise: the check is asyncio-specific, and Trio is the case. Trio does not run an asyncio loop, so `asyncio.get_running_loop()` raises there and a Trio-driven render resolves normally and blocks its scheduler. Not addressed here because Shiny, Starlette and FastAPI are all asyncio and I would rather not add a detection path with no test that exercises a real one behind it. If a Trio user hits it, `trio.lowlevel.current_task()` is the equivalent probe and it should be added then, with a Trio dependency in the test extras to prove it.

**Item 3** — leaving as is, for the reason you identify: it is the file's established convention, not a new deviation.

On not being able to run the suite: the numbers are `1075 passed` and `3 of 8 fail on main`, and CI has since gone green on all four Python versions plus lint and CodeQL.

---

Issue #298, opened 2026-07-29
check-bundle-freshness fails open indefinitely when the resolver is unreachable


jooyoungseo:
Raised during review of #291, which added `.github/workflows/check-bundle-freshness.yml`.

## The gap

The job asks `maidr.bundle_status()` for the published version and compares it to the bundled one. When resolution fails it deliberately passes:

```python
if status.published is None:
    # A resolver hiccup is not a drift signal.  Say so and pass;
    # failing here would train maintainers to ignore this job.
    print("::warning::Could not resolve the published version; skipping.")
    sys.exit(0)
```

That is right for a one-off hiccup. It is wrong for a *persistent* failure: if the resolver stays unreachable — npm outage, a change to the API shape, an expired hostname, a network policy change on the runner — the weekly job stays green forever while checking nothing. The bundle can then drift arbitrarily far and the one mechanism built to catch it before release reports success every Monday.

A green check that verifies nothing is worse than a red one, because it is indistinguishable from a real pass.

## Options

1. **Escalate after N consecutive skips.** Needs state across runs. Cheapest carrier is a workflow-run artifact or a tiny committed file; a repository variable via the API also works but needs a write token.
2. **Fail on skip, but only for scheduled runs.** A `workflow_dispatch` run stays lenient. Simple and stateless, but reintroduces exactly the noise the current comment is avoiding.
3. **Distinguish resolver-shape failures from network failures.** A `403`/`404`/JSON-shape error means our resolver code is wrong and should fail loudly; a timeout or DNS error means the network is down and should skip. This catches the most likely long-lived cause (an API change) without any cross-run state, and is probably the best value for the effort.
4. **Add a canary assertion.** Fail if *neither* endpoint answers, but pass if one does — turning the current all-or-nothing skip into a narrower one.

Option 3 looks like the right first step, optionally with option 1 later if a stateful counter proves worth the machinery.

## Note

Whatever we pick, `bundle_status()` currently collapses every resolution failure into `published=None`, so the workflow cannot yet tell these cases apart. Option 3 needs a way to surface *why* resolution failed — probably a separate accessor rather than widening `BundleStatus`, so the offline render path keeps its current shape.

## Related

- #291 — the PR that added the workflow
- #292 — whether `STALE_MINOR_GAP = 5` matches upstream's release cadence

PR #366, opened 2026-08-13
feat(deps): give the resolver's two silent failures a voice

### LINKED ISSUE #364: A broken bundle emits the mutable `@latest` URL with nothing said
# Bug Report

## Description

`_offline_version()` ends in a silent fallback:

```python
bundled = maidr_js_version()
if _is_valid_version(bundled) and bundled != _UNKNOWN_VERSION:
    return bundled

return LATEST_TAG
```

When `maidr/static/VERSION` is missing or unreadable, `maidr_js_version()` returns `_UNKNOWN_VERSION`, and every offline caller emits `maidr@latest/dist/maidr.js` instead.

That is the mutable dist-tag this whole module exists to stop emitting. jsDelivr serves it with a seven-day `max-age`, so a browser that loaded it before will replay a week-old bundle — the exact stale-cache behaviour #291 was filed to end, arrived at through a broken install rather than through a failed lookup.

Nothing says so. There is no warning and no log line; the URL is well-formed and the page works, so the only symptom is a user occasionally getting an old `maidr.js` for reasons nothing in the output explains.

## Steps to Reproduce

```python
from maidr.util import dependencies

dependencies.maidr_js_version.cache_clear()
dependencies.maidr_js_version = lambda: dependencies._UNKNOWN_VERSION  # a broken wheel
dependencies.reset_cdn_version_cache()

print(dependencies.bundled_cdn_url(dependencies.MAIDR_JS_FILENAME))
```

## Actual Behavior

```
https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js
```

No warning, no log record.

## Expected Behavior

The same URL — there is genuinely no better answer available offline — but **said out loud**, once per process, the way the other degradations in this module are. `_warn_once` already exists for exactly this shape of thing.

## Additional Information

**Pre-existing, and not a regression.** `bundled_cdn_url()` has had this silent tail since it was written; #363 only moved it into `_offline_version()` alongside the identical fallback in `get_cdn_version()`. Raised in review of #363 and deliberately left out of it, since fixing it there would have widened a scoped fix into a behaviour change on paths the PR did not otherwise touch.

**#363 makes it easier to reach.** Before that change the tail was only on `bundled_cdn_url()` — notebook tags and the offline path. It is now also where `get_cdn_version()` lands when a render happens on an event loop, so a Shiny app with a broken install reaches it on its ordinary render path.

**Which fallback is which matters here.** There are two ways to end up at `@latest` and they deserve different treatment:

| | cause | what it means |
| --- | --- | --- |
| lookup failed | offline, blocked, malformed response | expected, routine, already documented as the safe degradation |
| bundle unreadable | broken install, packaging fault | not routine — something is wrong with the installation |

The first is a normal operating condition and should stay quiet. The second is a fault, and the user is the only one who can fix it. Warning on both would train people to ignore the message; that is the argument for scoping the warning to the second.

Related: #291, #295, #296, #363.

jooyoungseo:
# Pull Request

## Description

Two silent failures on the CDN version path, one facing a user and one facing a monitor. They are separate commits and read best that way; they are together because they are the same module, the same theme, and about a hundred lines between them.

**Read as two commits, not one diff** — `426bfe4` is the first, `d814369` the second.

## Related Issues

Closes #364 and #298. #364 was raised in review of #363; #298 in review of #291.

---

## 1 — a broken bundle emits `@latest` in silence (#364)

`_offline_version()` ended at `LATEST_TAG` with nothing said when `maidr/static/VERSION` would not read. That is the mutable dist-tag this module exists to stop emitting: jsDelivr serves it with a seven-day `max-age`, so a browser that loaded it before replays a week-old `maidr.js`. The URL is well-formed and the page works, so the only symptom was someone occasionally being served an old bundle for reasons nothing in the output explained.

**Only the broken-install road warns:**

| | cause | treatment |
| --- | --- | --- |
| lookup failed | offline, blocked, malformed response | routine, documented safe degradation — quiet |
| bundle unreadable | broken install, packaging fault | a fault only the user can fix — speaks |

Warning on both would train people to ignore the message. The message names the *consequence* (the seven-day cache lifetime), because that is the part a reader cannot infer from a URL that looks perfectly valid. `_warn_once` keys on the unusable value, so an absent `VERSION` and one holding `not-a-version` are told apart, and a second, differently broken one stays audible.

**Nothing changes about what is emitted.** The URL was `@latest` and still is. The difference is that it is said out loud.

---

## 2 — the freshness job goes green forever when the resolver dies (#298)

`check-bundle-freshness` passes when the published version cannot be resolved, because a hiccup is not a drift signal. Right for a hiccup, wrong for a persistent failure: the job then stays green forever while checking nothing, and **a green check that verifies nothing is worse than a red one, because it is indistinguishable from a real pass.**

The issue offered four options and recommended the third — distinguish resolver-shape failures from network failures — noting the blocker: `_fetch_latest_version` collapses every failure into `None`, so the workflow cannot tell them apart.

`ResolverOutcome` is that distinction, as **a separate accessor rather than a wider `BundleStatus`** — exactly as the issue suggested, so the render path keeps its shape and its contract that a lookup failure cannot break a chart:

| bucket | what happened | who can fix it |
| --- | --- | --- |
| `unreachable` | timeout, refused connection, unresolvable name, blocked socket — nothing was read | nobody reading the job |
| `answered_badly` | HTTP error status, body that would not decode or parse, payload without the key, value that is not a version | us — and it will not fix itself |

An endpoint that changed shape looks exactly like `answered_badly` and looks like nothing else, which is why it is worth failing on.

**Two classification choices worth review:**

- **An unrecognised failure counts as `unreachable`**, not `answered_badly`. `pytest-socket` raises `SocketBlockedError`, which is not an `OSError`; calling an unknown failure "this code is wrong" would redden a scheduled check over a sandbox setting nobody can act on. The safe direction is the quiet one.
- **An endpoint the budget ran out before reaching lands in neither bucket.** It was never asked. Recording it as unreachable would report a spent budget as a network fault.

`reset_cdn_version_cache()` clears the outcome with the lookup it describes, so a monitor cannot read a stale verdict as the current one.

## Verification

**Every classification path, exercised against a stubbed `urlopen`:**

```
both endpoints time out (network down):        resolved=None unreachable=2 answered_badly=0
both endpoints answer 404 (path moved):        resolved=None unreachable=0 answered_badly=2
both answer with JSON missing the key:         resolved=None unreachable=0 answered_badly=2
both answer with unparseable bytes:            resolved=None unreachable=0 answered_badly=2
first is down, second answers correctly:       resolved='9.9.9' unreachable=1 answered_badly=0
first answers correctly (second never asked):  resolved='4.2.0' unreachable=0 answered_badly=0
```

**The workflow's own decision branch**, run outside CI so the YAML is exercised rather than read:

```
network down             -> SKIP   (both endpoints named)
endpoints 404            -> FAIL   (both endpoints named)
payload shape changed    -> FAIL   (both endpoints named)
healthy                  -> CHECK  (4.2.0)
```

| Gate | Result |
| --- | --- |
| `uv run pytest -q` | **1093 passed** |
| the 6 `#364` cases with that fix reverted | **3 fail** |
| `ruff check maidr/util maidr/__init__.py tests/core` | clean |
| `actionlint` on the changed workflow | clean |

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

## Additional Notes

**Why one PR and not two.** They were meant to be two, and the second landed on the branch before the first had merged. Splitting them now needs a force-push, which is not available here. The commits are separable and the review reads cleanly as two; I would rather say that than quietly present them as one change.

**Not addressed, from #298's option list.** Option 1 — escalate after N consecutive skips — still stands as the answer to a *network* that stays down, which this deliberately keeps quiet about. It needs cross-run state and is worth doing only if a persistent outage actually happens; the issue said as much, and this is the "best value for the effort" step it recommended first.

One fixture detail worth its comment rather than a puzzled reader later: `test_broken_bundle_version.py`'s teardown does **not** call `maidr_js_version.cache_clear()`. `monkeypatch` tears down after the fixture, so at that point the name still points at the test's stand-in — a plain function, with no `cache_clear` on it.


jooyoungseo:
Both nits taken in `dbbf7fd`, and the third is answered by reproducing it rather than reasoning about it.

**Nit 2 was the important one.** You are right that `is unreadable ('0.0.0')` reads as though the file contains those characters, which is the one thing it does not — and on a PR whose whole argument is "name the consequence clearly", that is not a nit I get to wave through. The two faults now read differently:

```
maidr: the bundled maidr.js version cannot be read (the VERSION file is missing or empty), so CDN URLs ...
maidr: the bundled maidr.js version is not a version ('not-a-version'), so CDN URLs ...
```

`maidr_js_version` returns the sentinel for missing, empty and unreadable alike, so that is as far as the distinction can honestly go — said in the comment, so nobody later reads the message as promising more resolution than it has.

**Nit 1** taken too: `bundled[:_MAX_WARNED_KEY_LEN]`, matching `_normalise_version_pin`. You are right that the risk is low because the value is local — but nothing bounds the length of a garbled file, and consistency here costs one slice.

**Nit 3 — reproduced, and two messages is the right answer.** With `use_cdn=False` and a wholly unreadable bundle, both do fire:

```
maidr: use_cdn=False was requested but the bundled maidr.js/css could not be read, so the notebook
       will load them from the CDN. Reinstall py-maidr to repair the bundle for offline use.
maidr: the bundled maidr.js version cannot be read (the VERSION file is missing or empty), so CDN
       URLs fall back to the mutable maidr@latest tag. jsDelivr serves that with a seven-day cache
       lifetime, so a browser may replay an old bundle. Reinstall py-maidr, or pin a version with
       maidr.set_cdn_version().
```

One root cause, but two consequences, and neither subsumes the other. The first says a promise was broken — you asked for no network and are getting the CDN. The second says the URL you are getting is the *mutable* one and may hand you a stale bundle. A reader who acts on only the first still has the second.

Combining them would also cost the case where they are not compound: a `VERSION` that is broken while `maidr.js` is fine reaches only the second, and folding it into the first would lose it there entirely.

**One thing that review could not have seen**: it ran against `426bfe4` only. `d814369` landed after it and adds the #298 half — `ResolverOutcome`, and the freshness workflow failing on a resolver that answers badly rather than skipping forever. The description covers both, and CI has gone green on all four Python versions since.

---

PR #367, opened 2026-08-13
fix(deps): stop emitting @latest on the failure path, and warn on a mistyped pin

### LINKED ISSUE #292: Re-check STALE_MINOR_GAP against upstream maidr.js release cadence
Follow-up from #291.

`warn_if_bundle_is_stale()` fires when the bundled `maidr.js` is behind the published release by a major version, or by `STALE_MINOR_GAP` (currently **5**) minor versions.

## Why 5 may be too tight

The threshold was picked to sit under the gap in the original report (#290 cited 3.66.1 bundled vs 3.74.0 published — 8 minors). But a reviewer pointed out that this cuts the other way: if upstream can ship 8 minors inside a single py-maidr release cycle, then 5 will be crossed routinely, and users will see the warning on plots that render perfectly well. A warning that fires in normal operation is one people learn to ignore.

## What to decide

1. Is 5 right, given upstream's actual release cadence? Someone with that history can judge it; I picked the number without it.
2. Or should staleness be derived from **bundle age** rather than version distance? "Your bundle is four months old" is more meaningful than "five minors behind" when cadence is uneven. That needs publish dates from the registry, which is a slightly larger change.

## Where it lives

- `STALE_MINOR_GAP` in `maidr/util/dependencies.py`
- Threshold behaviour is covered by the parametrized table in `tests/core/test_bundle_freshness.py` (`test_bundle_status_comparison`), including cases either side of the boundary — changing the constant means updating those expectations.
- `.github/workflows/check-bundle-freshness.yml` reads the same constant, so CI and the runtime warning stay in step automatically.

### LINKED ISSUE #294: set_cdn_version() raise on invalid version (warned, raise pending) + get_cdn_version() naming
Follow-up from #291.

`maidr.set_cdn_version("3.74")` (a typo — missing the patch component) currently logs a warning once and is otherwise ignored: the next URL build resolves normally, exactly as if no pin had been set.

## The argument for raising

That leniency is right for `MAIDR_CDN_VERSION`. An environment variable is ambient configuration that may be set by something outside the user's control, and crashing on it would be hostile.

But `set_cdn_version()` is an explicit function call with a bad argument, which is the textbook case for `ValueError`. As it stands the caller gets no return value, no exception, and a log line they may never see — so a typo silently does nothing.

## What to weigh

- It is a public-API behaviour change, so it wants a deliberate decision rather than being folded into a bug-fix PR — which is why it was left out of #291.
- The asymmetry would need documenting: strict for the function, lenient for the env var.
- `tests/core/test_cdn_version.py::test_invalid_pin_never_reaches_the_url` passes hostile values (`../../../evil`, `3.74.0 && rm -rf /`, full URLs) through `set_cdn_version()` and asserts they never reach the URL. If the function starts raising, those cases move to `pytest.raises` and the env-var path needs its own equivalent coverage.

## Where it lives

`set_cdn_version()` and `_normalise_version_pin()` in `maidr/util/dependencies.py`.

### LINKED ISSUE #295: Consider falling back to the bundled version, not @latest, when resolution fails

jooyoungseo:
# Pull Request

Three follow-ups from #291. Separate commits; read them that way.

## Related Issues

Closes #295. Closes #292. Closes #294 in part — see the note at the end, which I have **not** decided alone.

---

## 1 — a failed lookup emitted the mutable tag (#295)

`@latest` carries a seven-day `Cache-Control`, which is the whole of #290. Degrading to it when the lookup failed meant **the fix stopped applying in exactly the case it was written to survive**: an offline browser could still be handed a week-old build.

It now falls back to the bundled version — a real published release, immutable, and the same copy py-maidr would serve with `use_cdn=False`. The cost is that a network hiccup pins that page to the version you installed rather than the current one: quieter, but staler.

That trade was argued the other way when the fallback was written, on the grounds that `@latest` is byte-for-byte what users had before version resolution existed, so nobody ends up worse off. **Three other paths have since moved to the bundled answer** — notebook tags and `bundled_cdn_url()` via `_offline_version()` (#366), and a render on an event loop (#363) — and being the last one still emitting the mutable tag is not a place worth defending.

**An explicit `latest` pin is untouched.** Asking for the mutable tag is not a failure, and `MAIDR_CDN_VERSION=latest` remains the documented way to keep the CDN without the lookup. There is a case asserting exactly that, because it is the thing this change could plausibly have broken.

### On the ten tests that pinned the old behaviour

Each is updated in place, not deleted. Several had the fallback only *incidentally* — `test_timeout_is_a_total_budget_across_endpoints`, `test_budget_exhaustion_skips_remaining_endpoints`, `test_non_oserror_from_urlopen_does_not_escape` are about the budget, endpoint skipping and a blocked socket respectively, and merely happened to assert the resulting version. Those keep their subject; only the version assertion moves.

`test_offline_falls_back_to_latest_tag` is renamed, since its name was the claim.

---

## 2 — `set_cdn_version("3.74")` did nothing, quietly (#294)

A missing patch component logged a line and was otherwise ignored. The caller got no return value, no exception, and a **log** line they may never see, because logging is not configured by default.

It now raises a `FutureWarning` at the call, pointing at the caller's own line via `stacklevel=2`, and says a future major will raise `ValueError` instead. **Not raising today** — that would break a script quietly mistyping its pin and rendering fine, which is exactly the caller this is trying to reach. Same deprecation shape as the placeholder-CSS accessors in #334, and `FutureWarning` for the same reason: `DeprecationWarning` is silenced outside `__main__`, so a call from inside a Shiny app would never see it.

**The asymmetry is documented**, as the issue asked. `MAIDR_CDN_VERSION` stays lenient — ambient configuration a CI image may have set without the caller, where crashing would be hostile.

---

## 3 — `STALE_MINOR_GAP = 5`, measured (#292)

Picked without the release histories. It survives them, but the deciding number is not the one the issue reaches for first.

Upstream ships a minor every 5.1 days (median, 33 minors over 197 days). On its own that settles nothing. What settles it is **how many minors accumulate between the py-maidr releases that refresh the bundle**, over seventeen cycles:

```
median 1   mean 1.7   min 0   max 7
cycles reaching 3+:  6/17  (35%)
cycles reaching 5+:  1/17  (6%)
cycles reaching 8+:  0/17  (0%)
```

The one cycle reaching 5 is the 68-day gap with nothing released — exactly what the warning is for. So the worry (*"upstream ships 8 minors per cycle, so 5 fires routinely"*) does not happen. **3 would fire on a third of all cycles**, which is how a warning becomes noise.

On the issue's alternative — age instead of version distance — the docstring records why distance stays: both are proxies for *can this bundle draw my chart*, and since #358 `warn_if_bundle_cannot_render()` answers that directly, from the installed file, with no network.

---

## Verification

```
uv run pytest -q                                    1120 passed
tests/core/test_pin_validation.py against main        15 failed
test_a_failed_lookup_never_emits_the_mutable_tag      fails on main
uv run pytest -W error::FutureWarning               1120 passed  (no strays)
ruff check maidr/util tests/core                    All checks passed
```

The `-W error::FutureWarning` run matters: item 2 adds a warning to a call several existing tests make deliberately, and a run that merely *tolerates* it would not show whether it escapes somewhere it should not.

## Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

## Additional Notes

**#295 was closed by accident, by me.** #363's description contained the sentence *"This does not close #295"*, and GitHub's linked-issue parser read `close #295` inside it. A line written to say the issue was out of scope is what marked it done. Reopened, and noted there.

**#294's second item is left for you.** That issue's comment proposes renaming `get_cdn_version()` → `resolve_cdn_version()` with a deprecated alias, because `get_` reads like a cheap accessor over network I/O. It offers that and "document the I/O more loudly" as equal options, with no stated preference — and **#363 changed the premise**: `get_cdn_version()` no longer resolves on a thread running an event loop, so the name misleads less than it did, and `resolve_cdn_version` would newly mislead in the other direction for exactly the async callers the rename was meant to protect. A 50/50 call about a public symbol is yours rather than mine.

**Why `fix` and not `fix!`.** Item 1 changes what a URL says on a path users cannot easily observe, which is the reason #295 asked for it to be its own change rather than slipped into a review. It is not breaking: both the old and new URLs load a working `maidr.js`, and an explicit `latest` pin still gets `@latest`.

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -319,6 +319,14 @@ maidr.set_cdn_version("latest")   # skip resolution, emit the mutable @latest UR
 maidr.set_cdn_version(None)       # back to resolving automatically
 ```
 
+::: {.callout-note}
+## A mistyped pin warns; a mistyped environment variable does not
+
+`maidr.set_cdn_version("3.74")` — a missing patch component — raises a `FutureWarning` and is then ignored, so URLs resolve as if no pin had been set. A future major release will raise `ValueError` there instead.
+
+`MAIDR_CDN_VERSION=3.74` stays quiet apart from a log line. The asymmetry is deliberate: an explicit function call with a bad argument is the caller's mistake and worth interrupting them over, while an environment variable is ambient configuration that a CI image or a container base may have set without them, and crashing on it would be hostile.
+:::
+
 ::: {.callout-warning}
 ## These settings are process-wide, not per-session
 
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -284,7 +284,9 @@ This matters because the CDN serves the mutable `maidr@latest` URL with a seven-
 :::: {.callout-note}
 ## This performs one network request
 
-Resolving the version makes a single outbound HTTPS request, the first time a CDN URL is built for a *rendered plot*. It is cached for the rest of the process and falls back to the `@latest` URL if it fails — so a blocked or slow network costs you a one-time delay, never an error.
+Resolving the version makes a single outbound HTTPS request, the first time a CDN URL is built for a *rendered plot*. It is cached for the rest of the process and falls back to **the version bundled in your wheel** if it fails — so a blocked or slow network costs you a one-time delay, never an error, and never the mutable `@latest` URL whose seven-day cache lifetime this resolution exists to avoid.
+
+The trade is that a network hiccup pins that page to the version you installed rather than the current one. That is the quieter failure: the bundled version is a real published release, and it is the same copy py-maidr would serve with `use_cdn=False`.
 
 `import maidr` makes no request, even in a notebook: `init_notebook()` pins its tags to the version bundled in the wheel, which needs no lookup and is still immutable.
 
--- a/maidr/util/dependencies.py
+++ b/maidr/util/dependencies.py
@@ -390,6 +390,26 @@ def set_cdn_version(version: str | None) -> None:
     Takes precedence over the ``MAIDR_CDN_VERSION`` environment variable.
     A value that is neither a recognised tag nor a valid semver is
     ignored (with a warning) in favour of the normal resolution path.
+
+    Warns
+    -----
+    FutureWarning
+        When ``version`` is unusable.  Today the pin is ignored and the
+        next URL resolves normally; a future major release will raise
+        :class:`ValueError` instead.
+
+        The leniency is right for ``MAIDR_CDN_VERSION`` -- ambient
+        configuration may be set by something outside the caller's
+        control, and crashing on it would be hostile.  It is wrong here:
+        this is an explicit call with a bad argument, which is the
+        textbook case for ``ValueError``, and the caller currently gets no
+        return value, no exception, and a *log* line they may never see.
+        So a typo does nothing, visibly (#294).
+
+        Raising outright would break a script that has been quietly
+        mistyping its pin and rendering fine, so it goes through a
+        deprecation the way :func:`bundled_css_path` did.  The warning is
+        the behaviour change; the raise is the next major's.
     """
     global _cdn_version_override
     # A blank string is treated as ``None`` rather than as a malformed
@@ -400,7 +420,25 @@ def set_cdn_version(version: str | None) -> None:
         _cdn_version_override = None
         reset_cdn_version_cache()
         return
-    _cdn_version_override = str(version).strip()
+
+    candidate = str(version).strip()
+    # Checked here as well as on every URL build, because the two
+    # answer different questions. `_normalise_version_pin` asks "can I
+    # use this?" every time it needs a URL, and logs. This asks "did the
+    # caller just make a mistake?", once, at the point they made it --
+    # which is the only moment a stack trace points anywhere useful.
+    if _normalise_version_pin(candidate) is None:
+        warnings.warn(
+            f"maidr.set_cdn_version({version!r}) was given something that is "
+            f"neither a semver such as '3.74.0' nor {BUNDLED_TAG!r} or "
+            f"{LATEST_TAG!r}. The pin is ignored and URLs resolve as if it "
+            "had not been set. A future major release will raise ValueError "
+            "here instead.",
+            FutureWarning,
+            stacklevel=2,
+        )
+
+    _cdn_version_override = candidate
 
 
 def get_cdn_version() -> str:
--- a/maidr/util/dependencies.py
+++ b/maidr/util/dependencies.py
@@ -509,7 +509,22 @@ def get_cdn_version() -> str:
         return pin
     if _resolution_would_block():
         return _offline_version()
-    return _resolve_latest_version() or LATEST_TAG
+    # A lookup that failed falls back to the bundled version, not to
+    # `@latest`.  `@latest` is the mutable dist-tag whose seven-day
+    # `Cache-Control` is the whole of #290 -- so degrading to it meant the
+    # fix stopped applying in precisely the case it was meant to survive,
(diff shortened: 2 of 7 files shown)


jooyoungseo:
The ask about the numbers is the right one, and it deserved more than a re-run of my own script — so I re-derived them from **npm and PyPI directly**, which is a different data source from the GitHub releases API the first pass used, reached by a different code path.

```
source: npm registry + PyPI (not the GitHub releases API)
window: 2026-01-28 .. 2026-08-13
upstream releases 50   distinct minors 33
py-maidr releases 18

days between consecutive upstream minors:
  median 5.1  mean 6.2  min 0.0  max 18.9

upstream minors between consecutive py-maidr releases (17 cycles):
  median 1  mean 1.7  min 0  max 7
  reaching 3+: 6/17 (35%)
  reaching 5+: 1/17 (6%)
  reaching 8+: 0/17 (0%)
```

Every figure in the comment matches: 17 cycles, median 1, mean 1.7, max 7, and the three threshold counts. npm's `time` map and PyPI's `upload_time_iso_8601` are what the registries actually recorded at publish, so this is the stronger of the two derivations — the GitHub one measures when a release object was created, which can drift from when the artefact shipped.

One thing the cross-check settles that the first pass could not: py-maidr has **18 PyPI releases** in the window against 18 GitHub releases, so no release reached users without a tag or vice versa. A mismatch there would have quietly invalidated the cycle table, since a cycle is defined by consecutive releases.

Your second point — 19 lines of comment for one `int` — I agree with as a general worry, and it is why the block ends by naming *which* table to recompute rather than inviting the next person to redo the analysis. If a second constant needs this treatment, that is the moment to move it out to a docs page rather than the moment to shorten this one.

**This review ran against `2450355`, the docs-only commit.** Two more have landed since and the description now covers all three: `ef522f2` warns at the call when `set_cdn_version()` is given a bad version (#294), and `4364c95` stops the failed-lookup path emitting `@latest` (#295) — which is the one with real behaviour change, and where ten existing cases that pinned the old fallback are updated in place rather than deleted.

---

PR #368, opened 2026-08-13
feat(hexbin): read Axes.hexbin as a hexagonal bin lattice

### LINKED ISSUE #341: feat: support hist2d and hexbin

jooyoungseo:
Closes #341.

A hexbin is the standard answer to an overplotted scatter: bin the points into hexagons and encode the count as fill. Read that way it is a heatmap, and the navigation, braille and pitch all transfer. maidr.js has carried a `hexbin` trace since v4.2.0 and nothing here emitted one, so `ax.hexbin()` fell through to a static image.

Two things about how matplotlib builds the lattice do not survive a naive reading, and both are invisible until you look at the numbers. They are the whole reason this is not thirty lines.

## The emission order is not the reading order

`Axes.hexbin` builds its offsets like this:

```python
offsets[:nx1 * ny1, 0] = np.repeat(np.arange(nx1), ny1)     # aligned lattice
offsets[:nx1 * ny1, 1] = np.tile(np.arange(ny1), nx1)
offsets[nx1 * ny1:, 0] = np.repeat(np.arange(nx2) + 0.5, ny2)  # offset lattice
offsets[nx1 * ny1:, 1] = np.tile(np.arange(ny2), nx2) + 0.5
```

`repeat` on x and `tile` on y means consecutive offsets walk **up a column**, and the whole offset lattice comes after the whole aligned one. So grouping the points into rows is a permutation, not a reshape. For a `gridsize=3` lattice it comes out `[0, 2, 4, 6, 8, 9, 10, 1, 3, 5, 7]`.

The selector list has to be permuted with them. Left in document order, every bin past the first row boundary highlights someone else's hexagon while still announcing a real centre and a real count — the same defect as #316 and #350, in the trace where it is hardest to notice, because a hexbin announces centres rather than indices and so has nothing that would give it away.

Grouping is on the y centre by exact float equality, which is exact here rather than approximate: every centre in a row comes from the same `index * spacing + origin`, so a row's values are identical bit for bit, and the two lattices are half a spacing apart.

## The rows are ragged, and stay that way

The two lattices hold different numbers of bins, and `mincnt` or a `C` argument drops the empty ones — `gridsize=3` gives rows of 4, 3, 4. They are left ragged. Padding would put bins on the chart that were never drawn, and `MovableGrid.moveOnce` clamps a row change to the new row's length precisely because grids arrive like this:

> On ragged grids a row change can leave col past the new row's end

Rows ascend in y, so row 0 is the bottom — the frontend's UPWARD steps to the *next* row index, the same convention a heatmap follows.

## Every drawn bin is announced

With no `mincnt` matplotlib draws the empty cells too, so there is a hexagon on the chart for each. Dropping them would describe a lattice with holes a sighted reader does not see, and would put the selector list out of step with the DOM by however many were dropped. `HexbinTrace.mapToSvgElements` withdraws highlighting outright when the counts disagree, so that failure is total rather than partial.

Measured across every mode, offsets ↔ counts ↔ `<use>` stay 1:1:

| | offsets | array | `<use>` |
| --- | --- | --- | --- |
| default | 11 | 11 | 11 |
| `mincnt=1` | 9 | 9 | 9 |
| `bins="log"` | 11 | 11 | 11 |
| `bins=5` | 11 | 11 | 11 |
| `C=` | 9 | 9 | 9 |
| `marginals=True` | 11 | 11 | 11 |

## The colour axis is named for what the fill encodes

`hexbin` has two arguments that change what the number means:

- `C=` replaces the count with `reduce_C_function` applied to the values given for the points in the bin — a mean by default, routinely not an integer. Labelled `value`.
- a numeric `bins=` discretises the counts and colours each bin by *which interval* its count landed in. `get_array()` then holds an interval index: a three-point bin and a nine-point bin can both read 1. Labelled `count bin`.

Announcing either as a count would be wrong in the way that is hardest to catch — plausible number, sound chart, nothing contradicting it. `bins="log"` is **not** one of these: it only installs a log norm and leaves the array as raw counts, which I checked in the source rather than assumed. `z_label=` overrides all three.

## The collection comes from the patch

Rather than being searched for on the axes. `marginals=True` draws two further `PolyCollection`s, and a violin body or a `fill_between` band on the same axes is one as well, so the class does not identify the lattice while the call's return value does. The gid is assigned during extraction, as `AreaPlot` and `MultiLinePlot` already do, because a gid is otherwise only stamped at draw time and the schema is built first.

## Tests

Eight cases in `tests/core/test_hexbin_lattice.py`. The one that matters resolves the **emitted CSS against the real exported SVG** and checks each match is the hexagon whose centre and count that bin announces — nothing weaker distinguishes a correct mapping from an off-by-a-row one. `cssselect` is added to the `test` extra for it, alongside the `lxml` already used to build the SVG.

There is an explicit anti-vacuity guard: `test_the_reading_order_is_not_the_emission_order` fails if matplotlib ever starts emitting row by row, at which point the selector test would pass without proving anything.

Each mechanism was reverted and the named test confirmed to fail:

| reverted | fails |
| --- | --- |
| selectors left in document order | `each_selector_addresses_the_bin_it_describes`, `mincnt_...still_lines_up` |
| rows in descending y | `the_lattice_is_read_as_rows_bottom_first` |
| z label always `count` | `the_colour_axis_is_named_for_what_the_fill_encodes` |

`test_hexbin_is_not_registered_yet` in `test_hist2d.py` was written to fail on this day and be rewritten as a positive one. Done, along with the module docstring that drew the boundary.

Full suite: 1129 passed. Lint clean on every touched path.

## One thing found and not fixed here

While writing the docs example I hit a crash that has nothing to do with hexbin: a user-written `fig.colorbar(...)` registers a spurious `heat` layer and then **breaks `render()` outright** with an `ExtractionError`. It reproduces on `pcolormesh` and `scatter` with all hexbin code removed, so it predates this. Filed separately rather than folded in; the docs example simply does not call `fig.colorbar`.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,7 +41,7 @@ The `maidr/patch/` modules use `wrapt` to intercept matplotlib/seaborn plot call
 
 ### Supported Plot Types
 
-Defined in `maidr/core/enum/plot_type.py`: BAR, BOX, COUNT, DODGED, HEAT, HIST, LINE, PIE, SCATTER, STACKED, STEP, SMOOTH, CANDLESTICK, VIOLIN_KDE, VIOLIN_BOX.
+Defined in `maidr/core/enum/plot_type.py`: BAR, BOX, COUNT, DODGED, HEAT, HEXBIN, HIST, LINE, PIE, SCATTER, STACKED, STEP, SMOOTH, CANDLESTICK, VIOLIN_KDE, VIOLIN_BOX.
 
 Note that `maidr/plotly/` builds its MAIDR schema in Python and therefore needs its own
 handling per plot type, whereas `maidr/altair/` delegates entirely to the upstream
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -505,6 +505,48 @@ if cbar:
 plt.show() #<<
 ```
 
+### Hexbin Plot
+
+A hexbin is the usual answer to a scatter plot with too many points to read:
+the points are binned into hexagons and the count is shown as fill. maidr reads
+it as a lattice of counted cells, one row at a time.
+
+Hexagons tessellate by offsetting alternate rows by half a cell, so a bin's
+column index is not its position — bin 3 of one row and bin 3 of the next sit
+at different x. Each bin is therefore announced by its **centre** rather than
+by an index, and moving up or down keeps to the x you started from instead of
+carrying the index into a row where it means something else.
+
+```{python}
+#| fig-alt: Hexbin plot of two thousand points drawn from two overlapping clusters
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+import maidr #<<
+
+rng = np.random.default_rng(20260813)
+x = np.concatenate([rng.normal(-1, 0.8, 1200), rng.normal(1.5, 0.5, 800)])
+y = np.concatenate([rng.normal(0, 0.9, 1200), rng.normal(1.5, 0.6, 800)])
+
+fig, ax = plt.subplots(figsize=(7, 6))
+# `z_label` names the colour axis for maidr, so a bin reads as
+# "First measurement: -1.02, Second measurement: 0.31, Points: 47".
+# Without it the axis is named "count", which is what the fill encodes.
+hexes = ax.hexbin(x, y, gridsize=12, cmap="Blues", z_label="Points") #<<
+ax.set_xlabel("First measurement")
+ax.set_ylabel("Second measurement")
+ax.set_title("Where the points pile up")
+
+plt.show() #<<
+```
+
+`hexbin` has two arguments that change what the fill means, and maidr names the
+colour axis to match rather than calling everything a count: `C=` replaces the
+count with a reduction of the values you supply, and a numeric `bins=`
+discretises the counts so the colour shows *which interval* a bin landed in.
+Pass `z_label=` to name it yourself.
+
 ### Box Plot
 
 ```{python}
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -77,7 +77,7 @@ We currently support the following data visualization libraries in Python:, and 
 
 * Heatmap having one x (categorical) and one y (categorical) and z (numerical) axis.
 
-* Hexbin plot (`ax.hexbin()`) having one x (numerical) and one y (numerical) axis and a z axis carrying each hexagon's count. Read as a lattice of counted cells, one row at a time; because hexagons tessellate by offsetting alternate rows, each bin is announced by its centre rather than by a column index.
+* Hexbin plot (`ax.hexbin()`) having one x (numerical) and one y (numerical) axis and a z axis carrying what the fill encodes — each hexagon's count by default, the reduction of `C` when one is given, or the count's interval when `bins` is a number. Read as a lattice of counted cells, one row at a time; because hexagons tessellate by offsetting alternate rows, each bin is announced by its centre rather than by a column index.
 
 * Scatter plot having one x (numerical) and one y (numerical) axis.
 
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -77,6 +77,8 @@ We currently support the following data visualization libraries in Python:, and 
 
 * Heatmap having one x (categorical) and one y (categorical) and z (numerical) axis.
 
+* Hexbin plot (`ax.hexbin()`) having one x (numerical) and one y (numerical) axis and a z axis carrying each hexagon's count. Read as a lattice of counted cells, one row at a time; because hexagons tessellate by offsetting alternate rows, each bin is announced by its centre rather than by a column index.
+
 * Scatter plot having one x (numerical) and one y (numerical) axis.
 
 * Regression plot having one x (numerical) and one y (numerical) axis with scatter points and fitted regression line.
--- a/maidr/core/enum/maidr_key.py
+++ b/maidr/core/enum/maidr_key.py
@@ -43,6 +43,9 @@ class MaidrKey(str, Enum):
     # Scatter plot grid navigation keys.
     TICK_STEP = "tickStep"
 
+    # Hexbin keys. A bin carries its centre and how many points fell in it.
+    COUNT = "count"
+
     # Step plot keys. The per-point ordinal level name reuses LABEL above.
     STEP_DIRECTION = "stepDirection"
 
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -10,6 +10,11 @@ class PlotType(str, Enum):
     DODGED = "dodged_bar"
     ERRORBAR = "error_bar"
     HEAT = "heat"
+    #: A hexagonal bin lattice: the standard answer to an overplotted scatter.
+    #: Read as a grid of cells each carrying a count, which is a heatmap --
+    #: except that alternate rows are offset by half a cell, so a bin's column
+    #: index is not its position and the trace announces centres instead.
+    HEXBIN = "hexbin"
     HIST = "hist"
     LINE = "line"
     #: A filled band between a series and a baseline. Emitted for a single
--- a/maidr/core/figure_manager.py
+++ b/maidr/core/figure_manager.py
@@ -52,6 +52,7 @@ class FigureManager:
         PlotType.HIST: 1,
         PlotType.BOX: 1,
         PlotType.HEAT: 1,
+        PlotType.HEXBIN: 1,
         PlotType.COUNT: 1,
         PlotType.PIE: 1,
         PlotType.SMOOTH: 1,
--- a/maidr/core/plot/hexbinplot.py
+++ b/maidr/core/plot/hexbinplot.py
@@ -1,0 +1,167 @@
+from __future__ import annotations
+
+import uuid
+
+import numpy as np
+from matplotlib.axes import Axes
+from matplotlib.collections import Collection
+
+from maidr.core.enum import MaidrKey, PlotType
+from maidr.core.plot import MaidrPlot
+from maidr.exception import ExtractionError
+
+
+class HexbinPlot(MaidrPlot):
+    """
+    A hexagonal bin lattice read as a grid of counted cells.
+
+    ``Axes.hexbin`` answers an overplotted scatter by binning the points into
+    hexagons and encoding each bin's count as fill. Read that way it is a
+    heatmap, and the navigation, braille and pitch all transfer.
+
+    The one real difference is that a hex lattice **staggers alternate rows by
+    half a cell**, which is what lets the hexagons tessellate. Two consequences
+    shape everything below.
+
+    First, a bin's column index is not its position: bin 3 of one row and bin 3
+    of the next sit at different x. So each bin carries its own centre, and the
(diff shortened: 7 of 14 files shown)


jooyoungseo:
Checked all four rather than taking them on faith. Two turned out to be worth acting on, and one of them was more real than the review suspected.

**2 — the masked array is not hypothetical.** `get_array()` on a hexbin returns a `MaskedArray` today, in every mode:

```
default    n= 11 masked_type=True any_masked=False
mincnt=2   n=  9 masked_type=True any_masked=False
C          n=  9 masked_type=True any_masked=False
C+mincnt   n=  8 masked_type=True any_masked=False
bins=5     n= 11 masked_type=True any_masked=False
```

The mask is never *set* — matplotlib drops the empty bins through `good_idxs` before `set_array`, so there is nothing left to mask — which is precisely why one starting would go unnoticed. `np.asarray(masked)` returns the data underneath, so the bin would be announced with whatever was in that slot: a plausible number, a sound chart, nothing contradicting it. Now read with `np.ma.filled(..., nan)`.

The bin is still **emitted**, without a value rather than dropped. Dropping it would be worse than the fabricated count: the hexagon is drawn, `HexbinTrace.mapToSvgElements` matches the selector list against the DOM by length, and a one-short list withdraws highlighting from the entire lattice. `null` rather than `nan`, since JSON has no NaN literal and `json.dumps` would emit a bare `NaN` the browser's parser rejects. The trace already reads a non-finite count as "no value here" — it filters them out of the pitch range and announces `NaN`.

Driven directly by masking a bin, since no chart can produce one. Reverting to `np.asarray` fails that test.

**1 — log scales hold, and now say so.** `hexbin` takes its own `xscale`/`yscale`, which log-transform the input before binning — a second path through the centre arithmetic the exact-equality grouping depends on. The row grouping survives it:

| | offsets | distinct y | exact groups |
| --- | --- | --- | --- |
| default | 11 | 3 | 3, 4, 4 |
| `xscale="log"` | 11 | 3 | 3, 4, 4 |
| `yscale="log"` | 11 | 3 | 3, 4, 4 |
| both | 11 | 3 | 3, 4, 4 |

Worth a test rather than a note, because the failure would be quiet: a grouping that fell apart splits every row into singletons rather than raising.

**4 — fixed.** `docs/index.qmd` named the z axis for the count, which is only the default. It now says what the fill encodes in each of the three cases; `docs/examples.qmd` already carried the detail.

**3 — left as is, deliberately.** `z_label=""` falling through to the inferred label is the codebase's convention for a blank label, as you note. Respecting it literally would mean a bin announcing an empty axis name, which is not something a caller can want; if a way to suppress the axis is ever wanted it should be a distinct value, not the empty string.

Full suite 1133 passed, lint clean on every touched path.

---

jooyoungseo:
The log-scale test I added in the last commit failed on Python 3.9, and chasing it found that the previous commit's version of that test was **passing while the chart lied**. Review's point 1 was worth more than either of us thought.

`hexbin` bins in the transformed space. On matplotlib 3.10 the offsets come back in that space as well:

```
matplotlib 3.10.9
{}                 offsets=(11, 2) paths=1 array=11
   offsets x range: 0.304 to 11.232   data x range: 0.304 to 11.232
{'xscale': 'log'}  offsets=(11, 2) paths=1 array=11
   offsets x range: -0.518 to 1.05    data x range: 0.304 to 11.232
```

Those are log10 coordinates. A bin centred at x = 3.4 announces `0.53`. My test asserted the rows grouped cleanly — which they do, since `10 ** (j*sy + ymin)` is still one value per row — so it passed on exactly the property that was fine while the coordinates were wrong. Right structure, right counts, wrong numbers, and nothing in the output to contradict them.

CI then found the other half. On matplotlib 3.9.4 the same call returns **one path per hexagon and a single placeholder offset**:

```
matplotlib 3.9.4
{}                 offsets=(11, 2)  array=11
{'xscale': 'log'}  offsets=(1, 2)   array=11    offsets: [[0.0, 0.0]]
```

The hexagons are no longer congruent under a log axis, so each becomes its own path and the centres leave `get_offsets()` entirely. That mismatch is what the `len(values) != len(offsets)` guard was for, and it fired.

**So log scales are declined, not supported.** Un-transforming the 3.10 offsets would be an assumption about matplotlib's internals that the neighbouring release already breaks. The layer is not registered and the figure keeps the static image it had before this patch existed — the documented degradation, and the same thing `stackplot` does with a call it cannot read.

Keyed on `hexbin`'s own `xscale`/`yscale` rather than on the axis, which is a real distinction and now has its own test: an axes that was *already* log-scaled makes matplotlib bin linearly, and those offsets are honest data coordinates, so that chart still reads.

**Two other things fell out of the CI failure, both worth having.**

The decline had to move out of `ExtractionError` and into the patch. That exception is **fatal to the whole render** — it does not skip the layer, it takes the figure down. That is how one unreadable hexbin broke `test_plt_show_forwards_use_cdn_kwarg` and `test_plt_show_uses_module_default_when_no_kwarg`, which have nothing to do with hexbins: a figure left registered is rendered by any later `plt.show()`. Same shape as #369, and it is the reason that issue is about a crash rather than a stray layer.

And this file now closes its figures. Not housekeeping — the leak is what carried the failure into another module.

| | |
| --- | --- |
| Python 3.11 / matplotlib 3.10.9 | 1135 passed |
| Python 3.9 / matplotlib 3.9.4 | 1131 passed, 4 skipped |
| lint | clean |

---

jooyoungseo:
The bug is real. You asked for the sanity check you couldn't run, so here it is:

```
C only        dtype=float64  range=79.139 to 113.795  first4=[96.16, 79.139, 99.916, 113.795]
bins=5 only   dtype=int64    range=0.000 to 4.000     first4=[0, 1, 1, 2]
C + bins=5    dtype=int64    range=0.000 to 4.000     first4=[2, 0, 3, 4]
C + bins=log  dtype=float64  range=79.139 to 113.795  first4=[96.16, 79.139, 99.916, 113.795]
```

Exactly as you read it out of the source. A bin whose mean `C` is 113.8 was being announced as **"value: 2"** — and 2 is a perfectly sensible-looking number, which is the whole problem.

`bins` is read first now. I took the parenthetical too: the combined case is **`"value bin"`**, not `"count bin"`. Nothing was counted, so calling it a count bin would be its own small lie in the course of fixing a bigger one — and a reader told "count bin" would reasonably assume the underlying quantity was a frequency.

So four labels rather than three:

| `C` | `bins` | label |
| --- | --- | --- |
| — | — or `"log"` | `count` |
| — | numeric | `count bin` |
| given | — or `"log"` | `value` |
| given | numeric | `value bin` |

Reverting to the early `C` return fails `test_the_colour_axis_is_named_for_what_the_fill_encodes`. You had the diagnosis right there too: the test covered `C` and `bins` separately and never together, so there was no case for the precedence to be wrong in.

**On the two smaller notes.**

The `PLOT_TYPE_PRIORITY` entry — yes, deliberate, and thank you for checking rather than assuming. Absent, `.get(..., 0)` means a lone hexbin never becomes the figure's own type.

The visual highlight in a real browser — I want to be plain that this is **not** verified. The tests resolve the emitted CSS against the exported SVG and confirm each match is the hexagon whose centre and count that bin announces, which is what catches an off-by-a-row mapping; they do not drive a browser and do not observe an outline being painted. py-maidr has no browser-driven harness for me to add a case to, and standing one up is well outside this change. Worth saying rather than letting "the selector test passes" stand in for something it isn't — the same gap xability/maidr#814 names as this work's blind spot.

Full suite 1135 passed, lint clean.

---

jooyoungseo:
Both flagged items fixed, and the observations answered with what I could measure rather than with agreement.

**1 — backtick, fixed.** ` ``get_array()` ` → ` ``get_array()`` `.

**2 — lockfile churn, checked, and it is not drift.** Only `cssselect` was added; every other line is `uv` writing environment markers more explicitly on packages that already resolved to two versions:

```
$ git diff uv.lock | grep -cE '^[-+]name = '
2                     # cssselect 1.3.0 and 1.5.0, the two Python-version splits

-    { name = "jinja2" },
+    { name = "jinja2", marker = "python_full_version < '3.10'" },
```

No package version changed. `uv lock --check` passes, and CI uses `astral-sh/setup-uv@v5` unpinned, so its `uv` is at least as new as the one that wrote this. The lock's format `version`/`revision` is untouched — the diff starts inside the `altair` block, not at the header.

**3 — `z_label` popped before the guard now**, rather than after. You are right that it mirrors `heatmap.py`, and right that there is no live path: `Axes.hexbin` is the only entry point and nothing patched calls it. Popping first costs nothing and means there does not have to be one, so I would rather remove the trap than document it.

**4 — matplotlib versions, confirmed from the lock rather than assumed.** There are exactly two, and they are the two I tested:

```
matplotlib 3.9.4    python_full_version < '3.10'
matplotlib 3.10.9   python_full_version >= '3.10'
```

So the 3.9/3.10/3.11/3.12 matrix rows resolve to one or the other, and both representations — 3.9's one-path-per-hexagon-plus-placeholder-offset and 3.10's real-but-transformed offsets — are exercised by CI rather than only by me locally.

**5 — the exact-float grouping key. I think a tolerance would make this worse, not safer.** The two failures that actually occurred are both covered by narrower checks: the offsets/counts length guard caught 3.9's log representation, and the log decline caught 3.10's transformed one. A tolerance introduces a failure the exact test cannot have — merging two genuinely distinct rows on a fine lattice, where the row spacing is small and the epsilon is a guess. That collapses two rows into one, silently, and the reader loses a row of the chart with everything still self-consistent. Exact equality fails loudly instead: per-row noise would split every row into singletons, which `the lattice is read as rows, bottom first` catches immediately since a one-bin row cannot have an unsorted x and the row count would jump to the bin count. I would rather keep the sharp edge that fails visibly.

**6 — comment length.** Fair, and I do not have a principled counter beyond the content being load-bearing: the two facts that sink a naive implementation (lattice-major emission, ragged rows) are invisible in the code and would be re-derived by whoever edits it next. A PR description is not where someone editing `_get_selector` in a year will look. Happy to trim if the maintainers prefer the terser house style.

**7 — the colorbar issue is filed: #369.** Reproduced on `pcolormesh`, `scatter` and `hexbin` with all hexbin code removed, so it predates this.

Full suite 1135 passed, lint clean.

---

PR #370, opened 2026-08-14
fix(colorbar): a colorbar is a legend, not a second chart

### LINKED ISSUE #369: A user-written fig.colorbar() registers a phantom heat layer and breaks render()
Calling `fig.colorbar(...)` yourself — the ordinary way to put a colour scale next to a mappable — registers a second, bogus `heat` layer, and then `maidr.render()` fails outright.

```python
import matplotlib
matplotlib.use("Agg")
import numpy as np
import maidr

fig, ax = plt.subplots()
mesh = ax.pcolormesh(np.arange(12).reshape(3, 4))
fig.colorbar(mesh, ax=ax, label="L")

print([p.type.value for p in FigureManager.get_maidr(fig).plots])
FigureManager.get_maidr(fig)._flatten_maidr()
```

```
['heat', 'heat']
ExtractionError: Error extracting data for heat plot type from
<class 'matplotlib.collections.LineCollection'>
```

The colorbar draws onto its own axes through an entry point the heatmap patch wraps, so MAIDR registers it as a chart. It is not one — a colorbar is a legend for the chart beside it.

Two separate problems fall out, and the second is the serious one:

1. **A phantom navigable layer.** Even if extraction succeeded, the reader would be handed a second "heatmap" to page through that the figure does not contain.
2. **The render dies.** The extraction reaches a `LineCollection` (the colorbar's outline) where it expects a mappable, raises, and there is no chart at all — not a degraded one, none.

Not specific to heatmaps. Every plot type is affected, because the fault is in what the colorbar's own draw registers:

| the user's chart | layers registered | `render()` |
| --- | --- | --- |
| `ax.pcolormesh(...)` | `['heat', 'heat']` | `ExtractionError` |
| `ax.scatter(..., c=...)` | `['point', 'heat']` | `ExtractionError` |
| `ax.hexbin(...)` | `['hexbin', 'heat']` | `ExtractionError` |

So any chart that wants a colour scale — which is most charts that use colour to carry a third variable — cannot be rendered at all, as long as the caller adds the colorbar rather than letting seaborn do it.

That last clause is why this has stayed hidden. `sns.heatmap()` creates its colorbar *inside* the patched call, so the recursion guard suppresses it, and the documented heatmap example reaches for the existing one (`heatmap.collections[0].colorbar`) rather than making a new one. Every worked example in the docs happens to take the path that works.

### Reproduced

matplotlib 3.10.9, on `main` with all hexbin code removed, so it is not related to #341 — I hit it while writing that example and took the colorbar out of it rather than shipping a snippet that crashes.

### What a fix probably looks like

The colorbar's draw needs to be excluded from registration. Matplotlib gives a usable signal: the axes a colorbar lives on carries `ax._colorbar` (and the `Colorbar` holds a reference back), so the patch can decline an axes that belongs to a colorbar rather than trying to identify the artist. Worth checking against `Colorbar` construction directly instead of the private attribute if there is a public equivalent.

Whatever the mechanism, both halves need asserting: that the layer is not registered, **and** that a figure with a user-written colorbar renders. The second is the one a user notices.

jooyoungseo:
Closes #369.

A colorbar paints its gradient onto its own axes through the same entry points the heatmap patch wraps, so MAIDR registered it as a `heat` layer of its own. Two things followed, and the second is the one a user notices.

A **phantom layer**, first — a reader handed a second "heatmap" to page through that the figure does not contain.

And then the **render died**. Extraction reaches the colorbar's outline, a `LineCollection` where a mappable is expected, and raises `ExtractionError` — which is not confined to the layer that raised it. It takes the whole figure with it, so a chart that would have read perfectly well produced nothing at all.

## Not a heatmap problem

The fault is in what the colorbar's own draw registers, which is the same whatever it is a legend for. Before:

| the user's chart | layers registered | `render()` |
| --- | --- | --- |
| `ax.pcolormesh(...)` | `['heat', 'heat']` | `ExtractionError` |
| `ax.scatter(..., c=...)` | `['point', 'heat']` | `ExtractionError` |
| `ax.hexbin(...)` | `['hexbin', 'heat']` | `ExtractionError` |

So any chart wanting a colour scale — most charts that use colour to carry a third variable — could not be rendered at all, as long as the caller wrote the colorbar rather than letting seaborn do it.

That last clause is why this stayed hidden. `sns.heatmap()` creates its colorbar *inside* the patched call, where the recursion guard already suppressed it, and the documented heatmap example reaches for the existing one (`heatmap.collections[0].colorbar`) rather than making a new one. Every worked example in the docs happens to take the path that works. It turned up while I was writing the hexbin example for #368, which needed a colorbar and crashed.

## Where the guard goes, and why not where it looks like it should

`Colorbar._draw_all`, run inside the same recursion context every patched function already uses for the calls it makes internally.

The obvious alternative — test whether the target axes belongs to a colorbar — is wrong twice over. It would have to cover two routes that diverge well before the draw (`Figure.colorbar` steals space from the parent axes when given `ax=` and skips that entirely when handed a `cax`), and the attribute it would read is not set yet:

```
registered heat   | axes._colorbar set: False | _colorbar_info set: False
registered heat   | axes._colorbar set: False | _colorbar_info set: True
```

`ax._colorbar` is assigned *after* the draw that registers the layer. Measured, not assumed. Every route converges at `_draw_all`, which is why the guard sits there.

`_draw_all` is private, so the wrap is guarded: an upstream rename degrades to today's behaviour rather than stopping the package importing, and the tests below would catch it.

## Tests

Nine cases in `tests/core/test_colorbar_is_not_a_chart.py`, all nine failing with the guard removed.

The two that matter most are the pair: *no layer is added* and *the figure still renders*. The first alone would have passed on a fix that stopped registering the layer but left something else broken; the second is what a user actually experiences.

Two more guard the mechanism rather than the symptom, because the failure mode of a leaking context is silent and looks nothing like a colorbar bug:

- `test_the_chart_a_colorbar_belongs_to_is_unchanged` — the layer's data is byte-identical to the same chart drawn without a colorbar, so suppressing the legend suppressed nothing the chart says.
- `test_a_chart_drawn_after_a_colorbar_still_registers` — a second chart added afterwards is still picked up, so the context ends with the draw rather than outliving it.

And `test_an_explicitly_placed_colorbar_is_covered_too` pins the `cax=` route, which is the one an axes-attribute test would most plausibly have missed.

| | |
| --- | --- |
| Python 3.11 / matplotlib 3.10.9 | 1144 passed |
| Python 3.9 / matplotlib 3.9.4 | the new file's 9 passed |
| guard removed | 9 failed |
| lint | clean |

---

### DIFF
--- a/maidr/patch/__init__.py
+++ b/maidr/patch/__init__.py
@@ -4,6 +4,7 @@ from . import (  # noqa: F401
     barplot,
     boxplot,
     clear,
+    colorbar,
     errorbar,
     heatmap,
     hexbin,
--- a/maidr/patch/colorbar.py
+++ b/maidr/patch/colorbar.py
@@ -1,0 +1,64 @@
+from __future__ import annotations
+
+from typing import Any
+
+import wrapt
+from matplotlib.colorbar import Colorbar
+
+from maidr.core.context_manager import ContextManager
+
+
+def draw_quietly(wrapped, _, args, kwargs) -> Any:
+    """
+    Draw a colorbar without registering it as a chart of its own.
+
+    A colorbar paints its gradient onto its own axes through the very entry
+    points the heatmap patch wraps, so MAIDR was registering it as a second
+    ``heat`` layer. It is not a chart: it is the legend for the one beside it,
+    and the values it carries are already the ``z`` axis of that layer.
+
+    Two things went wrong, and the second is the one a user notices.
+
+    A phantom layer, first -- a reader handed a second "heatmap" to page
+    through that the figure does not contain. And then the render **died**:
+    extraction reaches the colorbar's outline, a ``LineCollection`` where a
+    mappable is expected, and raises. An ``ExtractionError`` is not confined
+    to its own layer; it takes the whole figure with it, so a chart that would
+    have read perfectly well produced nothing at all (#369).
+
+    Every route -- ``Figure.colorbar``, ``plt.colorbar``, an explicitly
+    supplied ``cax`` -- goes through ``Colorbar._draw_all``, which is why the
+    guard lives here rather than on a test for "is this axes a colorbar".
+    Attribute-sniffing would also have been wrong on the timing:
+    ``ax._colorbar`` is not assigned until after the draw that registers the
+    layer.
+
+    Parameters
+    ----------
+    wrapped : Callable
+        The original ``Colorbar._draw_all``.
+    _ : Any
+        The ``Colorbar`` wrapt bound the method to. Unused, and named for that.
+    args : tuple
+        Positional arguments matplotlib passed.
+    kwargs : dict
+        Keyword arguments matplotlib passed.
+
+    Returns
+    -------
+    Any
+        Whatever the wrapped method returned, unchanged.
+    """
+    with ContextManager.set_internal_context():
+        return wrapped(*args, **kwargs)
+
+
+# `_draw_all` is private, so a rename upstream must degrade to the previous
+# behaviour rather than stop the package importing. That is a phantom layer
+# and a broken render, which is bad -- but a package that cannot be imported
+# is worse, and the failure would be caught by the tests that pin this rather
+# than by a user.
+try:
+    wrapt.wrap_function_wrapper(Colorbar, "_draw_all", draw_quietly)
+except AttributeError:  # pragma: no cover - matplotlib renamed the method
+    pass
--- a/maidr/patch/colorbar.py
+++ b/maidr/patch/colorbar.py
@@ -8,9 +8,13 @@ from matplotlib.colorbar import Colorbar
 from maidr.core.context_manager import ContextManager
 
 
-def draw_quietly(wrapped, _, args, kwargs) -> Any:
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Both taken.

**Naming.** `draw_quietly` sat one underscore from `common._draw_quietly`, and you are right that a future reader grepping either name would land on the wrong one. `_suppress_registration` now, which says what it does rather than how it reads — and the docstring names the other one explicitly so the next person does not have to work out that they are unrelated.

**`plt.colorbar`.** Pinned. You are right that the reasoning covers it, and right that reasoning is not a test — it was the one route where a reader would reasonably wonder. Removing the guard now fails it alongside the other nine.

Ten cases, all ten failing with the guard removed. Full suite 1145 passed, lint clean.

On the docstring length: noted, and I do not have a strong defence beyond the same one as #368 — the timing fact about `ax._colorbar` is the reason the guard is where it is, and it is invisible in the code. Happy to trim toward the house style if the maintainers would rather.

---

Issue #339, opened 2026-08-10
feat: support area plots (stackplot, fill_between)


jooyoungseo:
## Description

`plt.stackplot`, `Axes.fill_between` and `Axes.fill_betweenx` are not patched, so area and stacked-area charts produce no MAIDR output.

## Scope

- [ ] New `maidr/patch/areaplot.py` patching `Axes.stackplot`, `Axes.fill_between`, `Axes.fill_betweenx`
- [ ] `PlotType.AREA` (and the stacked variant) in `maidr/core/enum/plot_type.py`
- [ ] `AreaPlot` under `maidr/core/plot/`, extracting the band boundaries from the returned `PolyCollection`
- [ ] Register `PolyCollection.draw` in `maidr/patch/highlight.py` so band segments can be highlighted
- [ ] Announce the series value and the cumulative total separately for the stacked case

## Note on `fill_between`

It is also commonly used to draw a confidence band around a line rather than an area chart in its own right. Those two cases want different readings, and the distinction is worth resolving here — a band around an existing line is better expressed as an interval on that line than as a standalone area layer.

## Blocked on

Core `area` trace support: xability/maidr#788.

## Difficulty

Medium.

jooyoungseo:
`stackplot` landed in #356. `fill_between` is the remaining half, and measuring it first is worth it, because it is not one chart type — it is three, and two of them would be actively wrong if read as the third.

Current state:

```
stackplot                    ['stacked_area']
fill_between (baseline)      not registered
fill_between (band)          not registered
fill_betweenx                not registered
plot + fill_between band     ['line']        <- the line only
```

That last row is the common one, and the most costly: a line chart with a confidence ribbon announces the line and says nothing at all about the band. A sighted reader sees the uncertainty; nobody else does.

## Three readings, not one

**1. `fill_between(x, y1)` — a band from the baseline to a curve.** This is an area chart, and it is the same claim a single `stackplot` band makes. `AREA`, one series. Straightforward, and the code for it already exists.

**2. `fill_between(x, lo, hi)` with a line drawn on the same axes.** This is an estimate and its interval, which is what `ERROR_BAR` is for. The core's point schema takes it directly:

```ts
interface ErrorBarPoint {
  x: number | string;
  y: number;      // the estimate
  yMin?: number;  // absolute lower bound
  yMax?: number;
}
```

[...]

PR #371, opened 2026-08-14
feat(area): read a baseline-to-curve fill_between as an area chart

jooyoungseo:
Part of #339.

`Axes.fill_between` was unregistered, so a filled area chart drew and read as a static image. `ax.stackplot()` already emitted an area layer for the same picture (#356), which made the gap arbitrary from a reader's side: **the same chart read or did not depending on which function drew it.**

```
                            before              after
fill_between(x, y)          not registered      area
fill_between(x, y, 0)       not registered      area
fill_betweenx(y, x)         not registered      area
fill_between(x, lo, hi)     not registered      not registered   (deliberate)
fill_between(x, y, 2)       not registered      not registered   (deliberate)
stackplot(x, y1, y2)        stacked_area        stacked_area
```

## The part that needed deciding rather than writing

`fill_between` draws two different charts, and only one of them is an area.

**`fill_between(x, y1)` fills from zero up to a curve.** That measures exactly what a one-series `stackplot` band measures — a magnitude per position, from a baseline the reader can assume. Registered.

**`fill_between(x, lo, hi)` draws the gap.** Its content is the distance between the edges, not the height of either. Read as an area it would announce `hi` as a magnitude and drop `lo` without saying so — a complete, confident description of a chart nobody drew. The honest reading is an estimate and its interval, and there is no estimate on the chart; the midpoint would be a number nothing drew. So it is left unregistered and the figure keeps the static image it had.

A **constant second edge** (`fill_between(x, y, 2)`) is the same problem in miniature: heights measured from somewhere the announcement would not mention. Declined too, and it is the case a bare zero-check would have let through.

## Why the arguments and not the artist

The same reason `stackplot`'s patch reads its series from the call: the drawn artist is a closed outline, running forward along the curve and back along the baseline with the endpoints repeated, so recovering the series from it means undoing the closure first.

`fill_betweenx` is the same chart turned over — the shared axis is `y` and the magnitude runs along `x` — which is why the arguments are read by parameter name rather than by position, since `x1` sits where `y1` does.

## Tests

Ten cases. Each mechanism reverted, with a named test failing:

| reverted | fails |
| --- | --- |
| baseline check | `a_band_between_two_curves_is_declined`, `a_constant_second_edge_is_declined_too`, `a_line_keeps_its_band_unannounced_rather_than_misannounced` |
| position/value length check | `a_scalar_curve_is_declined` |

**One thing worth flagging, because I wrote the test's rationale wrong first.** `test_a_stackplot_still_reads_as_a_stack` originally claimed the recursion guard was what stopped a stacked chart gaining an area layer per band. Removing the guard fails nothing: `stackplot` draws every band with two explicit edges (`fill_between(x, stack[i], stack[i + 1])`), so the baseline test declines them first and the guard never gets the chance. The test now asserts the outcome and says which mechanism actually fires — pinning it to the guard would have made it a description of matplotlib's `stackplot` rather than of this package. The guard stays as the convention every patch here follows.

| | |
| --- | --- |
| Python 3.11 / matplotlib 3.10.9 | 1155 passed |
| Python 3.9 / matplotlib 3.9.4 | the new file's 10 passed |
| lint | clean |

Docs updated: the supported-plots list, and an `Area Plot` example that also states plainly which bands are *not* read and links to #339.

## What stays open on #339

The band-around-a-line case — `ax.plot(...)` plus `ax.fill_between(x, y-e, y+e)`, the usual confidence ribbon. That is an `error_bar` reading (`{x, y, yMin, yMax}`, with `y` from the line), not an area one, and matching a band to a line is a heuristic: the line may be drawn after the band, there may be several, and the x may not line up. It should fail toward declining rather than toward guessing an estimate, and that is worth settling before writing. Analysis is on the issue.

What this PR does guarantee in the meantime is that such a figure does not silently gain a *wrong* layer — it reads as the line it always did.

---

### DIFF
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -736,6 +736,38 @@ ax.legend(loc="upper left")
 plt.show() #<<
 ```
 
+`ax.fill_between()` reads as an area too, when it draws one: `fill_between(x,
+y)` fills from zero up to a curve, which is the same chart a single
+`stackplot` band draws. `ax.fill_betweenx()` is the same picture turned over
+and reads the same way.
+
+```{python}
+#| fig-alt: Filled area chart of monthly rainfall
+
+import matplotlib.pyplot as plt
+
+import maidr #<<
+
+months = list(range(1, 13))
+rainfall = [78, 62, 71, 55, 48, 39, 31, 35, 52, 74, 89, 84]
+
+fig, ax = plt.subplots(figsize=(8, 4))
+ax.fill_between(months, rainfall, alpha=0.6, label="Rainfall") #<<
+
+ax.set_title("Monthly Rainfall")
+ax.set_xlabel("Month")
+ax.set_ylabel("Rainfall (mm)")
+
+plt.show() #<<
+```
+
+A band drawn **between two curves** — `fill_between(x, lower, upper)`, the
+usual way to shade a confidence interval — is a different chart, and maidr
+does not read it as an area. Its content is the gap between the edges rather
+than the height of either, so announcing it as an area would report the upper
+edge as a magnitude and drop the lower one. Those charts render as a static
+image for now; see [issue #339](https://github.com/xability/py-maidr/issues/339).
+
 ### Point Plot
 
 `sns.pointplot()` is the same reading from the other direction: it estimates a
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -79,6 +79,8 @@ We currently support the following data visualization libraries in Python:, and 
 
 * Hexbin plot (`ax.hexbin()`) having one x (numerical) and one y (numerical) axis and a z axis carrying what the fill encodes — each hexagon's count by default, the reduction of `C` when one is given, or the count's interval when `bins` is a number. Read as a lattice of counted cells, one row at a time; because hexagons tessellate by offsetting alternate rows, each bin is announced by its centre rather than by a column index.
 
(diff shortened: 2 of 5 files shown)


jooyoungseo:
The `where=` gap is real and you put your finger on exactly why it mattered — it was the one call shape that slipped past every decline condition while being the same kind of wrong those conditions exist to prevent.

Measured:

```
fill_between(x, y)                     -> 1 path
fill_between(x, y, where=y > 0)        -> 3 paths     (8-point series)
fill_between(x, y, where=all True)     -> 1 path
fill_between(x, y, where=all False)    -> 0 paths
```

Three bands announced as one continuous series, with every position reported as filled.

**Declined, not described partially.** Emitting only the masked-in points was the tempting middle road and is worse: an area layer is one continuous series, so a reader walking left to right would cross a gap without being told there was one. Silence about the whole chart beats confidence about a chart with holes in it.

**Keyed on what the mask says, not on the argument being present.** A mask that holds everywhere draws the single band the default draws, so declining on `where is not None` would refuse a chart identical to one already accepted. That has its own test.

The two smaller ones are both taken:

- **Array of zeros.** You are right that it is the same chart, and the asymmetry was mine rather than inherent. `_baseline_is_zero` now tests whether the edge is *all* zero rather than whether it is a scalar zero, so `fill_between(x, y, np.zeros_like(x))` reads as the area it draws.
- **`numpy.bool_` vs `bool`.** Cast.

**The `plt.fill_between` entry point** now has a test, matching what `lineplot.py` does for its own — you were right that it is the same bound method, and right that the repo tests both anyway.

**The `instance` fallback** I have left. You are right that `wrapt.wrap_function_wrapper(Axes, ...)` makes it unreachable for a bound call, but an unbound `Axes.fill_between(ax, ...)` passes `instance=None`, and the fallback degrades that to "return without registering" rather than to an `AttributeError`. Exotic, but the branch earns its keep at that price, and it keeps the shape identical to `stackplot`'s.

Fourteen tests now, `where=` failing without the new guard. Full suite 1159 passed on 3.11, the new file's 14 on 3.9, lint clean.

---

Issue #373, opened 2026-08-14
A seaborn colour probe registers as an area layer


jooyoungseo:
`seaborn.utils._default_color` resolves a default colour by *drawing* a throwaway artist, reading its face colour, and removing it again. Every branch ends in `scout.remove()` — it is a probe, never a chart:

```python
elif method.__name__ == "fill_between":
    kws = normalize_kwargs(kws, mpl.collections.PolyCollection)
    scout = method([], [], **kws)
    facecolor = scout.get_facecolor()
    color = to_rgb(facecolor[0])
    scout.remove()
```

Since #339 taught maidr to read `fill_between` as an area chart, that probe registers a MAIDR layer. `Axes.fill_between` cannot tell a real fill from a two-empty-arrays colour lookup, and `_default_color` runs before any seaborn-level patch has set a recursion context — so nothing suppresses it.

## Reproduction

`boxenplot` is the plainest case: an ordinary root-binding call, no exotic import.

```python
import matplotlib; matplotlib.use("Agg")
import matplotlib.pyplot as plt, seaborn as sns, pandas as pd, maidr
from maidr.core.figure_manager import FigureManager

df = pd.DataFrame({"g": list("aabbccaabbcc"), "v": [1,2,3,4,5,6,2,3,4,5,6,7]})
fig, ax = plt.subplots()
sns.boxenplot(data=df, x="g", y="v", ax=ax)
print([p.type.value for p in FigureManager.figs[fig].plots])
```

```
['area', 'line', 'point', 'point', 'point']
```

The leading `area` is the probe. Its stack, confirmed by instrumenting `FigureManager.create_maidr`:

```
seaborn/categorical.py:1947  in boxenplot
seaborn/utils.py:136         in _default_color
maidr/patch/fillbetween.py:222 in fill_between
maidr/patch/fillbetween.py:208 in _fill
```

The layer describes a fill of two empty arrays. It has no elements on the chart, so it announces a region that is not drawn and highlights nothing — the same shape of defect as #369, where a colorbar registered as a heatmap.

## Why the fill_between patch cannot decline it on its own

The probe passes `([], [])`, and an empty baseline is all-zero by every test that does not special-case emptiness. Adding "decline empty input" to `fillbetween.py` would fix this one call and leave the general hazard: `_default_color` also probes `plot`, `scatter` and `bar`, and each is patched or could become patched. The probe is the thing that should be suppressed, not each artist it happens to draw.

## Proposed fix

[...]

Issue #344, opened 2026-08-10
feat: support seaborn figure-level functions (catplot, relplot, displot)


jooyoungseo:
## Description

py-maidr patches axes-level seaborn functions (`barplot`, `boxplot`, `scatterplot`, `lineplot`, `histplot`, `kdeplot`, `heatmap`, `regplot`, `violinplot`, `countplot`). The figure-level wrappers are not patched:

- `sns.catplot` — categorical plots across a facet grid
- `sns.relplot` — scatter/line across a facet grid
- `sns.displot` — distributions across a facet grid

## Why they do not come for free

These build a `FacetGrid` and dispatch to the axes-level functions internally. Two problems follow:

1. `ContextManager` exists to stop a patched function that calls another patched function from registering twice — so the inner call may be suppressed exactly when it is the one carrying the data.
2. Even where panels do register, each becomes an independent figure rather than one faceted figure. The facet variable and its level — the entire reason the user wrote `catplot` instead of a loop — are never announced.

## Why it is worth doing

`catplot` / `relplot` / `displot` are what the seaborn documentation steers users toward. Someone following the official tutorial has a good chance of landing on an unsupported path.

## Scope

- [ ] Patch the three figure-level functions and register the `FacetGrid` as one MAIDR figure
- [ ] Emit a subplot grid with per-panel titles naming the facet variable and level
- [ ] Populate the figure-level `axes` labels, since facet panels share axes drawn at the figure margins
- [ ] Verify `ContextManager` suppresses the inner axes-level registrations rather than the outer one

## Difficulty

Medium-to-high — the facet plumbing is the work, not the data extraction.

jooyoungseo:
Measured what the figure-level functions actually do today, and the answer is not "nothing" — it is worse than nothing, and there is a single root cause behind this issue and #342.

## What a grid emits now

```
pairplot (3 vars, 9 panels)   18 layers: bar, dodged_bar, bar, dodged_bar, bar, dodged_bar,
                                          point x 12
jointplot(kind="scatter")     point, dodged_bar, dodged_bar
jointplot(kind="hex")         hexbin, dodged_bar, dodged_bar
```

Two things are wrong and both are invisible to a reader as anything other than nonsense: **every panel registers twice**, and the histograms register as **`bar` and `dodged_bar`** rather than `hist`.

Called directly, all of these are already correct:

```
sns.scatterplot(ax=ax)   ['point']
sns.histplot(ax=ax)      ['hist']
```

[...]

PR #374, opened 2026-08-14
fix(seaborn): patch boxplot and violinplot at both names too

jooyoungseo:
Follow-up to #372, working the three points from its review. All three were verified before acting on them, and two of them turned out to be corrections of claims I made.

## 1. `boxplot` and `violinplot` were left on the re-export

> it'd be worth either converting these two to `wrap_seaborn` for consistency, or explicitly noting in a follow-up issue that they're a known-untouched case

Converted. But not for consistency — measured, the gap is a real reading:

```
seaborn.violinplot              violin_box, violin_kde
seaborn.categorical.violinplot  area, line          <-- before
seaborn.boxplot                 box
seaborn.categorical.boxplot     area, box           <-- before
```

A violin announced as a **line chart** — not a degraded violin, a different chart — plus a phantom `area` layer.

Where the other nine were a *grid* fix, these two are not: no seaborn grid reaches them. `catplot` drives `_CategoricalPlotter` directly and never calls the module-level function, which the swept call counts confirm:

| grid | defining-module bindings reached |
|---|---|
| `pairplot` | `histplot`, `scatterplot` |
| `jointplot` | `histplot`, `scatterplot` |
| `relplot` | `scatterplot` |
| `lmplot` | `regplot` |
| `catplot` (box, violin, bar) | — none — |
| `displot` | — none — |

What did take the unpatched binding is ordinary user code: `from seaborn.categorical import violinplot`. Uncommon, silent, and severe when it happens — and an arbitrary hole with nine of eleven closed.

The phantom `area` comes from `seaborn.utils._default_color`, which resolves a default colour by drawing a throwaway artist through `ax.fill_between` and removing it again. With no seaborn-level patch there was no recursion context to suppress it. This PR removes it **only for these two functions**. The probe still registers one wherever no seaborn-level patch runs — `sns.boxenplot` is the plainest case, an ordinary root-binding call — filed separately as #373.

## 2. The docstring named the wrong grids

> Worth trimming that sentence to just `pairplot`, `jointplot`, `relplot`, `lmplot` (the ones actually affected by the binding fix), and leaving the catplot/displot caveat where the test file already states it precisely.

Correct, and the same error was in two places — `wrap_seaborn`'s docstring and the test module's — so both are fixed, with the measured table above recorded in the test file.

## 3. Every early return now warns

> A `warnings.warn(...)` in these branches would make that failure mode visible instead of only discoverable by a user noticing bad output

All four early returns share one consequence, and it is not a visible one: the function goes on being drawn by an unpatched path and the panel is read as the artists it drew — a histogram announced as a grouped bar chart. Nothing downstream can detect that, and no user can trace it back to here. One helper, `_warn_root_binding_only`, phrases the consequence; the "seaborn no longer exports it" case warns separately because it is worse (neither binding wrapped).

None of them fires on any seaborn MAIDR has been measured against. They exist for the release that moves a function or renames a module.

## Also correcting myself

I said on #344 that `boxplot` and `violinplot` were matplotlib-level only. That was wrong. Both are seaborn functions with exactly the same single-binding gap:

```
seaborn.boxplot      PATCHED   seaborn.categorical.boxplot      plain
seaborn.violinplot   PATCHED   seaborn.categorical.violinplot   plain
```

## Falsification

Reverting the two bindings to `wrapt.wrap_function_wrapper("seaborn", ...)` fails **4** assertions: `test_both_names_are_wrapped` for each of the two, and `test_a_categorical_plot_reads_the_same_from_either_binding` for each.

## Verification

`ruff check` clean on the changed files. Full suite: **1177 passed**.

---

### DIFF
--- a/maidr/patch/boxplot.py
+++ b/maidr/patch/boxplot.py
@@ -6,7 +6,7 @@ from matplotlib.axes import Axes
 from maidr.core.context_manager import BoxplotContextManager, ContextManager
 from maidr.core.enum import PlotType
 from maidr.core.figure_manager import FigureManager
-from maidr.patch.common import _draw_quietly, resolve_orientation
+from maidr.patch.common import _draw_quietly, resolve_orientation, wrap_seaborn
 
 
 @wrapt.patch_function_wrapper(Axes, "bxp")
@@ -35,7 +35,6 @@ def mpl_box(wrapped, _, args, kwargs) -> dict:
     return plot
 
 
-@wrapt.patch_function_wrapper("seaborn", "boxplot")
 def sns_box(wrapped, _, args, kwargs) -> Axes:
     # Set the internal context to avoid cyclic processing.
     with BoxplotContextManager.set_internal_context() as bxp_context:
@@ -57,6 +56,10 @@ def sns_box(wrapped, _, args, kwargs) -> Axes:
 
     # Return to the caller.
     return plot
+
+
+# Patch seaborn function.
+wrap_seaborn("boxplot", sns_box)
 
 
 def sns_infer_new_orient(wrapped, instance, args, kwargs) -> str:
--- a/maidr/patch/common.py
+++ b/maidr/patch/common.py
@@ -215,6 +215,40 @@ def common(
     return plot
 
 
+def _warn_root_binding_only(name: str, reason: str) -> None:
+    """
+    Say so when only the re-export could be wrapped.
+
+    Every early return in ``wrap_seaborn`` below has the same consequence and
+    it is not a visible one: seaborn's own grids call the defining module's
+    binding, so the function goes on being drawn by an unpatched code path
+    and the panel is read by the matplotlib-level patches alone. That is a
+    wrong reading rather than a missing one -- a histogram announced as a
+    grouped bar chart -- which nothing downstream can detect and no user can
+    trace back to here.
+
+    A warning is the only thing that makes it findable. None of these
+    branches fire against any seaborn MAIDR has been measured on; they exist
+    for the release that moves a function or renames a module.
+
+    Parameters
+    ----------
+    name : str
+        The function that could only be wrapped at ``seaborn.<name>``.
+    reason : str
+        What stopped the defining module from being reached, phrased to
+        complete "because ...".
+    """
+    warnings.warn(
+        f"maidr: patched seaborn.{name} but not the module that defines it, "
+        f"because {reason}. seaborn's figure-level grids -- pairplot, "
+        f"jointplot, relplot, lmplot -- call the defining module's binding, "
+        f"so a {name} panel inside one of them will be read as the artists it "
+        f"drew rather than as a {name}.",
+        stacklevel=3,
+    )
+
+
 def wrap_seaborn(name: str, wrapper: Callable) -> None:
     """
     Wrap a seaborn function at both of the names it answers to.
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Thanks — took two of the three nits.

**"that helper's body is untested by construction."** Fixed rather than accepted. `_warn_root_binding_only` is now driven directly in `test_the_root_binding_warning_says_what_breaks`, which asserts the *consequence* rather than the wording: it names the function, the reason handed in, and the four grids that take the defining-module binding — and asserts `catplot` and `displot` are **not** named. That last assertion is the one that matters, since naming them wrongly is the exact mistake this PR is correcting elsewhere. A message meant to be read once, years from now, by someone with no idea what a binding is, is worth pinning.

**The 3-line wrap.** Fits on one line; joined.

**The bare `UserWarning`.** Left alone, per your read. Worth recording why beyond "consistent precedent": a dedicated subclass buys independent filterability, and the case for it is that a user who has silenced the "figure fell back to a static image" warning would also silence this one. But these fire at import time and mean *maidr is now mis-reading your charts* — which is not something to make convenient to filter. If they ever do warrant a subclass it should be one that groups "a patch did not install", not one per call site.

Full suite after the change: **1178 passed**.

Merging once CI is green.

---

jooyoungseo:
Both findings are right, and the second one is a nice catch — it's the same mistake this PR was written to correct, reintroduced one layer down.

## 1. `defining == "seaborn"` was warning about the opposite of what happened

Correct, and it's backwards rather than merely noisy: if `__module__` is `seaborn`, then `seaborn.<name>` **is** the defining binding and `wrapt.wrap_function_wrapper(seaborn, name, wrapper)` one line above has done the whole job. The warning would send its reader looking for a module that is already patched.

Split out, and it now returns quietly. The branch is no longer unreachable either — a stand-in with `__module__ = "seaborn"` is installed on the seaborn module (removed again by `monkeypatch`), `wrap_seaborn` is run against it under `simplefilter("error")`, and the wrap is asserted. Coverage pragma dropped.

## 2. The message named grids that don't apply to the function that triggered it

> it would point users at grids that were never the relevant failure mode for these two functions, while omitting the one that actually matters

Exactly so, and worth naming what happened: the docstring said `catplot`/`displot` were affected when they aren't, this PR corrected that, and the warning I added in the same PR then hardcoded a *different* wrong grid list. Which grids are exposed genuinely varies per function — `pairplot`/`jointplot`/`relplot`/`lmplot` for the original nine, none at all for `boxplot`/`violinplot` — so any fixed list is wrong for someone.

Taken your second option: describe the risk, not the callers.

> Both names reach the same function and only the patched one is read as a violinplot: anything calling the other — one of seaborn's own figure-level grids, or a direct import from the defining module — gets a panel read by the matplotlib-level patches alone, and announced as the artists it drew.

`test_the_partial_patch_warning_names_no_grids` asserts both routes appear and that **no** grid is named. The docstring records why, so the next person to make it more specific has to argue with the reason rather than rediscover it.

## 3. The `original is None` branch

Pulled into `_warn_unpatched` and tested the same way, by calling `wrap_seaborn` on a name seaborn has never had. Also loses its pragma.

## Falsification

| reverted | result |
|---|---|
| root-defined branch warns again | 1 failed, 20 passed |
| missing-export warning removed | 1 failed, 20 passed |

Full suite: **1180 passed**.

---

PR #375, opened 2026-08-14
fix(seaborn): a colour probe is not a chart

### LINKED ISSUE #373: A seaborn colour probe registers as an area layer

jooyoungseo:
Closes #373.

`seaborn.utils._default_color` resolves a default colour by *drawing* a throwaway artist, reading its face colour, and removing it again. Every branch ends in `scout.remove()`:

```python
elif method.__name__ == "fill_between":
    kws = normalize_kwargs(kws, mpl.collections.PolyCollection)
    scout = method([], [], **kws)
    facecolor = scout.get_facecolor()
    color = to_rgb(facecolor[0])
    scout.remove()
```

It is a probe, never a chart. But it draws through `Axes.fill_between`, `Axes.plot`, `Axes.scatter` and `Axes.bar` — all patched — and it runs *before* any seaborn-level patch has set a recursion context, so nothing suppressed it. Since #339 taught MAIDR to read `fill_between` as an area chart, the probe started registering a layer describing a fill of two empty arrays.

## Measured, before → after

| | before | after |
|---|---|---|
| `rugplot` | **`ExtractionError`** | renders |
| `ecdfplot` | `line` | `step` |
| `stripplot` (3 groups) | 4 layers | 3 |
| `boxenplot` | `area, line, point, point, point` | `line, point, point, point` |

**`rugplot` is the #369 shape exactly.** Its ticks are a `LineCollection` MAIDR does not read, so the probe's layer was the *only* one registered — and reading its data raised `ExtractionError`, which is fatal to the whole render rather than to its own layer. A scatter with a rug over it, which is how `rugplot` is actually used, produced no HTML at all:

```
                                before                      after
scatter + rugplot overlay       ExtractionError             OK (19573 bytes)
rugplot alone                   ExtractionError             KeyError: no MAIDR found
```

Standalone `rugplot` is still not a supported chart, and now says so through the ordinary "no MAIDR found for figure" path instead of a data-extraction crash.

**`ecdfplot` is the one worth reading twice.** The layer count was already right and the numbers were already right — same `x: -Infinity, y: 0.0` first point before and after. The probe simply registered first, and its `ax.plot([], [])` carried no `drawstyle`, so the shared line pass settled on `line` for a curve drawn `steps-post`. An ECDF announced as a line chart, with nothing in the output to catch.

## Why suppress the function rather than teach `fill_between` to decline

`fill_between` could be taught to refuse `([], [])`. That fixes one branch and leaves the hazard: the probe also drives `plot`, `scatter` and `bar`, and the next patch added to any of them reintroduces this somewhere new. The probe is the thing that should be quiet, not each artist it happens to draw. Same reasoning as #370, which suppressed `Colorbar._draw_all` rather than testing "is this axes a colorbar".

## Why a sweep rather than `__module__`

`_default_color` is not a re-export like the plotting functions `wrap_seaborn` handles — it is a private helper pulled into three modules by name:

```
_default_color   __module__=seaborn.utils
                 bindings=['seaborn.categorical', 'seaborn.distributions',
                           'seaborn.relational', 'seaborn.utils']
```

So `__module__` names one binding out of four and the call sites take the other three. Sweeping the loaded `seaborn` modules covers every binding without a hard-coded table, and stays correct if seaborn switches to a qualified `utils._default_color(...)` call, since `seaborn.utils` is in the sweep too. The identity check is what makes it safe: a name that no longer resolves to the same object is left alone rather than wrapped by guess.

`test_no_seaborn_module_holds_an_unwrapped_probe` asserts the sweep left nothing behind, which is the check that fails if seaborn adds a fourth importer — a hard-coded table would not notice, and nothing downstream would report it.

## Falsification

Reverting the patch fails **9 of the 10** new assertions. The one that survives is `test_a_real_fill_between_is_untouched` — the control, asserting a fill a user actually asked for still reads as an area. Without it, "no phantom areas" would be satisfiable by reading no areas at all.

## Verification

`ruff check` clean. Full suite: **1190 passed**.

---

### DIFF
--- a/maidr/patch/__init__.py
+++ b/maidr/patch/__init__.py
@@ -15,10 +15,10 @@ from . import (  # noqa: F401
     pieplot,
     pointplot,
     scatterplot,
-    seaborn_probe,
     regplot,
     kdeplot,
     candlestick,
     mplfinance,
     violinplot,
+    seaborn_probe,
 )
--- a/maidr/patch/__init__.py
+++ b/maidr/patch/__init__.py
@@ -15,6 +15,7 @@ from . import (  # noqa: F401
     pieplot,
     pointplot,
     scatterplot,
+    seaborn_probe,
     regplot,
     kdeplot,
     candlestick,
--- a/maidr/patch/seaborn_probe.py
+++ b/maidr/patch/seaborn_probe.py
@@ -1,0 +1,113 @@
+"""Keep seaborn's colour probe from registering as a chart.
+
+``seaborn.utils._default_color`` resolves a default colour by *drawing* a
+throwaway artist, reading its face colour, and removing it again. Every
+branch ends in ``scout.remove()``::
+
+    elif method.__name__ == "fill_between":
+        kws = normalize_kwargs(kws, mpl.collections.PolyCollection)
+        scout = method([], [], **kws)
+        facecolor = scout.get_facecolor()
+        color = to_rgb(facecolor[0])
+        scout.remove()
+
+It is a probe, never a chart. But it draws through ``Axes.fill_between``,
+``Axes.plot``, ``Axes.scatter`` and ``Axes.bar`` -- all of which MAIDR
+patches -- and it runs *before* any seaborn-level patch has set a recursion
+context, so nothing suppressed it (#373).
+"""
+
+from __future__ import annotations
+
+import sys
+import warnings
+from typing import Any
+
+import wrapt
+
+from maidr.core.context_manager import ContextManager
+
+
+def _suppress_registration(wrapped, _, args, kwargs) -> Any:
+    """
+    Resolve a default colour without registering the artist it drew.
+
+    A layer from the probe describes a fill of two empty arrays. It has no
+    elements on the chart, so it announces a region that is not drawn and
+    highlights nothing -- the same shape of defect as #369, where a colorbar
+    registered as a second heatmap.
+
+    Suppressing the whole function rather than teaching each patch to decline
+    an empty call is the honest scope. ``fill_between`` could be taught to
+    refuse ``([], [])``, and that would fix the one branch and leave the
+    hazard: the probe also drives ``plot``, ``scatter`` and ``bar``, and the
+    next patch added to any of them reintroduces it somewhere new. The probe
+    is the thing that should be quiet, not each artist it happens to draw.
+
+    Parameters
+    ----------
+    wrapped : Callable
+        The original ``seaborn.utils._default_color``.
+    _ : Any
+        Unused -- the function is module-level, so wrapt binds nothing.
+    args : tuple
+        Positional arguments seaborn passed.
+    kwargs : dict
+        Keyword arguments seaborn passed.
+
+    Returns
+    -------
+    Any
+        The resolved colour, unchanged.
+    """
+    with ContextManager.set_internal_context():
+        return wrapped(*args, **kwargs)
+
+
+def _patch_default_color() -> None:
+    """
+    Wrap the probe at every seaborn module that holds a reference to it.
+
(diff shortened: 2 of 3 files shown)


jooyoungseo:
All three taken.

**1. Import placement.** Moved to the end of the list.

**2. The `sys.modules` snapshot assumption.** The sharpest of the three, and worth answering with more than a comment: the failure mode you describe — a lazily-importing seaborn leaving a binding unwrapped, with no error and a reintroduced phantom layer — would already be caught, just not by the test you'd expect. `test_the_known_call_sites_are_covered` indexes `sys.modules[module_name]` by name, so the moment `seaborn.categorical` is no longer loaded at import time it raises `KeyError` rather than passing. The sweep test would go quiet; its sibling would go red.

That was accidental rather than designed, so the docstring now states the assumption *and* names where the tripwire lives, so the next person to touch either test knows which one is load-bearing.

**3. The `pragma: no cover` warn branch.** Fixed rather than accepted, the same way #374's two branches were: `monkeypatch.delattr(seaborn.utils, "_default_color")`, then run the sweep and assert the warning. Re-running `_patch_default_color` is safe there precisely because the attribute is gone — it returns after warning, so nothing is double-wrapped, and `monkeypatch` puts the probe back. Pragma dropped.

Full suite: **1191 passed**.

Merging once CI is green.

---

PR #377, opened 2026-08-14
fix(bar): ask per axes which bar layer supersedes which

### LINKED ISSUE #376: A stacked bar anywhere in a figure deletes layers from every panel
`Maidr.plot_type` is a figure-wide "highest priority type seen so far", and `PLOT_TYPE_PRIORITY` gives `STACKED`/`DODGED` a 2 while every other type gets a 1. So one stacked bar anywhere in a figure sets it for the whole figure. `_flatten_maidr` then does this:

```python
# Handle DODGED/STACKED plots: only keep one plot per subplot position
# because GroupedBarPlot extracts all containers from the axes itself
if self.plot_type in (PlotType.DODGED, PlotType.STACKED):
    self._plots = self._merge_plots_by_subplot_position()
```

and the merge keeps `position_plots[0]` — the first plot registered at each position, whatever its type.

The intent is right: `GroupedBarPlot` reads every `BarContainer` on its axes, so the N registrations from N `ax.bar()` calls would otherwise emit N identical layers. But the condition asks a figure-wide question to solve a per-axes problem, and the resolution keeps an arbitrary member rather than the segmented one.

## 1. Layers vanish from panels that have no bar in them

```python
fig, axes = plt.subplots(1, 2)
axes[1].plot(cats, s0)
axes[1].scatter(cats, s1)
```

| panel 0 contains | panel 0 emits | **panel 1 emits** |
|---|---|---|
| `plot` + `scatter` | `['line', 'point']` | `['line', 'point']` |
| a stacked bar | `['stacked_bar']` | **`['line']`** |

Panel 1 is untouched between the two runs. Adding bars to panel 0 deletes its scatter layer. Nothing errors, and the remaining layer is correct — a reader is simply told the panel has one layer when it has two.

## 2. Which layer survives is registration order, not type

```python
ax.plot(cats, s0 + s1, label="total")          # registered first
ax.bar(cats, s0, bottom=np.zeros(3), label="s0")
ax.bar(cats, s1, bottom=s0, label="s1")
```

```
registered=['line', 'stacked_bar', 'stacked_bar']   emitted=['line']
```

The stacked bar chart — the thing the figure is *of* — is dropped, and the annotation line is all that is announced. Swap the two statements and you get `['stacked_bar']` with the line dropped instead.

## 3. matplotlib's own documented stacked bar renders nothing at all

This is the same bug reaching its worst outcome. The [matplotlib stacked-bar example](https://matplotlib.org/stable/gallery/lines_bars_and_markers/bar_stacked.html) omits `bottom` on the first call, as everyone does:

```python
ax.bar(labels, men_means, width, label='Men')
ax.bar(labels, women_means, width, bottom=men_means, label='Women')
```

```
registered=['bar', 'stacked_bar']   ->  ExtractionError
```

The first call registers a `BarPlot`; the second lifts the figure to `STACKED`, so the merge fires and keeps `position_plots[0]` — the `BarPlot`. `BarPlot._extract_bar_container_data` then flattens *every* container on the axes, finds 6 patches against 3 tick labels, and returns `None`. `ExtractionError` is fatal to the whole render rather than to its own layer, so the figure produces no HTML.

Passing `bottom=np.zeros(3)` on the first call makes it work, which is why this has survived: every test writes it that way.

```
stack with bottom=0 on the 1st       registered=['stacked_bar', 'stacked_bar']   OK  ['stacked_bar']
stack written the natural way        registered=['bar', 'stacked_bar']           ExtractionError
```

## Proposed fix

Decide the merge from the plots actually at each position rather than from a figure-wide type, and keep the segmented layer rather than the first:

- collapse a position's bar-family layers only when that position holds a `STACKED`/`DODGED` layer;
- keep the segmented layer, since it is the one whose extractor reads the whole axes;
- leave layers of other types alone — a line over a stacked bar is a second layer, not a duplicate.

That fixes all three cases above without needing `BarPlot` to change.

## Separate, and left out deliberately

Two plain `ax.bar()` calls with no `bottom` at all still raise `ExtractionError`, because `BarPlot` reads every container on the axes rather than the ones its own call drew. The merge above does not fire there (no segmented layer is ever registered), and what two overlapping bar layers *should* be announced as is a genuine question rather than an oversight. Worth its own issue.

jooyoungseo:
Closes #376.

`BarPlot` and `GroupedBarPlot` both read *every* `BarContainer` on their axes rather than the bars their own call drew, so a stacked chart built from three `ax.bar()` calls registers three layers that each describe the whole chart. One has to survive and the rest are duplicates — that much was already handled.

How the survivor was chosen was not. `Maidr.plot_type` is a **figure-wide** "highest priority type seen", and `PLOT_TYPE_PRIORITY` gives `STACKED`/`DODGED` a 2 against everything else's 1, so one stacked bar anywhere set it for the whole figure. `_flatten_maidr` then collapsed **every** subplot position down to `position_plots[0]` — the first layer registered there, whatever its type.

## Three things followed, and none of them errored

**A panel with no bars in it lost a layer.** Panel 1 is written identically in both rows below; the only difference is what panel 0 contains:

| panel 0 contains | panel 0 emits | **panel 1 emits** |
|---|---|---|
| `plot` + `scatter` | `['line', 'point']` | `['line', 'point']` |
| a stacked bar | `['stacked_bar']` | **`['line']`** |

**Which layer survived was registration order, not type.**

```
registered=['line', 'stacked_bar', 'stacked_bar']   emitted=['line']
```

The bar chart — the thing the figure is *of* — dropped, the annotation line kept. Drawing the line second dropped the line instead.

**matplotlib's own documented stacked bar rendered nothing at all.** The [gallery example](https://matplotlib.org/stable/gallery/lines_bars_and_markers/bar_stacked.html) omits `bottom` on the first call, as everyone does:

```python
ax.bar(labels, men_means, width, label='Men')
ax.bar(labels, women_means, width, bottom=men_means, label='Women')
```

That registers `BAR` then `STACKED`; the collapse fired and kept the `BAR`, whose extractor found six patches against three tick labels and raised `ExtractionError` — fatal to the whole figure. Writing `bottom=np.zeros(n)` on the first call avoided it, which is why every existing test wrote it that way and why this survived.

## The fix

The question is asked **per position** and answered **by type**: a position holding a segmented layer keeps the first of those and drops the other axes-wide bar layers there. Layers of any other type are left alone — a line over a stacked bar is a second layer, not a duplicate of it. `HIST` draws bars but reads its own container, so it is not in the family and is never dropped.

### Two details are load-bearing

**It runs before the artists are collected**, not only in `_flatten_maidr`. Reading `plot.elements` renders the layer, and the superseded `BarPlot` raised *there* — one step before the schema was ever built:

```
maidr/core/maidr.py:315  in <listcomp>          <- tagged_elements
maidr/core/plot/maidr_plot.py:295  in elements
maidr/core/plot/barplot.py:72  in _extract_plot_data
    raise ExtractionError(self.type, plot)
```

Collapsing only in `_flatten_maidr` would never have been reached. It also keeps one artist from appearing twice in that list, where `HighlightContextManager` resolves an artist by its **first** index and would hand every bar the dropped layer's selector id.

**`selector_ids` is filtered in step with `_plots`.** The two are paired by index in both directions — the artists are tagged with `selector_ids[i]`, and the schema stamps the same index into the layer's selector string. Dropping a layer without dropping its id shifts every id after it, so each surviving layer carries its neighbour's and the highlight lands on the wrong bar, with nothing raised. It matters here specifically because the dropped layer is often at index 0.

## Falsification

| reverted | failing tests |
|---|---|
| old merge restored (figure-wide gate + `position_plots[0]`) | 5 of 7 |
| keep `group[0]` instead of `segmented[0]` | 3 of 7 |
| `selector_ids` left unfiltered | 1 of 7 |

The two that survive the first revert are the control (the `bottom=np.zeros(n)` spelling that always worked) and the idempotence test.

## Left out deliberately

Two plain `ax.bar()` calls with no `bottom` at all still raise `ExtractionError`, because `BarPlot` reads the whole axes and no segmented layer is ever registered for this to act on. What two *overlapping* bar layers should be announced as is a genuine question rather than an oversight; noted on #376 for its own issue.

## Verification

`ruff check` clean. Full suite: **1198 passed**.

---

### DIFF
--- a/maidr/core/figure_manager.py
+++ b/maidr/core/figure_manager.py
@@ -25,8 +25,6 @@ class FigureManager:
     figs : dict
         A dictionary that maps matplotlib Figure objects to their corresponding
         Maidr instances.
-    PLOT_TYPE_PRIORITY : dict
-        Defines the priority order for plot types. Higher numbers take precedence.
 
     Methods
     -------
@@ -40,29 +38,6 @@ class FigureManager:
     """
 
     figs = {}
-
-    # Define plot type priority order (higher numbers take precedence)
-    PLOT_TYPE_PRIORITY = {
-        PlotType.BAR: 1,
-        PlotType.STACKED: 2,
-        PlotType.DODGED: 2,  # DODGED and STACKED have same priority
-        PlotType.LINE: 1,
-        PlotType.STEP: 1,
-        PlotType.SCATTER: 1,
-        PlotType.HIST: 1,
-        PlotType.BOX: 1,
-        PlotType.HEAT: 1,
-        PlotType.HEXBIN: 1,
-        PlotType.COUNT: 1,
-        PlotType.PIE: 1,
-        PlotType.SMOOTH: 1,
-        PlotType.CANDLESTICK: 1,
-        PlotType.VIOLIN_KDE: 1,
-        PlotType.VIOLIN_BOX: 1,
-        PlotType.ERRORBAR: 1,
-        PlotType.AREA: 1,
-        PlotType.STACKED_AREA: 2,
-    }
 
     _instance = None
     _lock = threading.Lock()
@@ -103,15 +78,22 @@ class FigureManager:
         """
         Retrieve or create a Maidr instance for the given Figure.
 
-        If a Maidr instance already exists for the figure, update its plot type
-        if the new plot type has higher priority (DODGED/STACKED > BAR).
+        A figure that already has one is returned as it is. It used to have
+        its ``plot_type`` raised here whenever a layer of higher priority was
+        registered -- DODGED and STACKED outranking everything else -- so that
+        ``_flatten_maidr`` could consult one figure-wide answer to decide
+        whether bar layers should be collapsed. That is what made a stacked
+        bar in one panel delete layers from every other panel, and the
+        question is asked per position now, so the table and the update it
+        served are both gone (#376).
 
         Parameters
         ----------
         fig : Figure
             The matplotlib figure to get or create a Maidr instance for.
         plot_type : PlotType
-            The plot type to set or update for the Maidr instance.
+            The type of the layer being registered. Used only when the figure
+            is new, to record what it started as.
 
         Returns
         -------
@@ -120,14 +102,6 @@ class FigureManager:
         """
         if fig not in cls.figs.keys():
             cls.figs[fig] = Maidr(fig, plot_type)
-        else:
-            # Update plot type if the new type has higher priority
-            maidr = cls.figs[fig]
-            current_priority = cls.PLOT_TYPE_PRIORITY.get(maidr.plot_type, 0)
-            new_priority = cls.PLOT_TYPE_PRIORITY.get(plot_type, 0)
-
-            if new_priority > current_priority:
-                maidr.plot_type = plot_type
         return cls.figs[fig]
 
     @classmethod
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -45,7 +45,7 @@ _AXES_WIDE_BAR_TYPES = frozenset(
 )
 
 #: The subset of the above that describes a whole segmented chart, and so is
-#: the layer to keep when a position holds more than one of the family.
+#: the layer to keep when one axes holds more than one of the family.
 _SEGMENTED_BAR_TYPES = frozenset({PlotType.DODGED, PlotType.STACKED})
 
 
@@ -403,9 +403,25 @@ class Maidr:
         )
 
     @staticmethod
-    def _subplot_position(plot: MaidrPlot) -> tuple[int, int]:
-        """The ``(row, col)`` cell a layer belongs to, defaulting to ``(0, 0)``."""
-        return (getattr(plot, "row_index", 0), getattr(plot, "col_index", 0))
+    def _layer_axes_key(plot: MaidrPlot) -> Any:
+        """
+        What decides whether two layers can supersede one another.
+
+        The axes itself, not the grid cell it sits in. The duplication this
+        resolves comes from the extractor reading every container on
+        ``self.ax``, so ``self.ax`` is the thing that makes two layers
+        descriptions of the same bars -- and ``ax.twinx()`` gives a *second*
+        axes at the *same* ``(row, col)``. Keyed by position, a stacked bar on
+        each side of a twinned pair collapsed to one layer and the right-hand
+        chart was announced nowhere.
+
+        Falls back to the grid cell for a layer that carries no axes, which is
+        no coarser than the position-keyed grouping this replaced.
+        """
+        ax = getattr(plot, "ax", None)
+        if ax is not None:
+            return id(ax)
+        return ("position", getattr(plot, "row_index", 0), getattr(plot, "col_index", 0))
 
     def _collapse_segmented_bar_layers(self) -> None:
         """
@@ -436,10 +452,11 @@ class Maidr:
           on the first call avoided it, which is why every test wrote it that
           way.
 
-        So the question is asked per position and answered by type: a position
+        So the question is asked per *axes* and answered by type: an axes
         holding a segmented layer keeps the first of those and drops the other
-        axes-wide bar layers there. Layers of any other type are left alone --
+        axes-wide bar layers on it. Layers of any other type are left alone --
         a line over a stacked bar is a second layer, not a duplicate of it.
+        See ``_layer_axes_key`` for why the axes and not the grid cell.
 
         Idempotent, and it has to be: it runs from ``_create_html_tag`` before
         the elements are collected *and* from ``_flatten_maidr``, which can be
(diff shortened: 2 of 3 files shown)


jooyoungseo:
All three verified before acting. Two land here; two go to a follow-up, with reproductions.

## 3. `plot_type` / `PLOT_TYPE_PRIORITY` are dead — done here

Correct, and it is dead code *this PR created*, so leaving it would be a trap of my own making. `PLOT_TYPE_PRIORITY` and the priority-raising in `_get_maidr` are gone.

The attribute itself stays, and the constructor parameter with it: `Maidr` is exported from `maidr/__init__.py`, so the signature is not mine to narrow on a bug fix. It now records what the figure started as, and says in its own docstring that it is descriptive and what consulting it cost.

## Nit — done here

`superseded` holds the layers themselves. You are right that `MaidrPlot` defines neither `__eq__` nor `__hash__`, so set membership is already identity — which is precisely what this wants, since two layers of the same type over the same bars are still two objects and only one is dropped. Noted in a comment so it does not read as an accident.

## 1 and 2 — verified, filed as #378

Both real, and worse than the review's phrasing suggested.

**The figure-wide gate** does not merely delete a layer, it deletes a *panel*:

```
registered: ['point', 'smooth', 'line']     # regplot in panel 0, ax.plot in panel 1
emitted   : [[['point', 'smooth']]]
```

The grid collapses from 1×2 to 1×1, because the surviving plots no longer carry the second column index. A regression line in one panel deletes an unrelated line chart in another *and* the panel it lived in.

**The id desync** only bites when the dropped layer is not last, which is why it has gone unnoticed. Constructing that case:

```
registered: ['line', 'point', 'smooth', 'smooth']    # line drawn first, and dropped
kept      : ['point', 'smooth', 'smooth']
ids issued: ['8eaae17d', '98d9751c', '9e1d623b', 'c06b6e45']
ids now   : ['8eaae17d', '98d9751c', '9e1d623b', 'c06b6e45']
each kept layer keeps its own id: False
```

Every survivor inherits its predecessor's id, so the outline lands one layer off.

Not folding them in here, though the temptation is real given they sit three lines below. They are a different family (smooth/line, not bar), they are pre-existing rather than introduced, and each deserves its own falsification table rather than riding in under a bar fix — which is the same reasoning that kept the `get_x_transformation` consolidation out of xability/r-maidr#160. #378 has both reproductions and proposes folding all three rules into the one pass this PR added, so there is one filter and one place the pairing invariant is stated.

Full suite after this round: **1198 passed**.

---

jooyoungseo:
The `twinx` edge case was worth more than a regression test — measured, it wasn't merely uncovered, it lost a chart.

```
registered  4 stacked_bar across 2 distinct axes, all at (0, 0)
emitted     1 layer, the left-hand one
```

`ax.twinx()` gives a **second axes at the same `(row, col)`**, so a stacked bar on each side of a twinned pair collapsed to one and the right-hand chart was announced nowhere.

So the grid cell was never the right key, and I'd picked it by inheriting it from the code I was replacing. The duplication this collapse resolves comes from the extractor reading every container on `self.ax` — which makes `self.ax` the thing that makes two layers descriptions of the same bars. Two axes in one cell are two charts. Keyed by axes, both survive with their own data:

```
emitted     ['stacked_bar', 'stacked_bar']
first y     10.0  (left)   60.0  (right)
```

It falls back to the grid cell for a layer carrying no axes, which is no coarser than the grouping it replaces. `test_a_twinned_axes_keeps_its_own_stacked_bar` asserts both layers by their **data**, not only by count, since two layers of the right type could still both be the left-hand chart. Reverting the key to the grid cell fails it.

Thank you for flagging it — "almost certainly the right call" turned out to be the one place it wasn't.

Your second observation is #378, filed with both reproductions.

Full suite: **1199 passed**.

---

jooyoungseo:
All three taken, and the `MplfinanceBarPlot` one was right in a way that matters.

## The type-based check was the wrong premise

Verified: `MplfinanceBarPlot._extract_plot_data` reads `self._custom_patches` — the volume patches handed to it — and only falls through to a container sweep when it has none. So the premise that justifies dropping a layer, *its extractor already describes every bar on this axes*, is not true of it, while `PlotType.BAR` said it was.

The premise is a property of the **extractor**, so it is now asked of the class:

```python
_AXES_WIDE_BAR_PLOTS = (BarPlot, GroupedBarPlot)
_SEGMENTED_BAR_PLOTS = (GroupedBarPlot,)
```

`MplfinanceBarPlot` and `HistPlot` subclass neither, so neither is eligible to be dropped — which is what I wanted the type sets to mean and not what they said. Thank you for catching it; the combination is unlikely enough that it would have sat there.

## The unreachable fallback

Correct — `MaidrPlot.__init__` takes a required `Axes` and assigns it, so every layer has one. Dropped rather than commented: a branch defending against a case that cannot arise is a claim that it can, and it would have gone on being carried and never exercised.

## The 89-char line

That was the fallback's `return`, so it went with it. Checked the rest: every line this PR adds is within 88, and the file's other over-length lines are pre-existing.

Full suite: **1199 passed**.

---

jooyoungseo:
Both taken.

## The missing regression test

You put your finger on exactly the right risk: *"someone 'simplifying' `_AXES_WIDE_BAR_PLOTS`/`_SEGMENTED_BAR_PLOTS` back to a `plot.type` check"* — which reads cleaner and is wrong, and which I would have been the one to do.

`test_a_bar_typed_layer_outside_the_family_is_not_dropped` drives the classification directly rather than standing up an mplfinance figure to reach it, since what is at risk is the classification and not the chart. It also asserts the duplicate stacked registration is *still* dropped, so it cannot pass by the collapse having quietly stopped working. Adding `MplfinanceBarPlot` to the family fails it:

```
falsify (mplfinance in the family): 1 failed, 8 passed
```

## The `id()` wrapper

Right, and it was worse than redundant — it made the key look like it needed explaining. `by_axes` is keyed by `plot.ax` now, with the reason in the docstring: `Axes` overrides neither `__eq__` nor `__hash__`, so it already groups by identity.

Full suite: **1200 passed**.

---

PR #379, opened 2026-08-14
fix(smooth): ask per axes which curve supersedes which line

### LINKED ISSUE #378: A smooth anywhere in a figure deletes every line layer, and shifts the selector ids
Raised in review of #377, and verified. Two defects, the same two the bar-layer collapse had in #376, one function call below the code that just fixed them.

```python
self._collapse_segmented_bar_layers()          # fixed in #377

self._plots = deduplicate_smooth_and_line(self._plots)   # drops LINE plots
...
self._plots = deduped_plots                              # drops duplicate-gid SMOOTH
...
for i, plot in enumerate(self._plots):
    ... self.selector_ids[i]
```

## 1. The gate is figure-wide, and the deletion is not

```python
def deduplicate_smooth_and_line(plots):
    has_smooth = any(getattr(plot, "type", None) == PlotType.SMOOTH for plot in plots)
    if has_smooth:
        return [plot for plot in plots if getattr(plot, "type", None) != PlotType.LINE]
    return plots
```

`any(...)` over the whole figure, then a filter over the whole figure. The intent is right — `regplot` registers a `LINE` from the `ax.plot` it draws internally *and* a `SMOOTH` for the same curve, so one is a duplicate — but that is true per axes, not per figure.

```python
fig, axes = plt.subplots(1, 2)
sns.regplot(data=frame, x="a", y="b", ax=axes[0])
axes[1].plot([1, 2, 3], [3, 2, 1])
```

```
registered: ['point', 'smooth', 'line']
emitted   : [[['point', 'smooth']]]
```

Panel 1's line is gone, and with it panel 1 — the grid collapses from 1×2 to 1×1, because the column index no longer appears among the surviving plots. A regression line in one panel deletes an unrelated line chart in another, and the figure loses a panel.

## 2. `selector_ids` is not filtered with `_plots`

`_plots` and `selector_ids` are paired by index in both directions: `_create_html_tag` tags each layer's artists with `selector_ids[i]`, and `_flatten_maidr` stamps the same index into that layer's selector string. Both dedup steps filter one list and not the other.

It only bites when the dropped layer is not last, which is why it has gone unnoticed:

```python
fig, ax = plt.subplots()
ax.plot([1, 2, 3], [3, 2, 1])          # registered FIRST, and dropped
sns.regplot(data=frame, x="a", y="b", ax=ax)
```

```
registered: ['line', 'point', 'smooth', 'smooth']
kept      : ['point', 'smooth', 'smooth']
ids issued: ['8eaae17d', '98d9751c', '9e1d623b', 'c06b6e45']
ids now   : ['8eaae17d', '98d9751c', '9e1d623b', 'c06b6e45']
each kept layer keeps its own id: False
```

Every surviving layer inherits its predecessor's id, so the highlight lands on the layer before the one the cursor is on. Nothing errors and the announced numbers are all correct — only the outline moves.

## Proposed fix

The same shape as #377: ask per position, and filter the ids in step.

- scope the smooth/line rule to a position — a `SMOOTH` supersedes the `LINE` layers **on its own axes**, not everywhere;
- fold the duplicate-gid `SMOOTH` rule into the same pass;
- drop the ids alongside the layers, once, rather than filtering `_plots` three separate times.

`_collapse_segmented_bar_layers` in `maidr/core/maidr.py` already does exactly this for the bar family and can absorb both rules, which would also mean one filter instead of three and one place where the pairing invariant is stated.

jooyoungseo:
Closes #378. Raised in review of #377, and the same shape of defect one function call below the code that PR fixed.

`regplot` draws its fit through `ax.plot`, so one curve registers both a `LINE` and a `SMOOTH`, and the line is the duplicate. That much was right. The question was asked of the whole figure:

```python
has_smooth = any(plot.type == PlotType.SMOOTH for plot in plots)
if has_smooth:
    return [plot for plot in plots if plot.type != PlotType.LINE]
```

*Any* smooth anywhere, then drop *every* line anywhere.

## 1. A fit in one panel deleted a line chart in another — and the panel

```python
fig, axes = plt.subplots(1, 2)
sns.regplot(data=frame, x="a", y="b", ax=axes[0])
axes[1].plot([1, 2, 3], [3, 2, 1])
```

| | emitted |
|---|---|
| before | `[[['point', 'smooth']]]` |
| after | `[[['point', 'smooth'], ['line']]]` |

Panel 1 lost its line, and with it panel 1: the grid's column count comes from the surviving layers' indices, and nothing left carried column 1. A 1×2 figure was announced as 1×1.

## 2. The selector ids shifted

Neither that filter nor the duplicate-gid one below it touched `selector_ids`, which is paired with `_plots` by index in **both** directions — the artists are tagged with `selector_ids[i]` and the schema stamps the same index into the layer's selector string. Every layer after a dropped one inherited its predecessor's id, so the highlight landed one layer off with nothing raised:

```
ids issued  fc8cef9b 99388429 171f9530 7c235364
before      fc8cef9b 99388429 171f9530 7c235364     <- 3 layers, 4 ids
after       99388429 171f9530 7c235364
```

It only bites when the dropped layer is not last, which is why it survived — and why the test draws the line *first*.

## The fix

All three rules — segmented bars, fitted curves, duplicate gids — now answer into **one** set and **one** lockstep filter, so there is one place the pairing invariant is stated instead of three places that could each forget it. `_collapse_segmented_bar_layers` becomes `_drop_superseded_layers`, and `maidr/util/dedup_utils.py` is deleted: leaving behind a helper whose whole shape is the bug invites it back.

### Two deliberate limits

**The gid rule stays figure-wide.** A `_smooth_gid` names one curve, and whether the same one can be registered against two axes is not something this change measured. Scoping it with the others would have been a behaviour change I could not justify; only its bookkeeping moves.

**A reference line on the *same* axes as a fit is still dropped.** Per-axes scoping cannot tell an annotation from the fit's own line — both are `LINE` on one axes. `test_a_deliberate_line_over_a_fit_is_kept` pins that, badly named on purpose so the day it is fixed the test fails and has to be rewritten rather than the behaviour quietly changing.

## Falsification

| reverted | failing tests |
|---|---|
| smooth/line rule back to figure-wide | 1 of 4 |
| `selector_ids` left unfiltered | 1 of 4 |

The two that survive each revert are the one asserting a fit still supersedes its own line (which must not break when the rule is scoped) and the pinned annotation case.

## Verification

`ruff check` clean. Full suite: **1204 passed**.

---


jooyoungseo:
Both nits taken, and one observation confirmed as deliberate.

## The gid path had no direct coverage

Right, and it was the one path in the diff without it. `test_one_curve_registered_twice_keeps_one_layer_and_one_id` stamps a shared gid on the fits two `regplot` calls register, then asserts one layer survives **and** that an id goes with each remaining layer — the half that was not filtered before. Falsified by dropping the layer without its id:

```
gid drops the layer but not the id:  1 failed, 4 passed
```

The gid is stamped rather than drawn, and the docstring says why: two `regplot` calls on one axes leave `_smooth_gid` as `None` on every layer, so no drawing I reached makes the rule fire on its own. What is under test is the filtering, not how a collision arises. (How many fits those two calls register is read rather than asserted — measured, it is three, which is a separate question and not this PR's.)

## `seen: set[str]`

Done.

## The early run in `_create_html_tag` — deliberate, and worth the callout you asked for

> Worth calling out explicitly if it wasn't deliberate, since it's a behavior change beyond the reported bug (in a good direction).

Deliberate, and it is the same mechanism #377 needed for a different reason. Before, the duplicate `LINE`'s artists were tagged for highlighting during SVG generation and its layer was then dropped from the schema — so those tags pointed at a layer that no longer existed. There is one pass now and it runs before tagging, so a layer that will not be announced is not tagged either. Strictly better, and it falls out of consolidating rather than being bolted on.

Full suite: **1205 passed**.

---

jooyoungseo:
The gid-timing observation is right, and sharper than "low confidence" suggests. Verified rather than reasoned through, including one thing that changes its status.

## The gid rule does see a later snapshot — and it is a live path

`_smooth_gid` is assigned during extraction, which is what reading `plot.elements` triggers, so the run in `_create_html_tag` sees every gid as `None` and only the run in `_flatten_maidr` can see a collision. Confirmed.

What I'd assumed otherwise, and checked: a real render **does** set a gid, so this is not a theoretical path.

```
after a real render:  [('point', None), ('smooth', 'maidr-2904e558-...')]
```

## The consequence is bounded, and bounded differently from the two bugs here

Your read of `_get_selector` is right and it is the load-bearing part: `SmoothPlot._get_selector` returns `g[id='<gid>'] path` when it has a gid, and that string carries no `maidr='true'` placeholder — so the `selector_ids[i]` substitution in `_flatten_maidr` is a **no-op** for it.

So a duplicate dropped on the second run leaves its artists carrying a tag no schema layer references, which is inert, while every surviving layer keeps its own id because the filter is lockstep. That is a stale tag, not the misattribution the other two rules caused.

> it means the 'one lockstep filter, so there's one place the pairing invariant is stated' framing is slightly optimistic for this one rule specifically

Fair, and the docstring now says so, so the framing doesn't outlive this thread.

## Both suggested tests added

**`test_two_axes_each_drop_only_their_own_duplicate`** — two independent fits, each with its own genuine `LINE`/`SMOOTH` pair. You're right that it's the shape the PR is named for, and right that per-group makes it safe by construction — which is exactly what the figure-wide version was assumed to be, so it gets its own test rather than being inferred.

**`test_dropping_twice_changes_nothing`** — the bar rule's idempotence check extended to the curve rules, which share the pass and so have to survive being asked twice in a single render.

I did not add the `render()`-level duplicate-gid repro. Reaching a genuine collision needs two `SmoothPlot`s to read the same `Line2D`, which no drawing I could construct produces — measured, two `regplot` calls on one axes give three smooths with `None` gids. Manufacturing one would test my manufacturing, not the code.

Full suite: **1207 passed**.

---

PR #381, opened 2026-08-14
fix(bar): read the bars this call drew, not every bar on the axes

### LINKED ISSUE #380: Two ax.bar() calls on one axes with no bottom render nothing
Deferred from #376 and promised there. Still reproduces on `main` after #377 and #379.

```python
fig, ax = plt.subplots()
ax.bar(cats, s0)
ax.bar(cats, s1)          # overlaid, no `bottom`
```

```
registered=['bar', 'bar']   ->  ExtractionError
```

`ExtractionError` is fatal to the whole render rather than to its own layer, so the figure produces no HTML.

## Why #377 does not reach it

`BarPlot._extract_bar_container_data` flattens **every** `BarContainer` on the axes, so each of the two layers finds 6 patches against 3 tick labels, returns `None`, and raises. #377 fixed the case where a *segmented* layer is registered, because that gives the collapse something to keep — here neither call passes `bottom`, so both register as plain `BAR`, no segmented layer exists, and nothing supersedes anything.

For contrast, on current `main`:

```
1 call: ax.bar(cats, s0)                          OK  ['bar']
2 calls, no bottom (overlaid)                     ** ExtractionError
stack written the natural way (no bottom on 1st)  OK  ['stacked_bar']
stack with bottom=0 on the 1st                    OK  ['stacked_bar']
bar then a second bar on twin axes                OK  ['bar', 'bar']
```

The twin-axes row is the tell: two `BarPlot`s are fine as long as they are on *different* axes, because then each sweeps only its own containers. The bug is the sweep, not the count.

## The part that is a real question, not an oversight

Two overlapping bar layers are genuinely ambiguous, which is why this was not folded into #377:

- If they are **two series meant to be compared** (the common reason to overlay with alpha), two layers is right, and each should describe only the bars its own call drew.
- If the second is meant to **overwrite** the first, one layer is right.
- matplotlib itself draws them overlapping and takes no position.

## Proposed fix

Scope `BarPlot` to its own artists rather than sweeping the axes, the way `violinplot.py` already snapshots `ax.lines` before its call and diffs after. That answers the ambiguity in the first way — two layers, each describing its own bars — which is the reading that loses no data and matches what is drawn.

The sweep exists for a reason worth preserving: seaborn 0.13 draws one layer as several `BarContainer`s, one per hue group. So the fix is not "read one container" but "read the containers *this call* added", which the patch site knows and the extractor does not. That is the same shape as the `lines_before` snapshot in `maidr/patch/violinplot.py`.

Not attempted here because it changes what `BarPlot` reads for every bar chart, which wants its own diff and its own falsification rather than riding along behind a merge fix.

jooyoungseo:
Closes #380. The case deferred from #376, now that #377 and #379 have landed.

`BarPlot` swept every `BarContainer` on its axes rather than the bars its own call drew, so two overlaid `ax.bar()` calls each found six patches against three tick labels, failed the count check, and raised `ExtractionError` — which is fatal to the whole figure rather than to its own layer.

```
2 calls, no bottom (overlaid)   before  ExtractionError
                                after   ['bar', 'bar']
                                        10/20/30  and  30/20/10
```

#377 fixed the neighbouring case, where one call passes `bottom` and so registers a segmented layer for the collapse to keep. Here neither does, both register as plain `BAR`, and nothing supersedes anything. The twin-axes row was always the tell — two `BarPlot`s are fine on *different* axes, because then each sweeps only its own containers. The bug is the sweep, not the count.

## The reading, which was the actual question

Two overlapping bar layers are genuinely ambiguous, which is why this was filed separately rather than folded into a merge fix. Two series drawn over one another with alpha are two series, so **two layers, each describing its own bars**, is the answer that loses no data and matches what is drawn. Asserted by data rather than by count, since two layers of the right type would also pass if both described the same six patches.

## How the layer learns which bars are its own

Only the matplotlib entry point knows. `common()` gains an optional `drawn_as`: the artist the call returned is handed to the layer under that keyword, **after** the draw so it cannot reach the wrapped function. `Axes.bar` uses it.

## What deliberately keeps sweeping

**seaborn.** It draws one bar layer as several containers, one per hue group, and registers it from `sns_bar` — where no single container is the answer. Narrowing that path would announce a hued bar chart as one group and silently drop the rest; `test_a_hued_seaborn_bar_still_reads_every_group` guards it by counting six bars rather than three.

**`GroupedBarPlot`.** A stacked layer describes every bar on its axes by design, and #377 depends on that. `test_a_stacked_chart_is_unchanged` is the regression guard.

`MaidrPlotFactory` now forwards its kwargs to `BarPlot`, which it already did for every other type.

## Falsification

| reverted | failing tests |
|---|---|
| `BarPlot` sweeps again | 1 of 4 |
| patch names no container | 1 of 4 |

The two that survive each revert are the single-call control and the seaborn/stacked guards — which is the point: those paths must not move.

## Verification

`ruff check` clean. Full suite: **1211 passed**.

---

### DIFF
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -37,9 +37,13 @@ from maidr.util.dependencies import (
 from maidr.util.environment import Environment
 from maidr.util.iframe_utils import wrap_in_iframe_matplotlib
 
-#: Layer classes whose extractor reads every ``BarContainer`` on its axes
-#: rather than the bars its own call drew, so two of them on one axes describe
-#: the same chart twice.
+#: Layer classes a segmented bar layer on the same axes can supersede.
+#:
+#: ``GroupedBarPlot`` reads every ``BarContainer`` on its axes, so it already
+#: describes whatever a ``BarPlot`` beside it would -- whether that ``BarPlot``
+#: swept the axes too (the seaborn path) or narrowed to the one container its
+#: own ``ax.bar()`` call drew (the matplotlib path, since #380). Either way the
+#: segmented layer covers it, which is what makes it superseded.
 #:
 #: Asked by class rather than by ``plot.type``, and that distinction is not
 #: pedantic: ``MplfinanceBarPlot`` also carries ``PlotType.BAR``, and it reads
@@ -433,11 +437,12 @@ class Maidr:
         """
         The bar layers a segmented layer on the same axes already describes.
 
-        ``BarPlot`` and ``GroupedBarPlot`` both read *every* ``BarContainer``
-        on their axes rather than the bars their own call drew, so a stacked
-        chart built from three ``ax.bar()`` calls registers three layers that
-        each describe the whole chart. The first segmented one survives; the
-        other axes-wide bar layers on that axes are duplicates of it.
+        ``GroupedBarPlot`` reads *every* ``BarContainer`` on its axes, so a
+        stacked chart built from three ``ax.bar()`` calls registers three
+        layers that each describe the whole chart. The first survives; the
+        rest are duplicates, and so is any ``BarPlot`` beside them -- whose
+        bars, whether it swept the axes or narrowed to its own call's
+        container, the segmented layer already covers.
 
         Layers of any other kind are left alone -- a line over a stacked bar
         is a second layer, not a duplicate of it.
@@ -536,11 +541,10 @@ class Maidr:
         """
         Drop every layer another layer on the same axes already describes.
 
-        ``BarPlot`` and ``GroupedBarPlot`` both read *every* ``BarContainer``
-        on their axes rather than the bars their own call drew, so a stacked
-        chart built from three ``ax.bar()`` calls registers three layers that
-        each describe the whole chart. One has to survive and the rest are
-        duplicates.
+        ``GroupedBarPlot`` reads *every* ``BarContainer`` on its axes, so a
+        stacked chart built from three ``ax.bar()`` calls registers three
+        layers that each describe the whole chart. One has to survive and the
+        rest are duplicates.
 
         Which one survives is the part that was wrong. This used to key off
         ``self.plot_type`` -- a *figure-wide* "highest priority type seen",
--- a/maidr/core/plot/barplot.py
+++ b/maidr/core/plot/barplot.py
@@ -13,6 +13,16 @@ from maidr.util.mixin import (
 )
 
 
+#: The keyword ``Axes.bar`` hands its own ``BarContainer`` to this layer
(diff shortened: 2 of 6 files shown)


jooyoungseo:
The stale-comment finding is the right call, and it's a defect I introduced rather than an inherited one — #377's rationale rests on a premise this PR makes false for the primary path, and I wrote both.

## The three comments

Corrected. You're right that it's docs-only: the superseding logic is class-based, not behaviour-based, so a narrowed `BarPlot` beside a `GroupedBarPlot` is still correctly dropped. But the *reason* changes, and that is the part the comments exist for. They now say the segmented layer covers the `BarPlot`'s bars **either way** — whether it swept the axes (seaborn) or narrowed to its own call's container (matplotlib, since this PR) — rather than resting on a sweeping premise that only half holds.

That distinction is exactly what a future reader debugging the collapse would need, and the old wording would have sent them looking for a sweep that no longer happens.

## `"_maidr_bars"` duplicated

Agreed, and for the reason you gave rather than tidiness: `kwargs.get` falls back to sweeping on a mismatch, so a typo would not raise — it would quietly restore the behaviour this PR removes. Now a shared `DRAWN_BARS`.

It lives in `maidr/core/plot/barplot.py` rather than beside `common.drawn_as`, which was my first instinct and a circular import: `maidr.patch` imports `maidr.core`, not the other way about. The constant's docstring records that, so the next person to move it somewhere tidier finds out why before the import graph tells them.

## `_own_containers` return type

Right — `extract_container` returns `None` when the axes holds no container list. Typed `list[BarContainer] | None`, with the docstring noting the caller already treats that as nothing to extract.

Full suite: **1211 passed**.

---

jooyoungseo:
Added the `barh` case — you're right that it was the one gap.

Worth stating why "it shares the same code path" is the argument *for* the test rather than against it: it stops being true one layer down. A horizontal bar's magnitude is its **width**, read through a different branch of `_extract_bar_container_data`, and the narrowing has to survive that branch too. `test_two_overlaid_barh_calls_read_their_own_bars_too` asserts on `x` rather than `y` for exactly that reason.

Reverting the narrowing now fails **2 of 5** rather than 1 of 4, which is the check that the new test is load-bearing rather than decorative.

On gating `drawn_as` to the plain-`BAR` branch: leaving it. It does reach `GroupedBarPlot`, which discards it — but the branch that decides `plot_type` runs *before* the draw, and `drawn_as` is consumed *after*, so gating it would mean re-deriving the decision or threading it through. That is more moving parts than the confusion it saves, and `DRAWN_BARS`'s docstring already says which layer reads it.

Full suite: **1212 passed**.

---

PR #383, opened 2026-08-14
fix(bar): announce a bar's position when no tick label names it

### LINKED ISSUE #382: A bar chart with a numeric x renders nothing
`BarPlot` pairs its bars with the labels on the categorical axis and raises when the two counts disagree:

```python
plot = [patch for container in plot for patch in container.patches]
if levels and len(plot) != len(levels):
    return None            # -> ExtractionError -> whole figure produces nothing
```

For a **categorical** x, matplotlib puts exactly one tick per category, so the counts match by construction. For a **numeric** x, the tick locator chooses its own breaks and the counts have no reason to agree:

```
                                               bars  labels  render
x = np.arange(len(labels))  [mpl's own recipe]    3      5    ** ExtractionError
x = np.arange(...), with set_xticks(x, labels)    3      3    ok
plain categorical strings                         3      3    ok
numeric x, barh                                   3      8    ** ExtractionError
numeric x, many bars (20)                        20      8    ** ExtractionError
```

`ExtractionError` is fatal to the whole render rather than to its own layer, so the figure produces no HTML at all.

## This is matplotlib's own documented shape

The [grouped bar chart example](https://matplotlib.org/stable/gallery/lines_bars_and_markers/barchart.html) opens with exactly this:

```python
x = np.arange(len(species))
ax.bar(x + offset, measurement, width, label=attribute)
```

It survives there only because that example goes on to call `ax.set_xticks(x + width, species)`, which coincidentally makes the counts line up. Drop that line — or draw a bar chart whose x really is numeric, which is ordinary — and there is no output.

Two more spellings of the same failure, both realistic:

```
bar, ticks removed after (set_xticks([]))    ** ExtractionError
bar, ticks removed before                    ** ExtractionError
```

Hiding tick marks is a common styling choice, and it currently deletes the chart.

## Why the count check exists

It is guarding something real: the labels are what a reader hears for each bar, and pairing three bars against five labels would announce the wrong name for every one of them. Raising is the wrong response, though — the labels are one *presentation* of x, not x itself. A bar drawn at x=0 with no tick beside it still has a position, and announcing `0` is honest where announcing nothing is not.

## Proposed fix

When the tick labels do not correspond one-to-one with the bars, fall back to the bars' own x positions rather than raising. `Rectangle.get_x() + get_width() / 2` is the centre matplotlib drew the bar at, which is the value the caller passed.

Keep the label path when the counts do agree — a categorical chart must still announce "Adelie", not `0`. So the rule is "labels when they line up, positions otherwise", and the count check becomes the thing that chooses between two readings rather than the thing that kills the figure.

Worth pinning both directions in tests: a categorical chart still announcing its names, and a numeric chart announcing its positions.

jooyoungseo:
Closes #382.

`BarPlot` paired its bars with the labels on the categorical axis and raised when the counts disagreed. For a **categorical** x, matplotlib puts exactly one tick per category, so they agree by construction. For a **numeric** x, the tick locator picks its own breaks and they have no reason to — and `ExtractionError` is fatal to the whole render rather than to its own layer, so the figure produced no HTML at all.

```
                                               bars  labels  render
x = np.arange(len(labels))  [mpl's own recipe]    3      5    raised
x = np.arange(...), + set_xticks(x, labels)       3      3    ok
plain categorical strings                         3      3    ok
numeric x, barh                                   3      8    raised
numeric x, many bars (20)                        20      8    raised
```

## The first row is matplotlib's own grouped bar chart

```python
x = np.arange(len(species))
ax.bar(x + offset, measurement, width, label=attribute)
```

It survives in [the gallery](https://matplotlib.org/stable/gallery/lines_bars_and_markers/barchart.html) only because the example goes on to call `ax.set_xticks(x + width, species)`, which makes the counts line up by accident. Drop that line, or draw a bar chart whose x really is numeric, and there is nothing. Hiding tick marks — an ordinary styling choice — deleted the chart for the same reason.

## The fix keeps the check and changes what it decides between

The count check was guarding something real: three bars against five labels would announce the wrong name for every bar. So it still decides — but between **labels when they line up** and **the bars' own drawn centres when they do not**, rather than between a reading and nothing.

Labels are one *presentation* of x, not x itself. A bar at x=0 with no tick beside it still has a position, and announcing `0` is honest where announcing nothing is not.

Positions are read off the rectangle and printed the way an axis prints them — `"0"` rather than `"0.0"`, while a bar really at 1.5 still says so.

## Two existing tests changed sides

`test_a_label_count_mismatch_raises_extraction_error` and `test_an_axis_with_no_tick_labels_raises_extraction_error` asserted the old raising. Both were **rewritten rather than deleted**, because neither was arguing that raising was correct:

- the first guarded an `ExtractionError` **rather than a `TypeError`** — failing cleanly instead of messily;
- the second guarded against emitting a **blank** label per bar.

Both guarantees survive the change, and the rewritten docstrings say what each was for so the next reader knows they were satisfied rather than overruled. They did their job by failing loudly.

## Falsification

| reverted | failing tests |
|---|---|
| no position fallback (raise as before) | 3 of 5 |
| positions used unconditionally | 2 of 5 |

The second matters as much as the first: using positions everywhere would announce a categorical chart as `0, 1, 2` and would read as a *working* chart — a worse defect than the one being fixed.

## Verification

`ruff check` clean. Full suite: **1217 passed**.

---

### DIFF
--- a/maidr/core/plot/barplot.py
+++ b/maidr/core/plot/barplot.py
@@ -80,11 +80,12 @@ class BarPlot(MaidrPlot, ContainerExtractorMixin, LevelExtractorMixin, DictMerge
     def _extract_plot_data(self) -> list:
         plot = self._own_containers()
         self._orientation = self._extract_orientation(plot)
+        patches = self._patches(plot)
         data = self._extract_bar_container_data(plot)
         if data is None:
             raise ExtractionError(self.type, plot)
 
-        levels = self._labels_for(plot, data)
+        levels = self._labels_for(patches, data)
 
         # A horizontal bar's magnitude runs along x and its label sits on y,
         # which is the layout the renderer reads for a horizontal layer. The
@@ -99,7 +100,7 @@ class BarPlot(MaidrPlot, ContainerExtractorMixin, LevelExtractorMixin, DictMerge
 
         return [{"x": x, "y": y} for x, y in combined_data]
 
-    def _labels_for(self, plot: list[BarContainer] | None, data: list) -> list[str]:
+    def _labels_for(self, patches: list[Rectangle], data: list) -> list[str]:
         """
         What to announce alongside each magnitude.
 
@@ -128,8 +129,8 @@ class BarPlot(MaidrPlot, ContainerExtractorMixin, LevelExtractorMixin, DictMerge
 
         Parameters
         ----------
-        plot : list of BarContainer, optional
-            The containers this layer describes.
+        patches : list of Rectangle
+            This layer's bars, already flattened out of their containers.
         data : list
             One magnitude per bar, used for its length.
 
@@ -143,7 +144,7 @@ class BarPlot(MaidrPlot, ContainerExtractorMixin, LevelExtractorMixin, DictMerge
         if levels and len(levels) == len(data):
             return levels
 
-        return [self._bar_position(patch) for patch in self._patches(plot)]
+        return [self._bar_position(patch) for patch in patches]
 
     def _bar_position(self, patch: Rectangle) -> str:
         """
--- a/maidr/core/plot/barplot.py
+++ b/maidr/core/plot/barplot.py
@@ -79,11 +79,11 @@ class BarPlot(MaidrPlot, ContainerExtractorMixin, LevelExtractorMixin, DictMerge
     def _extract_plot_data(self) -> list:
         plot = self._own_containers()
         self._orientation = self._extract_orientation(plot)
-        levels = self.extract_level(self.ax, self._level_key)
-
-        data = self._extract_bar_container_data(plot, levels)
+        data = self._extract_bar_container_data(plot, None)
         if data is None:
             raise ExtractionError(self.type, plot)
+
+        levels = self._labels_for(plot, data)
 
         # A horizontal bar's magnitude runs along x and its label sits on y,
         # which is the layout the renderer reads for a horizontal layer. The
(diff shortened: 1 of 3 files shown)


jooyoungseo:
All taken, and one of the "not a request" items turned out to deserve a fix.

## The dead `levels` parameter

Right, and it's the worst kind of leftover — a parameter whose docstring documents a rule it no longer enforces. Removed rather than deprecated. `_extract_bar_container_data` also flattens through `_patches` now, so there is one flattening implementation.

## `f"{centre:g}"` — fixed rather than noted

> worth being aware of if this is ever hit with large numeric x values

Worth more than awareness, I think: six significant figures then exponential, so a bar at x=1234567 announced `"1.23457e+06"`. That is lossy *and* hard to listen to, which is the specific thing this project exists to avoid — and a large x is not exotic, it is an index, an id, or a year.

Integers are now formatted exactly and only fractions fall back to `:g`, which lands the realistic large case squarely in the exact branch:

```
positions: ['0', '1234567', '2.5']
```

Reverting to plain `:g` fails the new test.

## Typing

`_bar_position(patch: Rectangle)` and `_patches(...) -> list[Rectangle]`.

## The sweep-path observation

Worth closing the loop, since I think #381 already answers it: after that change, `Axes.bar` narrows to the container its own call drew, so two independent `ax.bar()` calls each get their own layer with their own three bars rather than sharing a sweep. `test_two_overlaid_bar_calls_each_read_their_own_bars` asserts exactly that. The sweep now only serves the seaborn path, where every container genuinely *is* one layer — so "sequential positions across both calls' bars" has no path left to it. The `_own_containers` docstring says which path is which.

Full suite: **1218 passed**.

---

jooyoungseo:
Point 3 is real, and it means this PR's scope needs stating more carefully than I did.

## The stacked path has the same defect — verified

```
stacked, categorical x            ['stacked_bar', 'stacked_bar']   ok
stacked, NUMERIC x                ['stacked_bar', 'stacked_bar']   ** ExtractionError
stacked, numeric x + set_xticks   ['stacked_bar', 'stacked_bar']   ok
```

`GroupedBarPlot._extract_grouped_bar_data` carries its own copy of the count check, and it is the stacked half of the same matplotlib shape:

```python
x = np.arange(len(species))
ax.bar(x, first, label="first")
ax.bar(x, second, bottom=first, label="second")
```

Filed as #384 rather than folded in. Not out of tidiness — `GroupedBarPlot` pairs *per container* rather than once, and each point also carries a `z` naming its series, so "labels when they line up, positions otherwise" has to be decided per group and agree across them. That is a different fix wearing the same name, and it wants its own falsification.

What I should flag: #382 as I wrote it says "a bar chart with a numeric x renders nothing", which reads as covering both. It doesn't after this PR — the plain half is fixed and the stacked half is not. #384 records the remainder and links back, so closing #382 here does not overstate what landed.

## The `:g` caveat

You're right that the integer branch scopes the problem rather than solving it, and the docstring claimed more than it delivered. It now says what it does and why the rest is not a one-line fix: rounding fractions is not incidental — the centre is computed from the rectangle's geometry, so `0.1 + 0.2` arrives as `0.30000000000000004`, and printing that exactly would be worse than printing it short. Telling float noise from a real fraction needs more than a format string.

## The redundant flatten

Taken. `_extract_plot_data` flattens once and hands the patches to `_labels_for`.

Full suite: **1218 passed**.

---

Issue #385, opened 2026-08-14
A horizontal stacked bar (barh with left=) is read as two separate charts


jooyoungseo:
Found while testing #384's horizontal mirror.

`maidr/patch/barplot.py` decides a bar layer is stacked by looking for one keyword:

```python
if "bottom" in kwargs:
    bottom = kwargs.get("bottom")
    if bottom is not None:
        plot_type = PlotType.STACKED
```

`bottom` is how a **vertical** bar is stacked. The horizontal equivalent is `left`, and it is not checked — so the standard horizontal stacked bar

```python
ax.barh(positions, lower, left=np.zeros(n), label="lower")
ax.barh(positions, upper, left=lower,       label="upper")
```

arrives as **two independent `bar` layers**:

```
registered: ['bar', 'bar']
bar [{"x": 10.0, "y": "0"}, {"x": 20.0, "y": "1"}, {"x": 30.0, "y": "2"}]
bar [{"x": 30.0, "y": "0"}, {"x": 20.0, "y": "1"}, {"x": 10.0, "y": "2"}]
```

## What a reader loses

The numbers are right and the layer count is plausible, so nothing looks wrong. What is missing is the relationship: a reader is told there are two bar charts, and is not told the second sits **on top of** the first. The `stacked_bar` trace exists precisely to say that — it announces the segment's own value against the running total, which is the thing a stacked chart is *for*. Two plain bar layers announce two unrelated series that happen to share an axis.

It also means the segmented collapse never fires, so the two layers are kept separately — which is why this renders at all rather than raising.

## Not fixed by #383 / #384

Those are about which *labels* a bar announces, and after them both layers announce their positions correctly. This is about which *type* the layer is given, one step earlier, in the patch.

`tests/core/test_numeric_segmented_bar_positions.py::test_a_horizontal_stack_is_not_recognised_as_one` pins the current behaviour and is named for the defect, so the day this is fixed that test fails and has to be rewritten.

## Proposed fix

Read `left` alongside `bottom` in `maidr/patch/barplot.py`, since they are the same argument for the two orientations. Worth checking at the same time whether `_should_classify_as_dodged` has the mirrored gap — it reads `width` and numeric x positions, which are the vertical spellings of what a horizontal dodge does with `height` and numeric y.

PR #386, opened 2026-08-14
fix(bar): announce a segmented bar's position when no tick label names it

### LINKED ISSUE #384: A stacked or dodged bar chart with a numeric x renders nothing
Raised in review of #383 and verified. The same defect as #382, in the sibling class.

```
stacked, categorical x              ['stacked_bar', 'stacked_bar']   ok
stacked, NUMERIC x                  ['stacked_bar', 'stacked_bar']   ** ExtractionError
stacked, numeric x + set_xticks     ['stacked_bar', 'stacked_bar']   ok
```

`GroupedBarPlot._extract_grouped_bar_data` has its own copy of the count check that #383 replaced in `BarPlot`:

```python
for i, container in enumerate(plot):
    if len(level) != len(container.patches):
        return None            # -> ExtractionError -> whole figure produces nothing
```

Same cause: for a categorical x, matplotlib puts exactly one tick per category, so the counts agree by construction. For a numeric x the tick locator picks its own breaks, and they do not. `ExtractionError` is fatal to the whole render, so the figure produces no HTML.

`#383` fixes this for `BarPlot` — plain `ax.bar()`/`ax.barh()` — and deliberately does not reach here: `GroupedBarPlot` does not subclass `BarPlot`, and its structure differs enough that the fix is not a transplant. It pairs *per container* rather than once, and each point also carries a `z` naming its series, so "labels when they line up, positions otherwise" has to be decided per group and agree across groups.

## Why it matters as much as #382 did

This is the stacked half of matplotlib's own grouped-bar shape:

```python
x = np.arange(len(species))
ax.bar(x, first, label="first")
ax.bar(x, second, bottom=first, label="second")
```

which is how anyone writes a stacked chart over numeric positions, and it renders nothing unless `set_xticks(x, labels)` happens to make the counts line up.

## Proposed fix

The same shape as #383, adapted:

- decide the labels once for the layer rather than per container, since every container of a segmented layer shares the category axis;
- keep the tick labels when their count matches the bars of a container, and fall back to the bars' own drawn centres when it does not;
- keep the mismatch check as the thing that *chooses*, not the thing that raises.

Worth pinning both directions as #383 does: a categorical stacked chart still announcing its names, and a numeric one announcing its positions, with the `z` series names unaffected either way.

jooyoungseo:
Closes #384 — the segmented half of #382, raised in review of #383.

`GroupedBarPlot._extract_grouped_bar_data` carried its own copy of the count check that #383 replaced in `BarPlot`, so a stacked chart over numeric positions produced no HTML at all:

```
stacked, categorical x              ok
stacked, NUMERIC x                  ** ExtractionError
stacked, numeric x + set_xticks     ok
```

That is the stacked half of how anyone writes this:

```python
x = np.arange(len(species))
ax.bar(x, first, label="first")
ax.bar(x, second, bottom=first, label="second")
```

## Same rule, one difference that is why it wasn't a transplant

Tick labels when there is one per bar, the bars' own drawn centres otherwise — as in #383. But the labels are decided **once for the layer** rather than per container. Every series of a segmented chart shares one category axis, so a per-container answer could name a category in one series and a position in the next.

The per-container length check stays, and now guards only what `_labels_for` cannot: a container of a different length from its siblings, which there is nothing honest to pair with.

`z` comes from the legend and is untouched — asserted in **both** readings, because a reader who loses "upper" and "lower" cannot tell the series apart however good the categories are.

## Falsification

| reverted | failing tests |
|---|---|
| no position fallback (raise as before) | 2 of 5 |
| positions used unconditionally | 2 of 5 |

The second matters as much as the first: positions everywhere would announce a categorical stacked chart as `0, 1, 2` and read as a working chart.

## A boundary found while writing the horizontal mirror

`ax.barh(..., left=...)` is **not recognised as stacked at all** — the patch classifies on `"bottom" in kwargs` alone, and `left` is the horizontal spelling of the same argument. So a horizontal stack arrives as two independent `bar` layers:

```
registered: ['bar', 'bar']
```

The numbers are right and the layer count is plausible; what a reader is not told is that the second sits on top of the first. That is a wrong *reading* rather than a dead render, it is a step earlier than this PR (classification, not extraction), and it is filed as #385. `test_a_horizontal_stack_is_not_recognised_as_one` pins it and is named for the defect, so the day it is fixed that test fails and has to be rewritten — while still asserting the part this PR and #383 are responsible for: whatever the layers are called, each announces the bars it drew rather than raising.

## Verification

`ruff check` clean. Full suite: **1223 passed**.

---

### DIFF
--- a/maidr/core/plot/barplot.py
+++ b/maidr/core/plot/barplot.py
@@ -8,6 +8,7 @@ from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
 from maidr.exception import ExtractionError
 from maidr.util.mixin import (
+    BarPositionMixin,
     ContainerExtractorMixin,
     DictMergerMixin,
     LevelExtractorMixin,
@@ -24,7 +25,13 @@ from maidr.util.mixin import (
 DRAWN_BARS = "_maidr_bars"
 
 
-class BarPlot(MaidrPlot, ContainerExtractorMixin, LevelExtractorMixin, DictMergerMixin):
+class BarPlot(
+    MaidrPlot,
+    BarPositionMixin,
+    ContainerExtractorMixin,
+    LevelExtractorMixin,
+    DictMergerMixin,
+):
     def __init__(self, ax: Axes, **kwargs) -> None:
         super().__init__(ax, PlotType.BAR)
         self._orientation = "vert"
@@ -145,33 +152,6 @@ class BarPlot(MaidrPlot, ContainerExtractorMixin, LevelExtractorMixin, DictMerge
             return levels
 
         return [self._bar_position(patch) for patch in patches]
-
-    def _bar_position(self, patch: Rectangle) -> str:
-        """
-        The centre a bar was drawn at, as the axis would print it.
-
-        Read off the rectangle rather than the caller's argument, because the
-        caller's is not available here and the drawn centre is what the value
-        became. Whole numbers lose their trailing ``.0``: a bar at x=0 is at
-        ``"0"``, not ``"0.0"``, matching what a numeric axis shows.
-
-        Parameters
-        ----------
-        patch : Rectangle
-            One bar.
-
-        Returns
-        -------
-        str
-            The bar's position along its label axis.
-        """
-        if self._is_horizontal:
-            centre = patch.get_y() + patch.get_height() / 2
-        else:
-            centre = patch.get_x() + patch.get_width() / 2
-        if float(centre).is_integer():
-            return str(int(centre))
-        return f"{centre:g}"
 
     @staticmethod
     def _patches(plot: list[BarContainer] | None) -> list[Rectangle]:
--- a/maidr/core/plot/grouped_barplot.py
+++ b/maidr/core/plot/grouped_barplot.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from matplotlib.axes import Axes
 from matplotlib.container import BarContainer
+from matplotlib.patches import Rectangle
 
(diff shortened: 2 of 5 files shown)


jooyoungseo:
Took the duplication point rather than deferring it — this PR is what created the second caller, which is the project's own bar for extracting.

`BarPositionMixin` in `maidr/util/mixin/` now owns `_bar_position`, and both extractors mix it in. Net 42 lines removed, and there is exactly one source:

```
$ grep -rn "def _bar_position" maidr/
maidr/util/mixin/bar_position_mixin.py:25:    def _bar_position(self, patch: Rectangle) -> str:
```

The reason to share rather than spell it twice is recorded in its docstring, and it is specific rather than DRY-for-its-own-sake: the two copies would drift, the **horizontal branch** is precisely the kind of thing that gets corrected in one extractor and not the other, and a chart whose bars are announced at the wrong positions reads as a *working* chart. That is the failure mode this whole run has been about, so leaving two copies of the geometry seemed like a poor trade for a smaller diff.

Both readings unchanged:

```
stacked numeric : [['0', '1', '2'], ['0', '1', '2']]
plain numeric   : ['0', '1', '2']
```

Left alone: `_extract_orientation` / `_is_horizontal` / `_level_key`, which you also noted are duplicated. Those are about what the *layer* is rather than about a bar's geometry, and `GroupedBarPlot`'s `_level_key` carries its own reasoning about hue axes — folding them together is a bigger judgement about where the layer/geometry line sits, and it is not this bug's to make.

Full suite: **1223 passed**.

---

PR #387, opened 2026-08-14
fix(bar): read `left` as a stacked bar's baseline, as `bottom` is

### LINKED ISSUE #385: A horizontal stacked bar (barh with left=) is read as two separate charts

jooyoungseo:
Closes #385, found while writing #386's horizontal mirror.

A stacked bar is the one that says where its baseline is. `bottom` is how a **vertical** bar says it and `left` is how a **horizontal** one does — the same argument for the two orientations — and only the first was read:

```python
if "bottom" in kwargs:
    ...
    plot_type = PlotType.STACKED
```

So the standard horizontal stacked bar arrived as two independent `bar` layers:

```python
ax.barh(positions, lower, left=np.zeros(n), label="lower")
ax.barh(positions, upper, left=lower,       label="upper")
```

| | emitted |
|---|---|
| before | `['bar', 'bar']` |
| after | `['stacked_bar']` |

## What a reader was losing

The numbers were right and the layer count was plausible, so nothing looked wrong. What was missing is the relationship: a reader was told there are two bar charts, and was **not** told the second sits on top of the first. That is the whole content of a stacked chart, and the reason `stacked_bar` exists as a distinct trace rather than as a label — it announces the segment's own value against the running total.

## The pinning test did its job

`test_a_horizontal_stack_is_not_recognised_as_one`, added by #386 and named for the defect precisely so it would fail the day it was fixed, failed. Rewritten as `test_a_horizontal_stack_over_numeric_positions` — which is also the horizontal mirror #386 could not write, since the case did not reach `GroupedBarPlot` at all. It reads its label off `y` and its magnitude off `x`, through the branch of `BarPositionMixin` the vertical cases cannot reach.

## Falsification

| reverted | result |
|---|---|
| read `bottom` only | 1 of 6 fail |
| treat every bar as stacked | **23 of the suite fail** |

The second is the one worth having. `left` and `bottom` are absent from most bar calls, so a baseline check that fired unconditionally would announce every plain bar chart as a stack of one — and `test_a_plain_horizontal_bar_is_still_plain` is the control that names that directly.

## Verification

`ruff check` clean. Full suite: **1224 passed**.

---

### DIFF
--- a/maidr/patch/barplot.py
+++ b/maidr/patch/barplot.py
@@ -20,7 +20,8 @@ def bar(
 
     This function patches the bar plotting functions to identify whether the
     plot should be rendered as a normal, stacked, or dodged bar plot.
-    It uses the 'bottom' keyword to identify stacked bar plots. For dodged
+    It uses the 'bottom' keyword -- or 'left', which is how a horizontal bar
+    spells the same thing -- to identify stacked bar plots. For dodged
     plots, it uses robust detection logic that considers both width and
     context to avoid misclassifying simple bar plots with narrow widths as
     dodged plots. Seaborn's bar plots do not come through here — they are
--- a/maidr/patch/barplot.py
+++ b/maidr/patch/barplot.py
@@ -51,11 +51,16 @@ def bar(
     """
     plot_type = PlotType.BAR
 
-    # Check for stacked plots first (explicit bottom parameter)
-    if "bottom" in kwargs:
-        bottom = kwargs.get("bottom")
-        if bottom is not None:
-            plot_type = PlotType.STACKED
+    # A stacked bar is the one that says where its baseline is. `bottom` is
+    # how a vertical bar says it and `left` is how a horizontal one does --
+    # the same argument for the two orientations, and reading only the first
+    # meant `ax.barh(..., left=...)` arrived as two independent bar layers.
+    # The numbers were right and the layer count was plausible; what a reader
+    # was not told is that the second bar sits on top of the first, which is
+    # the whole content of a stacked chart (#385).
+    baseline = kwargs.get("bottom", kwargs.get("left"))
+    if baseline is not None:
+        plot_type = PlotType.STACKED
     else:
         # Extract width and align parameters
         if len(args) >= 3:
--- a/tests/core/test_numeric_segmented_bar_positions.py
+++ b/tests/core/test_numeric_segmented_bar_positions.py
@@ -178,3 +178,28 @@ def test_a_plain_horizontal_bar_is_still_plain() -> None:
     grid = FigureManager.get_maidr(fig)._flatten_maidr()["subplots"]
 
     assert [layer["type"].value for layer in grid[0][0]["layers"]] == ["bar"]
+
+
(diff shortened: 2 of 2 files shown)


jooyoungseo:
Both taken, and the "nice incidental fix" was worth turning into a deliberate one.

## The `bottom=None` side effect

You traced it exactly: `if "bottom" in kwargs` meant passing it explicitly as `None` — which matplotlib treats as omitting it — took the stacked branch's `else` away, so dodge detection never ran. The inner `is not None` guard kept the layer from being *called* stacked, so the result was a **plain** bar where a dodged one was drawn. Right numbers, wrong chart, nothing raised.

I hadn't spotted it, which means it was luck rather than design. `test_an_explicit_none_baseline_still_reaches_dodge_detection` makes it deliberate — restoring the key check fails it:

```
revert: key check, not value   1 of 7 fail
```

Worth being explicit that this is now behaviour I'm responsible for rather than a happy accident of the rewrite.

## The stale docstring

Fixed. It said the function uses the `bottom` keyword to identify stacked bars, which this PR makes half the story; it now names `left` as how a horizontal bar spells the same thing.

Full suite: **1225 passed**.

---

PR #389, opened 2026-08-14
fix(hist): do not read a 2D histogram as a bar histogram

### LINKED ISSUE #388: A bivariate histogram raises StopIteration and kills the render
`sns.histplot(x=..., y=...)` is a **2D** histogram — seaborn draws it as a `QuadMesh`, not as bars. MAIDR registers it as `hist` anyway, and extraction then dies with a bare `StopIteration`:

```
maidr/core/plot/histogram.py:47   in _extract_plot_data
maidr/util/mixin/extractor_mixin.py:31 in extract_container
StopIteration
```

Not an `ExtractionError` — a raw `StopIteration`, which tells a user nothing and is fatal to the whole figure.

```
2D histplot alone                 ['hist']            ** StopIteration
scatter + 2D histplot overlay     ['point', 'hist']   ** StopIteration
jointplot(kind='hist')                                ** StopIteration
jointplot(kind='kde')                                 ok
jointplot(kind='scatter')                             ok
```

**`sns.jointplot(kind="hist")` is a documented, mainstream call and produces no HTML at all.**

### 1. `extract_container` raises where it is documented to return

```python
if include_all:
    return [c for c in ax.containers if isinstance(c, container_type)]   # -> [] when none

return next(                                                            # -> StopIteration
    container for container in ax.containers if isinstance(container, container_type)
)
```

The two branches disagree: the list branch returns empty, the single branch raises. Every caller is written for the returning contract — `HistPlot._extract_bar_container_data` opens with `if plot is None or plot.patches is None: return None`, which cannot be reached because the `next()` above it raises first.

A bare `StopIteration` is never a useful failure. Inside a generator it is worse than useless: PEP 479 turns it into a `RuntimeError` with the original cause obscured.

### 2. A bivariate histogram should not register as `hist` at all

`hist` promises one bin per bar with a count. A 2D histogram has neither — it is a mesh of joint counts, closer to a heatmap. Registering it and then failing to extract is the #369 / #373 shape: a layer that promises a reading nothing can produce, and takes the figure with it when it cannot.

Fixing (1) alone would turn `StopIteration` into `ExtractionError`, which is still fatal. The layer has to not be registered.

## Proposed fix

- `extract_container` returns `None` when nothing matches, making the two branches agree and letting the callers' existing `None` handling work.
- The histogram patch declines to register when its call drew no `BarContainer`. A bivariate histogram is then simply unsupported — like `sns.kdeplot(x=..., y=...)`, which registers nothing today and renders as "no MAIDR found" rather than crashing — and, critically, **a supported chart sharing the axes still renders**.

The overlay row above is the one that matters most: a scatter that would read perfectly well produces nothing because a 2D histogram was drawn over it.

## Not in scope

Making 2D histograms *supported* is a separate feature. This is about not destroying the figure they appear in.

jooyoungseo:
Closes #388.

`sns.histplot(x=..., y=...)` is a **bivariate** histogram — seaborn draws it as a `QuadMesh` of joint counts, not as bars. MAIDR registered it as `hist` anyway, and extraction reached for the first `BarContainer` on an axes that has none, raising a bare `StopIteration`: not an `ExtractionError`, and fatal to the whole figure.

```
2D histplot alone               ['hist']            ** StopIteration
scatter + 2D histplot overlay   ['point', 'hist']   ** StopIteration
jointplot(kind='hist')                              ** StopIteration
jointplot(kind='kde')                               ok
jointplot(kind='scatter')                           ok
```

**`sns.jointplot(kind="hist")` is a documented, mainstream call and produced no HTML at all.** So did any supported chart sharing the axes — the scatter row is the one that matters, since it reads perfectly well and was destroyed by its neighbour.

### `extract_container` raised where it is documented to return

```python
if include_all:
    return [c for c in ax.containers if isinstance(c, container_type)]   # -> []

return next(                                                            # -> StopIteration
    container for container in ax.containers if isinstance(container, container_type)
)
```

The branches disagreed. Every caller is written for the returning contract — `HistPlot._extract_bar_container_data` opens with `if plot is None` — and the bare `next()` above it meant that could never run. A bare `StopIteration` is never a useful failure, and inside a generator PEP 479 turns it into a `RuntimeError` with the cause obscured. Both branches now answer "none of that type" with a value.

### But that only turns `StopIteration` into `ExtractionError`, which is still fatal

`hist` promises one bin per bar with a count, and a mesh has neither. So the patch declines to register when its call drew no `BarContainer` — asked of the **axes** rather than of a `y=` keyword, because "did this draw bars" is the question the extractor actually needs answered and the keyword is only seaborn's spelling of it.

A bivariate histogram is now simply unsupported, exactly as `sns.kdeplot(x=..., y=...)` already was — and the figure it appears in survives.

## After

```
scatter + 2D histplot overlay   ['point']    ok
jointplot(kind='hist')                       ok
jointplot(kind='kde')                        ok
jointplot(kind='scatter')                    ok
jointplot(kind='reg')                        ok
```

## Falsification

| reverted | failing tests |
|---|---|
| bare `next()` restored | 1 of 6 |
| bivariate registered again | 2 of 6 |

The mixin's contract is asserted **directly** as well as behaviourally, because once the patch stops registering the layer that reached it, no behavioural test would catch a return to `next()`.

Two controls guard the normal path: a plain 1D histogram, and `histplot(kde=True)` — the latter because `sns_hist` gained an early return, and the `smooth` registration runs *after* the point it skips.

## Not in scope

Making 2D histograms *supported*. This is about not destroying the figure they appear in.

## Verification

`ruff check` clean. Full suite: **1231 passed**.

---

### DIFF
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -43,12 +43,58 @@ def mpl_hist(
     return n, bins, plot
 
 
+def _drew_bars(plot) -> bool:
+    """
+    Whether the call that produced *plot* drew a histogram made of bars.
+
+    `sns.histplot(x=..., y=...)` is a **2D** histogram: seaborn draws it as a
+    ``QuadMesh`` of joint counts, not as bars. `hist` promises one bin per bar
+    with a count, which such a layer has neither of -- so registering it
+    promises a reading nothing can produce, and extraction then took the whole
+    figure down with it. `sns.jointplot(kind="hist")` produced no HTML at all,
+    and so did any supported chart that happened to share the axes (#388).
+
+    Asked of the axes rather than of the arguments, because "did this draw
+    bars" is the question the extractor actually needs answered, and a `y=`
+    keyword is seaborn's spelling of it rather than the thing itself.
+
+    Parameters
+    ----------
+    plot : Any
+        Whatever the patched call returned.
+
+    Returns
+    -------
+    bool
+        True when the axes holds at least one ``BarContainer``.
+    """
+    try:
+        ax = FigureManager.get_axes(plot)
+    except Exception:  # pragma: no cover - `common` already resolved this once
+        return False
+    return bool(ax is not None and ax.containers and any(
+        isinstance(container, BarContainer) for container in ax.containers
+    ))
+
+
 def sns_hist(wrapped, instance, args, kwargs) -> Axes:
     """
     Patch seaborn.histplot to register HIST and (if kde=True) SMOOTH layers for MAIDR.
+
+    A bivariate histogram is left unregistered rather than read wrongly; see
+    `_drew_bars`.
     """
+    if ContextManager.is_internal_context():
+        return _draw_quietly(wrapped, args, kwargs)
+
+    with ContextManager.set_internal_context():
+        drawn = _draw_quietly(wrapped, args, kwargs)
+
+    if not _drew_bars(drawn):
+        return drawn
+
     # Register the histogram as HIST as before
-    ax = common(PlotType.HIST, wrapped, instance, args, kwargs)
+    ax = common(PlotType.HIST, lambda *a, **k: drawn, instance, args, kwargs)
     # Only register KDE overlay as SMOOTH if kde=True was set
     kde_enabled = kwargs.get("kde", False)
     if kde_enabled:
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+from typing import Any
+
+import matplotlib.pyplot as plt
 import wrapt
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Finding 1 is right, and it is worse than "adjacent" — it is the exact failure this PR exists to prevent, reintroduced next to it. Verified:

```python
sns.barplot(data=df, x="g", y="v", ax=ax)      # leaves BarContainers
sns.histplot(data=df, x="v", y="w", ax=ax)     # bivariate: a QuadMesh
```

```
registered: ['bar', 'hist']
  bar   [{'x': 'a', 'y': 8.672...}, {'x': 'b', 'y': 10.193...}, ...]
  hist  [{'y': 8.672..., 'xMin': -0.4, 'xMax': 0.4}, ...]
```

The bivariate call answered True for someone else's artists, `extract_container` handed back the first container on the axes, and the `hist` layer described the **barplot's** bars with bin edges invented for them. The same three numbers announced twice — once correctly, once as a histogram that does not exist. Right numbers, wrong chart, nothing raised.

That is strictly worse than the crash I was fixing, so it is not a "known gap" I'd want to comment and leave.

## Fixed by diffing, as you suggested

Containers are snapshotted before the draw and diffed after — the same shape as #381's `drawn_as` and `violinplot.py`'s `lines_before`. `_prospective_axes` names the axes seaborn will use *without drawing anything*: an explicit `ax=` when given, otherwise `plt.gca()` and only when a figure already exists, so a call on a clean slate does not conjure one early.

Three tests, because the diff has two ways to be wrong and only one of them is the lie:

- `test_bars_already_on_the_axes_do_not_make_it_a_histogram` — the lie itself;
- `test_a_second_histogram_on_the_same_axes_still_registers` — new bars are new bars, so the diff must not over-decline;
- `test_a_histogram_drawn_without_an_explicit_axes_registers` — a snapshot that failed to find `plt.gca()` would be *empty*, which lets a stale container straight through while every explicit-`ax=` test still passes.

Reverting the snapshot to an empty set fails 1 of 9.

## Finding 2

`_drew_bars(plot: Any, before: set[int])`, plus annotations on the two new helpers.

Full suite: **1234 passed**.

---

jooyoungseo:
All three taken, and the first is a correction of my own comment rather than a change of mind.

## 1. The broad `except` — and a wrong pragma

You asked whether it is really unreachable. It isn't, and my comment saying so was simply false: `_drew_bars` runs **before** `common()`, not after, so nothing has resolved `get_axes` on that object yet. And `FigureManager.get_axes` does raise — it walks a container's children with a bare `next()`:

```python
elif isinstance(artist, BarContainer):
    return next(
        child_artist.axes for child_artist in artist.get_children()
        if isinstance(child_artist.axes, Axes)
    )
```

So the guard is load-bearing. Narrowed to `StopIteration`, with a comment that says what it is for: declining loses nothing, because `common` calls the same function a line below and would raise the same way.

## 2. `id()` reuse

Fixed rather than flagged — the snapshot holds the containers themselves. As you say, an id is unique only while its object is alive, so an id snapshot could be matched by an unrelated container allocated at a freed address, making a genuinely new histogram look pre-existing and be declined.

There is a second reason a `set` would have been wrong, which I found while making the change: `BarContainer` extends `tuple`, so it hashes **by value** — two equal containers would collapse into one. A list plus identity comparison avoids both; `n` here is the number of containers on one axes, so the quadratic scan costs nothing.

## 3. `ax` read from kwargs only

Verified and then asserted rather than commented:

```
ax kind: KEYWORD_ONLY
first positional-capable params: ['data', 'kwargs']
```

`test_seaborn_still_takes_ax_by_keyword` fails if a future seaborn allows it positionally. A comment would not have been enough here, because the failure is silent in exactly the way that matters: every test in the file passes `ax=` as a keyword and would go on passing while the snapshot came back empty and let a stale container through.

Full suite: **1235 passed**.

---

PR #391, opened 2026-08-14
fix(plotly): read a stacked bar chart as stacked

### LINKED ISSUE #390: plotly: a stacked bar chart is announced as grouped, or as two separate charts
`maidr/plotly/plotly_maidr.py` classifies multi-trace bar figures from `layout.barmode`:

```python
barmode = layout.get("barmode", "group")
...
if len(bar_traces) > 1 and barmode in ("group", "stack"):
    plot_type = PlotType.DODGED if barmode == "group" else PlotType.STACKED
```

Two things are wrong with that, and both produce a confident, wrong reading rather than an error.

## 1. The default is wrong, and inverts the relationship

Plotly's default `barmode` is **`relative`**, which *stacks*. The line above defaults to `"group"`, which *dodges*.

```
                             plotly draws              maidr says
no barmode                   relative (default)        ['dodged_bar']
barmode=relative             relative                  ['bar', 'bar']
barmode=stack                stack                     ['stacked_bar']
barmode=group                group                     ['dodged_bar']
barmode=overlay              overlay                   ['bar', 'bar']
```

Row 1 is the severe one. A reader is told the bars sit **side by side** when plotly drew them **on top of each other** — so every segment's value means something different from what is announced, and the totals a stacked chart is read for do not exist in the reading at all. Nothing errors and the numbers are all present.

## 2. `relative` is not in the tuple at all

Row 2: asking for stacking *explicitly*, the way `px.bar(color=...)` does, falls straight through to two independent `bar` layers. The stacking is simply lost.

`px.bar(df, x="g", y="v", color="h")` — the ordinary way to draw a stacked bar chart in plotly express — sets `barmode='relative'` and lands here.

## The signal is declared, so no inference is needed

Unlike the matplotlib side, where stacking has to be inferred from a `bottom=`/`left=` argument (#385), plotly states it in the layout. This is a lookup table that is missing a row and has the wrong fallback.

## Proposed fix

- Default to `"relative"`, matching plotly, rather than `"group"`.
- Treat `relative` as stacked — it is plotly's own name for a stack that allows negative values to go below the axis.
- Leave `overlay` reading as separate layers: overlaid bars genuinely are separate, drawn over one another rather than combined.

Worth pinning every barmode in a table test, since the failure is a silent mislabel and the set is small and enumerable.

## Related

`px.area` has the same shape of gap in a different place: it declares itself with `stackgroup='1'` on a `Scatter` trace, which the adapter does not read, so an area chart is announced as `line`. Filed separately.

jooyoungseo:
Closes #390.

The plotly bar classification is a lookup on `layout.barmode`. It was missing a row *and* carrying the wrong fallback:

```python
barmode = layout.get("barmode", "group")
...
if len(bar_traces) > 1 and barmode in ("group", "stack"):
```

| | plotly draws | maidr said |
|---|---|---|
| no barmode | `relative` (default) | **`['dodged_bar']`** |
| `barmode=relative` | `relative` | `['bar', 'bar']` |
| `barmode=stack` | `stack` | `['stacked_bar']` |
| `barmode=group` | `group` | `['dodged_bar']` |
| `barmode=overlay` | `overlay` | `['bar', 'bar']` |

## Row 1 inverts the relationship rather than losing it

Plotly's default `barmode` is `relative`, which **stacks**. The code defaulted to `group`, which **dodges**. A reader was told the bars sit side by side when plotly drew them on top of each other — so every segment means something other than what was announced, and the totals a stacked chart is read *for* are absent from the reading entirely. Nothing errored.

## Row 2 is how plotly express writes a stacked bar chart

`px.bar(df, x="g", y="v", color="h")` sets `barmode='relative'`. Asking for stacking explicitly fell straight through to two independent `bar` layers.

## The fix

`relative` is plotly's own name for a stack that lets negative values run below the axis, so it joins `group` and `stack` as a mode that **combines** traces. `overlay` deliberately does not: those bars are drawn over one another rather than joined, and separate layers is the honest reading.

Unlike the matplotlib side, where stacking has to be inferred from a `bottom=`/`left=` argument (#385), plotly **states** this in the layout — so this is a table to correct rather than a heuristic to design. Every value plotly accepts is now enumerated in a parametrized test, since three of the five rows were wrong in a way nothing downstream could detect and the set is small.

The default is asserted against plotly's own reported value rather than against a literal, so the two cannot drift apart silently.

## Falsification

| reverted | failing tests |
|---|---|
| default back to `group` | 2 of 8 |
| `relative` not merged | 4 of 8 |

Plus a control: a single-trace figure carries the stacking default too, so reading `barmode` alone would turn every plain bar chart into a stack of one.

## Related, filed separately

`px.area` declares itself with `stackgroup='1'` on a `Scatter` trace, which the adapter does not read — so an area chart is announced as `line`.

## Verification

`ruff check` clean. Full suite: **1243 passed**.

---

### DIFF
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -83,6 +83,20 @@ def _is_domain_trace(trace: dict) -> bool:
     return trace.get("type") in _DOMAIN_TRACE_TYPES
 
 
+#: Plotly's own default when a figure sets no ``barmode``. It stacks -- and
+#: it is the value `px.bar(color=...)` leaves behind, so it is the ordinary
+#: way a stacked bar chart arrives rather than an exotic one.
+_PLOTLY_DEFAULT_BARMODE = "relative"
+
+#: The barmodes under which plotly *combines* several bar traces into one
+#: chart, and so the ones MAIDR merges into a single layer. ``relative`` is
+#: plotly's name for a stack that lets negative values run below the axis.
+#:
+#: ``overlay`` is deliberately absent: those bars are drawn over one another
+#: rather than joined, so separate layers is the honest reading.
+_COMBINED_BARMODES = frozenset({"group", "stack", "relative"})
+
+
 class PlotlyMaidr:
     """
     Handles rendering Plotly figures as accessible MAIDR HTML.
@@ -230,8 +244,11 @@ class PlotlyMaidr:
         Groups traces by their subplot position (axis pair), then applies
         merging rules within each group:
 
-        * Multiple bar traces with ``barmode='group'`` or ``'stack'`` are
-          merged into a single :class:`PlotlyGroupedBarPlot`.
+        * Multiple bar traces that plotly combines -- ``barmode`` of
+          ``'group'``, ``'stack'`` or ``'relative'`` -- are merged into a
+          single :class:`PlotlyGroupedBarPlot`. ``'overlay'`` is not combined:
+          those bars are drawn over one another rather than joined, so they
+          stay separate layers.
         * Scatter/lines traces are split by *renderer* first — SVG traces
           apart from canvas-painted ``scattergl`` ones — because a layer's
           selector list is all-or-nothing, so a mixed layer could only claim a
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Both taken. One correction to the diagnosis, though — it doesn't change the fix, but it would be a misleading thing to leave in the thread.

## The `to_dict()` claim

> using a default that plotly's own `to_dict()` never lets you reach, since the key is present with value `None`

Measured, and it's the other way round — the key is **absent**:

```
barmode in to_dict layout: False | value: None
layout keys: ['template']
```

So `.get("barmode", "group")` really did return `"group"`, which is exactly what produced the `dodged_bar` mislabel. If the key had been present as `None`, the old code would have fallen through to `['bar', 'bar']` instead — a lost relationship rather than an inverted one, and a milder bug than the one measured.

The `or` still earns its place, just for the other reason: it covers an explicit `None` should a future plotly start exporting one. That is now what the comment says, along with the fact that no valid barmode is falsy — which was your nit, and fair.

## The optional-dependency guard

Right, and I'd missed the convention. `plotly` is an optional extra, so without `importorskip` a minimal install fails at **collection** rather than skipping — which takes the whole run down, not just this file. Guarded for both `plotly` and `pandas`, matching `conftest.py`, with a comment naming the failure it prevents.

Full suite: **1243 passed**.

---

PR #393, opened 2026-08-14
feat(plotly): read a 100% stacked bar as one

### LINKED ISSUE #338: feat: support 100% stacked bar (stacked_normalized_bar)
## Description

The MAIDR JS core defines `TraceType.NORMALIZED = 'stacked_normalized_bar'` and `TraceFactory` routes it through `SegmentedTrace`, so the runtime has supported it for some time. `maidr/core/enum/plot_type.py` has no matching member:

```python
class PlotType(str, Enum):
    BAR = "bar"
    ...
    STACKED = "stacked_bar"
    STEP = "step"
```

There is `STACKED` but no normalized variant, so py-maidr cannot emit one. Today only the Vega-Lite adapter produces this trace type — it is unreachable from Python.

## Why it is not just "a stacked bar with different numbers"

In a 100% stacked bar the segment value is a share of its own category, not an absolute count, and the totals are equal by construction. Reading it as `stacked_bar` would announce proportions as if they were counts and would suggest every category has the same total. The distinct type is what keeps the announcement honest.

## Scope

- [ ] `NORMALIZED = "stacked_normalized_bar"` in `PlotType`, with a `_DISPLAY_NAMES` entry ("100% stacked bar")
- [ ] Detect normalization in the bar patch / `grouped_barplot.py`
- [ ] Announce the share, and carry the underlying count where it is recoverable
- [ ] Tests

## Related

The same gap exists in r-maidr; a matching issue is filed there.

## Difficulty

Low.

jooyoungseo:
Closes #338.

## The defect

`layout.barnorm` is plotly's own switch for normalising each stack to a common total — `'percent'` scales to 100, `'fraction'` to 1. Either way the segment values are *shares of their category* rather than counts. MAIDR did not read it, so such a chart arrived as `stacked_bar`.

```
                          plotly draws           maidr said
barmode=stack             a stack of counts      stacked_bar          ok
barmode=stack             a stack of shares      stacked_bar          ** wrong
  + barnorm=percent       summing to 100
barmode=stack             a stack of shares      stacked_bar          ** wrong
  + barnorm=fraction      summing to 1
```

What a reader loses is not the numbers, which are announced either way, but what they *are*. A `stacked_bar` invites the reading that each segment is a count and that the categories happen to total the same; `stacked_normalized_bar` says the totals are equal by construction and the parts are proportions. Every category totalling exactly 100 reads as a striking finding about the data when it is a property of the chart.

## Why this was reachable at all

The MAIDR core has carried the type for some time:

```ts
// maidr/src/type/grammar.ts:1435
NORMALIZED = 'stacked_normalized_bar'
```

`PlotType` simply had no member to emit it with, so the type was unreachable from Python. This adds it, and reads plotly's declaration into it.

## Lookup, not heuristic

I had earlier flagged this issue as needing a heuristic I was not comfortable writing — something like "every category totals 1.0, so this must be normalised". That would name a chart from a coincidence in its data, and a chart whose parts genuinely happen to sum to 1 would be relabelled.

It turns out no heuristic is needed, because plotly states it:

```
layout.barnorm = 'percent' | barmode = 'stack'
to_dict layout keys: ['barmode', 'barnorm', 'template']
```

`barnorm` is present in `to_dict()` when set, so this is the same shape as #391's `barmode` — a lookup table, read from what the library declares.

matplotlib and seaborn have **no** equivalent declaration. A user normalises the data themselves and calls `ax.bar(bottom=...)`, and nothing in the artists says whether that was normalisation or arithmetic. Those paths are deliberately untouched; plotly is where this can be read honestly.

## Two decisions worth stating

**The dodge check stays ahead of the normalisation check.** `barnorm` normalises a *stack*, and a dodge has none to normalise — plotly ignores it there. Ordered the other way, `barmode='group' + barnorm='percent'` would announce side-by-side bars as shares of a total the chart never draws, which is plausible-looking and wrong.

**Membership of the normalising set, not key presence.** Plotly's own "off" value is the empty string, so `barnorm=""` is how a figure says *not* normalised after something set it. `"barnorm" in layout` would call that normalised — the same bug shape as the `"bottom" in kwargs` test fixed in #387.

## Behaviour

| `barmode` | `barnorm` | emitted |
|---|---|---|
| — | — | `stacked_bar` |
| `stack` | — | `stacked_bar` |
| `stack` | `percent` | `stacked_normalized_bar` |
| `stack` | `fraction` | `stacked_normalized_bar` |
| `relative` | `percent` | `stacked_normalized_bar` |
| — | `percent` | `stacked_normalized_bar` |
| `group` | `percent` | `dodged_bar` |
| `group` | — | `dodged_bar` |
| `stack` | `""` | `stacked_bar` |

Both axes of a small closed set, so every combination is named rather than left to a representative case — including the rows where `barnorm` is set and must *not* change the answer.

## Falsification

Each mechanism reverted in turn, to confirm the tests fail for the reason claimed rather than passing alongside it.

| reverted to | failures |
|---|---|
| `normalised = False` (stop reading `barnorm`) | 4 |
| normalisation checked ahead of the dodge | 2 |
| `"barnorm" in layout` (key presence, not value) | 1 |
| `NORMALIZED = "normalized_bar"` (wrong wire value) | 5 |
| display name removed | 1 |

Restored, `tests/plotly/test_plotly_barnorm.py` is 11 passed; the full suite is **1254 passed** (1243 before, plus these 11).

`ruff check` is clean on all three files. `ruff format` reports drift in `plotly_maidr.py`, but it predates this change — the file at `HEAD` fails the same check, and no reformat hunk touches the lines added here — so it is left for a separate pass rather than mixed into this diff.

---

### DIFF
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -28,6 +28,7 @@ class PlotType(str, Enum):
     PIE = "pie"
     SCATTER = "point"
     STACKED = "stacked_bar"
+    NORMALIZED = "stacked_normalized_bar"
     STEP = "step"
     SMOOTH = "smooth"
     CANDLESTICK = "candlestick"
@@ -63,6 +64,7 @@ _DISPLAY_NAMES = {
     PlotType.HEAT: "heatmap",
     PlotType.HIST: "histogram",
     PlotType.SCATTER: "scatter",
+    PlotType.NORMALIZED: "100% stacked bar",
     PlotType.STACKED: "stacked bar",
     PlotType.STACKED_AREA: "stacked area",
     PlotType.VIOLIN_BOX: "violin",
--- a/maidr/plotly/grouped_bar.py
+++ b/maidr/plotly/grouped_bar.py
@@ -6,7 +6,14 @@ from maidr.plotly.plotly_plot import PlotlyPlot, as_list
 
 
 class PlotlyGroupedBarPlot(PlotlyPlot):
-    """Extract data from multiple Plotly bar traces (dodged or stacked).
+    """Extract data from multiple Plotly bar traces (dodged, stacked or
+    normalised).
+
+    The class does not branch on *plot_type*: every combination hands the
+    traces' own ``x``/``y``/``fill`` through unchanged, and the type decides
+    only how the MAIDR core reads them. Which one applies is worked out from
+    ``layout.barmode`` and ``layout.barnorm`` by
+    :meth:`~maidr.plotly.plotly_maidr.PlotlyMaidr._extract_plots`.
 
     Parameters
     ----------
@@ -15,7 +22,8 @@ class PlotlyGroupedBarPlot(PlotlyPlot):
     layout : dict
         The Plotly figure layout.
     plot_type : PlotType
-        Either ``PlotType.DODGED`` or ``PlotType.STACKED``.
+        ``PlotType.DODGED``, ``PlotType.STACKED`` or
+        ``PlotType.NORMALIZED``.
     """
 
     def __init__(
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -88,6 +88,12 @@ def _is_domain_trace(trace: dict) -> bool:
 #: way a stacked bar chart arrives rather than an exotic one.
 _PLOTLY_DEFAULT_BARMODE = "relative"
 
+#: The values `layout.barnorm` takes when plotly normalises each stack to a
+#: common total -- its own switch for a 100% stacked bar. `percent` scales to
+#: 100 and `fraction` to 1; both mean the segment values are *shares of their
+#: category* rather than counts, which is what the distinct trace type says.
+_NORMALISING_BARNORMS = frozenset({"percent", "fraction"})
+
 #: The barmodes under which plotly *combines* several bar traces into one
 #: chart, and so the ones MAIDR merges into a single layer. ``relative`` is
 #: plotly's name for a stack that lets negative values run below the axis.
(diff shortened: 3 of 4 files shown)


jooyoungseo:
Both nits taken, in `5caf87b`.

**1. `grouped_bar.py` docstring** — you're right, and the useful part turned out to be *why* the list can grow without the class changing. Reworded to name all three types and to record that `PlotlyGroupedBarPlot` never branches on the value: it passes `x`/`y`/`fill` through unchanged, and `_extract_plots` is where the choice is made. That is the fact that made this PR a one-line classification change rather than a new extractor.

**2. Express-level test** — added, but the spelling you suggested doesn't exist:

```
px.bar  barnorm param: False
px.bar(frame, x=..., y=..., color=..., barnorm="percent")
  ** TypeError: bar() got an unexpected keyword argument 'barnorm'
```

`px.histogram` takes `barnorm`; `px.bar` does not. So the express route to a 100% stacked bar is `px.bar(...).update_layout(barnorm="percent")`, and that is what the new test drives — the point you were making still holds, since what matters is that real usage lands as `relative` + `percent` in `to_dict()` for the lookup to fire. Verified it does.

**A separate gap this turned up.** `px.histogram` *does* take `barnorm`, and that path is worse off than the one this PR fixes:

```python
px.histogram(frame, x="g", color="h", barnorm="percent")
#   layout: barmode='relative', barnorm='percent'
#   traces: ['histogram', 'histogram']
#   maidr : ['hist', 'hist']
```

Two independent histogram layers, with nothing saying they are stacked at all — let alone normalised. Same shape as #390, on the histogram path rather than the bar one, and it needs its own design because a `histogram` trace carries no `y` (plotly bins in the browser). Out of scope here; I'll file it separately rather than widen this PR.

`tests/plotly/` is 353 passed with the additions.

---

Issue #395, opened 2026-08-14
fix(plotly): a box plot's highlight selector counts boxes, not the DOM slots they occupy


jooyoungseo:
Two defects in `PlotlyBoxPlot._get_selector`, both measured in Chromium against real plotly output. Found while adding candlestick support (#343).

## The code

```python
# maidr/plotly/box.py
num_boxes = max(len(self._outlier_counts), 1)
for i in range(num_boxes):
    n = i + 1
    box_sel = f"{prefix}.boxlayer > g:nth-child({n}) > path.box"
```

The index runs `1..number-of-boxes`. Plotly's `g.boxlayer` children are **one group per trace**, not one per box, so the two only agree when the figure holds exactly one single-box trace.

## Defect 1 — a trace with several categories

One `go.Box` trace with three categories draws all three boxes inside *one* `g.boxlayer` child:

```python
go.Box(x=["a"]*15 + ["b"]*15 + ["c"]*15, y=values)
```

```
.subplot.xy .boxlayer > g:nth-child(1) > path.box     3     ** all three at once
.subplot.xy .boxlayer > g:nth-child(2) > path.box     0
.subplot.xy .boxlayer > g:nth-child(3) > path.box     0
.subplot.xy .boxlayer path.box                        3
```

So navigating to box "a" highlights all three, and boxes "b" and "c" highlight nothing. Nothing errors; the audio, braille and text are all correct, so there is no other signal that the highlight is wrong.

## Defect 2 — a neighbour that shares the layer

A `candlestick` trace draws into `g.boxlayer` too, and emits its own `path.box`. Declared before a box, it takes slot 1:

```python
go.Figure([go.Candlestick(...), go.Box(y=values)])
```

```
.subplot.xy .boxlayer > g:nth-child(1) > path.box     4     ** the candlestick's four bars
```

The box plot's selector resolves to the candlestick. A `go.Violin` does **not** cause this — it draws into `g.violinlayer` — and neither does a scatter; measured, the boxlayer trace types are `{box, candlestick}`.

## Why now

Defect 2 was unreachable in practice while candlesticks produced no layers at all. #343 makes them a supported combination, so a figure mixing the two is now something a user can reasonably build.

Defect 1 is independent of that and is the more common case: `go.Box(x=..., y=...)` with a categorical `x` is the ordinary way to draw a grouped box plot.

## Suggested shape

[...]

Issue #343, opened 2026-08-10
feat(plotly): violin, candlestick and smooth traces are missing from the Python binding


jooyoungseo:
## Description

`maidr/plotly/` covers bar, box, grouped bar, heatmap, histogram, line, multi-box, multi-line, pie, scatter and step. It does not cover **violin**, **candlestick** or **smooth**, all three of which the matplotlib/seaborn path already supports and all three of which the browser-side Plotly adapter in maidr.js handles.

## Why this is confusing for users

The same figure is accessible or not depending on which plotting library was used, with nothing explaining the difference. A user who moves a working seaborn violin plot to Plotly loses accessibility and gets no warning.

As `CLAUDE.md` notes, `maidr/plotly/` builds its MAIDR schema in Python and therefore needs its own handling per plot type — unlike `maidr/altair/`, which delegates to the upstream Vega-Lite adapter. So this cannot be inherited; it has to be written.

## Scope

- [ ] `maidr/plotly/violin.py` emitting the `violin_kde` + `violin_box` layer pair, matching the matplotlib path
- [ ] `maidr/plotly/candlestick.py` for `go.Candlestick` (and `go.Ohlc`)
- [ ] Smooth / trendline traces (`px.scatter(trendline=...)`)
- [ ] Register all three in `maidr/plotly/plotly_plot_factory.py`
- [ ] Selector generation consistent with the existing Plotly traces

## Reference

The browser-side adapter already derives these from Plotly's `_fullData` / `calcdata`, which is a useful map of where each value lives.

## Difficulty

Medium.

jooyoungseo:
Working this issue. Measured the current state and the feasibility of each of the three parts, since they turn out to differ a lot.

## Current behaviour

All of these render HTML and load MAIDR successfully. What arrives is an empty shell — no layers, nothing to navigate, and no error saying why:

```
go.Candlestick(...)                     layers: []
go.Ohlc(...)                            layers: []
go.Violin(...)                          layers: []
go.Violin(..., box_visible=True)        layers: []
```

## Smooth / trendline is already covered

`px.scatter(trendline="ols")` **works today**:

```python
px.scatter(df, x="x", y="y", trendline="ols")
#   maidr: ['line', 'point']
```

[...]

PR #396, opened 2026-08-14
feat(plotly): read candlestick and OHLC charts

jooyoungseo:
Part of #343 — the candlestick half. Violin and the smooth/trendline naming are left open; reasoning in [this comment](https://github.com/xability/py-maidr/issues/343#issuecomment-5295569176).

## The defect

`maidr/plotly/` builds its schema in Python and needs its own handling per trace type. It had none for `candlestick` or `ohlc`, and neither errored — both fell through to `PlotlyPlotFactory`, which returned `None`:

```
go.Candlestick(...)   ->  layers: []
go.Ohlc(...)          ->  layers: []
```

The HTML still rendered and MAIDR still loaded. What arrived was an empty shell: a chart with nothing to navigate, no announcement, and no error saying why.

## Transcription, not reconstruction

Both trace types state every number they draw — `open`, `high`, `low` and `close` are arrays on the trace itself. Nothing here is derived from the drawing.

Three things are deliberately **not** emitted:

- **`trend` and `volatility`** — `Candlestick`'s constructor in the core maps every point to one carrying its own, computed from OHLC. A copy emitted here could only ever be a second opinion that is discarded, and would rot silently if the core's rule changed. The matplotlib emitter omits them for the same reason.
- **`volume`** — no plotly OHLC trace carries one, and the core's section list (`['volatility', 'open', 'high', 'low', 'close']`) has no volume section, so there is nothing to invent.

`ohlc` and `candlestick` are one MAIDR type. Plotly draws a body-and-wick for one and a bar-with-ticks for the other; a reader is told open, high, low and close either way.

## The selectors were measured, not reasoned about

Plotly puts a candlestick in the DOM in three ways a plausible selector gets wrong. All counts below are from Chromium against real plotly output, 4 candles.

**1. The rangeslider.** Plotly gives a candlestick chart one *by default*, and it holds a complete second copy of the plot:

```
.trace.boxes .box                                      8    ** two per candle
.subplot.xy .boxlayer > .trace.boxes:nth-child(1) …    4
```

The duplicate lives at `g.infolayer > g.rangeslider-container > g.rangeslider-rangeplot.xy > … > g.boxlayer.mlayer.rangeplot`. That ancestor carries `xy` but **not** `subplot`, so the existing `.subplot.<id>` prefix excludes it. Setting `xaxis_rangeslider_visible=False` drops the unscoped count to 4, confirming the extra four are the slider's.

**2. `go.Box` shares the layer.** It lands in the same `g.boxlayer` and draws its own `path.box`:

```
one box + one candlestick:
  .subplot.xy .trace.boxes path.box                       5    (1 + 4)
  .subplot.xy .boxlayer > .trace.boxes:nth-child(1) …     1    the box
  .subplot.xy .boxlayer > .trace.boxes:nth-child(2) …     4    the candlestick
```

So the position is an index among `{box, candlestick}` traces, in declaration order — verified both ways round. A `go.Violin` (`g.violinlayer`) and a scatter (`g.scatterlayer`) do **not** shift it, also verified.

**3. `ohlc` is a different layer entirely.** It draws into `g.ohlclayer > g.trace.ohlc > path`, so `boxlayer` selectors match nothing for it.

Every emitted selector was then checked back against the DOM and resolves to exactly 4 — one per candle — in all six cases (single, two candlesticks, box-before, candle-before, no-rangeslider, second subplot).

## Where the code lives

`_extract_plots` builds these rather than the factory, for the same reason a lone line is built there: only it knows the trace's position among its layer-mates. The factory keeps a branch for direct construction with position 0, matching the comment convention the lines and pie branches already use.

## Two defects this turned up

Filed rather than folded in:

- **#395** — `PlotlyBoxPlot._get_selector` numbers boxes `1..n` from its own data, but `g.boxlayer` children are one group per *trace*. A three-category box trace has all three boxes in one child, so box 1 highlights all three and boxes 2–3 highlight nothing (measured 3 / 0 / 0). And a candlestick declared first takes slot 1, so the box's selector resolves to the candlestick.
- **xability/maidr#881** — the browser-side adapter's `.trace.boxes .box` has the rangeslider problem described above.

## Falsification

Each mechanism reverted in turn, against the 15 tests in the new file.

| reverted to | failures |
|---|---|
| feature removed entirely (both `_extract_plots` and factory branches) | 14 |
| `layer_position` counting every trace, not layer-mates | 2 |
| box-family set narrowed to `{candlestick}` | 1 |
| `.subplot` prefix dropped (rangeslider) | 8 |
| one DOM shape for both trace types | 1 |
| read to the longest series instead of the shortest | 1 |
| `volatility` emitted alongside | 2 |

Removing only the `_extract_plots` branch fails just 2 — the factory fallback still builds the layer, correctly, minus the positions. That is why the row above removes both.

Full suite **1270 passed** (1255 on `main` after #393, plus these 15). `ruff check` clean on all four files.

---

### DIFF
--- a/maidr/plotly/candlestick.py
+++ b/maidr/plotly/candlestick.py
@@ -1,0 +1,191 @@
+from __future__ import annotations
+
+from maidr.core.enum.plot_type import PlotType
+from maidr.plotly.plotly_plot import PlotlyPlot, as_list
+
+#: The two plotly trace types that carry an OHLC series. They differ only in
+#: how plotly draws a bar -- a filled body for ``candlestick``, a tick on
+#: either side of a vertical range for ``ohlc`` -- and not at all in what the
+#: numbers mean, so both are read as the same MAIDR layer.
+OHLC_TRACE_TYPES = frozenset({"candlestick", "ohlc"})
+
+#: The DOM layer each trace type draws into, measured in a browser rather than
+#: assumed: a ``candlestick`` shares plotly's box machinery (``g.boxlayer`` >
+#: ``g.trace.boxes`` > ``path.box``) while an ``ohlc`` has a layer of its own
+#: (``g.ohlclayer`` > ``g.trace.ohlc`` > ``path``).
+_LAYER_OF = {"candlestick": "boxlayer", "ohlc": "ohlclayer"}
+_GROUP_OF = {"candlestick": ".trace.boxes", "ohlc": ".trace.ohlc"}
+_MARK_OF = {"candlestick": "path.box", "ohlc": "> path"}
+
+#: Trace types that share ``g.boxlayer`` and so count towards a candlestick's
+#: position within it. A ``go.Violin`` draws into ``g.violinlayer`` and a
+#: scatter into ``g.scatterlayer``, so neither shifts the count.
+_BOXLAYER_TRACE_TYPES = frozenset({"box", "candlestick"})
+
+
+def is_ohlc_trace(trace: dict) -> bool:
+    """
+    Return whether *trace* carries an open/high/low/close series.
+
+    Parameters
+    ----------
+    trace : dict
+        A plotly trace dictionary.
+
+    Returns
+    -------
+    bool
+        ``True`` for a ``candlestick`` or ``ohlc`` trace.
+    """
+    return trace.get("type") in OHLC_TRACE_TYPES
+
+
+def layer_position(traces: list[dict], trace: dict) -> int:
+    """
+    Return *trace*'s zero-based position among its DOM layer-mates.
+
+    Plotly appends one ``<g class="trace ...">`` per trace to the layer its
+    type draws into, in the order the traces were declared, so a trace's
+    position among the traces sharing that layer is its ``nth-child`` index.
+    Counting all traces instead would skip a child for every scatter or violin
+    beside it and scope the selector to nothing.
+
+    Parameters
+    ----------
+    traces : list of dict
+        Every trace on the subplot, in declaration order.
+    trace : dict
+        The trace to locate.
+
+    Returns
+    -------
+    int
+        The zero-based position among traces drawing into the same layer.
+    """
+    kind = trace.get("type")
+    if kind == "candlestick":
+        mates = _BOXLAYER_TRACE_TYPES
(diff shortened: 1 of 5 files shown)


jooyoungseo:
All three taken, in `0afab79`.

**Line length — you were right, and my claim was weaker than it sounded.** That line is 92 characters:

```
tests/plotly/test_plotly_candlestick.py:135 (92)
```

`ruff check` passed because **`E501` is not in the selected rules** — the default selection is `E4,E7,E9,F`, which excludes line length entirely. So "ruff check is clean" was true and did not mean what I implied it meant; the configured `line-length = 88` is only enforced by `ruff format`, which this repo does not run in CI. Wrapped, and I checked the other four files the same way (`awk 'length > 88'`) rather than trusting the linter for it again.

**Factory branch coverage.** Added three tests to `test_plotly_plot_factory.py`, following the file's convention: `test_candlestick_trace`, `test_ohlc_trace_is_the_same_type`, and `test_candlestick_assumes_it_is_the_only_one` — the last mirroring `test_pie_assumes_it_is_the_only_one`, since it is the same "sees one trace, defaults to slot 0" reasoning. The `ohlc` one asserts the selector reaches `.ohlclayer`, so the branch that varies the DOM shape by trace type is pinned there too rather than only through `_extract_plots`.

**Mid-series gap.** Added. Confirmed it takes the branch you identified — the per-candle `except`, not the shortest-array count:

```python
open=[1.0, None, 3.0], high=[2.0, 3.0, 4.0], low=[0.0, 1.0, 2.0], close=[1.5, 2.5, 3.5]
#   -> value: ["d1", "d3"],  open: [1.0, 3.0]
```

The candle with the gap drops out and `d3` keeps its own date. Carrying the gap forward instead would shift every later candle onto the wrong date, which is the failure worth pinning.

**#395** — agreed, and to be explicit: it is pre-existing rather than introduced here. A candlestick has always occupied a `boxlayer` slot whether or not MAIDR read it; what this PR changes is that a box-plus-candlestick figure is now something a user would reasonably build.

Full suite **1274 passed**; `ruff check` clean and no line over 88 in any file this PR touches.

---

PR #397, opened 2026-08-14
feat(plotly): read violin plots

jooyoungseo:
Part of #343 — the violin half, following #396's candlestick half. Only the smooth/trendline naming is left after this, and [it is already handled](https://github.com/xability/py-maidr/issues/343#issuecomment-5296079563).

## The defect

`_extract_plots` had no branch for the `violin` trace type, so a violin fell through to `PlotlyPlotFactory`, which returned `None`:

```
go.Violin(...)                      layers: []
go.Violin(..., box_visible=True)    layers: []
```

Nothing errored. The HTML rendered and MAIDR loaded; what arrived was an empty shell with nothing to navigate and no error saying why.

A violin is now announced as two layers, matching the matplotlib path and the browser-side plotly adapter: `violin_box` summarises the distribution, `violin_kde` is the shape the chart draws. Both are built from **one** list of violins, so row *i* of the box and curve *i* of the KDE cannot come to mean different violins.

## The hard part: the curve is not in the figure

Plotly runs the kernel density estimate **in the browser**. A plotly figure carries the sample, never the curve — so unlike candlestick, this cannot be a transcription.

I flagged this on the issue earlier as the thing that might make violin dishonest to implement, the same way `vioplot`'s `sm::sm.density` blocked xability/r-maidr#132. A KDE computed with library defaults is not wrong in any way a reader could detect; it simply describes a shape the chart does not draw.

That concern turned out not to apply, and I checked rather than assumed. Plotly's rules are recoverable from the bundle:

```js
bandwidth  = max(1.059 * min(sd, iqr / 1.349) * n**-0.2, (max - min) / 100)
span       = [min - 2 * bandwidth, max + 2 * bandwidth]      // spanmode "soft"
nIntervals = ceil((span[1] - span[0]) / (bandwidth / 3))
kde(x)     = (1 / (n * bandwidth)) * sum(phi((x - xi) / bandwidth))
```

with `sd` the **sample** standard deviation and quartiles by the **Hazen** rule — neither of which is the obvious default. And, crucially, the port is *checkable*: plotly puts its own `bandwidth`, `span` and `density` in `calcdata`.

Measured in Chromium against real plotly output, eight sample sizes (n = 7, 11, 12, 13, 17, 23, 41, 43):

```
   n   pts   max |Δdensity| rel   max |Δt| abs
   7    35            4.61e-16       0.00e+00
  11    27            1.86e-15       1.78e-15
  12    37            5.17e-16       0.00e+00
  13    30            8.28e-16       1.78e-15
  17    39            5.59e-16       0.00e+00
  23    64            8.69e-16       0.00e+00
  41    41            1.08e-15       8.88e-16
  43    48            5.91e-15       3.55e-15
```

Point counts **identical** in every case. The residual is summation order between numpy and JavaScript, not the rule — which is why `test_plotly_violin_stats.py` compares to a tolerance rather than `==`, and says so. `plotly_violin_reference.json` carries three of those samples with plotly's own numbers so CI pins the agreement without a browser.

The tolerance is `1e-12`: three orders above the worst observed noise, and four below the ~1.3e-2 that reverting the sample standard deviation to the population one moves the bandwidth by. There is also a control test asserting numpy's default quantile rule *disagrees* by far more than the tolerance, so the quartile test cannot be passing by coincidence.

## The data

**One-sided curve, `density` not `width`, no pixel coordinates.** Verified against the plotly-specific JS path rather than inferred:

```ts
// src/adapters/plotly/extractor.ts, buildViolinKdeLayer
const point: ViolinKdePoint = { x: label, y: sample.t, density: sample.v };
```

The matplotlib path emits `width` and both sides because it walks a drawn `PolyCollection` outline; plotly's `calcdata.density` *is* the curve, so there is no second side to walk. Both fields are optional on `ViolinKdePoint` and the JS reader falls back `density ?? width ?? 0`. `svg_x`/`svg_y` are omitted because plotly lays the chart out in the browser — a coordinate emitted here would be a guess, and a wrong one moves the highlight to where the value is not.

**The box draws to the data, not to a Tukey fence, and separates no outliers.** That is what a plotly violin draws — the curve beside it already covers the tails — so splitting points off would announce a distinction the chart does not make.

**Categories keep first-appearance order.** Plotly's default `categoryorder` is `trace`, not sorted. There is a test where the first category sorts last, so a sorted grouping cannot pass.

## Three cases where "empty" means different things

These pull against each other, and getting one wrong breaks another. All three measured in Chromium:

| the sample | plotly draws | so MAIDR |
|---|---|---|
| all values **equal** | a degenerate violin, one `path.violin` | announces it, and it **takes** an index |
| all values **missing** | nothing at all, no element | skips it, and it takes **no** index |
| the trace is **hidden** | nothing at all, no group | skips it, and it takes **no** group slot |

I got the first two wrong initially, in opposite directions. Skipping the all-equal case took a visible category out of the reading; counting the all-missing one left the real category's selector pointing at a shape that does not exist. `stats is None` turns out to be exactly the set plotly drops, which is what makes one predicate serve both.

The hidden case was worse: the trace was **announced** — name, quartiles, whole curve — for something the reader cannot see. A legend click sets `visible="legendonly"`, so it is reached by ordinary use.

## Selectors

Every one re-verified in Chromium and resolving to exactly one element, across seven figure shapes.

```
categorical trace:   .violinlayer > g:nth-child(1) > :nth-child(1 of path.violin)
                     .violinlayer > g:nth-child(1) > :nth-child(2 of path.violin)
separate traces:     .violinlayer > g:nth-child(1) > :nth-child(1 of path.violin)
                     .violinlayer > g:nth-child(2) > :nth-child(1 of path.violin)
```

`:nth-child(N of path.violin)` rather than `path.violin:nth-child(N)` because a trace drawing its inner box puts `path.box` among the same siblings; the `of` form counts only violins.

**The box layer emits `BoxSelector` objects, not strings** — the frontend reads `selectors[i].min`, `selectors[i].q2` and so on. Every section shares one selector because plotly draws whiskers, quartile box and median as a single `path.box`; `mean` is separate because plotly draws it as its own path, named `path.mean` inside a box and `path.meanline` without one. A layer with neither a box nor a mean line emits no selectors rather than a set that resolves to nothing — but a mean line alone is enough, since `box_visible` is off by default and withholding the set would leave the common case unreachable.

## Falsification

Sixteen mechanisms reverted in turn against the 50 tests in the two new files.

| reverted to | failures |
|---|---|
| the `_extract_plots` branch removed | 19 |
| numpy's default quantile method | 9 |
| span at 3 bandwidths instead of 2 | 6 |
| points spaced by half a bandwidth | 6 |
| a flat list of box selector strings | 5 |
| population standard deviation | 3 |
| the degenerate-sample case dropped | 2 |
| no horizontal reversal | 2 |
| the hidden-trace guard removed | 3 |
| selectors emitted with no box drawn | 2 |
| sorted categories | 1 |
| trace slot not tracking rendered traces | 1 |
| `mean` sharing the box selector | 1 |
| index by sample rather than by drawn violin | 1 |
| `_ordered` reversing in place | 1 (was **0**) |
| one element name for the mean | 1 (was **0**) |
| `linspace` instead of `span[0] + i·step` | **0** |

**Three initially survived, and chasing them found real defects.**

*In-place reversal* passed every test. `render()` asks for the data first and the selectors second, so reversing in place reverses once for the data and back for the selectors — each list internally consistent, only the pairing wrong: `data[0]='B'` carrying trace A's selector. Now asserted directly.

*One element name for the mean* passed every test — because the `path.meanline` branch was **unreachable**: I was emitting no selectors at all when no box was drawn, which is plotly's default. A mean line the chart visibly draws could never be highlighted. Fixed, then the revert fails.

*`linspace`* produces **bit-identical** output on the reference samples — max absolute difference `0.000e+00`. No test can distinguish it, and my code comment claimed a difference that does not exist. Corrected the comment rather than inventing a test for a distinction that is not there.

## Also in this diff

- `subplot_css_prefix` lifted to module level in `plotly_plot.py` (the method delegates to it). A layer built from several traces needs the prefix *before* it has a plot to ask.
- `test_unsupported_trace_skipped` used `go.Violin` as its stand-in for "unsupported". It now uses `go.Sunburst`, with a sibling test asserting the violin beside a bar *is* read. The factory's equivalent likewise, plus a test recording that a violin returning `None` from the factory means "not mine to build", not "not supported".

## Filed, not folded in

- **#398** — the matplotlib `ViolinBoxPlot` rounds its statistics to 4 decimals, which is *absolute*: a dataset below 1e-4 has every quartile announced as `0.0`. Raised as a parity question in review; the plotly path emits raw floats deliberately.
- **#399** — the merged candlestick path has the same hidden-trace gap this PR fixes for violins.

Full suite **1326 passed**. `ruff check` clean, and no line over the configured 88 in any file touched.


jooyoungseo:
First two taken in `113aa2d`. The third I'd rather not match, and the reason is worth stating.

**1. `MaidrKey.MEAN`** — done. It was the one key in that dict written as a literal.

**2. Import style** — done, and you're right that there was no reason for it: `is_violin_trace` at module level with its three neighbours from the same module imported inside the branch was just an artefact of the order I wrote them in. All four are hoisted now. (The `step_shape` cycle at the top of `plotly_plot.py` is genuine and stays where it is.)

**3. Rounding parity — declining, because the matplotlib side is wrong here rather than a standard to match.**

`round(x, 4)` is *absolute*, not significant, so it does not mean "four decimals of precision" — it means "nothing below 1e-4 survives":

```
scale      raw q1                       round(q1, 4)
units      1.7901224999999998           1.7901
millis     0.0017901225                 0.0018
micros     1.7901225e-06                0.0
```

A dataset in micrograms, seconds-to-failure, or any unit where the values sit below 1e-4 has **every quartile announced as 0.0** — min, q1, q2, q3 and max all identical, so the box reads as a flat line and the distribution disappears entirely. That is the kind of failure this PR's whole approach is trying to avoid, and copying it for parity would spread it to a second backend.

Emitting the raw float is not lossy in the other direction either: the frontend formats for announcement, so what a reader hears is decided there rather than by how many digits the producer threw away.

The parity concern is real though, so I've filed it the other way round — **#398** — proposing the matplotlib path drop its rounding, with the reproduction above. Happy to take it on next if that reads right to you; I kept it out of this diff because it changes existing matplotlib output and deserves its own review rather than riding along with a plotly feature.

Thanks for the `maidr.js` cross-check on the density schema — I had verified it against the adapter source in the `maidr` repo, and independent confirmation from the bundled build is a better check than mine since that is what actually ships.

Full suite **1317 passed**; `ruff check` clean.

---

jooyoungseo:
One more fix in `bede900`, found via a review on the JS side rather than here — worth flagging since it changes what this PR emits.

**A hidden violin was being read.** `collect_violins` checked the trace type but not `visible`, and a `visible=False` / `"legendonly"` trace draws no group in `g.violinlayer` at all. Two failures, and the first is the worse:

```python
go.Figure([
    go.Violin(y=..., name="hidden", visible=False),
    go.Violin(y=..., name="shown"),
])

# was:  labels ['hidden', 'shown']
#       selectors [... g:nth-child(1) ...,  ... g:nth-child(2) ...]
# now:  labels ['shown']
#       selectors [... g:nth-child(1) ...]
```

The hidden violin was **announced** — its name, its quartiles, its whole density curve — describing a shape the reader cannot see, with nothing saying it is not drawn. That is the confidently-wrong failure this PR's whole approach is built to avoid, and I had it in the diff.

Second, it advanced the group index, so the violin that *is* drawn took `nth-child(2)` of a layer with one child and its selector matched nothing — the highlight silently stopping, with audio and text still correct.

A legend click sets `visible="legendonly"` and re-renders, so this is reached by ordinary use rather than an exotic figure.

Tests added, parametrized over both `False` and `"legendonly"`, plus one asserting a figure of only-hidden violins emits no layers at all rather than an empty pair. Falsified: removing the guard fails all three.

The same gap exists in the **already-merged** candlestick path from #396 — `layer_position` there has no visibility check either. Filed as **#399** with the reproduction rather than fixed here, since this PR is about violins and that code is on `main`. Its severity is lower: a candlestick's selector matches nothing, but nothing false is announced.

Credit where it is due — this came from the reviewer on xability/maidr#882, who spotted that my hand-rolled counting loop in the JS adapter was missing the `visible` guard every other counting helper in that file already had.

Full suite **1320 passed**; `ruff check` clean.

---

jooyoungseo:
Both findings were real bugs. Fixed in `29e1894`, and chasing the first one turned up a third.

## 1. The box selectors were the wrong shape — confirmed

You were right, and checking the JS side settled it. `BoxSelector` in the grammar is an object:

```ts
export interface BoxSelector {
  lowerOutliers: string[];  min: string;  iq: string;
  q2: string;  max: string;  upperOutliers: string[];
  mean?: string;  q1?: string;  q3?: string;
}
```

and the browser-side plotly adapter's `buildViolinBoxSelectors` builds exactly that. A flat list of strings points at nothing, and — as you say — nothing in the emitted schema would show it: the layer is present, the statistics are right, and only the highlight silently fails to resolve.

Now returns the dict, with every section sharing one selector because plotly draws whiskers, quartile box and median as a single `path.box`. Added `test_the_box_selectors_are_box_shaped`, which is the assertion that was missing.

## 2. The mean had no selector — confirmed, and the DOM differs by case

Also right. Plotly draws the mean as its own path, and the element's name depends on whether there is a box to draw it in. Measured in Chromium:

```
box_visible + meanline_visible  ->  path.violin, path.box, path.mean
meanline_visible only           ->  path.violin, path.meanline
box_visible only                ->  path.violin, path.box
```

So `mean` gets its own selector, naming `path.mean` or `path.meanline` as appropriate.

**This exposed a third bug, in my first attempt at the fix.** I initially emitted no selectors at all when no violin drew a box — matching the JS adapter, which returns `undefined` in that case. But `box_visible` is **off by default**, so that is the common case, and it made the `path.meanline` branch unreachable: a mean line that plotly visibly draws could never be highlighted. My falsification caught it — reverting `path.mean if has_box else path.meanline` to a single name failed **zero** tests, which is what dead code looks like.

Now a mean line alone is enough to emit the set. The box sections then name a `path.box` the figure has none of and resolve to nothing — the same graceful miss a boxless violin already takes when it shares a layer with one that has a box — while the mark that *is* drawn stays reachable.

## 3. The all-NaN category — you were right to doubt it

I had tested the all-*equal* case in Chromium and not the all-*missing* one, and the two pull opposite ways:

```
all values equal    ->  plotly draws a degenerate violin,  one path.violin
all values missing  ->  plotly draws nothing at all,       no element
```

Measured on a trace with one all-missing category and one real one: `calcdata` has a single entry and the group holds **one** `path.violin`. My index counted samples, so the real category came out at `:nth-child(2 of path.violin)` — matching nothing, exactly the off-by-one you predicted.

The index now counts *drawn* violins. `stats is None` turns out to be precisely the set plotly drops, since it is returned only for a sample with no finite values — the all-equal case keeps its stats because plotly keeps drawing it.

## On the `density` vs `width` fork

Fair to ask for it explicitly rather than inferred from tone. It was verified against the plotly-specific JS path, not assumed:

```ts
// src/adapters/plotly/extractor.ts, buildViolinKdeLayer
const point: ViolinKdePoint = { x: label, y: sample.t, density: sample.v };
```

One side, `density` rather than `width`, and `svg_x`/`svg_y` set only when the pixel coordinates are available. The matplotlib path emits `width` and both sides because it walks a drawn `PolyCollection` outline; plotly's `calcdata.density` *is* the curve, so there is no second side to walk. Both fields are optional on `ViolinKdePoint` and the JS reader falls back `density ?? width ?? 0`, so the two vocabularies are both understood. I've said so in the PR description now.

## Nit

Fixed — the orientation expression is one line.

## Falsification, updated

| reverted to | failures |
|---|---|
| flat list of strings | 5 |
| no selectors when no box is drawn | 2 |
| `mean` sharing the box selector | 1 |
| index by sample rather than by drawn violin | 1 |
| one element name for the mean | 1 (was **0** — see above) |

Every emitted selector re-verified in Chromium across four figure shapes (box+mean, categorical box+mean, all-missing category, mean-only): **each resolves to exactly one element**.

Full suite **1326 passed**; `ruff check` clean, no line over 88.

Thank you — three real defects from one review, and the second one only surfaced because the first was worth chasing properly.

---

PR #400, opened 2026-08-14
fix(plotly): do not read a trace the chart does not draw

### LINKED ISSUE #399: fix(plotly): a hidden trace still takes a candlestick's DOM slot
Same gap a reviewer caught in the browser-side adapter (xability/maidr#882). Filing rather than folding it into an open PR, since the code is already on `main` via #396.

## The code

```python
# maidr/plotly/candlestick.py
for candidate in traces:
    if candidate is trace:
        return position
    if candidate.get("type") in mates:
        position += 1
```

It checks the trace type but not `visible`.

## What happens

A `visible: False` or `"legendonly"` trace draws **no group** in `g.boxlayer` at all — measured in Chromium, the layer holds a single child for the candlestick beside it. Counting it puts the candlestick at the second slot of a one-child layer:

```python
go.Figure([
    go.Box(y=values, name="b", visible=False),   # draws nothing
    go.Candlestick(...),                          # boxlayer's only child
])

# emitted:
#   .subplot.xy .boxlayer > .trace.boxes:nth-child(2) path.box    -> matches 0
# should be:
#   .subplot.xy .boxlayer > .trace.boxes:nth-child(1) path.box    -> matches 4
```

Both `False` and `"legendonly"` reproduce it.

The audio, braille and text stay correct — only the highlight silently stops appearing, with nothing in the output to say why. And this is reachable by ordinary use rather than an exotic figure: clicking a legend entry sets `visible="legendonly"` and re-renders.

## Reproduction

```python
import numpy as np, plotly.graph_objects as go

rng = np.random.default_rng(3)
K = dict(x=["d1", "d2"], open=[1, 2], high=[2, 3], low=[0, 1], close=[1.5, 2.5])

for hidden in (False, "legendonly"):
    fig = go.Figure([
        go.Box(y=list(rng.normal(size=20)), name="b", visible=hidden),
        go.Candlestick(name="A", **K),
    ])
    layers = [x for r in PlotlyMaidr(fig)._flatten_maidr()["subplots"]
              for c in r for x in c.get("layers", [])]
    print(hidden, [x["selectors"] for x in layers if x["type"].value == "candlestick"])
#   False        ['... .trace.boxes:nth-child(2) path.box']
#   legendonly   ['... .trace.boxes:nth-child(2) path.box']
```

## Fix

Skip hidden traces in `layer_position`:

```python
if candidate.get("visible") in (False, "legendonly"):
    continue
```

The violin path added in #397 had the same gap and is fixed there — worse in that case, because a hidden violin was also *announced*, describing a shape the reader cannot see. A candlestick trace that is itself hidden is a separate question worth deciding at the same time: it should presumably not be registered at all rather than registered with a selector that matches nothing.

- [ ] `layer_position` skips hidden traces
- [ ] Decide whether a hidden candlestick trace should register a layer at all
- [ ] Tests for both `False` and `"legendonly"`

## Difficulty

Easy.

jooyoungseo:
Closes #399, which turned out to be one symptom of a figure-wide defect rather than a candlestick one.

## The defect

`visible=False` and `visible="legendonly"` both tell plotly to draw nothing, and plotly obeys completely — it renders **no group at all**. Measured in Chromium, two traces with one hidden:

```
bar      2 declared -> 1 group drawn   ["trace bars"]
scatter  2 declared -> 1 group drawn   ["trace scatter traced00a3b"]
pie      2 declared -> 1 group drawn   ["trace"]
box      2 declared -> 1 group drawn   (from the #397 investigation)
```

`_extract_plots` read `to_dict()["data"]` without asking, so **every** hidden trace became a layer:

```
hidden bar + shown bar     ['stacked_bar']   ** and merged into a stack
hidden scatter             ['point']
hidden box                 ['box']
hidden pie                 ['pie']
hidden histogram           ['hist']
```

Nothing errored. A reader was told about series that are not on the chart — their values, their categories, their names — with nothing saying the series is switched off.

**The bar row is the worst.** Plotly's default `barmode` stacks, so the hidden trace was *merged* with the visible one: a plain one-series bar chart announced as a `stacked_bar` of two, the invisible series contributing segments nobody can see and totals that do not exist. Not a lost relationship but an invented one — the same shape as #390, arrived at from a different direction.

**Underneath that**, every selector scoped by position among layer-mates — candlestick, violin, pie — counted the hidden trace as holding a slot it does not have, so the drawn trace's selector pointed at a group that does not exist and matched nothing. The audio, braille and text stayed correct, so only a sighted reader could tell the highlight had stopped.

A legend click sets `visible="legendonly"` and re-renders, so this is reached by ordinary use.

## How it was found

A reviewer on xability/maidr#882 spotted the missing `visible` guard in the JS adapter's candlestick counting. I fixed it there, checked whether the Python binding had the same gap, and filed #399 for the candlestick case. Writing the fix for #399 I checked the other trace types out of habit — and found all of them affected, plus the bar merge, which is worse than the case originally reported.

## The fix

One filter where the traces enter:

```python
traces = [trace for trace in fig_dict.get("data", []) if is_drawn(trace)]
```

rather than a guard in each branch. Every downstream reader — bar merging, line grouping, pie and candlestick positions, the violin pair, the subplot grid — then sees only what was drawn, and none of them has to remember to ask. That matters because the per-branch version is exactly what failed here: the guard was easy to add for one type and forget for the next six.

The per-type guard added for violins in #397 is removed in favour of it, and `is_drawn` moves to `plotly_plot.py` since it is no longer candlestick-specific.

`visible` is read by **membership** rather than truthiness, because `visible=True` is a value a figure may set explicitly and `not True` would be the wrong answer — the same shape as `barnorm` in #393.

## Falsification

| reverted to | failures |
|---|---|
| no filter at all | 23 |
| `visible=True` by truthiness rather than membership | 13 |
| `False` only, not `"legendonly"` | 12 |

The middle row is the one worth noting: reading truthiness passes every test where a trace is hidden and fails the control where one is explicitly visible, which is why that control is there.

## Tests

`tests/plotly/test_plotly_hidden_traces.py`, 25 tests. The main one is parametrized across seven trace types × both spellings of hidden, deliberately rather than left to a representative case — the filter is meant to be type-blind, and the per-type guard it replaces is what went wrong.

Plus the two specific consequences (the phantom stack, the shifted position), the pie and violin position cases, a figure of only-hidden traces (which must render rather than raise), and a control that an ordinary figure is untouched — including the order its branches emit in.

Full suite **1355 passed**. `ruff check` clean, no line over 88 in any file touched.

---

### DIFF
--- a/maidr/plotly/candlestick.py
+++ b/maidr/plotly/candlestick.py
@@ -53,7 +53,9 @@ def layer_position(traces: list[dict], trace: dict) -> int:
     Parameters
     ----------
     traces : list of dict
-        Every trace on the subplot, in declaration order.
+        The subplot's *drawn* traces, in declaration order. A hidden trace
+        gets no group in the layer, so passing one would push this trace's
+        index onto a group that does not exist.
     trace : dict
         The trace to locate.
 
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -15,6 +15,7 @@ from maidr.plotly.candlestick import is_ohlc_trace, layer_position
 from maidr.plotly.plotly_plot import (
     PlotlyPlot,
     domain_interval,
+    is_drawn,
     subplot_css_prefix,
 )
 from maidr.plotly.violin import (
@@ -300,7 +301,15 @@ class PlotlyMaidr:
         """
         fig_dict = self._fig.to_dict()
         layout = fig_dict.get("layout", {})
-        traces = fig_dict.get("data", [])
+        # Dropped once, here, rather than in each branch below. Plotly renders
+        # no group at all for a hidden trace, so one that is `visible=False`
+        # or `"legendonly"` is not on the chart: reading it announces marks
+        # nobody can see, and letting it occupy a position pushes its
+        # neighbours' selectors onto groups that do not exist. Filtering at
+        # the source means every downstream reader -- bar merging, line
+        # grouping, pie and candlestick positions, the subplot grid -- sees
+        # only what was drawn, and none of them has to remember to ask.
+        traces = [trace for trace in fig_dict.get("data", []) if is_drawn(trace)]
         # Plotly's own default is `relative`, which stacks. Defaulting to
         # `group` here meant a figure that plotly drew stacked was announced
         # as *dodged* -- not a lost relationship but an inverted one, telling
--- a/maidr/plotly/plotly_plot.py
+++ b/maidr/plotly/plotly_plot.py
@@ -605,6 +605,42 @@ def domain_interval(box: Any, key: str) -> tuple[float, float]:
         return whole
 
 
+#: The values of ``trace.visible`` that mean plotly drew nothing. ``False``
+#: hides a trace outright; ``"legendonly"`` is what clicking a legend entry
+#: sets, so a reader reaches it by ordinary use rather than an exotic figure.
+_HIDDEN = (False, "legendonly")
+
+
+def is_drawn(trace: dict) -> bool:
+    """
+    Return whether plotly drew this trace at all.
+
+    Plotly renders no group whatsoever for a hidden trace -- measured across
+    bar, scatter, pie, box and violin: two traces with one hidden produce a
+    single group in the layer. So a hidden trace must be dropped before
+    anything reads it, for two reasons at once.
+
+    It must not be *announced*: describing it tells a reader about marks that
+    are not on the chart, with nothing saying the series is switched off. And
+    it must not take a *slot*: every selector scoped by position among its
+    layer-mates would be pushed onto a group that does not exist, and a
+    selector matching nothing loses the highlight silently while the audio,
+    braille and text stay correct.
+
+    Parameters
+    ----------
+    trace : dict
+        A plotly trace dictionary.
+
+    Returns
+    -------
+    bool
+        ``True`` unless ``visible`` says otherwise. An absent ``visible`` is
+        plotly's default of drawn.
+    """
+    return trace.get("visible") not in _HIDDEN
+
+
 def subplot_css_prefix(xaxis_name: str, yaxis_name: str) -> str:
     """
     Return the CSS prefix scoping selectors to one subplot.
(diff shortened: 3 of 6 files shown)


jooyoungseo:
Taken, in `0b34a86` — and it was the right nit rather than an optional one.

The parametrization covers seven types *because* the filter is meant to be type-blind and the per-type guard it replaces is what went wrong. Leaving `ohlc` to be covered by its `candlestick` sibling was the single place the file assumed a neighbour stood in for a type, which is the assumption it exists to refuse. It also draws into `g.ohlclayer` rather than the shared `boxlayer`, so it is a genuinely different DOM path.

Now 27 tests in that file; full suite **1357 passed**.

Two things in your review worth recording, since neither was in my description:

**The hidden-pie grid case.** You're right that filtering before `axis_groups` also fixes `_subplot_domain_starts` — a hidden pie no longer reserves a row/column via its `domain` rectangle. I had not traced that path, and it is the same reasoning `_PLACED_BY_DOMAIN` already encodes: a trace maidr draws no layer for occupies no cell, since reserving one both adds an empty cell to tab through and shifts every renderable subplot beside it. So the filter lands consistently with a decision the file had already made for a different reason.

**No other unfiltered path.** Thanks for checking `fig_dict["data"]` / `to_dict()["data"]` across `maidr/plotly/` — that is the check that decides whether "filter at the source" actually holds, and it is better coming from someone who was not the one who put the filter there.

And agreed that `pie.py`'s "invisible" is a different concept — a wedge with no positive value left to draw, not `trace.visible`. Deliberately untouched.

---

jooyoungseo:
Taken, in `4b2b2b5`. Pinned exactly:

```python
assert layer["selectors"] == ".pielayer > .trace:nth-child(1) > .slice > path.surface"
```

You're right that it was weaker than its siblings, and worse than merely inconsistent: a substring check against `str()` of a possibly-list-valued field can pass on a coincidence elsewhere in the string. In a file whose whole purpose is catching a position that silently points at the wrong group, that is a test which looks like coverage and is not.

Full suite **1357 passed**; `ruff check` clean.

Thanks for tracing the flow rather than taking the description's word for it — particularly confirming that the catch-all "Remaining traces" loop is covered too. That is the branch an untyped trace like a histogram reaches, and "one filter at the entry point" is only true if *every* path downstream of it is fed from the filtered list. Checking that independently is worth more than my having asserted it.

---

Issue #402, opened 2026-08-14
fix(plotly): an explicit histogram bin size announces bins plotly never draws


jooyoungseo:
Found while measuring #401, and independent of it: the defect is identical on both orientations. Filed separately for that reason.

## What happens

`_compute_bin_edges` has two paths. The autobin port is exact — I compared it against Plotly.js for 14 figure shapes and every bin edge and count matched. The **explicit bin size** path does not, in two ways, and both add bins that carry a count of zero and no drawn element.

### 1. The `end` fallback overshoots by one bin

```python
fig = go.Figure([go.Histogram(x=V, xbins=dict(size=2))])   # V ~ N(0,1), n=60
```

| | bins |
|---|---|
| Plotly.js `calcdata` | `(-4,-2)=4  (-2,0)=27  (0,2)=27  (2,4)=2` |
| py-maidr | `(-4,-2)=4  (-2,0)=27  (0,2)=27  (2,4)=2  ` **`(4,6)=0`** |

`end` is computed as `math.ceil(arr.max() / size) * size + size`. With `max = 3.323, size = 2` that is `4 + 2 = 6`; plotly stops at `4`. The `+ size` has no counterpart in plotly's arithmetic — `start` uses the matching `math.floor(arr.min() / size) * size` with no such term, and it agrees with plotly exactly.

### 2. An explicit `start`/`end` wider than the data is taken verbatim

```python
go.Histogram(x=V, xbins=dict(start=-10, end=10, size=1))
```

Plotly's *resolved* spec keeps `start: -10, end: 10`, but its `calcdata` holds 7 bins running `-3` to `4` — the data's own extent, on the caller's grid. py-maidr emits all 20, so thirteen empty bins are announced.

This is not plotly discarding empty bins in general. Interior ones survive: a bimodal sample with a hole in the middle gives 7 bins from both, the two empty ones included, agreeing exactly.

```
gap sample: uniform(0,3) x40  +  uniform(9,12) x40
both:  (0,2)=28  (2,4)=12  (4,6)=0  (6,8)=0  (8,10)=12  (10,12)=28
```

So the rule plotly follows is: **bins span the data on the caller's grid; empty bins outside the data range are not emitted, empty bins inside are.**

## Why it matters

[...]

PR #403, opened 2026-08-14
fix(plotly): read a histogram's bins from the axis it bins

### LINKED ISSUE #401: fix(plotly): a horizontal histogram announces nothing at all
Found while measuring #394. Filed separately because it is a different defect with a much smaller fix, and #394's grouping work would otherwise inherit the same blind spot.

## What happens

`px.histogram(y=...)` is the ordinary way to draw a histogram with the bins running up the y axis. Plotly sets `orientation='h'` and puts the sample in the trace's `y`:

```python
import numpy as np, pandas as pd, plotly.express as px

rng = np.random.default_rng(3)
frame = pd.DataFrame({"v": rng.normal(0, 1, 60)})

fig = px.histogram(frame, y="v")
print(PlotlyMaidr(fig)._flatten_maidr()["subplots"][0][0]["layers"][0])
```

```json
{
  "type": "hist",
  "axes": {"x": {"label": "count"}, "y": {"label": "v"}},
  "data": [],
  "selectors": ".subplot.xy .trace.bars .point > path"
}
```

A layer, correctly typed, with both axes correctly named — and nothing in it. The labels are right because they come from `layout`, not from the trace, so the emptiness is not visible from the schema's metadata: it looks like a histogram of nothing rather than a histogram that failed to read.

The bars are drawn. Reading `gd.calcdata` back out of Plotly.js in Chromium for the same figure gives 13 points with `orientation: 'h'`, so this is purely a read failure on our side.

## Cause

`PlotlyHistogramPlot._extract_plot_data` reads `x` and gives up when it is absent:

https://github.com/xability/py-maidr/blob/main/maidr/plotly/histogram.py#L176-L180

```python
raw_x = self._trace.get("x", None)
if raw_x is None:
    return []
```

A horizontal histogram carries `y` instead. Everything downstream — `_compute_bin_edges`, the autobin port, `_extract_categorical_data` — is orientation-agnostic arithmetic that would work unchanged on the other array; only the pick and the emitted key names differ.

## Scope

The sample should be read from whichever of `x`/`y` the trace carries, chosen by `orientation` rather than by which key happens to be present, and the bin bounds emitted against that axis (`yMin`/`yMax` with the count on `x`) so the announced extent matches the axis the bins are drawn along.

Both single and multi-trace horizontal histograms are affected; #394 covers what multi-trace figures additionally get wrong about stacking.

## Difficulty

Small.

jooyoungseo:
Closes #401.

A histogram bins one axis and counts into the other. `PlotlyHistogramPlot` read the sample from `x` and returned early when it was absent, so every horizontal histogram emitted a layer with an empty `data` list:

```json
{
  "type": "hist",
  "axes": {"x": {"label": "count"}, "y": {"label": "v"}},
  "data": [],
  "selectors": ".subplot.xy .trace.bars .point > path"
}
```

Correctly typed, both axes correctly named — and nothing to navigate. The labels are right because they come from `layout` rather than the trace, so nothing in the schema's metadata shows the emptiness. It reads as a histogram of nothing rather than one that failed to read.

## Which axis is binned

Not always stated. `px.histogram(y=...)` writes `orientation` onto the trace; `go.Histogram(y=...)` writes nothing and leaves Plotly.js to infer it. Rather than guess the rule, I read `gd._fullData[i].orientation` back out of Chromium for each shape:

| trace | plotly's orientation | binned |
|---|---|---|
| `go.Histogram(x=v)` | `v` | `x` |
| `go.Histogram(y=v)` | `h` | `y` |
| `px.histogram(y="v")` | `h` | `y` |
| `go.Histogram(x=cats, y=vals, histfunc="sum")` | `v` | `x` |
| `go.Histogram(y=cats, x=vals, histfunc="sum")` | `v` | `x` |
| `go.Histogram(x=v, orientation="h")` | `h` | `y` |

The last row settles the precedence, and it is the one I would have got wrong by reasoning: plotly honours the explicit attribute and bins the **absent** `y`, drawing an empty trace rather than falling back to `x`. So an explicit `orientation` wins outright, and only in its absence do the arrays decide.

## The bin spec follows the binned axis

`_compute_bin_edges` read `xbins`/`nbinsx` for every trace. Measured, plotly discards the other axis's spec outright rather than falling back to it:

| figure | plotly draws |
|---|---|
| `go.Histogram(y=v, ybins=dict(size=2))` | 4 bins of width 2 |
| `go.Histogram(y=v, xbins=dict(size=2))` | 13 bins of width 0.5 — **autobinned, spec ignored** |
| `go.Histogram(y=v, nbinsx=4)` | 13 bins — **ignored** |
| `go.Histogram(x=v, ybins=dict(size=2))` | 13 bins — **ignored** |

So reading `xbins` for everything did both halves wrong on a horizontal trace: honoured a spec plotly throws away, missed the one it uses.

Categorical samples move with it. Plotly draws string data as a count bar chart, and on a horizontal one the labels belong on `y`; left on `x` they would be announced as counts and the counts as labels.

## Verification

I built a like-for-like harness rather than comparing by eye: the same figure objects serialized to JSON, drawn by the real `plotly.min.js` in Chromium, and `gd.calcdata[0]` compared elementwise against the emitted layer. A bin is `[p - size/2, p + size/2)` with `size` from `gd._fullData[0]`.

**14 of 18 shapes agree exactly** — every bin bound and every count, on both orientations. Before this change all 9 horizontal shapes emitted nothing at all.

The 4 that still disagree are all on the explicit-bin-size path, and all fail *identically* on `x` and `y`, which is how I know they are not mine: an extra empty bin past the end (`xbins=dict(size=2)`), and empty bins kept when an explicit `start`/`end` is wider than the data. Filed as **#402** with the measurements. This PR does not fix it, and does extend its reach — a horizontal trace with `ybins` now takes that path, where before it emitted nothing.

Worth recording one hypothesis I had and killed: I first read those 4 as plotly discarding empty bins. It does not. A bimodal sample with a hole in the middle gives 7 bins from both sides, the two empty ones included, agreeing exactly. Only bins outside the data's own extent are dropped. My original ad-hoc check had hardcoded the bin width and looked convincing.

## Falsification

Each mechanism reverted in turn, against the 20 new tests:

| reverted | failures |
|---|---|
| `binned_axis` always returns `"x"` (the original bug) | 13 |
| explicit `orientation` ignored, arrays only | 2 |
| bin spec always read off `x` | 4 |
| `orientation` key dropped from the schema | 7 |
| categorical data always on `x` | 1 |

## Checks

- `uv run pytest tests/` — **1377 passed**
- `ruff check` on both changed files — clean. The repo has 52 pre-existing findings on `main`; the count is unchanged.
- No line over 88 characters in either file (checked directly — `ruff` does not select `E501`).

---

### DIFF
--- a/maidr/plotly/histogram.py
+++ b/maidr/plotly/histogram.py
@@ -290,16 +290,23 @@ class PlotlyHistogramPlot(PlotlyPlot):
             )
         return data
 
-    def _extract_categorical_data(self, x: list) -> list[dict]:
+    def _extract_categorical_data(self, values: list) -> list[dict]:
         """Count occurrences of categorical values and return bar-format data.
 
         Mirrors how seaborn ``countplot`` produces ``type: "bar"`` schemas.
         The plot type is switched from HIST to BAR so the JS side renders
         it as a bar chart with proper categorical navigation.
+
+        Parameters
+        ----------
+        values : list
+            The binned axis's sample -- ``trace["y"]`` on a horizontal trace,
+            ``trace["x"]`` on a vertical one. Named for the role rather than
+            the axis, since either can be the one that holds it.
         """
         # Preserve order of first appearance.
         counts: dict[str, int] = {}
-        for val in x:
+        for val in values:
             key = str(val)
             counts[key] = counts.get(key, 0) + 1
 
--- a/maidr/plotly/histogram.py
+++ b/maidr/plotly/histogram.py
@@ -166,44 +166,126 @@ def _auto_shift_bins(
     return bin_start
 
 
+#: Plotly's spelling of a horizontal trace.
+_HORIZONTAL = "h"
+
+
+def binned_axis(trace: dict) -> str:
+    """Return which of ``x``/``y`` plotly bins for *trace*.
+
+    A histogram bins one axis and counts into the other, and which one is not
+    always stated: ``px.histogram(y=...)`` writes ``orientation`` onto the
+    trace, but ``go.Histogram(y=...)`` writes nothing and lets Plotly.js infer
+    it. Reading only ``x`` therefore left every horizontal histogram with no
+    data at all (#401).
+
+    The rule below is Plotly.js's own, measured rather than assumed --
+    ``gd._fullData[i].orientation`` read back out of Chromium for each shape:
+
+    ==================================================  ===========  ======
+    trace                                               orientation  binned
+    ==================================================  ===========  ======
+    ``go.Histogram(x=v)``                               ``v``        ``x``
+    ``go.Histogram(y=v)``                               ``h``        ``y``
+    ``px.histogram(y="v")``                             ``h``        ``y``
+    ``go.Histogram(x=cats, y=vals, histfunc="sum")``    ``v``        ``x``
+    ``go.Histogram(y=cats, x=vals, histfunc="sum")``    ``v``        ``x``
+    ``go.Histogram(x=v, orientation="h")``              ``h``        ``y``
+    ==================================================  ===========  ======
+
+    So an explicit ``orientation`` wins outright, and only in its absence does
+    the presence of the arrays decide. The last row is the one that settles
+    the precedence: plotly honours the attribute and bins the *absent* ``y``,
+    drawing an empty trace rather than falling back to ``x``.
+
+    Parameters
+    ----------
+    trace : dict
+        A ``histogram`` trace dict.
+
+    Returns
+    -------
+    str
+        ``"x"`` or ``"y"``.
+    """
+    orientation = trace.get("orientation")
+    if orientation is not None:
+        return "y" if orientation == _HORIZONTAL else "x"
+    if trace.get("y") is not None and trace.get("x") is None:
+        return "y"
+    return "x"
+
+
 class PlotlyHistogramPlot(PlotlyPlot):
     """Extract data from a Plotly histogram trace."""
 
     def __init__(self, trace: dict, layout: dict, **kwargs: str) -> None:
         super().__init__(trace, layout, PlotType.HIST, **kwargs)
+        self._binned = binned_axis(trace)
+
+    @property
+    def _horizontal(self) -> bool:
+        """Whether the bins run along the y axis."""
+        return self._binned == "y"
+
+    def render(self) -> dict:
+        """Add ``orientation`` to the base schema.
+
+        The core reads a histogram's bin bounds from ``xMin``/``xMax`` or from
+        ``yMin``/``yMax`` depending on this, so a horizontal layer that did not
+        carry it would have its counts announced as the bin range.
+        """
+        schema = super().render()
+        schema[MaidrKey.ORIENTATION] = "horz" if self._horizontal else "vert"
+        return schema
 
     def _get_selector(self) -> str:
         return f"{self._subplot_css_prefix()}.trace.bars .point > path"
 
     def _extract_plot_data(self) -> list[dict]:
-        raw_x = self._trace.get("x", None)
-        if raw_x is None:
+        raw = self._trace.get(self._binned, None)
+        if raw is None:
             return []
-        x = as_list(raw_x)
+        values = as_list(raw)
 
         # Detect categorical (string) data — Plotly renders these as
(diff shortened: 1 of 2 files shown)


jooyoungseo:
One thing worth adding, because it decides whether this PR is the whole fix or only half of it: **the JS side already handles a horizontal histogram.** No companion change in `xability/maidr` is needed.

I checked rather than assumed. `Histogram extends AbstractBarPlot` and branches on `this.orientation` in all three places that matter — `description` (bin range, stats, data-table headers and row order), `text` (the announced range), and the `barValues` the bar base class builds. `resolveOrientation` already lists `TraceType.HISTOGRAM` among the oriented types, with tests.

To be sure that reading was right I put a throwaway jest spec against `Histogram` with a hand-built `orientation: 'horz'` layer, in the exact shape this PR now emits:

```ts
data: [
  { y: 1, x: 4,  yMin: 0, yMax: 2, xMin: 0, xMax: 4 },
  { y: 3, x: 33, yMin: 2, yMax: 4, xMin: 0, xMax: 33 },
]
```

All three assertions pass unmodified:

- bin range → `"0 to 4"`, read from `yMin`/`yMax`
- min/max value → `4` / `33`, read from `x`
- data table → headers `['v', 'count', 'Bin Min', 'Bin Max']`, first row `[1, 4, 0, 2]`

So the bin lands on the axis the reader navigates and the count on the other, which is what the emitted keys were chosen for. The spec was a probe, not a deliverable — it asserts existing behaviour that already has coverage through `orientation.test.ts`, so I deleted it rather than adding a test that pins nothing new.

That leaves the emitter as the only place this was broken, which is what the diff changes.

---

jooyoungseo:
Thanks — all three points addressed, and the middle one turned out to be bigger than the review framed it.

## Both cosmetic points, applied (`239a2ad`)

**Parameter rename.** `_extract_categorical_data(self, x)` → `values`, with a docstring naming it for the role rather than the axis. You are right that it reads as a contradiction now that it receives `trace["y"]` half the time.

**RNG stability.** Rather than accept the low risk I removed the dependency: the 60-point sample is written out as a literal instead of regenerated from `default_rng(3)`. Worth doing because of *how* it would fail — the expectations are what Plotly.js drew for those particular numbers, so a numpy that changed the `PCG64` stream would leave every bin edge and count describing a sample the test no longer uses. It would simply start failing, with nothing in the file to say why.

## `histfunc` — filed as #405, and it is not only `sum`

I measured it before filing rather than taking the framing, and it widened twice.

**Every aggregating mode is ignored, not just `sum`.** Categorical, `x=list("abcabcabc")`, `y=[1..9]`:

| `histfunc` | plotly draws | py-maidr says |
|---|---|---|
| `count` *(default)* | `3, 3, 3` | `3, 3, 3` ✓ |
| `sum` | `12, 15, 18` | `3, 3, 3` |
| `avg` | `4, 5, 6` | `3, 3, 3` |
| `max` | `7, 8, 9` | `3, 3, 3` |
| `min` | `1, 2, 3` | `3, 3, 3` |

**And it is not confined to the categorical path.** Numeric `x` with an explicit bin grid:

| `histfunc` | plotly draws | py-maidr says |
|---|---|---|
| `sum` | `30, 70, 110` | `2, 2, 2` |
| `avg` | `15, 35, 55` | `2, 2, 2` |

That second row is the worse of the two. On the categorical path a reader gets `3` where the chart shows `12` — plainly a different quantity. On the numeric path they get `2` where the chart shows `30`, in a layer whose bin bounds are all correct, so nothing else in the announcement looks wrong.

Agreed it stays out of scope here. One note for whoever takes #405: the value array is not "whichever one is not categorical". This PR's precedence table has `go.Histogram(x=cats, y=vals, histfunc="sum")` and `go.Histogram(y=cats, x=vals, histfunc="sum")` both resolving to `orientation: v`, so plotly bins `x` in **both** — `binned_axis` already answers the question and the other array is the values.

## Re-checked

- `uv run pytest tests/` — **1377 passed**
- `ruff check` on both files — clean
- `ruff format --diff` — the test file is clean. `histogram.py` shows 18 diff lines, identical with and without this branch's changes, so no drift added.
- No line over 88 characters.

---

Issue #405, opened 2026-08-14
fix(plotly): histfunc is ignored, so an aggregating histogram is read as counts


jooyoungseo:
Raised in review on #403 as a pre-existing gap on the `histfunc="sum"` path. Measuring it before filing turned up two things that widen it beyond that framing, so recording them here rather than in the PR thread.

## What happens

`histfunc` selects what a histogram bar *measures*. py-maidr calls `np.histogram` (and, on the categorical path, counts occurrences), so every mode is announced as a count.

### It is not only `sum`, and not only the categorical path

Categorical, `x=list("abcabcabc")`, `y=[1..9]`:

| `histfunc` | plotly draws | py-maidr says |
|---|---|---|
| `count` *(default)* | `3, 3, 3` | `3, 3, 3` ✓ |
| `sum` | `12, 15, 18` | `3, 3, 3` |
| `avg` | `4, 5, 6` | `3, 3, 3` |
| `max` | `7, 8, 9` | `3, 3, 3` |
| `min` | `1, 2, 3` | `3, 3, 3` |

Numeric, `x=[0.1,0.4,1.2,1.7,2.3,2.8]`, `y=[10..60]`, `xbins=dict(start=0, end=3, size=1)`:

| `histfunc` | plotly draws | py-maidr says |
|---|---|---|
| `sum` | `30, 70, 110` | `2, 2, 2` |
| `avg` | `15, 35, 55` | `2, 2, 2` |

So the numeric binning path has it too: the bins are right and the values are the bin populations rather than the aggregate. Only `count` is correct, and only because it is what we already compute.

The numeric row is the worse of the two. On the categorical path a reader gets `3` where the chart shows `12` — plainly a different number. On the numeric path they get `2` where the chart shows `30`, in a layer whose bin bounds are all correct, so nothing else in the announcement looks wrong.

## Ground truth

`gd.calcdata[0][i].s` after `Plotly.newPlot` in Chromium, for the figures above.

## Scope

`histfunc` picks the aggregate; `y` (or `x`, on a horizontal trace — see #401) supplies the values to aggregate; the binning stays exactly as it is. `count` keeps its current path since there is nothing to aggregate.

[...]

PR #406, opened 2026-08-14
fix(plotly): emit the histogram bins plotly actually draws

### LINKED ISSUE #402: fix(plotly): an explicit histogram bin size announces bins plotly never draws

jooyoungseo:
Closes #402.

Three defects on the explicit-bin-size path, all identical on both orientations. An empty bin is not a harmless extra row: plotly draws no `.point` element for one, and the layer's selector resolves positionally, so a phantom bin shifts the highlight of every bin after it — and a *leading* phantom shifts all of them. Same desync class as #354 and #378.

## 1. Empty edge bins were emitted

Plotly emits bins from the first that holds an observation to the last, and keeps every empty bin between them. It does not emit the empty ones outside that span.

```python
go.Histogram(x=[-2.8, -1.2, 0.3, 1.1, 2.4, 3.3], xbins=dict(start=-10, end=10, size=1))
```

Plotly's *resolved* spec keeps `start: -10, end: 10` and draws **7** bins, `-3` to `4`. py-maidr emitted **20**, thirteen of them phantom and all thirteen before the first real bin.

## 2. The bin start skipped plotly's anti-clustering shift

Only the autobin path ran `_auto_shift_bins`; an explicit `size` took a bare `floor(min / size) * size`.

| figure | plotly draws from | py-maidr said |
|---|---|---|
| `go.Histogram(x=[0, 1, 2, 3, 4], xbins=dict(size=2))` | `-0.5` | `0` |
| `go.Histogram(x=[-4, -3, 0, 1], xbins=dict(size=2))` | `-4.5` | `-4` |
| `go.Histogram(x=[0.5, 1.5, 4.0], xbins=dict(size=2))` | `-1.0` | `0` |

Every value in the first case is an integer, so on a grid starting at 0 they all sit exactly on bin edges; plotly moves the grid half a unit down. The result is that each value was announced in a *different bin than the one drawn around it*, with the bin count unchanged — invisible to anything checking lengths.

This one wasn't in the issue. I found it only because fixing (1) made me test values sitting exactly on the grid. The existing `_auto_shift_bins` reproduces all three of plotly's starts exactly; it was simply never called from this branch.

## 3. The `end` fallback was expressed on a grid the bins do not use

`ceil(max / size) * size + size` → one bin past the last value, measured from `start`.

**This change is not observable, and I want to be explicit about that** — its falsification row returns 0 failures. Both forms reach at least the upper edge of the bin holding the maximum, and the trim from (1) removes the surplus either way. I kept it because once the shift is applied, `ceil(max / size) * size` is computed on a different origin than the bins actually use, which reads as a bug even where it isn't. I could not write a test that distinguishes the two, and did not invent one.

## The rule, and the reading it replaced

I had "bins span the data on the caller's grid, clamped to `[start, end)`". It fits every window wider than the data. A window narrower on **both** sides refutes it:

`xbins=dict(start=-1, end=2, size=1)` on that same sample → plotly draws **2** bins; clamping predicts **3**. Data exists below `start`, so clamping keeps bin `(-1, 0)` — and plotly drops it, because nothing landed *in* it. The out-of-window values are discarded rather than piled into the edge bin.

What survives is just: **trim empty bins from both ends, keep every one between them.** That also subsumes (3)'s overshoot without it needing separate arithmetic.

Two behaviours that were *already* correct and are easy to break while fixing this — data outside an explicit window being dropped rather than clipped, on each side — now have tests of their own.

## Verification

The same harness as #403: the figure objects serialized to JSON, drawn by the real `plotly.min.js` in Chromium, and `gd.calcdata[0]` compared elementwise against the emitted layer.

**29 of 29 shapes now agree** — every bin bound and every count, both orientations. Before: **16 of 29**.

That 16 is worth stating precisely, because I first estimated it at 25 and was wrong by a wide margin. Six of the thirteen disagreements emitted the *right number of bins on the wrong grid*, so a comparison by length alone would have called them equal and I would have reported a much smaller defect than the one that was there.

## Falsification

| reverted | failures |
|---|---|
| no trim — emit every bin | 3 |
| trim every empty bin, interior included | 6 |
| no bin-start shift on the explicit-size path | 4 |
| shift applied even when `start` is given | 4 |
| old `end` fallback (`ceil + size`) | **0** — see (3) above |

## Checks

- `uv run pytest tests/` — **1396 passed** (up 19)
- `ruff check` on both changed files — clean
- `ruff format --diff` — test file clean; `histogram.py` shows 18 lines, identical with and without this branch
- No line over 88 characters

---

### DIFF
--- a/maidr/plotly/histogram.py
+++ b/maidr/plotly/histogram.py
@@ -108,10 +108,26 @@ def _auto_shift_bins(
     Exact port of Plotly.js ``autoShiftNumericBins`` from
     ``src/plots/cartesian/axes.js``.
 
+    *bin_start* need only be **some** multiple of *dtick* within one *dtick*
+    of the true unshifted start -- not one particular formula. The two callers
+    rely on that: the autobin path passes
+    ``ceil(data_min / dtick) * dtick - dtick`` and the explicit-size path
+    ``floor(data_min / size) * size``. Those agree except when ``data_min`` is
+    itself a multiple of the width, where they differ by exactly one bin, and
+    the branches below then correct both to the same answer -- ``near_edge``
+    is trivially true of a seed equal to ``data_min``, so the no-shift
+    fallback cannot fire for it either.
+
+    That convergence is a property of the branches rather than of the
+    arithmetic that produced the seed, so it is worth keeping deliberate:
+    ``test_plotly_histogram_bins.py`` asserts the two seeds land on the same
+    start across a spread of samples and widths, which would otherwise be
+    rediscovered by hand the next time this function is touched.
+
     Parameters
     ----------
     bin_start : float
-        Initial bin start (a round multiple of *dtick*).
+        A multiple of *dtick* within one *dtick* of the unshifted start.
     data : np.ndarray
         Raw data values.
     dtick : float
@@ -327,8 +343,7 @@ class PlotlyHistogramPlot(PlotlyPlot):
         count_min, count_max = bounds[counted]
 
         data = []
-        for i in range(first, last + 1):
-            count = counts[i]
+        for i, count in enumerate(counts[first : last + 1], start=first):
             low = float(bin_edges[i])
             high = float(bin_edges[i + 1])
             data.append(
--- a/maidr/plotly/histogram.py
+++ b/maidr/plotly/histogram.py
@@ -170,6 +170,55 @@ def _auto_shift_bins(
 _HORIZONTAL = "h"
 
 
+def _occupied_span(counts: np.ndarray) -> tuple[int | None, int | None]:
+    """Return the first and last bin index that anything landed in.
+
+    Plotly emits bins from the first that holds an observation to the last,
+    and keeps every empty bin between them. It does **not** emit the empty
+    ones outside that span, however the grid came to reach past the data
+    (#402).
+
+    The rule is exactly that -- trim the ends, keep the middle -- and it took
+    a figure narrower than its data on both sides to establish it. The reading
+    it replaced was "span the data, clamped to the caller's ``[start, end)``",
+    which fits every wider window and predicts one bin too many here:
+
+    ==========================================  ==============  ============
+    ``xbins`` on ``[-2.8, -1.2, .3, 1.1, 2.4, 3.3]``  Plotly.js       clamping
+    ==========================================  ==============  ============
+    ``start=-1, end=2, size=1``                 ``(0,1) (1,2)`` adds ``(-1,0)``
+    ==========================================  ==============  ============
+
+    Data exists below ``start``, so clamping keeps bin ``(-1, 0)``; plotly
+    drops it because nothing landed *in* it, discarding the out-of-window
+    values rather than piling them into the edge bin.
+
+    Trimming subsumes the other half of #402 as well. The explicit-size path
+    computed ``end`` one bin past plotly's, and that surplus bin is empty and
+    at the end, so it goes without the arithmetic needing to be reasoned about
+    separately.
+
+    Empty bins matter beyond the announced values: plotly draws no ``.point``
+    element for one, and the layer's selector resolves positionally, so a
+    phantom bin shifts the highlight of every bin after it.
+
+    Parameters
+    ----------
+    counts : np.ndarray
+        Per-bin counts, in bin order.
+
+    Returns
+    -------
+    tuple[int | None, int | None]
+        Inclusive first and last occupied index, or ``(None, None)`` when
+        every bin is empty.
+    """
+    occupied = np.flatnonzero(counts)
+    if not occupied.size:
+        return None, None
+    return int(occupied[0]), int(occupied[-1])
+
+
 def binned_axis(trace: dict) -> str:
     """Return which of ``x``/``y`` plotly bins for *trace*.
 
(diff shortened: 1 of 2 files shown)


jooyoungseo:
The seed-divergence catch is a good one, and I measured it rather than taking the trace on faith — it holds.

## The two seeds converge, 0 of 48

Across eight samples chosen to hit the divergent condition and six widths each:

| sample | why it is here |
|---|---|
| `[0, 1, 2, 3, 4]` | `data_min` is a multiple; all-integer branch |
| `[-4, -3, 0, 1]` | same, shifting the other way |
| `[0.5, 1.5, 4.0]` | near-edge branch, not all integers |
| `[2.0, 3.7, 5.1, 8.9]` | `data_min` a multiple, nothing else is |
| `[0.0, 2.0, 4.0, 6.0]` | every value on an edge |
| `[-6.0, -4.0, -2.0, 0.0]` | the same, negative |
| `[1.0, 1.0000001]` | no spread to speak of |
| `[0.001, 0.002, 0.004]` | sub-unit, where `msexp` changes decade |

Widths `0.001, 0.5, 1, 2, 3, 5`. **0 of 48 combinations diverge.** So your reading of both branches self-correcting is right, including the point that the plain `return bin_start` fallback cannot fire when the seed equals `data_min` — `near_edge(data_min)` is trivially true there.

## Pinned, not just commented (`b06e38f`)

You suggested a one-line comment. I did that, but a comment is exactly what a refactor walks past, and your own framing — "could silently break" — is the argument for a test. So both:

- `_auto_shift_bins`'s docstring now states the actual contract: the seed need only be **some** multiple of the width within one width of the true unshifted start, not one particular formula. The parameter description said "a round multiple of *dtick*", which is true of both seeds and does not say they may differ.
- `TestBothCallersSeedTheShiftEquivalently` asserts the two seeds land on the same start across that whole grid. If someone rewrites the branches so the convergence stops holding, 48 parametrised cases say so.

I did not extract a shared seed helper. That would remove the coupling by making the two paths pass the same value — but they are ports of two *different* pieces of Plotly.js arithmetic, and collapsing them would make the port harder to check against upstream than the invariant is to state.

## Also applied

- `for i, count in enumerate(counts[first : last + 1], start=first)`.
- `histfunc` is on the radar: filed as #405 while working #403. It is broader than `sum`/`avg` — every aggregating mode is ignored, and the numeric binning path has it too, where a reader gets `2` for a bar the chart draws at `30` inside a layer whose bin bounds are all correct.

## Re-checked

- `uv run pytest tests/` — **1444 passed** (up 48 from the new parametrisation)
- The Chromium comparison still **29 of 29**, so neither change moved an emitted bin
- `ruff check` clean, `ruff format --diff` clean, no line over 88

---

PR #408, opened 2026-08-14
fix(plotly): aggregate a histogram's bars the way histfunc does

### LINKED ISSUE #405: fix(plotly): histfunc is ignored, so an aggregating histogram is read as counts

jooyoungseo:
Closes #405.

`histfunc` selects what a bar *measures*. `_extract_plot_data` called `np.histogram`, which returns bin populations, and never read the attribute — so every aggregating mode was announced as a count.

## Two ways it is wider than the review's framing

**Not only `sum`.** Categorical, `x=list("abcabcabc")`, `y=[1..9]`:

| `histfunc` | plotly draws | py-maidr said |
|---|---|---|
| `count` *(default)* | `3, 3, 3` | `3, 3, 3` ✓ |
| `sum` | `12, 15, 18` | `3, 3, 3` |
| `avg` | `4, 5, 6` | `3, 3, 3` |
| `min` | `1, 2, 3` | `3, 3, 3` |
| `max` | `7, 8, 9` | `3, 3, 3` |

**Not only the categorical path.** Numeric binning has it too, and that case is worse: a reader gets `2` where the chart draws `30`, inside a layer whose bin bounds are all correct, so nothing else in the announcement looks wrong.

## Three things I could not have reasoned out

Each was measured against `gd.calcdata` rather than derived, and each would have produced a plausible-looking wrong answer.

**1. `avg`/`min`/`max` drop an empty bin; `count`/`sum` keep it as zero.** Over a sample with a two-bin gap in the middle, `count` and `sum` give six points and the other three give **four** — and the two dropped are *interior*, so this is not #402's edge trim reaching further. There is no average of nothing to announce.

**2. Any `histnorm` brings those bins back as zeros.** `avg` alone gives four points; `avg` with any of the four norms gives six. Measured across all three functions × all four norms — uniform. So the two attributes do **not** compose as one step after the other, and an implementation that applied them in sequence would be wrong for exactly the figures that use both.

**3. A non-numeric value is dropped, not counted as zero.** The value array is not guaranteed numeric — `go.Histogram(y=cats, x=vals, histfunc="sum")` bins `x` and hands the *category strings* to the aggregate, and plotly renders it without complaint where a plain `float()` conversion raised. With `y=[1,"z",3,4,"w",6,7,8,9]`, category `b` holds `['z','w',8]`:

| | plotly | if strings were zero |
|---|---|---|
| `avg` | **8** | 2.667 |
| `min` | **8** | 0 |

Both readings look reasonable; only the measurement separates them.

## Which array supplies the values

Settled by `binned_axis`, not by which array is numeric. Plotly resolves **both** two-array spellings to `orientation: v` and bins `x` in each, so:

- `go.Histogram(x=cats, y=vals, histfunc="sum")` → bins `cats`, sums `vals` ✓
- `go.Histogram(y=cats, x=vals, histfunc="sum")` → bins `vals`, sums `cats` → plotly draws two bars of **zero**

The second is a chart nobody wants and it is the chart the call describes, so it is announced rather than second-guessed. A horizontal aggregating histogram needs an explicit `orientation="h"` — there is a test for each.

That last point cost me a wrong test: I wrote the horizontal cases as "swap the arrays", which is the natural reading and is not what plotly does. I had already measured the correct rule earlier in the same session and still reached for the wrong one when writing the test. It failed, which is what the tests are for.

## Verification

Same harness — figures drawn by the real `plotly.min.js` in Chromium, `gd.calcdata[0]` compared elementwise. **60 of 60 shapes agree**, both orientations, nineteen of them new `histfunc` cases. The harness now also handles the categorical layer shape (a `bar` layer with no bin bounds), which it previously could not compare at all.

## Falsification

| reverted | failures |
|---|---|
| value array = the numeric one rather than the other axis | 22 |
| ignore `histfunc` entirely (the bug) | 20 |
| keep empty bins for `avg`/`min`/`max` | 15 |
| drop empty bins even under `histnorm` | 12 |
| categorical path ignores `histfunc` | 9 |
| treat a non-number as zero | 2 |

## Checks

- `uv run pytest tests/` — **1517 passed** (up 49)
- `ruff check` clean; `ruff format` applied
- No line over 88 characters

---

### DIFF
--- a/maidr/plotly/histogram.py
+++ b/maidr/plotly/histogram.py
@@ -185,6 +185,123 @@ def _auto_shift_bins(
 #: Plotly's spelling of a horizontal trace.
 _HORIZONTAL = "h"
 
+#: ``histfunc`` values that aggregate a second array rather than counting.
+#: ``count`` is plotly's default and is what the bin populations already are.
+_AGGREGATING = frozenset({"sum", "avg", "min", "max"})
+
+#: Of those, the ones with no answer for a bin nothing landed in. Plotly emits
+#: no point at all for such a bin under these, where ``count`` and ``sum`` emit
+#: a zero -- measured in Chromium over a sample with a two-bin gap in the
+#: middle: ``count`` and ``sum`` return six points, ``avg``/``min``/``max``
+#: return four, the interior empties dropped rather than zeroed.
+_UNDEFINED_WHEN_EMPTY = frozenset({"avg", "min", "max"})
+
+
+def aggregate_bins(
+    assignment: np.ndarray, values: np.ndarray, n_bins: int, histfunc: str
+) -> tuple[np.ndarray, np.ndarray]:
+    """Reduce each bin's values the way ``histfunc`` does.
+
+    ``histfunc`` decides what a bar *measures*. Ignoring it announced the bin
+    populations for every mode, so a chart drawn at ``12, 15, 18`` was read out
+    as ``3, 3, 3`` (#405).
+
+    Parameters
+    ----------
+    assignment : np.ndarray
+        Bin index per observation; ``-1`` for observations outside every bin,
+        which plotly discards rather than clipping into an edge bin.
+    values : np.ndarray
+        The value array, aligned with *assignment*.
+    n_bins : int
+        How many bins the grid holds.
+    histfunc : str
+        One of ``sum``, ``avg``, ``min``, ``max``.
+
+    Returns
+    -------
+    tuple[np.ndarray, np.ndarray]
+        The per-bin values, and a mask of which bins hold an observation.
+        Bins outside the mask carry ``0`` and are the caller's to drop or
+        keep, since ``count`` and ``sum`` announce a zero there while the
+        other three announce nothing.
+    """
+    aggregated = np.zeros(n_bins, dtype=float)
+    present = np.zeros(n_bins, dtype=bool)
+
+    for index in range(n_bins):
+        in_bin = values[assignment == index]
+        # A value that is not a number is not an observation, rather than an
+        # observation of zero. Measured: a bin holding ``['z', 'w', 8]``
+        # averages to **8**, not 8/3, and its min and max are 8 as well --
+        # so the strings are dropped, not counted. A bin left with nothing
+        # numeric is empty, which the caller then treats exactly as it treats
+        # a bin nothing landed in.
+        in_bin = in_bin[np.isfinite(in_bin)]
+        if not in_bin.size:
+            continue
+        present[index] = True
+        if histfunc == "sum":
+            aggregated[index] = float(in_bin.sum())
+        elif histfunc == "avg":
+            aggregated[index] = float(in_bin.mean())
+        elif histfunc == "min":
+            aggregated[index] = float(in_bin.min())
+        else:  # max
+            aggregated[index] = float(in_bin.max())
+
+    return aggregated, present
+
+
+def as_numeric(values: list) -> np.ndarray:
+    """Coerce a value array to floats, with non-numbers as ``nan``.
+
+    The value array is not guaranteed numeric: ``go.Histogram(y=cats,
+    x=vals, histfunc="sum")`` bins ``x`` and hands the *category strings* to
+    the aggregate. Plotly does not error on that -- it drops them -- so a
+    plain ``np.array(..., dtype=float)`` would raise where the chart renders.
+
+    ``nan`` rather than ``0`` because that is the measured behaviour:
+    :func:`aggregate_bins` drops them, and a bin holding ``['z', 'w', 8]``
+    averages to 8.
+    """
+    coerced = np.empty(len(values), dtype=float)
+    for index, value in enumerate(values):
+        try:
+            coerced[index] = float(value)
+        except (TypeError, ValueError):
+            coerced[index] = np.nan
+    return coerced
+
+
+def value_array(trace: dict, binned: str) -> list | None:
+    """The array ``histfunc`` aggregates, or ``None`` when there is none.
+
+    A histogram that aggregates carries both arrays: one is binned and the
+    other supplies the values. Which is which is not "the numeric one" --
+    ``go.Histogram(x=cats, y=vals, histfunc="sum")`` and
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Good catch, and it reached further than the crash you named. Fixed in `9805635`.

## Reproduced first

Both paths, exactly as you predicted:

```
go.Histogram(x=[1..9], y=[10..50], histfunc="sum")
  -> IndexError: boolean index did not match indexed array along axis 0;
     size of axis is 5 but size of corresponding boolean axis is 9

go.Histogram(x=list("abcabcabc"), y=[1,2,3], histfunc="sum")
  -> IndexError: ... size of axis is 3 ... boolean axis is 9
```

Uncaught, out of `_extract_plot_data`, on figures plotly renders without complaint.

## What plotly does, and why `min(len(...))` alone is not the whole fix

You suggested guarding with `min(len(arr), len(raw_values))` before calling `aggregate_bins`. That stops the crash, but it would still have left the layer wrong — plotly truncates **before binning**, not just before aggregating:

| figure | plotly draws | binning all of `x` gives |
|---|---|---|
| `x=[1..9], y=[10..50], histfunc="sum"` | 3 bins spanning 1→5: `10, 50, 90` | 2 bins spanning 1→9 |
| `x=[1..5], y=[10..90], histfunc="sum"` | identical — symmetric | — |
| `x=cats(9), y=[1,2,3], histfunc="sum"` | `a:1  b:2  c:3` | — |

So the arrays are paired positionally and *both* cut to the shorter length.

## And it is not confined to aggregating

This is the part I would have missed by guarding only the call you pointed at. Plotly cuts the arrays **whatever `histfunc` says**:

```
go.Histogram(x=[1..9], y=[10..50], histfunc="count")
  plotly:      3 bins  (-0.5,1.5)=1  (1.5,3.5)=2  (3.5,5.5)=2
  py-maidr:    2 bins  (-0.5,4.5)=4  (4.5,9.5)=5
```

Same for a trace with both arrays and no `histfunc` at all. So the **count path was already wrong** for these figures on `main` — different bins *and* different counts than the chart draws — before any of this PR's aggregation existed. Your note that "none of the new tests exercise unequal-length arrays" was right in a stronger sense than intended: nothing did.

## The fix

`paired_arrays(trace, binned)` returns both arrays cut to their common length, and is the single place either is read. Truncating where they are read rather than at each point of use is what makes the bins and the values agree by construction, instead of by two call sites remembering to slice the same way — which is precisely how the original defect arose.

Seven new tests: the helper in isolation (both directions, single-array untouched), the binning span, the `count` path, the crash case, and the categorical equivalent.

## Your other two points

**O(bins × n) in `aggregate_bins`** — acknowledged and left as is. A vectorised pass would help only at bin counts far past what plotly's autobin produces (it caps at ~`2·stdev/n^0.4` spacing), and the loop is the version that reads as the four aggregate definitions it implements. Worth revisiting if a profile ever points here; I would rather not trade that legibility on a prediction.

**Dead `present = counts > 0`** — moved into the `raw_values is None` branch so the two are symmetric.

## Re-checked

- `uv run pytest tests/` — **1524 passed** (up 7)
- Chromium comparison still **60 of 60**, plus four new mismatched-length shapes checked against `gd.calcdata` directly
- `ruff check` clean, no line over 88

One note on the diff: `ruff format` wanted to reformat two spots in `histogram.py` that predate this branch (`n ** 0.4`, and a multi-line `if` in `_auto_shift_bins`). I reverted both — the repo does not enforce `ruff format` in CI, and a formatter sweep mixed into a bug fix makes the review harder. The diff against `main` now touches nothing outside the functions this PR changes.

---

PR #411, opened 2026-08-14
feat(plotly): read an area chart as an area rather than a line

### LINKED ISSUE #392: plotly: an area chart is announced as a line chart
Noted while sweeping the plotly binding for #390. Filed with an honest severity: this is a **parity gap, not a wrong reading** — unlike #390, which inverted a relationship.

`px.area` produces a `Scatter` trace carrying `stackgroup='1'`, which is how plotly declares an area. The adapter has no area handling at all — `maidr/plotly/` has `bar`, `box`, `heatmap`, `histogram`, `line`, `pie`, `scatter`, `step` and their multi variants, and no `area.py` — so the trace falls through to `line`.

```
px.area                    ['line']
px.area(color=...)         ['line']    stackgroups: ['1', '1']
```

## What a reader actually loses, and what they don't

**The numbers are right.** Plotly stores each series' *own* values in `trace.y` and does the stacking in the browser at render time, so a stacked area announces `1, 2, 3` and `10, 20, 30` — the series values, not running totals. That is the opposite of the matplotlib/ggplot2 hazard called out in xability/maidr#788 and xability/r-maidr#134, where the built data holds the cumulative top and reading it straight through announces totals as though they were values.

**What is missing is the name and the total.** A filled, stacked area is announced as a line chart, so a reader is not told the bands are filled, that they stack, or what the running total at each x is — which is the thing `stacked_area` exists to say and the reason someone draws this chart rather than a multi-line one.

So: incomplete rather than false. Worth fixing, not urgent in the way #388 or #390 were.

## Scope

- `maidr/plotly/area.py` emitting `area` / `stacked_area`, matching what the matplotlib path emits (#339, #356).
- Classify on `stackgroup` — already read in `maidr/plotly/step_shape.py` for a different purpose, so the signal is known to be reliable and present.
- Register in `plotly_plot_factory.py`, and decide the layer type in `PlotlyMaidr._extract_plots` alongside the bar `barmode` table.
- Announce the running total for the stacked case, since plotly does not supply it and the browser computes it.

The last point is the only real design work: the totals have to be computed from the traces in stack order, and the stack order is `stackgroup` membership in trace order.

## Related

Belongs with #343, which tracks the plotly binding's missing types (violin, candlestick, smooth). Adding `area` to that list.

jooyoungseo:
Closes #392.

`px.area` produces a `Scatter`, and the only thing separating it from a line is `stackgroup`. The adapter had no area handling at all — `maidr/plotly/` carried bar, box, heatmap, histogram, line, pie, scatter and step, and no `area` — so every one of them fell through to `line`.

```
px.area                    ['line']
px.area(color=...)         ['line']    stackgroups: ['1', '1']
```

## Severity, honestly

The numbers were already right. Plotly keeps each series' own values in `trace.y` and stacks in the browser, so a stacked area announced `1, 2, 3` and `10, 20, 30` rather than running totals — the opposite of the matplotlib/ggplot2 hazard where the built data holds the cumulative top.

What was missing is the name and the relationship: a filled, stacked band announced as a line tells a reader neither that the bands are filled nor that they stack, which is the reason someone draws this chart rather than a multi-line one.

## `stackgroup` is structural, and plotly's calcdata confirms which number is which

| | `fill` | calcdata keys |
|---|---|---|
| `stackgroup` set | `tonexty` | `s` (band's own value) **and** `y` (running total) |
| plain line | `none` | no `s` at all |

For series of `1,2,3` and `10,20,30`, the second's `y` comes back `11,22,33` while its `s` stays `10,20,30`. So `trace["y"]` — which the Python side has — is the band's own value and needs no un-accumulating. The total stays the core's to derive, as it is for `stackplot`; emitting it here would give the trace two sources that could disagree.

## Two rules measured rather than assumed

**Traces stack only within their own `stackgroup`.** Given `one` and `two` on one subplot, each series' calcdata `y` equals its own `s` — nothing accumulates across them. Two groups therefore become two layers.

**An empty `stackgroup` means "do not stack"**, not "stack with the unnamed group". A plain scatter resolves to `stackgroup: ""`, so the test is truthiness rather than presence — there is a falsification row for exactly that.

## The trap this had

An area passes *every* structural test for a connected line. Left in the line grouping it is emitted **twice** — once as its own layer and once inside the multi-line one. It is split out before any of that machinery sees the group; reverting that split fails 4 tests.

## Also

Adds `PlotType.NORMALIZED_AREA` for `groupnorm="percent"`/`"fraction"`. The JS grammar already carries `stacked_normalized_area` and py-maidr had no counterpart, so a 100% stacked area had nowhere to go.

## Selector verification

Every emitted selector resolved against real Plotly.js output in Chromium: **9 of 9 matched exactly one element**, including the mixed area-and-line figures in both orders, where the two layers compete for `scatterlayer` positions.

## One existing expectation changed rather than broke

`test_a_stacked_modeless_chart_is_a_line_at_any_size` asserted `LINE` for a `stackgroup` trace. It predates area support and was guarding against the trace being announced as loose markers — `AREA` is a stricter reading of the same figure, not a regression. Renamed to `..._is_connected_at_any_size`, and the anti-`SCATTER` property it existed for is now asserted explicitly beside it.

## Falsification

| reverted | failures |
|---|---|
| no area handling at all (the bug) | 10 |
| a lone band is still `stacked_area` | 7 |
| leave areas in the line grouping | 4 |
| one layer for every stackgroup at once | 3 |
| ignore `groupnorm` | 3 |
| an empty `stackgroup` counts as an area | 1 |

## Checks

- `uv run pytest tests/` — **1570 passed** (up 24)
- `ruff check` on every changed file — clean. The tree reports 7 findings, identical with and without this branch.
- No line over 88 characters

---

### DIFF
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -25,6 +25,12 @@ class PlotType(str, Enum):
     #: a line layer announces one number per point with nothing to say which
     #: of those two it is.
     STACKED_AREA = "stacked_area"
+    #: A stacked area rescaled so every position totals the same, the area
+    #: counterpart of :attr:`NORMALIZED`. Distinct for the same reason: a band
+    #: is a *share* of its position's total rather than a magnitude, and read
+    #: as a plain stack the equal totals look like a property of the data
+    #: rather than of the chart.
+    NORMALIZED_AREA = "stacked_normalized_area"
     PIE = "pie"
     SCATTER = "point"
     STACKED = "stacked_bar"
@@ -67,6 +73,7 @@ _DISPLAY_NAMES = {
     PlotType.NORMALIZED: "100% stacked bar",
     PlotType.STACKED: "stacked bar",
     PlotType.STACKED_AREA: "stacked area",
+    PlotType.NORMALIZED_AREA: "100% stacked area",
     PlotType.VIOLIN_BOX: "violin",
     PlotType.VIOLIN_KDE: "violin",
 }
--- a/maidr/plotly/area.py
+++ b/maidr/plotly/area.py
@@ -1,0 +1,144 @@
+"""Plotly area charts, which arrive as scatter traces carrying a stackgroup.
+
+`px.area` produces a `Scatter`, and the only thing separating it from a line
+is `stackgroup`. The adapter had no area handling at all, so every one of them
+fell through to `line` -- a reader was not told the bands are filled, that they
+stack, or what the running total at each x is, which is the reason someone
+draws this chart rather than a multi-line one (#392).
+"""
+
+from __future__ import annotations
+
+from maidr.core.enum.maidr_key import MaidrKey
+from maidr.core.enum.plot_type import PlotType
+from maidr.plotly.plotly_plot import PlotlyPlot, as_list
+from maidr.plotly.step_shape import is_scatter_family_trace
+
+#: The values ``groupnorm`` takes when plotly rescales a stack to a common
+#: total -- its own switch for a 100% stacked area, and the counterpart of
+#: ``barnorm`` on the bar path.
+_NORMALISING_GROUPNORMS = frozenset({"percent", "fraction"})
+
+
+def is_area_trace(trace: dict) -> bool:
+    """Whether plotly fills this trace down to a baseline.
+
+    ``stackgroup`` is the whole signal, and it is structural rather than a
+    display string: plotly stacks traces that share one and leaves traces with
+    an empty one alone. Measured in Chromium -- a trace with ``stackgroup``
+    set resolves to ``fill: "tonexty"`` and its calcdata carries an ``s`` key
+    holding the series' own value, while a plain line resolves to
+    ``fill: "none"`` and has no ``s`` at all.
+
+    Parameters
+    ----------
+    trace : dict
+        The plotly trace dictionary.
+
+    Returns
+    -------
+    bool
+        True when the trace is drawn as a filled band.
+    """
+    return bool(trace.get("stackgroup")) and is_scatter_family_trace(trace)
+
+
+def area_stack_groups(traces: list[dict]) -> list[list[dict]]:
+    """Split area traces into the stacks plotly actually accumulates.
+
+    Traces stack only with others sharing their ``stackgroup``. Two groups on
+    one subplot are two independent stacks, and measured that way: given
+    ``stackgroup='one'`` and ``stackgroup='two'``, each series' calcdata ``y``
+    equals its own ``s`` -- nothing accumulated across the two -- where within
+    one group the second series' ``y`` is the running total.
+
+    Groups come back in first-seen order so the emitted layers follow plotly's
+    own trace order rather than a sorted one.
+
+    Parameters
+    ----------
+    traces : list[dict]
+        The subplot's area traces, in declaration order.
+
+    Returns
+    -------
+    list[list[dict]]
+        One list per stack group.
+    """
+    groups: dict[str, list[dict]] = {}
+    for trace in traces:
+        groups.setdefault(str(trace.get("stackgroup")), []).append(trace)
+    return list(groups.values())
+
+
+def area_plot_type(traces: list[dict]) -> PlotType:
+    """Which area type a stack group is.
+
+    A lone band has nothing stacked on it, so it is a plain ``area`` -- the
+    same distinction the matplotlib path draws for a single ``stackplot``
+    band. ``groupnorm`` then separates a stack from a normalised one, exactly
+    as ``barnorm`` does for bars.
+    """
+    if any(t.get("groupnorm") in _NORMALISING_GROUPNORMS for t in traces):
+        return PlotType.NORMALIZED_AREA
+    if len(traces) > 1:
+        return PlotType.STACKED_AREA
+    return PlotType.AREA
+
+
+class PlotlyAreaPlot(PlotlyPlot):
+    """One stack of filled bands.
+
+    Emits each band's **own** value, with the series name alongside, matching
+    what :class:`~maidr.core.plot.areaplot.AreaPlot` produces for
+    ``stackplot``. The running total is the core's to derive: a stacked area
+    trace already sums its bands, and emitting a total here would give it two
+    sources that could disagree.
+
+    Plotly agrees about which number is which. Its calcdata carries both --
+    ``s`` for the band's own value and ``y`` for the running total -- and for
+    two series of ``1,2,3`` and ``10,20,30`` the second's ``y`` comes back
+    ``11,22,33``. So ``trace["y"]``, which the Python side has, is the band's
+    own value and needs no un-accumulating. That is the opposite of the
+    matplotlib and ggplot2 hazard, where the built data holds the cumulative
+    top and reading it straight through announces totals as values.
+    """
+
+    def __init__(
+        self,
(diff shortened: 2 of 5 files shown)


jooyoungseo:
Both points check out against the sibling code. Fixed in 2a0f62a, with one correction to the severity of the second.

### 1. Missing constructor guards — confirmed, fixed

Verified rather than assumed: `PlotlyMultiLinePlot` (`multiline.py:45-54`) and `PlotlyStepPlot` (`step.py:54-63`) both refuse an empty trace list and call `_validate_scatter_positions`. `PlotlyAreaPlot` did neither, and is built identically — `traces[0]` in `__init__`, `scatter_positions` used positionally in `_get_selector` — so it was exposed to the same silent wrong-`nth-child` failure the guard was written for.

Both siblings also do something the review didn't mention and that I'd missed for the same reason: they copy the two lists rather than aliasing them, with a comment naming the failure (a caller mutating its list afterwards changes this layer's selectors on the next render). Added that too, with a test that mutates the caller's list after construction.

### 2. Unconditional `z` — confirmed as an inconsistency, but not a mis-announcement

Reproduced: `px.area(d, x="x", y="y")` with no `color=` produces a trace with `name: ''`, and every point came out `{'x': 1, 'y': 1, 'z': ''}`. The sibling guards are exactly where the review says — `PlotlyPlot._line_series_with_positions` uses `if name:`, `AreaPlot._extract_plot_data` uses `if label:`.

Where I'd temper the framing: I checked the consumer before treating it as a behavioural bug, and the JS side already guards `z` with truthiness everywhere it reads it — `zData = point.z ? { z: {...} } : {}` (`src/model/line.ts:423`) and `this.points[i][0]?.z || \`l${i + 1}\`` (`:403`), inherited by `AreaTrace` from `LineTrace`. So `""` was already degrading to precisely the absent-key behaviour; nothing was announced wrongly. It's schema noise — a key whose value carries nothing — rather than a wrong reading, and it's worth removing on the consistency argument alone, which is what I've done.

Answering the review's open question directly: no, the always-present `z` was not intentional and the JS side does not treat it specially for area layers.

### Verification

- 9 new tests (7 on the guards, 2 on `z` present/absent). Suite **1579 passed**, up from 1570.
- Falsification: dropping the constructor guards fails **6 of the 7** guard tests — the seventh is the positive well-formed case, which correctly still passes. Emitting the blank `z` again fails **1**. Restored: 33/33 in the file.
- `ruff check` clean on both changed files; no line over 88 (checked explicitly, since ruff doesn't select `E501` here).

---

jooyoungseo:
Checked all three in the browser before touching anything. Two were real, one had a false premise — and the browser check found that the second was **worse** than described. Fixed in e83f031.

### 1. Mixed SVG/WebGL in one `stackgroup` — premise is false, but it exposed a real bug of mine

Plotly does not allow it, at either level:

- **plotly.py 6.7.0**: `scattergl` has no `stackgroup` attribute. `go.Scattergl(x=..., y=..., stackgroup="one")` raises `Bad property path: stackgroup`. Confirmed with `'stackgroup' in dir(...)` → `scatter: True`, `scattergl: False`.
- **plotly.js**: ignores the key on a raw dict that carries it anyway. Measured in Chromium with an svg and a gl trace both declaring `stackgroup: "one"` — the gl trace came back `fill: "none"`, no `stackgroup` in `_fullData`, `s: [null]`, nothing accumulated, while the svg trace beside it stacked normally.

So a mixed-renderer area stack cannot exist and the `renderer_groups` split would be machinery for an unreachable state.

But the finding was still worth raising, because it was built on `test_scattergl_counts_too` — **my test, asserting something plotly cannot produce.** The actual defect was the mirror image of the suggested fix: `is_area_trace` was *more* permissive than plotly, so a hand-built `scattergl` dict with a `stackgroup` would have been announced as a filled, stacked band where plotly draws a plain line — and, since a WebGL layer carries no selectors, would have taken the highlight with it. `is_area_trace` now excludes WebGL and the test asserts `False`.

### 2. Empty band — confirmed, and it *is* #316's misalignment

Reproduced the phantom series exactly as described. The review concluded "this isn't a *repeat* of #316's misalignment — the pairing happens to stay index-aligned since neither list is filtered". That holds only if the DOM keeps a slot for the empty trace, so I measured it. It doesn't:

```
3 traces (middle one empty) → scatterlayer holds 2 .trace.scatter nodes
nth-child(1) → 1 element   nth-child(2) → 1 element   nth-child(3) → 0
```

Plotly gives an empty trace no DOM node at all, so everything after it shifts up one. The emitted selectors would have handed band 2's selector to band 3 and left band 3 pointing at nothing — the misalignment, not just a phantom entry.

`_extract_plot_data` and `_get_selector` now both come from `_drawn_line_series`, which drops an undrawn band from data and selectors together, and the hand-rolled loop is gone.

**One part I did not fix here, deliberately.** Dropping the band is only half of it: the positions that survive are the *declared* indices, and plotly numbers the DOM by what it draws. That is pre-existing and not area-specific — the same figure built from `mode="lines"` traces already emits `nth-child(1)` and `nth-child(3)` into a two-node DOM on `main`. One layer cannot fix it, because an empty trace in another layer shifts this one too; it has to be a compaction pass where `position_of` is built. Filed as **#412**, with a test here asserting the current wrong value so it fails for whoever takes it rather than being rediscovered.

### 3. Stepped area — confirmed, filed as #413

Verified: `go.Scatter(stackgroup="one", line=dict(shape="hv"))` now emits `PlotType.AREA` with no `stepDirection`. Worth noting it was already incomplete before this PR in the other direction — it kept the step and lost the area — so neither classification has ever carried both facts. Filed rather than folded in, as suggested.

### Minor points

All taken: added the class-level `Parameters` section, and the `fill` local is gone entirely with the hand-rolled loop (the shared helper calls it `name`). The import ordering point is moot now — `area` no longer imports `MaidrKey` or `as_list`, and ruff confirms no unused imports.

### Verification

- Suite **1582 passed** (was 1579). 36 in the area file.
- Falsification: restoring the WebGL acceptance fails **1**; restoring the hand-rolled extraction fails **3**.

On the review's closing note about not being able to execute the suite — the numbers above are from actual runs, and the browser measurements from real `plotly.min.js` in Chromium via Playwright, not from documentation.

---

jooyoungseo:
Took both of the cheap ones in 25c161b; the other two are already accounted for.

**4 — `is_area_trace` ignores `mode`.** Right that it was unpinned, and it's the one of the four that could actually regress: "an area needs a line" is a reasonable-sounding narrowing that would drop a real filled band back onto the scatter path. Now asserted for `mode="markers"` with a `stackgroup`.

**3 — `NORMALIZED_AREA.display_name`.** Added, alongside `STACKED_AREA`. Worth having: the wire value is `stacked_normalized_area`, which is not a phrase to say out loud to anyone.

**1 — `area_plot_type` counts declared rather than drawn traces.** Confirmed, and leaving it. A stack of one real band and one empty one announces `stacked_area`. I'd rather it stayed consistent with `PlotlyMultiLinePlot`, which doesn't recompute its own plurality either, than have the area path alone start second-guessing the author's declared structure — a band that is empty in this render may not be in the next, and the chart is still authored as a stack.

**2 — stepped area.** Filed as #413 before this comment, with the note that it was incomplete in the other direction before this PR too (it kept the step and lost the area), so neither classification has ever carried both facts.

Suite **1584 passed**, 38 in the area file. Merging once CI is green.

---

Issue #394, opened 2026-08-14
fix(plotly): a normalised stacked histogram is read as independent histograms


jooyoungseo:
Found while working #338. Filed separately because the fix needs its own design.

## What happens

`px.histogram` accepts `barnorm` directly (unlike `px.bar`), so this is the ordinary way to draw a 100% stacked histogram in plotly express:

```python
px.histogram(frame, x="g", color="h", barnorm="percent")
```

```
layout: barmode='relative', barnorm='percent'
traces: ['histogram', 'histogram']
maidr : ['hist', 'hist']
```

Two independent histogram layers. A reader is told nothing about the bars being stacked, let alone normalised — the same class of defect as #390, on the histogram path rather than the bar one.

Without `barnorm` it is still wrong in the same direction:

```python
px.histogram(frame, x="g", color="h")          # barmode='relative' -> plotly stacks
#   maidr: ['hist', 'hist']                    # announced as two separate distributions
```

So the severity is not confined to normalisation: **any** multi-series `px.histogram(color=...)` is drawn as a stack and read as independent layers.

## Why #338 and #390 do not cover it

Both work on `bar` traces. `_extract_plots` collects `bar_traces` by `t.get("type") == "bar"`, and a `histogram` trace is a different type that never enters that branch — it falls through to `PlotlyPlotFactory`, which builds one `PlotlyHistogramPlot` per trace with no knowledge of its neighbours.

## Why it needs its own design

A `bar` trace states its `y`. A `histogram` trace does **not**: it carries only the raw `x` sample, and plotly bins it *in the browser*. So the segment heights a stacked histogram would need are not present in the Python figure at all, and reproducing plotly's binning here would be inference rather than transcription — exactly the trap #338 avoided by reading `barnorm` instead of guessing from the data.

Worth scoping deliberately. Some options, roughly in increasing order of ambition:

- [ ] Read `layout.barmode`/`barnorm` for histogram traces and at least emit the correct **relationship** between the layers, even if the values stay per-series
- [ ] Reproduce plotly's binning (`nbinsx`, `xbins`, and its auto rule) in Python and emit real stacked segments
- [ ] Decline to merge and say so, rather than announcing a stack as independent layers

## Reproduction

```python
import numpy as np, pandas as pd, plotly.express as px
from maidr.plotly.plotly_maidr import PlotlyMaidr

rng = np.random.default_rng(3)
frame = pd.DataFrame({"g": list("abc") * 10, "h": ["x", "y"] * 15})

[...]

PR #414, opened 2026-08-15
fix(plotly): announce a barnorm layer's shares rather than its counts

### LINKED ISSUE #409: fix(plotly): a barnorm layer announces counts under a "100% stacked" type
Found while implementing #394, and deliberately left out of it because the same divergence is already shipped on the **bar** path — fixing it for histograms alone would make the two disagree with each other.

## What happens

`layout.barnorm` tells plotly to rescale each stack position to a common total. Since #338/#393 the layer is correctly typed `stacked_normalized_bar`, and the values under it are the raw ones.

**Bars:**

```python
df = pd.DataFrame({"c": ["a","a","b","b"], "g": ["x","y","x","y"], "v": [3,1,2,6]})
px.bar(df, x="c", y="v", color="g").update_layout(barnorm="percent")
```

| | series x | series y |
|---|---|---|
| Plotly.js draws | `75, 25` | `25, 75` |
| py-maidr announces | `3, 2` | `1, 6` |

**Histograms** (#394's path, same rule):

| | series 0 | series 1 |
|---|---|---|
| Plotly.js draws | `100, 25` | `75, 100` |
| py-maidr announces | `4, 1` | `3, 2` |

The type says the bars are rescaled to a common total; the values are the untouched inputs. A reader adding them up does not get 100, and comparing two categories gets their counts rather than the shares the chart is drawn to show — which is the one thing a 100% stacked chart exists to say.

Not as severe as #404 was: there the axis label and the value contradicted each other inside one layer. Here the *type* and the values do, which is a weaker signal but the same class.

## Why it is one issue and not two

`PlotlyGroupedBarPlot` passes `trace["y"]` through unchanged, and `PlotlyGroupedHistogramPlot` matches it on purpose. Whichever way this resolves, both should move together — so this issue covers both, and neither path should be changed alone.

## A detail that only matters if we emit shares

Plotly **drops** a bin whose stack total is zero. Measured on a stacked histogram with a trailing empty bin: the raw layer has a `0` there and plotly's `calcdata` has no point at all, because the share would be `0/0`. So emitting shares is not simply "divide by the position total" — it also changes which points exist, and the count has to follow.

That is why this is not a two-line change and why it is filed rather than folded into #394.

## Ground truth

`gd.calcdata[i][j].s` after `Plotly.newPlot` in Chromium, for both figures above.

## Difficulty

Small-to-medium. The arithmetic is one pass over the stack positions; the care is in the zero-total rule and in keeping the two paths in step.

jooyoungseo:
Closes #409.

`layout.barnorm` rescales each stack position to a common total, so the bars a reader sees are *shares of their category*. Since #338/#393 the layer was typed `stacked_normalized_bar` for it — and the values underneath were the untouched inputs.

| | series x | series y |
|---|---|---|
| Plotly.js draws | `75, 25` | `25, 75` |
| py-maidr announced | `3, 2` | `1, 6` |

The type said the bars are proportions; the numbers were the tallies. Adding them up did not reach 100, and comparing two categories gave their counts rather than the shares — which is the one thing a 100% stacked chart exists to say.

## Why the producer, and not the core

I checked before assuming, because the opposite convention exists in this codebase. `SegmentedTrace` handles `stacked`, `dodged` and `stacked_normalized_bar` with one class and normalises nothing, so a normalised layer is expected to arrive already carrying shares. Both r-maidr paths already do:

- **base R** — `base_r_stacked_bar_layer_processor.R` says so outright: "the values are already the drawn shares, since base R has no `position = "fill"` and the author normalised the matrix before calling `barplot()`."
- **ggplot2** — `position = "fill"` builds its data in 0..1, and `ggplot2_stacked_bar_layer_processor.R` already handles the exact hazard: "the value the chart draws is a share and NOT the number sitting in the user's data frame… a normalized layer has to take the geometry route and subtract."

So the plotly path was the outlier, not the standard-setter. Contrast `AreaTrace`, which *does* compute its own stack totals and is therefore deliberately fed raw values — the two conventions coexist because the JS side differs per trace type, so this had to be established rather than assumed.

## Measured, not read

Every rule comes from `gd.calcdata[i][j].s` after `Plotly.newPlot` in Chromium. Two are not what the documentation suggests, and I only found them by measuring:

**The denominator depends on `barmode`.**

| input | `relative` (default) | `stack` |
|---|---|---|
| `[3, -1]` | `100, -100` | `150, -50` |
| `[3, -1, 6]` | `33.3, -100, 66.7` | `37.5, -12.5, 75` |

`relative` draws positive and negative as two stacks growing away from the baseline and normalises each against its own total; `stack` pools them. My first reading was "sign-grouped always", and the `stack` column refuted it.

**Under `stack` the denominator is the *absolute value* of the signed sum.** `[0, -4]` draws `0, -100`. A plain signed sum of `-4` would have made the second bar `+100`. This one is easy to get wrong and produces a sign-flipped chart when you do.

**A zero-total position is a gap that keeps its `x`.** The issue predicted plotly "has no point at all"; measured, the point survives with `x` intact and `s` null. That is better than predicted — the point count never changes, so nothing downstream has to be renumbered, and it maps onto the gap concept `SegmentedTrace` already has ("a category is only a gap in the total when every segment in it is one").

Also measured: a null contributes nothing to the denominator and stays null rather than becoming a zero share; positions are matched by category value, so a ragged series contributes nothing where it has no bar.

## Scope

Bars and histograms move together, which is what #409 asks for — `PlotlyGroupedHistogramPlot` matched `PlotlyGroupedBarPlot` on purpose, so changing one alone would make them disagree. A test asserts the two produce identical shares from the same tallies.

Horizontal bars normalise `x` rather than `y`, since the category and the magnitude swap axes; normalising `y` would have rescaled the categories.

**Dodged is deliberately left alone.** `barnorm` makes every bar its own stack under `barmode="group"`, so plotly draws four 100s. But the layer is typed `dodged_bar`, which claims no normalisation, so the type and the values do not contradict each other the way #409 describes. Emitting four 100s would be faithful to the geometry and useless to a reader; emitting the counts keeps the information and matches the announced type. Neither is fully right, so it is pinned with that reasoning rather than quietly half-fixed.

## Verification

- **Suite 1621 passed** (1584 on `main` + 37 new).
- The new module reproduces **13 of 13** Chromium-measured cases exactly before any wiring.
- Falsification: without the bar path **5** fail; without the histogram path **4**; without the sign split **7**; without the pooled absolute value **1**.
- `test_the_values_are_the_counts_rather_than_the_shares`, pinned in #394 so whoever took this would see a failure rather than a discovery, failed exactly as intended and is updated.
- `ruff check` clean; no line over 88 (checked separately, since ruff does not select `E501` here).

One note on process: this PR's test file already existed from #393, and my first pass overwrote it. Caught before committing — the diff is **279 additions, 0 deletions**, and all five original tests are still present and passing.

---

### DIFF
--- a/maidr/plotly/barnorm.py
+++ b/maidr/plotly/barnorm.py
@@ -1,0 +1,177 @@
+"""Rescaling a stack to a common total, the way ``layout.barnorm`` draws it.
+
+``barnorm`` tells plotly to rescale every stack position to a common total, so
+the bars a reader sees are *shares of their position* rather than the numbers
+in the data frame. The layer is typed ``stacked_normalized_bar`` for it, and
+the values underneath were the untouched inputs -- the type and the values
+contradicting each other (#409).
+
+Every rule here was measured against ``gd.calcdata[i][j].s`` after
+``Plotly.newPlot`` in Chromium rather than read from the documentation, and
+two of them are not what the documentation would suggest. See
+``stack_shares`` for the denominators.
+
+The MAIDR core does not do this arithmetic for us, and that is the settled
+convention rather than an oversight: ``SegmentedTrace`` handles ``stacked``,
+``dodged`` and ``stacked_normalized_bar`` with one class and normalises
+nothing, so a normalised layer is expected to arrive already carrying shares.
+Both r-maidr paths do exactly that -- base R because the author normalised the
+matrix before calling ``barplot()``, ggplot2 because ``position = "fill"``
+builds its data in 0..1 -- which is what makes py-maidr's plotly path the
+outlier rather than the standard-setter. Contrast ``AreaTrace``, which *does*
+compute its own stack totals, and so is deliberately fed raw values.
+"""
+
+from __future__ import annotations
+
+import math
+from typing import Any, Hashable, Sequence
+
+#: The values ``layout.barnorm`` takes when plotly rescales a stack. ``percent``
+#: scales each position to 100 and ``fraction`` to 1; anything else -- ``None``,
+#: ``""``, an unrecognised string -- leaves the bars alone.
+_NORMALISING_BARNORMS: dict[str, float] = {"percent": 100.0, "fraction": 1.0}
+
+#: The one barmode that pools positive and negative into a single stack. Every
+#: other combining mode -- ``relative``, which is plotly's default and what
+#: ``px.bar`` leaves behind -- stacks the two sides separately.
+_POOLED_BARMODE = "stack"
+
+
+def barnorm_scale(barnorm: Any) -> float | None:
+    """The multiplier a ``barnorm`` setting scales each position to.
+
+    Parameters
+    ----------
+    barnorm : Any
+        ``layout.barnorm``, which may be absent, empty or unrecognised.
+
+    Returns
+    -------
+    float or None
+        ``100.0`` for ``percent``, ``1.0`` for ``fraction``, and ``None``
+        when plotly normalises nothing -- in which case the caller should
+        emit the values unchanged rather than scaling by 1.
+    """
+    if not isinstance(barnorm, str):
+        return None
+    return _NORMALISING_BARNORMS.get(barnorm)
+
+
+def _measured(value: Any) -> bool:
+    """Whether a value takes part in a total.
+
+    A ``None`` or non-finite entry is a gap: plotly leaves it out of the
+    denominator and leaves it undefined in the output. Measured -- a stack of
+    ``[None, 4]`` came back ``[None, 100]``, so the null neither contributed
+    to the total nor became a zero share.
+    """
+    return isinstance(value, (int, float)) and not isinstance(value, bool) and (
+        math.isfinite(value)
+    )
+
+
+def stack_totals(
+    series: Sequence[Sequence[tuple[Hashable, Any]]], barmode: Any
+) -> dict[Hashable, dict[bool, float]]:
+    """Sum each stack position, the way the active barmode stacks it.
+
+    Two denominators, and which applies is the barmode's doing:
+
+    ``relative`` -- plotly's default, and what ``px.bar`` leaves behind --
+    draws the positive and negative bars as two stacks growing away from the
+    baseline, and normalises each against its own total. So the denominator
+    is the sum of the *absolute* values sharing a sign. Measured: ``[3, -1]``
+    comes back ``100, -100``, not ``75, -25``.
+
+    ``stack`` pools both signs into one stack, so the denominator is the
+    absolute value of the signed sum, applied to every value at the position.
+    Measured: ``[3, -1, 6]`` comes back ``37.5, -12.5, 75`` -- a denominator
+    of 8. The absolute value is load-bearing rather than cosmetic: ``[0, -4]``
+    comes back ``0, -100``, and a plain signed sum of ``-4`` would have made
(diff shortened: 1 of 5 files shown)


jooyoungseo:
All three taken in 7b0ffc5. The second one was a fair hit against this PR's own standard — I asserted the composition order in a docstring and never measured it.

### 2 — `histnorm` + `barnorm` composition, now measured

And the first figure I reached for would have given me a false confirmation. Two equal-sized series make `histnorm="percent"` divide each by the same total, so the shares are identical whichever order the two are applied in — every `histnorm` value returned `75/33.3, 25/66.7` and it looked like the claim was verified. It wasn't; the case just couldn't distinguish the hypotheses.

With six observations against three, the order is visible:

| | series x | series y |
|---|---|---|
| raw counts | `4, 2` | `1, 2` |
| `barnorm` alone | `80, 50` | `20, 50` |
| `histnorm="percent"` alone | `66.7, 33.3` | `33.3, 66.7` |
| **both** | `66.7, 33.3` | `33.3, 66.7` |

`barnorm` operating on the raw counts would leave `histnorm` unable to change the result at all, and it plainly does. The docstring was right; it just hadn't earned it. Implementation reproduces all three rows exactly, and dropping `apply_histnorm` now fails 3 tests.

### 1 — `_measured` readability

Taken; it's an early return now rather than a wrapped boolean.

### 3 — `barmode` default indirection

Taken, with the reasoning recorded rather than just the equivalence. Testing `!= "stack"` and `_PLOTLY_DEFAULT_BARMODE = "relative"` spell the same default from opposite directions, and they agree only because these are the two states that can reach here — `group` never does, since a dodged layer is never `NORMALIZED`. The comment says a third combining mode would have to be added in both places, which is the failure the note was pointing at.

Suite **1624 passed** (was 1621), 52 in this file.

---

jooyoungseo:
This landed after I merged, and the main point is right — I'll fix it rather than leave it.

**Duplicated `_NORMALISING_BARNORMS` — confirmed, and worse than a naming smell.** Verified both: `plotly_maidr.py:109` holds `frozenset({"percent", "fraction"})` used at `:426` to pick `PlotType.NORMALIZED`, and `barnorm.py:33` holds `{"percent": 100.0, "fraction": 1.0}` used by `barnorm_scale`. Same name, same keys, two modules, no shared source.

The failure mode you describe is exactly #409 coming back: a layer typed `stacked_normalized_bar` by one file whose values are left as raw counts by the other, contradicting each other silently. That is the specific class of defect this PR existed to remove, so leaving two places to edit is the wrong ending for it. The classification should derive from `barnorm_scale(...) is not None` and the frozenset should go.

Doing that as its own PR rather than appending it here, since this one is merged and the change is a different logical unit. Tracking it now so it does not get lost.

**Pairs built unconditionally in `grouped_bar.py`** — fair, and I'll take it in the same change: `scale` is knowable before the loop, so the non-normalised path can skip building the tuples entirely.

**Dodged-histogram analog to `TestDodgedIsLeftAlone`** — agreed it is low priority, and I'd add that both paths route through `_combined_type()`, so the existing bar test already covers the shared decision. I'll add it anyway if it costs one figure, since "shares a helper today" is the assumption that ages worst.

Thanks for the catch — it's the kind that only shows up on a re-read of the whole diff rather than the changed lines.

---

PR #415, opened 2026-08-15
fix(violin): emit box statistics raw instead of rounded to four decimals

### LINKED ISSUE #398: fix(violin): rounding box statistics to 4 decimals flattens small-magnitude data
Raised during review of #397, filed separately because it changes existing matplotlib output.

## The code

```python
# maidr/core/plot/violin_box_plot.py
MaidrKey.MIN.value: round(stats[MaidrKey.MIN.value], 4),
MaidrKey.Q1.value:  round(stats[MaidrKey.Q1.value], 4),
MaidrKey.Q2.value:  round(stats[MaidrKey.Q2.value], 4),
MaidrKey.Q3.value:  round(stats[MaidrKey.Q3.value], 4),
MaidrKey.MAX.value: round(stats[MaidrKey.MAX.value], 4),
...
record[MaidrKey.MEAN.value] = round(stats[MaidrKey.MEAN.value], 4)
```

## Why it bites

`round(x, 4)` is **absolute**, not significant. It does not mean "four decimals of precision"; it means "nothing below 1e-4 survives".

```
scale      raw q1                       round(q1, 4)
units      1.7901224999999998           1.7901
millis     0.0017901225                 0.0018
micros     1.7901225e-06                0.0
```

So for any dataset whose values sit below about 1e-4 — micrograms, molar concentrations, failure probabilities, seconds in a fast benchmark — **every** statistic is announced as `0.0`. `min`, `q1`, `q2`, `q3` and `max` all come out identical, so the box reads as a flat line at zero and the distribution disappears from the announcement entirely.

Nothing errors. The chart is drawn correctly, the layer is present, the count of violins is right. Only the numbers are gone.

At the other end, a dataset in millions keeps 4 decimals of noise that no one can hear.

## Reproduction

```python
import matplotlib
matplotlib.use("Agg")
import matplotlib.pyplot as plt, numpy as np, pandas as pd, seaborn as sns
import maidr

rng = np.random.default_rng(0)
frame = pd.DataFrame({"g": ["a"] * 40, "v": rng.normal(2e-6, 3e-7, 40)})

fig, ax = plt.subplots()
sns.violinplot(data=frame, x="g", y="v", ax=ax)

layers = FigureManager.get_maidr(fig)._flatten_maidr()["subplots"][0][0]["layers"]
box = next(l for l in layers if l["type"].name == "VIOLIN_BOX")
print(box["data"][0])
#   {'z': 'a', 'lowerOutliers': [], 'min': 0.0, 'q1': 0.0, 'q2': 0.0,
#    'q3': 0.0, 'max': 0.0, 'upperOutliers': []}
```

## Suggested fix

Drop the rounding and emit the raw float. Formatting for announcement is the frontend's job — it is what decides how many digits a reader actually hears, and doing it there adapts to the magnitude instead of assuming one.

The plotly violin path added in #397 already emits raw floats for this reason, so removing the rounding would also bring the two backends into agreement rather than away from it.

- [ ] `maidr/core/plot/violin_box_plot.py` — the six `round(..., 4)` calls
- [ ] Check whether `boxplot.py` and the other summary emitters do the same thing
- [ ] A test pinning a small-magnitude sample, since the failure is silent

## Difficulty

Easy, but it changes existing output, so it wants its own review rather than riding along with a feature.

jooyoungseo:
Closes #398.

`round(x, 4)` sounds like precision and is not. It is absolute, so it does not keep four significant digits — it discards everything below `1e-4`.

| scale | raw q1 | `round(q1, 4)` |
|---|---|---|
| units | `1.7901224999999998` | `1.7901` |
| millis | `0.0017901225` | `0.0018` |
| micros | `1.7901225e-06` | **`0.0`** |

Reproduced exactly as filed, at `2e-6`:

```python
{'z': 'a', 'lowerOutliers': [], 'min': 0.0, 'q1': 0.0, 'q2': 0.0,
 'q3': 0.0, 'max': 0.0, 'upperOutliers': []}
```

All five statistics collapse onto one number, so the box reads as a flat line at zero and the distribution stops being announced. This is the session's recurring shape: nothing errors, the chart draws correctly, the layer is present and the violin count is right — only the meaning is gone. Micrograms, molar concentrations, failure probabilities and fast-benchmark seconds all live below that floor.

At the other end, a dataset in millions carried four decimals of noise no one can hear.

## Why raw

How many digits a reader hears is the frontend's decision, and it is the only place that can make it relative to the magnitude rather than assuming one.

Emitting raw also settles two disagreements this file was on the wrong side of, both checked rather than assumed:

- **Within the same record** — `lowerOutliers` and `upperOutliers` sit beside these keys and were never rounded, so a truncated quartile could already appear next to a full-precision outlier drawn from the same data.
- **Across the codebase** — I grepped every `round(..., 4)` in `maidr/`: these six were the only ones. `boxplot.py` emits `cap["min"]` and `whisker["q1"]` unrounded, and the plotly violin path added in #397 emits raw floats. The remaining `round()` calls elsewhere are coordinate and tick lookups, not announced statistics.

That covers the issue's second checklist item: no other summary emitter does this.

## Verification

- **Suite 1631 passed** (1624 + 7 new). **No existing test depended on the rounded values** — worth stating explicitly, since this changes shipped output and a silent expectation would have been the risk.
- Falsification: restoring the six `round(..., 4)` calls fails **5 of the 7** new tests. The two that still pass are the ordering assertions, which rounding preserves — that is the point of the other five.

One note on the tests: my first version of `test_full_precision_now_reaches_the_schema` asserted `a != round(a, 4) or a == approx(round(a, 4))`, which is a tautology and passes either way. Replaced with an assertion that at least one of the five differs from its own four-decimal rounding, which actually fails when the rounding returns.

---

### DIFF
--- a/maidr/core/plot/violin_box_plot.py
+++ b/maidr/core/plot/violin_box_plot.py
@@ -111,20 +111,15 @@ class ViolinBoxPlot(MaidrPlot):
 
         Notes
         -----
-        The statistics are emitted raw. They were rounded to four decimals,
-        which sounds like precision and is not: ``round(x, 4)`` is absolute,
-        so nothing below ``1e-4`` survives it. For data in micrograms, molar
-        concentrations, failure probabilities or seconds of a fast benchmark,
-        every one of ``min``/``q1``/``q2``/``q3``/``max`` came out ``0.0`` and
-        the box read as a flat line at zero -- silently, since the chart drew
-        correctly and only the numbers were gone. At the other end a dataset
-        in millions carried four decimals of noise no one can hear.
+        Emitted raw. Rounding to four decimals here is absolute, not
+        significant, so data below ``1e-4`` collapsed to five identical
+        ``0.0``\\ s and the box read as a flat line at zero (#398).
 
-        How many digits a reader hears is the frontend's decision, and it is
-        the only place that can make it relative to the magnitude. Emitting
-        raw also settles two disagreements this file was on the wrong side
-        of: the outliers beside these keys were never rounded, and neither
-        the matplotlib box plot nor the plotly violin path rounds at all.
+        The frontend already handles this correctly and needs the raw value
+        to do it: ``defaultFormat`` rounds to two decimals for the
+        announcement, then falls back to three significant digits when that
+        would erase a non-zero value. Rounding here defeated that guard
+        before the number reached it.
         """
         # Tag artists and register elements for SVG highlighting.
         self._tag_artists()
--- a/maidr/core/plot/violin_box_plot.py
+++ b/maidr/core/plot/violin_box_plot.py
@@ -108,6 +108,23 @@ class ViolinBoxPlot(MaidrPlot):
         list[dict]
             One dict per violin with keys: ``z, lowerOutliers, min,
             q1, q2, q3, max, upperOutliers, mean?``.
+
+        Notes
+        -----
+        The statistics are emitted raw. They were rounded to four decimals,
+        which sounds like precision and is not: ``round(x, 4)`` is absolute,
+        so nothing below ``1e-4`` survives it. For data in micrograms, molar
+        concentrations, failure probabilities or seconds of a fast benchmark,
+        every one of ``min``/``q1``/``q2``/``q3``/``max`` came out ``0.0`` and
+        the box read as a flat line at zero -- silently, since the chart drew
+        correctly and only the numbers were gone. At the other end a dataset
+        in millions carried four decimals of noise no one can hear.
+
+        How many digits a reader hears is the frontend's decision, and it is
+        the only place that can make it relative to the magnitude. Emitting
+        raw also settles two disagreements this file was on the wrong side
+        of: the outliers beside these keys were never rounded, and neither
+        the matplotlib box plot nor the plotly violin path rounds at all.
         """
         # Tag artists and register elements for SVG highlighting.
         self._tag_artists()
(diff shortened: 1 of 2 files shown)


jooyoungseo:
All three taken in e376834. The third was the one worth having — I asserted the frontend's behaviour in the PR body without checking it, and checking makes the case stronger rather than weaker.

### The frontend does format, and it already guards exactly this

`src/util/format.ts` — `defaultFormat`:

```js
const rounded = Number(value.toFixed(DEFAULT_MAX_DECIMALS));   // 2
if (rounded !== 0) {
  return `${rounded}`;
}
// Rounding erased the value (0.00012 -> 0). Announcing `0` for something that
// is not zero is worse than the extra digits, so keep it visible instead.
return `${Number(value.toPrecision(DEFAULT_SIGNIFICANT_DIGITS))}`;   // 3
```

So the announcement is not raw floats: fractional values are spoken at two decimals, and when that would erase a non-zero value it falls back to three significant digits. It is magnitude-aware, and its own comment describes the identical defect this PR fixes — "announcing 0 for something that is not zero is worse than the extra digits."

That reframes the change. The Python-side rounding was not merely redundant with the frontend; it was **defeating a guard that already existed**, because the frontend can only rescue a small value if it receives one. `round(2e-6, 4)` handed it a `0.0`, at which point nothing downstream can tell "genuinely zero" from "rounded away".

Two more things that fell out of reading it, both relevant to the bloat concern:

- The same docstring notes that **rounding stops at the announcement** — sonification, braille and extrema all read the underlying numbers. So the raw floats are not just harmless there, they are what keeps a bar's pitch and a braille cell's position accurate. Rounding at the producer degraded those silently too.
- On embedded-JSON size: `boxplot.py` and the plotly violin path already emit raw floats, so this makes the payload uniform rather than adding a new cost. If size ever matters it should be solved once at serialisation for every emitter, not by one path rounding its own numbers away.

### The two style nits

Both taken: `layer["type"] is PlotType.VIOLIN_BOX` to match the sibling tests, and the `Notes` block trimmed from narrative prose to the fact worth keeping — which is now the frontend-guard point above rather than the weaker "formatting is the frontend's job".

Suite still 1631, 7 in the new file, `ruff` clean.

---

PR #416, opened 2026-08-15
refactor(plotly): read the normalising barnorm values from one place

jooyoungseo:
Follow-up to review feedback on #414, which landed after that PR merged. The reviewer was right and this is the fix.

## The problem

`plotly_maidr.py` and `barnorm.py` each held their own copy of the two values `layout.barnorm` takes when plotly rescales a stack — under **the same name** in two modules:

```python
# plotly_maidr.py:109        -> decides the layer TYPE
_NORMALISING_BARNORMS = frozenset({"percent", "fraction"})

# barnorm.py:33              -> decides whether values are RESCALED
_NORMALISING_BARNORMS: dict[str, float] = {"percent": 100.0, "fraction": 1.0}
```

They agreed, so nothing was broken. But editing one without the other is exactly how #409 comes back: a `barnorm` that one recognised and the other did not would type the layer `stacked_normalized_bar` while leaving its values the raw counts — the type and the numbers contradicting each other again, and silently. That is the defect these two readings were reconciled to end, so leaving two places to edit was the wrong ending for it.

## The fix

The classification now asks `barnorm_scale(...) is not None`, so there is one list.

## Tests

Two, both aimed at the drift rather than at the values:

- **The equivalence, over the whole reachable set.** For every `barnorm` plotly will accept, "typed `stacked_normalized_bar`" and "gets a scale" must be the same boolean. Worth noting what I found writing it: `update_layout(barnorm="nonsense")` **raises** — plotly validates the enum — so an unrecognised value cannot arrive from a real figure at all. My first version parametrized over junk strings and failed for that reason. The reachable set is `percent`, `fraction`, `""`, `None`; the junk cases moved to a separate test of `barnorm_scale` directly, since it is also reachable from a hand-built layout dict where nothing validates.
- **A guard against a second copy returning**, asserted on the module rather than by grepping text.

Falsification: reintroducing a divergent copy (dropping `fraction` from the classifier) fails **3** tests.

## Also here

The same review noted `grouped_bar._extract_plot_data` built its `(position, value)` tuples for every layer and discarded them unless the layer was normalised. `scale` is knowable before the loop, so the ordinary un-normalised path now skips them entirely.

And a dodged-**histogram** test, the analog of the existing `TestDodgedIsLeftAlone` for bars. It is redundant today because both route through `_combined_type()` — which is the point: "shares a helper today" is the assumption that ages worst, and the two paths are already pinned against each other elsewhere for the same reason.

## Verification

- **Suite 1641 passed.**

---

### DIFF
--- a/maidr/plotly/grouped_bar.py
+++ b/maidr/plotly/grouped_bar.py
@@ -55,6 +55,11 @@ class PlotlyGroupedBarPlot(PlotlyPlot):
         traces' own numbers, because that is what plotly draws and what the
         type claims. See :mod:`maidr.plotly.barnorm`.
         """
+        # Resolved before the loop so the ordinary, un-normalised layer does
+        # not build the `(position, value)` tuples only to discard them.
+        scale = barnorm_scale(self._layout.get("barnorm"))
+        normalising = scale is not None and self.type == PlotType.NORMALIZED
+
         data: list[list[dict]] = []
         pairs: list[list[tuple]] = []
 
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Both taken in a49ce94.

The second is the one worth acting on rather than waving through: the dodged-histogram test existed to catch the bar and histogram paths drifting apart, and burying it inside `TestTheEmittedHistogramLayer` meant that purpose was legible only from a comment. It is now `TestDodgedHistogramsAreLeftAloneToo`, sitting directly beside `TestDodgedIsLeftAlone`, so the pairing is visible from the structure — which matters more for this particular test than most, since the thing it guards is precisely that the two paths stop being able to be read as one.

The stray double blank line went with it, and the `xbins` dict is hoisted to a local so both traces share one spec rather than repeating it past the line limit.

62 tests in the file, `ruff` clean, no line over 88.

---

PR #417, opened 2026-08-15
fix(plotly): address a box by its trace group and its position within it

### LINKED ISSUE #395: fix(plotly): a box plot's highlight selector counts boxes, not the DOM slots they occupy

jooyoungseo:
Closes #395.

A box needs **two** indices and was given one. Plotly puts one `<g>` in the `boxlayer` per trace and draws that trace's boxes as direct `path.box` children of it, so a categorical `go.Box` puts all of its boxes inside a single group. Numbering them as though each were its own group made box 1 match *every* box in the trace and boxes 2..n match nothing.

The DOM, measured in Chromium:

| figure | trace groups | children of each group |
|---|---|---|
| one `go.Box`, 3 categories | **1** | `path.box ×3`, `g.points ×3` |
| three `go.Box` traces | 3 | `path.box`, `g.points` |
| `go.Candlestick` + `go.Box` | 2 | group 1 is the candlestick's 4 boxes |

## Three failures, not one

**1. One trace, several categories.** Emitted `g:nth-child(1..3)` into a DOM with one group. Now one group, three `:nth-child(n of path.box)` positions inside it.

**2. A candlestick sharing the layer.** `layer_position` widened its search to boxlayer-mates only when the trace *was* a candlestick:

```python
if kind == "candlestick":
    mates = _BOXLAYER_TRACE_TYPES        # {"box", "candlestick"}
else:
    mates = frozenset({kind})            # a box counts only boxes
```

So a candlestick counted the boxes beside it, while a box ignored the candlestick and claimed the group the candlestick had already taken. Both draw into `g.boxlayer`, so each shifts the other. Tests assert both directions, and that a `go.Violin` (`violinlayer`) and a scatter (`scatterlayer`) shift neither.

**3. Multi-box with categories — the one the issue didn't cover.** `PlotlyMultiBoxPlot` looped over *traces* rather than boxes:

```
2 traces × 2 categories → 4 boxes of data, 2 selectors
```

Half the boxes addressed nothing at all, and since the frontend pairs selector *i* with box *i*, the two that did exist were pointing at the wrong boxes. The per-trace box count is now recorded during extraction, because a flat list of stats no longer says which trace contributed what.

## The outlier groups

The part that decides whether outliers are addressable: plotly emits one `g.points` per box, **including an empty one for a category with no outliers**, and the *j*-th pairs with the *j*-th box. That empty group is load-bearing — it keeps the pairing positional rather than a count of which categories happen to have any. Measured with 1, 0 and 2 outliers across three categories:

| box | box x | points in *j*-th `g.points` |
|---|---|---|
| 1 | 150 | 1 |
| 2 | 397 | 0 |
| 3 | 644 | 2 |

## Scope note

A lone box is now built in `_extract_plots` rather than left to `PlotlyPlotFactory`, for the reason the candlestick branch there already documents: the factory sees one trace and cannot know what shares its layer. The factory default stays `0`, which is the only assumption available to it.

## Verification

- **15 of 15** box selectors resolve to exactly one element in Chromium — across a categorical trace, three plain traces, two categorical traces, both candlestick orderings, and the mixed-outlier figure. All selectors within a figure are distinct; outlier selectors match only the categories that have outliers.
- **Suite 1659 passed** (1641 + 18 new). Notably **no existing test failed**, which is itself the finding: the old selectors were never pinned, so the defect was live and silent.
- Falsification: reverting the two-index selector fails **4**; restoring the `layer_position` asymmetry fails **2**; looping over traces again fails **2**.

---

### DIFF
--- a/maidr/plotly/box.py
+++ b/maidr/plotly/box.py
@@ -14,19 +14,48 @@ _logger = logging.getLogger(__name__)
 
 def _build_box_selector(
     prefix: str,
-    nth_child: int,
-    box_sel: str,
+    group: int,
+    box: int,
     lower_count: int,
     upper_count: int,
 ) -> dict:
     """Build a ``BoxSelector``-compatible dict with split outlier selectors.
 
+    A box needs *two* indices, not one. Plotly gives each trace one
+    ``<g>`` in the ``boxlayer`` and draws that trace's boxes as direct
+    ``path.box`` children of it, so a categorical trace puts every one of
+    its boxes inside a single group. Numbering boxes as though each were
+    its own group -- which this did -- made box 1 match all of them and
+    boxes 2..n match nothing (#395).
+
+    Measured in Chromium: one ``go.Box`` over three categories produces one
+    ``.trace.boxes`` group holding three ``path.box`` children and three
+    ``g.points`` children, and the *j*-th of each pair up positionally --
+    including an empty ``g.points`` for a category with no outliers, which
+    is what keeps the pairing positional rather than a count of which
+    categories happen to have any.
+
     Plotly renders outlier ``path.point`` elements in value-sorted order
-    (ascending).  Lower outliers come first, upper outliers last.  We use
+    (ascending). Lower outliers come first, upper outliers last. We use
     CSS ``:nth-child(An+B of S)`` to address each group separately, the
     same technique matplotlib uses.
+
+    Parameters
+    ----------
+    prefix : str
+        The subplot CSS prefix.
+    group : int
+        One-based position of the trace's ``<g>`` among the ``boxlayer``'s
+        children -- ``layer_position`` + 1, so a candlestick sharing the
+        layer is counted.
+    box : int
+        One-based position of this box among its own trace's boxes.
+    lower_count, upper_count : int
+        How many outliers fall below and above the whiskers.
     """
-    base = f"{prefix}.boxlayer > g:nth-child({nth_child}) .points"
+    group_sel = f"{prefix}.boxlayer > g:nth-child({group})"
+    box_sel = f"{group_sel} > :nth-child({box} of path.box)"
+    base = f"{group_sel} > :nth-child({box} of g.points)"
 
     if lower_count > 0:
         lower_sel = [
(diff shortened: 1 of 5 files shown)


jooyoungseo:
The bug report doesn't hold — but thank you for flagging the CI discrepancy rather than assuming, because that's exactly what distinguishes it.

### The lone-box branch does register its plot

The excerpt in the review stops one line early. `maidr/plotly/plotly_maidr.py:611-623`:

```python
elif len(box_traces) == 1:
    from maidr.plotly.box import PlotlyBoxPlot

    plot = PlotlyBoxPlot(
        box_traces[0],
        layout,
        layer_position=layer_position(group_traces, box_traces[0]),
        **axis_kwargs,
    )
    plot.row_index = row
    plot.col_index = col
    self._plots.append(plot)          # 622
    merged.update(id(t) for t in box_traces)   # 623
```

Independently confirmed rather than argued from the source: a candlestick-then-box figure produces **1** BOX layer, not 2. A fall-through would have built a second one via the factory, so the count is the discriminator. And its selector is `g:nth-child(2)`, which the factory-built fallback could not produce since it would carry the default `0`.

That also answers "why did CI run green" — `test_a_box_after_a_candlestick_is_in_the_second_group` asserts `nth-child(2)` and passes, which is only possible if the branch's plot is the one that survives. Your trace of the consequences was right in every respect; only the premise was off.

### Pinned anyway

The failure you described is real if that branch ever loses those two lines, and nothing asserted it. Now something does: `TestTheLoneBoxIsBuiltOnce` checks exactly one BOX layer for the candlestick figure and for a lone box, plus that the surviving layer is the one that knows its group. **Introducing the fall-through fails 2 tests.** Worth having — the duplicate would announce the same boxes twice and be invisible otherwise.

### On the browser verification not being in a test

This is the nit I'd rank highest, and it's a real limit rather than an oversight. The Python suite has no browser, so the 15/15 resolution check is a separate measurement; the committed tests assert the emitted strings. I've added `TestTheDomContractTheSelectorsAssume` to name the two facts the shape rests on — a box addressed *inside* its group rather than as one, and outliers scoped to an indexed `g.points` rather than reached as a descendant `.points` — so the specific regressions have somewhere to fail. That catches the shape drifting back, which is what actually happened here; it does not catch plotly changing its DOM out from under us, and I'd rather say so than imply otherwise.

### Minor

`layer_position` is keyword-only now, matching `PlotlyCandlestickPlot`.

On docstring style: I'll leave the measured-behaviour notes in place. The narrative parts are recording *why* a selector has the shape it has — facts measured from a renderer, which a `Parameters`/`Returns` block has nowhere to put and which are the reason this defect took three attempts to find. `_build_box_selector` keeps its `Parameters` section, so the argument contract is still in the conventional form.

Suite **1664 passed**, 23 in this file. `ruff` clean, no line over 88.

---

jooyoungseo:
Thanks for re-tracing it — and for confirming `ohlc` correctly stays out of `_BOXLAYER_TRACE_TYPES`, which is the part of the symmetry fix I was least certain wouldn't over-reach.

**The `.trace.boxes` doc reference — taken, in 32a102e.** You're right that it reads as though the selector filtered on that class. Fixed, and I made the note say the thing that actually matters rather than just dropping the mention: the group is matched by *position* and has to be, because a candlestick's group in the same layer is built the same way. A class would not tell the two apart — which is precisely why `layer_position` has to count both types. That turns a misleading detail into the reason the neighbouring fix exists.

**`layer_positions` not keyword-only — leaving it, and your own note is the reason.** It matches `scatter_positions` in `multiline.py` and `step.py`, which is the convention for the plural list-of-positions parameter; `layer_position` singular now matches `PlotlyCandlestickPlot`. Two conventions, each internally consistent, and the plural one has two existing sites against this one. Making it keyword-only would align it with its singular sibling at the cost of diverging from the three other plural parameters, which seems like the worse trade.

Everything else stands. CI green across 3.9–3.12, 23 tests in the new file, suite 1664.

Merging.

---

Issue #412, opened 2026-08-15
fix(plotly): a trace after an empty one highlights the wrong element


jooyoungseo:
Found while fixing review feedback on #411, and deliberately left out of it: this is pre-existing on `main` and shared by the line, step and area layers, so no one layer can fix it.

## What happens

A scatter-family trace with no points gets **no DOM node at all** from plotly — it does not get an empty one. Every later trace therefore moves up a position in the `scatterlayer`, while py-maidr keeps numbering selectors by the trace's *declared* index.

```python
fig = go.Figure([
    go.Scatter(x=[1, 2, 3], y=[1, 2, 3], mode="lines", name="a"),
    go.Scatter(x=[],        y=[],        mode="lines", name="empty"),
    go.Scatter(x=[1, 2, 3], y=[7, 8, 9], mode="lines", name="c"),
])
```

Emitted on `main` today:

```
.subplot.xy .scatterlayer > .trace.scatter:nth-child(1) path.js-line
.subplot.xy .scatterlayer > .trace.scatter:nth-child(3) path.js-line
```

Measured in Chromium, the rendered `scatterlayer` holds **two** `.trace.scatter` nodes:

| selector | resolves to |
|---|---|
| `nth-child(1)` | line `a` ✅ |
| `nth-child(2)` | line `c` |
| `nth-child(3)` | **nothing** ❌ |

So the second surviving line is announced correctly but highlights nothing. Same class as #316 — correct audio, braille and text, no visible highlight, no warning.

## Why one layer cannot fix it

`_drawn_line_series` already does half the job: it drops the empty series from `data` and from the position list together, which is what #316 added it for. What it cannot do is renumber the positions that survive — they are indices into the *subplot's* scatter traces, and an empty trace belonging to a **different** layer shifts this layer's traces just as much. The compaction has to happen where `position_of` is built in `PlotlyMaidr._extract_plots`, which is the only place that sees every scatter trace on the subplot.

## Scope

Affects `PlotlyMultiLinePlot`, `PlotlyStepPlot` and `PlotlyAreaPlot` identically, since all three number from the same shared map. Verified for lines (above) and for areas; steps share the code path.

## Pinned, not silent

[...]

PR #419, opened 2026-08-15
fix(plotly): generate the positions plotly generates when x is omitted

### LINKED ISSUE #418: fix(plotly): a trace written without x announces nothing
Found while measuring #412. More severe than that one and independent of it, so filed separately.

## What happens

`x` is optional in plotly. Omit it and plotly generates `0, 1, 2, …` — which is how most quick plots are written:

```python
go.Figure([go.Scatter(y=[1, 2, 3], mode="lines")])   # layer emitted, 0 points
go.Figure([go.Scatter(y=[1, 2, 3], mode="markers")]) # layer emitted, 0 points
go.Figure([go.Bar(y=[3, 1, 2])])                     # layer emitted, 0 points
```

All three produce a layer of the right **type** carrying **no data at all**. The chart is drawn, the layer is present, the type is right, and there is nothing to navigate.

## Ground truth

`gd.calcdata[0]` after `Plotly.newPlot` in Chromium:

| figure | `calcdata` x | `calcdata` y | drawn marks |
|---|---|---|---|
| `go.Scatter(y=[1,2,3], mode="lines")` | `[0, 1, 2]` | `[1, 2, 3]` | 1 `path.js-line` |
| `go.Bar(y=[3,1,2])` | `[0, 1, 2]` | `[3, 1, 2]` | 3 bars |

So plotly supplies the positions and draws normally. py-maidr drops the series because it pairs the two arrays with `zip(as_list(x), as_list(y))`, and `as_list(None)` is `[]`, so the zip yields nothing.

## It also hides a second trace

```python
go.Figure([
    go.Scatter(y=[1, 2, 3], mode="lines", name="a"),
    go.Scatter(x=[1, 2, 3], y=[4, 5, 6], mode="lines", name="b"),
])
```

emits **one** series for a two-line chart. The selector happens to be right here (`nth-child(2)`, and both traces are drawn so declaration order holds), but a reader is told the chart has one line.

## Scope

Affects the scatter family and bars at least — the `zip`-two-arrays shape is used across `maidr/plotly/`. Worth auditing every extractor rather than fixing the first three: the same pattern appears in the line, step, area, bar and histogram paths, and each will need the generated positions when its position array is absent.

The fix is to supply `range(len(values))` when the position array is missing, which is what plotly itself does. Care needed on the horizontal cases, where the *value* array is `x` and the generated one is `y`.

## Difficulty

Small per site, but it is several sites and wants a shared helper rather than five copies — the two-copies-drift problem in #416 is recent.

jooyoungseo:
Closes #418. Found while measuring #412, and more severe than it.

`x` is optional in plotly. Omit it and the trace is drawn at `0, 1, 2, …` — which is how most quick plots are written:

```python
go.Figure([go.Scatter(y=[1, 2, 3], mode="lines")])   # layer emitted, 0 points
go.Figure([go.Scatter(y=[1, 2, 3], mode="markers")]) # layer emitted, 0 points
go.Figure([go.Bar(y=[3, 1, 2])])                     # layer emitted, 0 points
```

All three produced a layer of the right **type** carrying **no data at all**. The chart is drawn, the layer is present, the type is correct, and there is nothing to navigate. `go.Bar(y=[3, 1, 2])` is about as ordinary as plotly gets, so this was not a corner case.

The cause is one line of pairing: `zip(as_list(x), as_list(y))`, where `as_list(None)` answers `[]` — documented behaviour that "saves each caller a null check", and which here silently emptied the result.

**It also hid the traces beside it.** Two lines where one omits `x` announced **one** line, so a reader was told the chart had half its content.

## Ground truth

`gd.calcdata[0]` after `Plotly.newPlot` in Chromium — measured, not read from the docs:

| figure | calcdata x | calcdata y | drawn marks |
|---|---|---|---|
| `go.Scatter(y=[1,2,3], mode="lines")` | `[0, 1, 2]` | `[1, 2, 3]` | 1 `path.js-line` |
| `go.Bar(y=[3,1,2])` | `[0, 1, 2]` | `[3, 1, 2]` | 3 bars |

So generating `range(len(values))` reproduces what plotly does rather than inventing a convention. A test pins that the positions start at **zero** for exactly that reason — 1-based or the data's own index would both be plausible-looking and wrong.

## One helper, not five copies

The same two-array pairing appears in the bar, scatter, line, grouped-bar and shared line-series paths. `paired_axes(trace, value_key, position_key)` is the single place, since two copies of one rule is what #416 had just finished removing.

Two decisions worth stating:

- **A short position array is left short.** Plotly pairs the two positionally and draws only as far as the shorter reaches, so padding would invent points it never drew. Truncating is its behaviour, not damage to repair here.
- **The value axis is a parameter**, because a horizontal trace carries its magnitudes on `x` and the generated array is `y`.

## Verification

- **Suite 1683 passed** (1664 + 19 new). No existing test failed — again the finding rather than the reassurance: nothing pinned that an `x`-less trace should announce anything.
- Falsification: removing the generation fails **12 of 19**.
- `ruff check` clean across `maidr/plotly/`; no line over 88.

An explicit `x` is untouched — categorical, numeric, and grouped-bar cases are all pinned, including that `[0, 0, 0]` is not mistaken for a missing array.

---

### DIFF
--- a/maidr/plotly/bar.py
+++ b/maidr/plotly/bar.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
-from maidr.plotly.plotly_plot import PlotlyPlot, as_list
+from maidr.plotly.plotly_plot import PlotlyPlot, paired_axes
 
 
 class PlotlyBarPlot(PlotlyPlot):
@@ -15,8 +15,7 @@ class PlotlyBarPlot(PlotlyPlot):
         return f"{self._subplot_css_prefix()}.trace.bars .point > path"
 
     def _extract_plot_data(self) -> list[dict]:
-        x = as_list(self._trace.get("x"))
-        y = as_list(self._trace.get("y"))
+        x, y = paired_axes(self._trace)
 
         return [
             {MaidrKey.X: self._to_native(xv), MaidrKey.Y: self._to_native(yv)}
--- a/maidr/plotly/grouped_bar.py
+++ b/maidr/plotly/grouped_bar.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
 from maidr.plotly.barnorm import barnorm_scale, stack_shares
-from maidr.plotly.plotly_plot import PlotlyPlot, as_list
+from maidr.plotly.plotly_plot import PlotlyPlot, paired_axes
 
 
 class PlotlyGroupedBarPlot(PlotlyPlot):
@@ -64,8 +64,7 @@ class PlotlyGroupedBarPlot(PlotlyPlot):
         pairs: list[list[tuple]] = []
 
         for trace in self._traces:
-            x_vals = as_list(trace.get("x"))
-            y_vals = as_list(trace.get("y"))
+            x_vals, y_vals = paired_axes(trace)
             fill = trace.get("name", "")
             horizontal = self._horizontal(trace)
 
--- a/maidr/plotly/line.py
+++ b/maidr/plotly/line.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
-from maidr.plotly.plotly_plot import PlotlyPlot, as_list
+from maidr.plotly.plotly_plot import PlotlyPlot, paired_axes
 from maidr.plotly.step_shape import renders_through_webgl
 
 
@@ -66,8 +66,7 @@ class PlotlyLinePlot(PlotlyPlot):
         return [self._scatter_line_selector(self._scatter_position)]
 
     def _extract_plot_data(self) -> list[list[dict]]:
-        x = as_list(self._trace.get("x"))
-        y = as_list(self._trace.get("y"))
+        x, y = paired_axes(self._trace)
         name = self._trace.get("name", "")
 
         line_data = []
--- a/maidr/plotly/plotly_plot.py
+++ b/maidr/plotly/plotly_plot.py
@@ -208,8 +208,7 @@ class PlotlyPlot(ABC):
         drawn_positions: list[int] = []
 
         for trace, position in zip(traces, positions):
-            x_values = as_list(trace.get("x"))
-            y_values = as_list(trace.get("y"))
+            x_values, y_values = paired_axes(trace)
             name = trace.get("name", "")
 
             series: list[dict] = []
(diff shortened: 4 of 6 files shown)


jooyoungseo:
Two corrections to what I first pushed, in d77dec7. Both are mine, and one was caught by the existing suite rather than by me.

### 1. It only fixed half the defect

`paired_axes` took a `value_key`, defaulting to `y`. A horizontal bar carries its magnitudes on `x` and needs `y` supplied, so this still announced nothing:

```python
go.Figure([go.Bar(x=[3, 1, 2], orientation="h")])   # 0 points
```

I had tested the `value_key` parameter on the helper directly and never driven a horizontal figure through the pipeline, which is exactly the gap a unit test can hide. Measured:

| trace | calcdata x | calcdata y | drawn |
|---|---|---|---|
| `go.Bar(x=[3,1,2], orientation="h")` | `3,1,2` | `0,1,2` | 3 bars |
| `go.Scatter(x=[3,1,2], mode="lines")` | `3,1,2` | `0,1,2` | 1 line |

The helper is symmetric now and takes no axis argument at all. Naming one of them "the value axis" made every caller answer an orientation question it does not otherwise ask, and getting that answer wrong is what produced the half-fix.

### 2. It conflated an absent array with an empty one — and the suite caught it

Two existing tests failed: `test_a_trace_with_no_y_counts_as_empty_too[step]` and `[multiline]`. **They were right and my change was wrong.**

Plotly draws the two cases differently, which I would not have guessed:

| trace | calcdata | drawn marks |
|---|---|---|
| `y` **absent** | x `[0,1]`, y `[0,1]` | 1 line / 4 bar marks |
| `y: []` **explicitly empty** | `[null]` | **0** |

`as_list` answers `[]` for both, so the raw key is the only thing that separates them. Generating for an empty array would have invented points for a trace plotly leaves blank — and worse, put back the phantom-series misalignment that #316 removed and that those very tests exist to hold.

The condition is now `trace.get("y") is None` rather than `not ys`, and the distinction has its own tests pointing at the alignment file, since it is the subtle part and the reason the naive version looks correct.

### Verification

- **Suite 1689 passed**, including the two alignment tests that failed on the first pass.
- Falsification: dropping the horizontal half fails **4**; conflating absent with empty fails **4**, two of them the pre-existing alignment tests.
- `ruff` clean across `maidr/plotly/` and `tests/plotly/`; no line over 88.

Worth stating plainly: the first version of this PR would have shipped a half-fix plus a regression, and the thing that stopped it was a test written for an unrelated issue. That is the argument for the pinned tests in this codebase doing real work.

---

jooyoungseo:
The axes bug is real and is fixed in 0fc4eb6. Good catch — it's the third correction to this PR and the second one I'd have shipped.

### `_extract_axes_data` still read the raw arrays — confirmed

Reproduced before touching it, and the reproduction is the interesting part: it **only shows with `dtick` set**. Without a tick step the grid precondition fails anyway, so a fixed and an unfixed trace both emit a bare `label` and look identical:

```
y-only  : x={'label': 'X'}                    # no dtick — looks fine
x and y : x={'label': 'X'}                    # …and so does the control
```

With `dtick`, the divergence appears:

| trace | before | after |
|---|---|---|
| `go.Scatter(y=[1,2,3])` | `{'label': 'X'}` | `{'label': 'X', 'min': 0.0, 'max': 2.0, 'tickStep': 1.0}` |
| `go.Scatter(x=[1,2,3])` | `{'label': 'Y'}` | `{'label': 'Y', 'min': 0.0, 'max': 2.0, 'tickStep': 1.0}` |

So the layer announced points at `0, 1, 2` while its axes reported no range for them — grid navigation silently off. That the control case looks identical without `dtick` is exactly why my tests missed it, and the new ones set `dtick` deliberately.

### The redundant guards — taken

`range(len([]))` is empty, so `and xs` never changed an answer. Removed; an extra condition implies a distinction that isn't there.

### The mixed normalised stack — measured, and it's already correct

This one I checked rather than assumed, and the answer inverts the concern. Plotly does not try to line generated integers up with real category labels either — it places `0, 1` as two **new** categories after `a` and `b`:

```
plotly:  p [(0, 100), (1, 100)]      # categories 'a', 'b'
plotly:  q [(2, 100), (3, 100)]      # generated 0, 1 → positions 2, 3
```

Four bars at four positions, each alone in its stack, each at 100%. py-maidr emits exactly that (`['a', 'b', 0, 1]`, all `100.0`). The keys differing is the correct reading, because the positions genuinely differ. Pinned with the reasoning, since it looks like a mis-key and is not.

### Verification

- **Suite 1695 passed**, 31 in the new file.
- Falsification: reading the raw arrays again fails **3**.
- `ruff` clean across `maidr/plotly/` and `tests/plotly/`.

Tally on this PR: a half-fix (horizontal), a regression (absent vs empty, caught by #316's tests), and this inconsistency (axes range). All three were found by measuring or by existing tests rather than by my reading of the diff, which is worth noting about the diff.

---

PR #420, opened 2026-08-15
fix(plotly): number scatter positions by what plotly draws

### LINKED ISSUE #412: fix(plotly): a trace after an empty one highlights the wrong element

jooyoungseo:
Closes #412.

A scatter-family trace with nothing to plot gets **no DOM node at all** — not an empty one. Every trace after it moves up a position in the `scatterlayer`, while selectors were numbered by the trace's *declared* index:

```python
go.Figure([
    go.Scatter(x=X, y=[1,2,3], mode="lines", name="a"),
    go.Scatter(x=[],  y=[],     mode="lines", name="empty"),
    go.Scatter(x=X, y=[7,8,9], mode="lines", name="c"),
])
```

| emitted | resolves to |
|---|---|
| `nth-child(1)` | line `a` ✅ |
| `nth-child(3)` | **nothing** ❌ |
| *(`nth-child(2)`)* | *line `c`, unaddressed* |

So the second surviving line was announced correctly and highlighted nothing. Correct audio, braille and text; no visible highlight; no warning — the session's recurring shape.

## It's the empty sibling of a rule already here

`is_drawn` implements exactly this for *hidden* traces (#400), with a docstring that names both consequences: such a trace "must not be announced" and "must not take a slot". An empty trace ends the same way — no group in the layer — so positions now count what plotly **draws** rather than what was declared.

`_draws_marks` asks through `paired_axes`, so it agrees with the extraction after #418: a trace omitting one axis **is** drawn, because plotly generates the missing array for it, and only a trace with nothing left after pairing is absent from the layer. The two changes would contradict each other otherwise.

## Undrawn traces are numbered after the drawn ones, not skipped

Their indices are never rendered — `_line_series_with_positions` drops the series and its position together — but the layer classes validate the list they're handed, and a gap or duplicate fails that check. Counting them from the end keeps every index unique and correct-by-construction, which is precisely what the WebGL numbering a few lines above already does, for the same reason.

## One fix, three paths

Lines, steps and areas all read the same map, so this fixes them together — and a per-layer fix was never possible, because an empty trace in one layer shifts the traces in another. That was the reason #412 was filed separately from #392 rather than folded in.

`test_a_band_after_an_empty_one_is_still_numbered_by_declaration`, pinned in #411 to assert the wrong-but-current value so whoever took this would see it fail, **failed exactly as intended** and is updated.

## Verification

- **10 of 10** selectors resolve to exactly one element in Chromium: empty trace first, in the middle, two in a row, an area stack, and an empty trace beside a hidden one (both rules composing).
- **Suite 1708 passed** (1694 + 13 new + the updated pin).
- Falsification: restoring declaration numbering fails **7**.

---

### DIFF
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -18,6 +18,7 @@ from maidr.plotly.plotly_plot import (
     PlotlyPlot,
     domain_interval,
     is_drawn,
+    paired_axes,
     subplot_css_prefix,
 )
 from maidr.plotly.violin import (
@@ -95,6 +96,38 @@ def _is_domain_trace(trace: dict) -> bool:
         ``type`` at all is a ``scatter``, which is cartesian.
     """
     return trace.get("type") in _DOMAIN_TRACE_TYPES
+
+
+def _draws_marks(trace: dict) -> bool:
+    """Whether this trace puts any geometry in its layer.
+
+    The empty sibling of :func:`~maidr.plotly.plotly_plot.is_drawn`, which
+    answers the same question for a *hidden* trace. Both end the same way:
+    plotly renders no group for it, so every trace after it shifts up one in
+    the layer and a selector numbered by declaration lands on the wrong
+    element or on none (#412, and #400 for the hidden case).
+
+    Measured in Chromium -- three line traces with an empty one in the middle
+    produce two ``.trace.scatter`` nodes, and ``nth-child(2)`` resolves to
+    the *third* trace.
+
+    Asked through ``paired_axes`` so it agrees with the extraction: a trace
+    that omits one axis is drawn, because plotly generates the missing array
+    for it, and only a trace with nothing left after pairing is absent from
+    the layer.
+
+    Parameters
+    ----------
+    trace : dict
+        A plotly trace dictionary.
+
+    Returns
+    -------
+    bool
+        True when plotly gives this trace a group of its own.
+    """
+    xs, ys = paired_axes(trace)
+    return bool(xs) and bool(ys)
 
 
 #: Plotly's own default when a figure sets no ``barmode``. It stacks -- and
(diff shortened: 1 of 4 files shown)


jooyoungseo:
All three taken in 14d8a42, and the test suggestion found a bug — in my helper rather than in the code.

### Placement — moved

You're right that it belongs next to `is_drawn`. `draws_marks` is that rule's empty sibling and asks its question through `paired_axes`, which lives there too. Keeping the three together says they are one idea — *what plotly puts in a layer* — rather than two rules in two files. Now public in `plotly_plot.py`, since it crosses a module boundary.

On the "third spelling" point: I'd rather leave `_line_series_with_positions`'s inline `if series:` alone for now. It filters *after* building the points, so it also catches a series that pairs to nothing for reasons the predicate doesn't model; collapsing them would couple the two. If a third caller appears that's the moment to unify.

### The gl coverage — added, and it caught something

Worth having for exactly the reason you gave: a WebGL layer emits no selectors, so nothing downstream complains if those indices go malformed, and the validation happens at construction where only a test reaches it.

Writing it turned up a defect in my own test helper: it assumed every layer carries a `selectors` key. A WebGL layer **omits it entirely** — `render()` drops the key rather than emitting an empty list, which is the documented honest answer for a layer with no highlightable geometry. So my helper raised `KeyError` where the code was behaving correctly. Fixed, with the reason recorded.

### The lone empty line — confirmed, filed as #421

```
go.Figure([go.Scatter(x=[], y=[], mode="lines", name="lonely")])
→ layers: 1 | type: LINE | data: [[]] | selectors: ['...nth-child(1) path.js-line']
```

A line layer with no points and a selector that can never resolve. And you're right that it's untouched by this PR: with only one trace there's nothing after it to shift, so the numbering fix has no bearing — the layer's *existence* is the problem, not its index.

Also worth noting the multi-trace paths already get this right: `test_every_trace_empty_emits_no_selectors_at_all` pins that an all-empty layer emits `data == []` and omits the selector key. A lone empty trace should reach the same answer and doesn't. Filed with that comparison, and with a note to check the equivalent for other trace types rather than fixing only the line case.

Suite **1710 passed**, 15 in the new file. `ruff` clean, no line over 88.

---

PR #422, opened 2026-08-15
fix(plotly): build no layer for a scatter trace that draws nothing

### LINKED ISSUE #421: fix(plotly): a subplot whose only trace is empty emits a phantom layer
Raised in review on #420 and confirmed. Pre-existing and untouched by that PR, whose fix is about *numbering* rather than about whether a lone empty trace should produce a layer at all.

## What happens

```python
go.Figure([go.Scatter(x=[], y=[], mode="lines", name="lonely")])
```

emits:

```
layers: 1
  type: LINE | data: [[]] | selectors: ['... .trace.scatter:nth-child(1) path.js-line']
```

A `line` layer carrying one empty series, and a selector that resolves to nothing — plotly draws no node for an empty trace, as measured for #412.

So a reader is offered a line chart to navigate that has no points in it, and a highlight that can never appear.

## Why #420 does not cover it

The single-line branch in `PlotlyMaidr._extract_plots` builds `PlotlyLinePlot` directly, and that class does not go through `_drawn_line_series` — the helper that drops a series and its position together. With only one trace there is nothing after it to shift, so the numbering fix has no effect here; the layer's *existence* is the problem rather than its index.

The multi-trace paths are already correct: `test_every_trace_empty_emits_no_selectors_at_all` in `tests/plotly/test_plotly_empty_series_alignment.py` pins that a layer whose traces are all empty emits `data == []` and omits the selector key entirely. A lone empty trace should reach the same answer and does not.

## What a fix needs

Either route the single-line branch through `_drawn_line_series` like its siblings, or decline to build a layer for a trace that draws nothing. The second is closer to what `is_drawn` already does for hidden traces — "it must not be announced" — and would also cover the equivalent lone-empty cases for other trace types, which are worth checking at the same time (`go.Bar(x=[], y=[])` and friends).

Worth confirming what the JS side does with a layer whose `data` is `[[]]` before choosing: if it already treats it as empty, this is cosmetic; if it announces a navigable chart, it is not.

## Difficulty

Small, but it changes whether a layer exists, so it wants its own review rather than riding along with a numbering change.

jooyoungseo:
Closes #421.

```python
go.Figure([go.Scatter(x=[], y=[], mode="lines", name="lonely")])
→ layers: 1 | type: LINE | data: [[]] | selectors: ['...nth-child(1) path.js-line']
```

A layer holding one empty series and a selector that can never resolve.

## It does not merely announce nothing — it throws

The issue asked to check the JS side before choosing a fix. I did, by constructing the layer directly in the core:

```
TypeError: Cannot read properties of undefined (reading 'z')

  423 |       zData = point.z ? { z: { label: zLabel, value: point.z } } : {};
      at LineTrace.text        (src/model/line.ts:423:21)
      at LineTrace.state       (src/model/abstract.ts:507:18)
```

`point` is `this.points[this.row]?.[this.col]` — `undefined` for an empty series. The optional chaining stops one step short of the dereference; the guard is on the *property*, not the *point*.

A throw out of trace construction propagates out of `new Figure(...)`, so this takes the **whole render** rather than degrading to one silent layer. Filed as **xability/maidr#905** for the core half — a producer can always emit an empty series, and the core shouldn't crash on one. This PR is the producer half.

## Two attempts, because the first was wrong

Excluding undrawn traces from the groupings **alone** didn't work: they fell through to the fallback factory, which built the same layer. Worse, a drawn-plus-empty figure went from 1 layer to **2**, since the empty one became its own. They have to be marked handled as well — which is what `merged` is for, and how every other deliberate skip in `_extract_plots` works.

## Scoped to the scatter family, deliberately

`draws_marks` reads `x`/`y`. A pie carries neither and draws perfectly well, so asking it of every trace would drop **every pie in the figure**. A bar with no data is likewise left to its own path. Both are pinned, so widening the predicate later is a decision rather than a side effect.

| figure | layers before | after |
|---|---|---|
| lone empty line / markers / area / step | 1 (crashes core) | **0** |
| drawn + empty | 1 | 1 |
| pie | 1 | 1 |
| bar with no data | 1 | 1 |

## Verification

- **Suite 1718 passed** (1710 + 8 new).
- Falsification: not excluding them fails **10 of the 23** tests in the file.

---

### DIFF
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -347,11 +347,27 @@ class PlotlyMaidr:
             # line grouping it would be emitted twice, once as its own layer
             # and once inside the multi-line one. Split out here, before any
             # of the line/step machinery sees the group.
-            area_traces = [t for t in group_traces if is_area_trace(t)]
+            # A trace that draws nothing forms no layer. Plotly gives it no
+            # group, so there is nothing to announce and nothing to highlight,
+            # and the core does worse than ignore such a layer: a series with
+            # no points makes `LineTrace.text` dereference an undefined point
+            # and throw, which propagates out of `Figure` and takes the whole
+            # render with it (#421, and xability/maidr#905 for the core half).
+            #
+            # The multi-trace paths already reach this answer inside
+            # `_drawn_line_series`, which drops an undrawn series and its
+            # position together. Excluding them here means a lone one reaches
+            # it too, rather than bypassing it through the single-trace
+            # branches below.
+            area_traces = [
+                t for t in group_traces if is_area_trace(t) and draws_marks(t)
+            ]
             connected_traces = [
                 t
                 for t in group_traces
-                if is_connected_line_trace(t) and not is_area_trace(t)
+                if is_connected_line_trace(t)
+                and not is_area_trace(t)
+                and draws_marks(t)
             ]
             box_traces = [
                 t for t in group_traces if t.get("type") == "box"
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Nit is real and fixed in `1fce692`. Both branches of that ternary evaluated to `data`, so the normalisation it reads as doing never happened — the assertion was only ever checking whatever shape it was handed. It now checks both explicitly:

```python
assert data, "a layer carries no data at all"
series = data if isinstance(data[0], list) else [data]
assert all(s for s in series), "a layer carries an empty series"
```

The added `assert data` matters more than the branch it fixes: in the flat single-series shape, "empty" *is* the whole list, and the old form would have walked zero elements and passed.

23 of 23 still pass.

---

PR #423, opened 2026-08-15
feat(plotly): keep the step convention of a filled staircase

### LINKED ISSUE #413: fix(plotly): a stepped area loses its step semantics
Raised in review on #411 and confirmed. Narrow, and outside #392's scope, so filed rather than folded in.

## What happens

A trace can be both a staircase and a filled band — `line.shape` and `stackgroup` are independent attributes:

```python
go.Scatter(x=[1, 2, 3], y=[1, 2, 3], stackgroup="one", line=dict(shape="hv"))
```

Since #411, `is_area_trace` is tested before `is_step_trace`, so this is diverted into a plain `area` layer:

```
type: PlotType.AREA | keys: [axes, data, id, selectors, title, type]
```

No `stepDirection`. Before #411 it fell into `step_traces` and kept the step information, but was then announced as a line rather than an area — so neither classification has ever been complete. This is a gap that changed shape rather than a regression: one of the two facts was always dropped.

## Why it matters

A staircase and an interpolated line make different claims about what happens *between* samples. A step says the value holds and then jumps; an interpolated line says it slid. Reading a filled staircase as a smooth band tells a reader the wrong thing about every interval.

## What a fix needs

The two facts are orthogonal, so the layer has to carry both — an area layer that also reports its `stepDirection`, rather than a fourth mutually-exclusive plot type. Worth checking what the JS side does with `stepDirection` on an area layer before deciding whether this is a producer-side or a core change; `AreaTrace` extends `LineTrace`, and the step handling lives on the line side.

## How common

Uncommon but real — a stacked step area is the standard way to draw a cumulative count that changes at discrete events. Low priority relative to the charts that announce nothing.

jooyoungseo:
Closes #413. This is the producer half; the core half landed as xability/maidr#902.

## The defect

`line.shape` and `stackgroup` are independent plotly attributes, so a trace can be both a staircase and a filled band:

```python
go.Scatter(x=[1, 2, 3], y=[1, 2, 3], stackgroup="one", line=dict(shape="hv"))
```

Since #411 `is_area_trace` is tested before `is_step_trace`, so that became a plain `area` layer with no `stepDirection`. Before #411 it fell into the step grouping and kept the direction but was announced as a line. Neither classification was ever complete — one of the two facts was always dropped.

A staircase and an interpolated line make different claims about what happens *between* samples: a step says the value holds and then jumps, a line says it slid. A stacked step area is the standard way to draw a cumulative count that changes at discrete events, so reading it as a smooth band misdescribes every interval of it.

## The fix

The two facts are orthogonal, so the layer carries both rather than becoming a fourth mutually-exclusive plot type. The area type still says the band is filled and how it stacks; `stepDirection` rides alongside it.

```
stacked step area hv     type=STACKED_AREA     stepDirection='hv'
lone step area vh        type=AREA             stepDirection='vh'
hvh                      type=AREA             stepDirection='mid'
vhv                      type=AREA             (absent)
plain area               type=AREA             (absent)
mixed stack              type=STACKED_AREA     (absent)
plain step line          type=STEP             stepDirection='hv'   ← unchanged
```

**A mixed stack says nothing.** This is the one place the area path cannot copy the step path. `PlotlyStepPlot` layers are *split* by direction before construction, so a step layer's traces always agree. A stack cannot be split that way: the bands of one `stackgroup` are one stack, and separating them would leave the core summing part of it and announcing that as the total. So a stack whose bands disagree stays whole and withholds the key rather than describing one of its bands wrongly.

That includes the asymmetric case — one band `hv`, the other authoring no shape at all. "Nothing" is a real disagreement, because plotly interpolates that band, so the layer must not adopt its neighbour's direction.

**One resolver, two callers.** `PlotlyStepPlot._resolve_direction` and the new area path asked the same question, so the logic moved to `shared_step_direction` in `step_shape.py` and both read it. The step path's answers are unchanged and pinned by `TestTheStepLayerIsUnchanged`.

## Verification

Full suite **1739 passed**, `ruff check` clean on every file touched, plus 21 new tests in `tests/plotly/test_plotly_step_area.py`.

Falsification — each mechanism reverted in turn:

| reverted | failures |
|---|---|
| area layer stops emitting `stepDirection` | **6 of 21** |
| resolver takes the first shape it finds instead of requiring agreement | **4** (across the step and area suites) |

## What a reader gets, end to end

`stepDirection` on an area layer is read by `AreaTrace` since maidr#902, which announces the convention under "Step direction" in the description for every variant.

The *highlight* needs one more piece, and it is not in this change: a staircase's rendered path is drawn with `H`/`V` commands, which the core's path parser did not read at all — so a plotly step chart of any kind got no highlight, filled or not. Found while measuring this issue's geometry, filed as xability/maidr#907, fixed in xability/maidr#908. Nothing here depends on that landing; this layer is correct either way, and the highlight starts working when #908 ships.

---

### DIFF
--- a/maidr/plotly/area.py
+++ b/maidr/plotly/area.py
@@ -9,9 +9,14 @@ draws this chart rather than a multi-line one (#392).
 
 from __future__ import annotations
 
+from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
 from maidr.plotly.plotly_plot import PlotlyPlot
-from maidr.plotly.step_shape import is_scatter_family_trace, renders_through_webgl
+from maidr.plotly.step_shape import (
+    is_scatter_family_trace,
+    renders_through_webgl,
+    shared_step_direction,
+)
 
 #: The values ``groupnorm`` takes when plotly rescales a stack to a common
 #: total -- its own switch for a 100% stacked area, and the counterpart of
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Merged. Agreed on the duplicated four lines — I left them deliberately. The two `render()` overrides share a shape but not a reason: the step layer's `None` is a guard against a caller that skipped `group_by_direction`, while the area layer's is an expected outcome for a stack that genuinely mixes conventions. Folding them together would put one comment on two different arguments. The part worth sharing — *how* the direction is resolved — is already shared as `shared_step_direction`, and that is where a third caller would land.

---

PR #424, opened 2026-08-15
feat(plotly): read a plotly.express trendline as a smooth layer

jooyoungseo:
Closes the last open sub-item of #343, after the candlestick (#343a) and violin (#343b) halves.

## The defect

`px.scatter(..., trendline="ols")` appends a second `scatter` trace carrying the fit. Measured, the two traces differ in nothing structural:

```
data       {"mode": "markers", "name": "", "showlegend": false, "marker": {"color": "#636efa", …}, "type": "scatter"}
trendline  {"mode": "lines",   "name": "", "showlegend": false, "marker": {"color": "#636efa", …}, "type": "scatter"}
```

So the fit was merged into the multi-line layer and read as one more series of the user's data. A blind reader was told a model's prediction was a measurement.

## Why I changed my mind about the signal

I twice declined this on the grounds that the only signal is `hovertemplate` — a *display* string, unlike the structural `stackgroup` the area path reads. That premise was wrong, and I should have checked it before asserting it: this package **already** resolves exactly this question from a display string. `maidr/core/enum/smooth_keywords.py` matches a matplotlib artist's `label` — equally a display string, equally user-settable — to find seaborn's regression lines, and has since long before the plotly path existed. Reading plotly's own generated template is the same trade in the same place, so declining it was inconsistent rather than conservative.

## The rule

What I did keep from the objection is that a keyword scan is too loose. The detector matches the *shape* plotly generates rather than a word: the fit's name in bold as the whole opening of the template.

```python
_TRENDLINE_HOVERTEMPLATE = re.compile(r"^<b>[^<>]*\btrendline\b[^<>]*</b>", re.I)
```

Measured across all five modes `px` offers, every one writes that opening:

| `trendline=` | `hovertemplate` opens with |
|---|---|
| `ols` | `<b>OLS trendline</b>` |
| `lowess` | `<b>LOWESS trendline</b>` |
| `rolling` | `<b>Rolling mean trendline</b>` |
| `ewm` | `<b>Exponentially Weighted mean trendline</b>` |
| `expanding` | `<b>Expanding mean trendline</b>` |

Anchored and bounded, so prose a user writes about their own chart — "This is the line of best fit", "Regression: y = 2x", "see the trendline above" — does not reclassify it. Six such strings are pinned as non-matches.

## Behaviour

```
scatter + ols          SMOOTH  series=[10]   SCATTER  series=10
scatter + lowess       SMOOTH  series=[10]   SCATTER  series=10
coloured + ols         SMOOTH  series=[5,5]  SCATTER ×2
px.line (no trendline) LINE
hand-built line        LINE
line + trendline       LINE + SMOOTH
```

The fit is separated *from* the data, not instead of it — a reader navigates the measurements and the model as two things. Splitting happens before the line branches because a layer carries one type for every series it holds: a fit sharing a layer with the lines beside it could only be `line`, or make them all `smooth`.

All fits share one layer, since `color=` fits one trend per group and those are the same kind of thing navigated together — exactly as the multi-line layer holds the series they were fitted to. `PlotlyMultiLinePlot` gained an optional `plot_type` (defaulting to `LINE`) rather than growing a near-duplicate class; that mirrors the core, where `SmoothTrace extends LineTrace` and overrides only `plotType`.

## Verification

Full suite **1768 passed**, `ruff check` clean, plus 29 new tests.

Every selector resolved against real Plotly.js output in Chromium — **4 of 4 matched exactly one element**, across a lone trendline (`nth-child(2)`) and a pair fitted per colour group (`nth-child(2)`, `nth-child(4)`). A lone fit goes through the multi-series path rather than the single-line one so it too is position-scoped; the unscoped form would also match the scatter's own elements.

Falsification:

| reverted | failures |
|---|---|
| trendlines not split out of `line_traces` | **12 of 29** |
| detector loosened to a bare `trend\|fit\|regression` scan | **4** — all the prose cases |

---

### DIFF
--- a/maidr/plotly/area.py
+++ b/maidr/plotly/area.py
@@ -9,9 +9,14 @@ draws this chart rather than a multi-line one (#392).
 
 from __future__ import annotations
 
+from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
 from maidr.plotly.plotly_plot import PlotlyPlot
-from maidr.plotly.step_shape import is_scatter_family_trace, renders_through_webgl
+from maidr.plotly.step_shape import (
+    is_scatter_family_trace,
+    renders_through_webgl,
+    shared_step_direction,
+)
 
 #: The values ``groupnorm`` takes when plotly rescales a stack to a common
 #: total -- its own switch for a 100% stacked area, and the counterpart of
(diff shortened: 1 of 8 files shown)


jooyoungseo:
Housekeeping note taken — fixed in `657e5f9`. #413 landed as a squash (#423), so this branch's own copy of that commit is not an ancestor of `main` and kept showing in the diff even though the content is identical. Merging `main` in resolves it as a no-op for those files. The diff is now the trendline change alone:

```
maidr/plotly/multiline.py             |  10 +-
maidr/plotly/plotly_maidr.py          |  45 ++++++-
maidr/plotly/trendline.py             |  69 +++++++++++
tests/plotly/test_plotly_trendline.py | 223 ++++++++++++++++++++++++++++++++++
```

1768 tests still pass after the merge.

On the residual false-positive risk you flagged — agreed, and I want to name the asymmetry that makes it acceptable rather than just noting the precedent. The two failure directions are not equally bad here:

- **False negative** (today): a fitted curve is announced as `line`. A trendline *is* a line, so the reader loses the fact that it is a model rather than a measurement — real, but not a false claim about the geometry.
- **False positive**: a user's own series is announced as `smooth`. That *is* a false claim — `SmoothTrace` tells the reader the curve is a fit.

So the anchoring is doing load-bearing work, not just tidiness: it is what keeps the worse direction rare. A user who hand-writes `hovertemplate="<b>My trendline</b>…"` on real data still trips it, and that case is genuinely unreachable from structure — but they have had to write plotly's exact generated shape to get there, which is a much narrower door than "mentions a fit somewhere".

Worth adding that the same is true of the `SMOOTH_KEYWORDS` precedent, and rather more loosely: it matches bare `"fit"` anywhere in a matplotlib artist's label. This rule is the stricter of the two.

---

PR #428, opened 2026-08-16
fix(scatter): read the collection each scatter call drew

### LINKED ISSUE #426: A multi-category strip/swarm/boxen plot announces its first category three times and the rest never
## Problem

`ScatterPlot._extract_plot_data()` calls `extract_collection(self.ax, PathCollection)`, which is documented as returning **the first** collection of that type:

```python
# maidr/util/mixin/extractor_mixin.py:266
# We assume only one collection of each type is present to avoid plot clutter,
# even though multiples are technically possible.
return next(
    collection
    for collection in ax.collections
    if isinstance(collection, collection_type)
)
```

seaborn's categorical scatter plots break that assumption: they draw **one `PathCollection` per category**. Each one registers its own `ScatterPlot`, and every one of them re-extracts collection **0**.

## Measured

`stripplot` over 90 rows in 3 categories:

```
PathCollections on the axes: 3   sizes: [30, 30, 30]
distinct first-x per collection: [0.0066, 0.9831, 1.926]   ← the three categories

layers registered: 3
distinct layer payloads: 1                                  ← all three are category "a"
```

Same shape for `swarmplot` (3 × 30) and `boxenplot` (3 × 16).

So a reader of a three-category strip plot:

- hears category **a** three times, once per layer switch, with nothing to tell the layers apart;
- **never hears categories b or c at all** — 60 of the 90 drawn points are absent from the schema.

This is the failure mode worth prioritising: the chart is not empty and does not error. It presents a complete, plausible reading that is wrong about two thirds of the data, and nothing in the output signals that anything is missing.

## Also visible in the same measurement

The x values are the numeric category positions (`0.0066`, `0.9831`, `1.926` — jittered by `stripplot`) while the axis is labelled `g`. Even the one category that is announced is announced as `g is 0.0066` rather than `g is a`. That is a second defect and probably a separate fix; filing it here only so the measurement is not lost.

## Expected

Each registered scatter layer reads the collection it was registered for, so a three-category plot announces three distinct categories and all 90 points.

## Not affected

A plain `plt.scatter` / `sns.scatterplot` draws one `PathCollection`, so the single-collection case — which is what the existing tests cover — is unchanged either way. That is why this has stayed invisible.

jooyoungseo:
Closes #426.

## The defect

`ScatterPlot._extract_plot_data()` took the **first** `PathCollection` on the axes. `extract_collection` states the assumption outright:

```python
# We assume only one collection of each type is present to avoid plot clutter,
# even though multiples are technically possible.
```

A layer is one collection only while nothing draws two, and two things routinely do.

## Measured, before

```
sns.stripplot (3 cats)   collections=3  sizes=[30,30,30]  layers=3  distinct payloads=1
sns.swarmplot (3 cats)   collections=3  sizes=[30,30,30]  layers=3  distinct payloads=1
sns.boxenplot (3 cats)   collections=3  sizes=[16,16,16]  layers=3  distinct payloads=1
two ax.scatter calls     collections=2  sizes=[10,80]     layers=2  → layer sizes [10, 10]
```

So a reader of a three-category strip plot heard category **a** three times, once per layer switch with nothing to tell the layers apart, and **never heard b or c** — 60 of the 90 drawn points absent from the schema entirely.

This is the failure mode worth prioritising over a crash: the chart is not empty and does not error. It presents a complete, plausible reading that is wrong about two thirds of the data, and nothing in the output signals anything is missing.

**It is not a seaborn issue.** Two plain `ax.scatter()` calls fail identically — the second's 80 points were never announced and the first's 10 were announced twice. That is why the fix is at the matplotlib entry point.

## The fix

The patch hands the layer the collection its own call returned, through `drawn_as` — the mechanism #380 introduced for `Axes.bar`, reused rather than reinvented.

`seaborn.scatterplot` is wrapped through the same function and returns an `Axes`, not a collection, so the type guard sends it to the existing sweep. That is **correct** for it, and asserted rather than assumed: measured, it draws every point as a single `PathCollection` even under `hue`, so the sweep finds exactly the right one. If that ever changes, the test fails rather than the reading going quietly wrong.

One thing had to be repaired before the handover could arrive: `MaidrPlotFactory` dropped `**kwargs` on the `SCATTER` branch alone —

```python
return ScatterPlot(single_ax)      # every neighbouring branch passes **kwargs
```

— so the first version of this fix changed nothing at all, and the probe still showed `distinct=1`. Worth naming because the same silent drop would defeat any future per-layer handover to a scatter.

## Measured, after

```
stripplot: layers=3  first-x per layer=[0.04, 0.958, 2.053]   distinct payloads=3
swarmplot: layers=3  first-x per layer=[0.0,  1.0,   1.968]   distinct payloads=3
boxenplot: layers=3  first-x per layer=[0.0,  1.0,   2.0]     distinct payloads=3
two ax.scatter calls -> layer sizes [10, 80]
```

## Verification

Full suite **1786 passed** (1776 + 10 new), `ruff check` clean on every changed file — the 52 findings it reports repo-wide are identical in count on `main`.

Falsification:

| reverted | failures |
|---|---|
| layer sweeps for collection 0 again | **8 of 10** |
| factory drops `**kwargs` on the SCATTER branch | **8 of 10** |

The 2 survivors in both runs are exactly the single-collection non-regression cases — a plain `plt.scatter`, and `sns.scatterplot` under `hue`. Those are the shapes the existing tests already covered, which is why this stayed invisible, so they are the ones a fix has to leave alone.

The strongest assertion is not "the layers differ" but that the announced points are exactly the drawn points, once each — a layer set that were distinct and still missed a category would pass the weaker form.

## Noted, not fixed here

The same measurement shows a categorical scatter's x announced as its numeric tick position (`g is 0.0066`) rather than its label (`g is a`). Recorded in #426; it is a separate fix.

---

### DIFF
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -88,7 +88,7 @@ class MaidrPlotFactory:
         elif PlotType.PIE == plot_type:
             return PiePlot(single_ax, **kwargs)
         elif PlotType.SCATTER == plot_type:
-            return ScatterPlot(single_ax)
+            return ScatterPlot(single_ax, **kwargs)
         elif PlotType.DODGED == plot_type or PlotType.STACKED == plot_type:
             return GroupedBarPlot(single_ax, plot_type, **kwargs)
         elif PlotType.SMOOTH == plot_type:
--- a/maidr/core/plot/scatterplot.py
+++ b/maidr/core/plot/scatterplot.py
@@ -34,7 +34,9 @@ class ScatterPlot(MaidrPlot, CollectionExtractorMixin):
         # single `PathCollection` of every point even under `hue`, so the
         # sweep finds exactly the right one.
         own_points = kwargs.get(DRAWN_POINTS, None)
-        self._own_points = own_points if isinstance(own_points, PathCollection) else None
+        self._own_points = (
+            own_points if isinstance(own_points, PathCollection) else None
+        )
 
     def _get_selector(self) -> str | list[str]:
         return ["g[maidr='true'] > g > use"]
--- a/maidr/core/plot/scatterplot.py
+++ b/maidr/core/plot/scatterplot.py
@@ -11,9 +11,30 @@ from maidr.exception import ExtractionError
 from maidr.util.mixin import CollectionExtractorMixin
 
 
+#: The keyword ``Axes.scatter`` hands its own ``PathCollection`` to this layer
+#: under. Named once and imported at both ends rather than spelled twice:
+#: ``kwargs.get`` falls back to sweeping the axes on a mismatch, so a typo
+#: would not raise -- it would quietly restore the behaviour #426 removed.
+#:
+#: Lives here rather than beside ``common.drawn_as`` because ``maidr.patch``
+#: imports ``maidr.core`` and not the other way about.
+DRAWN_POINTS = "_maidr_points"
+
+
 class ScatterPlot(MaidrPlot, CollectionExtractorMixin):
-    def __init__(self, ax: Axes) -> None:
+    def __init__(self, ax: Axes, **kwargs) -> None:
         super().__init__(ax, PlotType.SCATTER)
+        # The collection this layer's own call drew, when the patch could say.
+        # `None` falls back to the first collection on the axes, which is the
+        # right answer only while a layer *is* one collection.
+        #
+        # Guarded on the type rather than on presence: `seaborn.scatterplot`
+        # is patched through the same wrapper and returns an `Axes`, not the
+        # collection. Falling back is correct there -- measured, it draws a
+        # single `PathCollection` of every point even under `hue`, so the
+        # sweep finds exactly the right one.
+        own_points = kwargs.get(DRAWN_POINTS, None)
+        self._own_points = own_points if isinstance(own_points, PathCollection) else None
 
     def _get_selector(self) -> str | list[str]:
         return ["g[maidr='true'] > g > use"]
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Thanks — took all three. Two acted on, one I have to push back on, because acting on it would break things.

**Line length** — confirmed 89, reflowed in `c9eec76`.

**Test location** — I checked before moving it. `tests/core/plot/` is where the per-plot-type tests live (`test_areaplot`, `test_boxplot_orientation`, `test_errorbar`, `test_regplot_confidence_band`, `test_stepplot`, `test_violinplot_orientation`, …); `test_bar_reads_its_own_bars.py` at `tests/core/` is the outlier of the pair, not the pattern. Leaving it where it is.

**The `HIST` branch — this one is load-bearing, and passing `**kwargs` there is a crash.**

I set out to file the follow-up you suggested and tried to reproduce the latency first. `HistPlot.__init__(self, ax)` accepts no `**kwargs` at all, and `maidr/patch/histogram.py:175` registers through `common()`, which forwards *the user's plotting kwargs*. So the branch is dropping them on purpose.

My first check was misleading and worth recording, since it is the check anyone verifying this would reach for: `plt.hist(data, bins=20)` came back **OK** with `**kwargs` passed. That path is `histogram.py:43`, `FigureManager.create_maidr(ax, PlotType.HIST)` — no kwargs forwarded, so it never exercises the risk. The seaborn path does:

```
HIST branch changed to HistPlot(single_ax, **kwargs):

  plt.hist(bins=20)        -> OK                    ← the misleading one
  sns.histplot(bins=20)    -> TypeError: HistPlot.__init__() got an unexpected keyword argument 'x'
  sns.histplot(kde=True)   -> TypeError: HistPlot.__init__() got an unexpected keyword argument 'x'
```

So the rule the factory actually follows is not "some branches forgot `**kwargs`" — it is that a branch passes them exactly when its class accepts them. `BarPlot` and now `ScatterPlot` take `**kwargs`; `HistPlot` does not. The `SCATTER` branch was a real bug only because I had just given `ScatterPlot` a `**kwargs` it could not receive.

If `HistPlot` ever needs a `drawn_as`-style handover, the change is to give it `**kwargs` first — the factory line follows from that, not the other way round.

---

Issue #252, opened 2025-10-13
feat: Support Strip Plot


Bhoomika2905:
# Feature Request

## Description

Currently, our visualization module supports only a limited subset of Seaborn Cartesian plots. To enhance analytical and visualization capabilities, we should add support for the `sns.stripplot()` function.  
A **strip plot** displays individual data points for one or more categories, often with jittering to prevent overlap.  
It is particularly useful for visualizing raw data distributions alongside summary plots like boxplots or violin plots.

## Proposed Solution

Use Seaborn’s built-in `sns.stripplot()` as the plotting layer. It should handle arguments like `x`, `y`, `hue`, and `data`, and integrate smoothly with existing visualization styles and configurations.

## Additional Context

Below is a simple example demonstrating how a strip plot works:

```python
import seaborn as sns
import matplotlib.pyplot as plt

# Load sample dataset
tips = sns.load_dataset("tips")

# Create a strip plot
sns.stripplot(x="day", y="total_bill", hue="sex", data=tips, jitter=True, dodge=True)

plt.title("Seaborn Strip Plot Example")
plt.xlabel("Day of the Week")
plt.ylabel("Total Bill ($)")
plt.tight_layout()
plt.show()
```

## Checklist

- [ x ] I have searched for similar feature requests and confirmed that this is a new request.
- [ x ] I have provided a clear and concise description of the feature.
- [ x ] I have explained the motivation behind this feature request.
- [ x ] I have outlined a proposed solution or ideas for implementing this feature.
- [ x ] I have provided any additional context or screenshots if applicable.

jooyoungseo:
Measured on `main` at `c07b98c`: a strip plot already reads, so this looks closeable without code.

```python
df = pd.DataFrame({"day": ["Mon", "Mon", "Tue", "Tue", "Wed", "Wed"],
                   "tip": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]})
sns.stripplot(df, x="day", y="tip")
```

```
scatter  [{"x": 0.0, "y": 1.0, "xLabel": "Mon"}, {"x": 0.0, "y": 2.0, "xLabel": "Mon"}]
scatter  [{"x": 1.0, "y": 3.0, "xLabel": "Tue"}, {"x": 1.0, "y": 4.0, "xLabel": "Tue"}]
scatter  [{"x": 2.0, "y": 5.0, "xLabel": "Wed"}, {"x": 2.0, "y": 6.0, "xLabel": "Wed"}]
[...]

Issue #251, opened 2025-10-13
feat: Support Swarm Plot


Bhoomika2905:
# Feature Request

## Description

Currently, our visualization module supports only a limited subset of Seaborn Cartesian plots. To enhance analytical and visualization capabilities, we should add support for the `sns.swarmplot()` function.  
A **swarm plot** is similar to a strip plot but arranges points to avoid overlap, providing a clear visualization of all data points. It is particularly useful for displaying the distribution of individual observations within categories while preserving readability.

## Proposed Solution

Use Seaborn’s built-in `sns.swarmplot()` as the plotting layer. It should handle arguments such as `x`, `y`, `hue`, and `data`, and follow the same design and theming consistency as other supported plots.

## Additional Context

Below is a simple example demonstrating how a swarm plot works:

```python
import seaborn as sns
import matplotlib.pyplot as plt

# Load sample dataset
tips = sns.load_dataset("tips")

# Create a swarm plot
sns.swarmplot(x="day", y="total_bill", hue="sex", data=tips, dodge=True)

plt.title("Seaborn Swarm Plot Example")
plt.xlabel("Day of the Week")
plt.ylabel("Total Bill ($)")
plt.tight_layout()
plt.show()
```

## Checklist

- [ x ] I have searched for similar feature requests and confirmed that this is a new request.
- [ x ] I have provided a clear and concise description of the feature.
- [ x ] I have explained the motivation behind this feature request.
- [ x ] I have outlined a proposed solution or ideas for implementing this feature.
- [ x ] I have provided any additional context or screenshots if applicable.

jooyoungseo:
Measured on `main` at `c07b98c`: a swarm plot already reads, so this looks closeable without code.

```python
df = pd.DataFrame({"day": ["Mon", "Mon", "Tue", "Tue", "Wed", "Wed"],
                   "tip": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]})
sns.swarmplot(df, x="day", y="tip")
```

```
scatter  [{"x": 0.0, "y": 1.0, "xLabel": "Mon"}, {"x": 0.0, "y": 2.0, "xLabel": "Mon"}]
scatter  [{"x": 1.0, "y": 3.0, "xLabel": "Tue"}, {"x": 1.0, "y": 4.0, "xLabel": "Tue"}]
[...]

PR #438, opened 2026-08-16
feat: read seaborn's categorical charts as their data rather than as how they were drawn

### LINKED ISSUE #253: feat: Support Boxen Plot
# Feature Request

## Description

Currently, our visualization module supports only a limited subset of Seaborn Cartesian plots. To enhance analytical and visualization capabilities, we should add support for the `sns.boxenplot()` function. 
A **boxen plot** (also known as a letter-value plot) is an enhanced box plot that provides more detailed information about data distribution, particularly in large datasets. It shows additional quantiles beyond the quartiles, giving a deeper view of the tails and overall distribution shape.

## Proposed Solution

Use Seaborn’s built-in `sns.boxenplot()` as the plotting layer. It should handle arguments such as `x`, `y`, `hue`, and `data`, and align with the same styling and theming standards as other plots in the visualization module.

## Additional Context

Below is a simple example demonstrating how a boxen plot works:
```python
import seaborn as sns

# Load sample dataset
tips = sns.load_dataset("tips")

# Create a boxen plot
sns.boxenplot(x="day", y="total_bill", hue="sex", data=tips)

plt.title("Seaborn Boxen Plot Example")
plt.xlabel("Day of the Week")
plt.ylabel("Total Bill ($)")
plt.tight_layout()
plt.show()
```

## Checklist

- [ x ] I have searched for similar feature requests and confirmed that this is a new request.
- [ x ] I have provided a clear and concise description of the feature.
- [ x ] I have explained the motivation behind this feature request.
- [ x ] I have outlined a proposed solution or ideas for implementing this feature.
- [ x ] I have provided any additional context or screenshots if applicable.

jooyoungseo:
Closes #253. Part of #251, #252.

Two commits, one theme: a seaborn categorical chart was announcing something the *renderer* invented as though it were a measurement. Each is independently reviewable and independently falsified.

---

## 1. `feat(boxen)` — a letter-value plot read as its scaffolding (#253)

`sns.boxenplot` draws the box plot's five-number summary generalised to a variable-depth ladder of quantiles, so a large sample's tails stay legible instead of collapsing into a whisker and a cloud of dots. The depth is the whole point — it is what a boxen has and a box plot does not.

MAIDR had no type that could hold a variable-depth ladder, so the chart fell through to the generic matplotlib patches. What came out was not a partial reading. Measured on three categories of 200 observations:

```
line  : 3 series of 2 points, each series one value repeated twice
point : 3 layers, x = 0.0 / 1.0 / 2.0
```

The line layer was the three median segments, so the chart **announced itself as a line chart** and said each median twice. The point layers were the outliers alone, at numeric slots rather than at `a`/`b`/`c`. Every rung of every ladder — the entire chart — was absent, and nothing suggested anything was missing.

maidr 4.3.0 shipped `TraceType.BOXEN`, which carries a median, a ladder of `{p, lo, hi}` rungs and whatever fell beyond the deepest one, and announces each rung as the percentile it actually is. `BoxenPlot` emits it. Verified `boxen` is in the pinned 4.3.0 bundle's vocabulary, so the new `PlotType` raises no staleness warning.

Extraction reads what was drawn: seaborn renders each ladder as a `PatchCollection` of `2k - 1` boxes whose edges along the value axis are exactly the letter values it computed. Four decisions follow:

- **Rungs come from the boxes, each consumed once**, not from a set of their shared edges. Adjacent boxes meet at a quantile but hold *separately computed* floats, and when those differ in the last bits both survive the set. Measured that way, one 200-sample group came out with five rungs instead of four, `lo` repeated across two of them, and the widest rung a copy of its neighbour — all plausible in the payload.
- **The median is matched by containment**, not by position in `ax.get_lines()`. Order works on a chart seaborn drew alone, but a threshold drawn *before* the boxen takes index 0 and shifts every median onto the wrong artist.
- **Orientation comes from the median segment** — the axis it runs along is the category axis — covering `orient=`, `x=`/`y=` and seaborn's dtype inference with one rule.
- **A hue level is identified by inverting seaborn's dodge arithmetic**, not by walking left to right, so a category missing a level does not rename every ladder after it.

Only the tail probability naming each rung is inferred, from seaborn's own construction: `LetterValues.__call__` builds its percentiles as `0.5 ** arange(k + 1, 1, -1)`. `k_depth="full"` is the one setting that departs from it — seaborn overwrites the outermost pair with 0 and 100 — and the nominal `p` is kept there deliberately: the core drops a rung whose `p` is outside `(0, 0.5)`, and dropping it would take the sample's range out of the reading and leave `Min value`/`Max value` reporting the *second* deepest quantile. At the depths `"full"` reaches the two differ by well under one observation.

The patch suppresses the inner `Axes.plot` and `Axes.scatter` calls through the existing internal context — not tidying, since those two produced the reading this replaces.

**Falsification** (disjoint sets, so the five mechanisms are independent rather than one tested five times):

|---|---|
| rungs from deduplicated edges | 4 |
| no internal context in the patch | 1 |
| orientation hardcoded to vertical | 2 |
| median matched positionally | 1 (`before` case only) |
| hue index by left-to-right walk | 1 |

The positional-median case is worth naming: my first version of that test appended the threshold *after* the boxen, which leaves the medians at the indices they were already at — so it passed under the reverted code and exercised nothing. Drawing the reference first is what makes it load-bearing, and both orders are now parametrized.

---

## 2. `fix(scatter)` — a strip plot announced its jitter (part of #251, #252)

`sns.stripplot` scatters each point sideways by a random offset so overlapping observations stay separable; `sns.swarmplot` runs a packing algorithm to the same end. Neither offset is an observation — but the offset is what `get_offsets` returns, so it is what was announced. Measured on a three-category strip plot whose x axis is labelled `g` and whose ticks read `a`, `b`, `c`:

```json
{"x": -0.03994198890364789, "y": 0.1257302210933933}
{"x":  0.06297755824455262, "y": -0.1321048632913019}
{"x": -0.07398610980531882, "y": 0.6404226504432821}
```

A reader was given a precise number for a quantity that does not exist, where the chart says a name. For a strip plot that number is literally random — run the same script twice and it changes.

**The chart's shape went with it.** `ScatterTrace` groups points into columns by exact `x` equality, so 90 jittered points became 90 columns of one point each instead of 3 columns of 30. Column navigation stepped through individual observations and never through categories — the axis a strip plot exists to compare across.

A point is now reported at its nearest tick when, and only when, the axis really is a category axis. The test for that is the one #353 established (matplotlib leaves a `UnitData` only on an axis it mapped strings onto), so a numeric axis is untouched — which matters, because there the drawn position *is* the value and snapping would round every measurement onto a gridline while the payload still looked like a scatter. Both axes are asked, since categories sit on y when the chart is turned on its side; asking about x alone was itself the #353 defect.

This does **not** put the category's name in the payload. `ScatterPoint.x` is typed `number` and `ScatterTrace` subtracts x values to sort, index columns and resolve highlights, so a string would give an unstable sort and a broken index rather than an announcement. Filed as xability/maidr#927. The position is the half fixable here, and it is the half that carries the navigation — which is why #251/#252 stay open.

**Falsification:** returning the drawn coordinate unchanged fails 3 tests; snapping to every tick rather than to a category's fails 5, including the numeric-axis guard.

**One existing assertion changed rather than being added to.** #426's `test_every_drawn_point_is_announced_exactly_once` compared the announced pairs against the raw drawn offsets — the behaviour being removed. It now identifies an observation by its `y`, which is the measurement, and the `x` claim is a case of its own. The count and uniqueness assertions that actually guard #426 are untouched, and reverting #426's fix still fails them.

---

## Verification

- `uv run pytest` — **1862 passed**
- `ruff check maidr/ tests/` — 7 errors, all pre-existing `F401` re-exports in `__init__.py` files, none in the files touched here
- Tests assert the boxen ladder against `np.percentile` re-derived from the letter-value definition rather than against a stored payload, because the failure being replaced looked entirely plausible: the way to tell a ladder that was *read* from one that was *assembled* is whether its rungs are the quantiles seaborn computed.


jooyoungseo:
Both flagged items reproduced and fixed in `290bf30`. Both were real, and both were the failure mode this PR is about — a reading that is wrong and sounds complete.

## 2. Blended-transform lines — confirmed

`axhline(0.0, xmin=.15, xmax=.35)` over two categories:

```
medians read : [0.0,    -0.1934]
true medians : [0.0484, -0.1934]
```

The threshold was read as category `a`'s median. Your reasoning about why the existing test passed is exactly right: its `axhline` spanned the whole axes, so x ran 0→1, straddled both categories and was contained in neither. A property of the fixture, not of the check.

Now asks `line.get_transform() is self.ax.transData` — the same test `MultiLinePlot` uses to keep a reference line out of a line layer (#434). It stays a question about the transform rather than the shape, so a genuinely flat data line is still a median.

## 1. `gap=` — confirmed, and worse than "near slot boundaries"

Four hue levels:

```
gap=0.0   a, p   a, q   a, r   a, s
gap=0.3   a      a, q   a, r   a
gap=0.6   a      a, p   a, s   a
```

At 0.3 the outer ladders lose their level name; at **0.6 they are given the wrong one** — `a, p` against a ladder the chart drew as `a, q`. The error compounds outward from the middle, so it is the extremes that go first.

Rather than correcting the pitch, the arithmetic is gone: every ladder's offset from its nearest tick, deduplicated and sorted, **is** the dodge lattice, and a level is its offset's rank among them. No pitch, slot width or gap enters it. Collecting offsets across the whole chart rather than per category also survives a category missing a level, so long as another category draws it — the offsets belong to the dodge, not to one slot's data.

This made extraction two passes: naming needs the lattice, and the lattice is only visible once every centre is known. A ladder genuinely cannot say which level it is from its own position alone — which is the fact the old formula was working around.

Your suggestion of a 3+ level test was the right instinct; the parametrized case covers 4 levels at gaps 0.0 / 0.3 / 0.6.

## 3. `common()` — comment added

Correct that the reason is the before/after collection diff, which `common()` has no hook for. Now stated in the patch so the next reader doesn't have to derive it.

## Verification

|---|---|
| the transform check | 1 |
| pitch back to the widest box | 4 (all three `gap` cases **and** the original two-level one, which the old code passed) |

That last row is the useful one: the two-level test was green under the broken formula, so it was not pinning what it appeared to.

`uv run pytest` — **1872 passed**. `ruff check` clean on touched files.

---

jooyoungseo:
Correction to my previous comment: item 3 was **not** in `290bf30` — I wrote that it was before checking, and it wasn't. It is in `d38e21b` now.

The reason recorded in the patch: `common()` draws inside the internal context and registers in one step, and this layer needs the axes' collections from *before* the call, which there is no hook for between the two halves. Widening `common()` for one caller would put a before/after snapshot in the path of every patch that doesn't need one.

Items 1 and 2 are as described and were in `290bf30`.

---

jooyoungseo:
All three addressed in `93e33b2`. The first was a real defect, not just a documented limitation.

## Data-space reference line — reproduced

```
ax.plot([-0.1, 0.1], [0.5, 0.5])   # then sns.boxenplot(..., x="g", y="v")

medians read : [0.5,    -0.1934]
true medians : [0.0484, -0.1934]
```

You were right that this is the same class as the `axhline` case, and right that the docstring was accepting it rather than closing it. It's closed now, and it turns out containment was the wrong test all along rather than an insufficient one.

**What separates a median is span, not position.** seaborn draws a median *across* the ladder, so its endpoints are the ladder's own extent on the category axis:

```
seaborn's medians : -0.4 → 0.4   and   0.6 → 1.4     (widest box is exactly 0.8)
the user's segment: -0.1 → 0.1
```

A median is now a segment that runs the ladder end to end on one axis and holds a single value inside it on the other. That is also orientation agnostic, so it still runs before orientation is known — which was the reason containment was reached for in the first place.

Reverting to containment fails the new test and nothing else.

## Style items

- `typing.List/Optional/Tuple` → builtin generics, matching the rest of `maidr/core/plot/` under `from __future__ import annotations`.
- `matplotlib.pyplot` moved to module scope — there was no deliberate reason, so no comment was warranted, just the move.
- The `id()` snapshot now records the invariant you asked about: every collection counted stays referenced by `ax.collections` for the whole window, so none can be freed and have its address reused by a collection seaborn is about to add.

## On the `1e-6` clustering tolerance

Noted rather than changed. The offsets being clustered are the *same* lattice position arriving from different categories, so they differ only in the last bits — orders of magnitude below the tolerance — while distinct dodge positions are separated by a fraction of a category width. There is no realistic `gap`/`width` combination that brings those within 1e-6 of each other, since `gap` shrinks boxes and never moves centres. If a report ever does mention mislabelled levels under an unusual combination, this is the line to look at.

`uv run pytest` — **1873 passed**.

---

jooyoungseo:
All four addressed in `4dddc0a`.

## 1. Silent partial extraction

Agreed, and it's the same failure arriving from the other side — every other guard here stops a *wrong* reading, this one a *short* one. A chart announcing two distributions where three were drawn sounds exactly as complete as a correct one. It warns now, naming both counts.

Also asserted: a complete chart warns about **nothing**. A warning on every ordinary chart is how a real one gets missed.

## 3. Legend-ordering coupling — verified rather than commented

You were right that this is load-bearing and silent if it breaks. Rather than only noting it, `hue_order` turns out to move both together, which makes it testable:

```
hue_order=["r","q","p"]   legend ['r','q','p']   z  'a, r'  'a, q'  'a, p'
hue_order=["q","p","r"]   legend ['q','p','r']   z  'a, q'  'a, p'  'a, r'
```

So the assumption is now pinned as well as written down. Reversing the level list fails 6 tests — every naming test, including the plain two-level one.

**Worth flagging: my first version of that test was broken.** It reused the two-level hue fixture and overwrote its hue column with a longer one, which truncated — each category got two of three levels, four ladders were drawn instead of six, and it failed for a reason unrelated to what it checks. It has its own balanced fixture now, documented as to why balance matters.

## 2. Per-ladder recomputation

Done — `_tick_labels` and `get_legend` belong to the axes, not to any one ladder, and extraction was already two passes, so they're read once and passed in.

## 4. Unused binding

Removed.

## Verification

|---|---|
| the partial-extraction warning | 1 |
| hue level list reversed | 6 |

`uv run pytest` — **1877 passed**. `ruff check` clean on touched files.

---

jooyoungseo:
Both points from this round handled in `0ad6aba`.

## The `_ladders()` fallback — removed

You're right, and the framing is the useful part: it was unreachable *and* it was precisely the pairing two commits of this change went to remove, so keeping it armed a trap that would fire silently. Gone. With no collections there are no ladders and `_extract_plot_data` raises.

Verified rather than reasoned:

```python
BoxenPlot(ax).render()   # on an axes seaborn drew
→ ExtractionError: Error extracting data for boxen plot type ...
```

Previously that returned a plausible-looking layer.

## `path.get_extents()` and data space — honest answer

I can't claim what you asked for. `pyproject.toml` floors seaborn at `>=0.12`, and I verified only against the installed **0.13.2**; 0.12 predates the `_CategoricalPlotter` rewrite, so its construction path may well differ and I have not measured it.

What I can say is that the failure mode you describe is loud rather than silent here. The tests assert every rung against `np.percentile` re-derived from the letter-value definition, so display-space bounds would be off by orders of magnitude and fail immediately — they would not produce a plausible-but-wrong ladder. That is the property worth having, but it only fires on whatever seaborn CI installs. A pinned-floor job would be the real answer and this PR does not add one.

## On the duplicated narrative

Fair, and I'd rather note the tradeoff than argue it away: the measured examples do appear in the PR body, the commit message and the docstring. My reasoning is that the commit message and PR body are read once at review time while the docstring is read whenever someone next touches the code, and the thing most easily lost is *why* a check exists — the `axhline` case looks like defensive paranoia until you see the numbers it produced. But you're right that it is more prose to keep in sync, and a future behaviour change here means updating it in more than one place.

## On the changed `#426` assertion

Agreed it is easy to skim past, which is why it's called out in that commit's message too — an edit to an existing regression test deserves more scrutiny than a new one, not less. The count and uniqueness assertions that actually guard #426 are untouched, and reverting #426's fix still fails them.

`uv run pytest` — **1877 passed**.

---

jooyoungseo:
Following up on the `path.get_extents()` question from the last round — I said I couldn't claim anything about the supported seaborn range and would need to measure. I measured, and it turned up something bigger than the extents.

## `import maidr` does not survive seaborn 0.12 at all

`pyproject.toml` declares `seaborn>=0.12`. On 0.12.2:

```
AttributeError: type object '_CategoricalPlotter' has no attribute 'plot_boxes'
```

`maidr/patch/boxplot.py` branches on `min_version = "0.12"`, but `_CategoricalPlotter.plot_boxes` arrived with the categorical rewrite in **0.13**. So 0.12 takes the 0.13 branch and `wrapt` raises at import time. (The `< 0.12` branch is broken too — it passes the version *string* where the wrapper function goes — but nothing reaches it.)

Filed as #441. Pre-existing, nothing to do with this PR, and it means the declared floor is not real.

## What that means for the extents question

It answers it in a way I didn't expect: **0.13 is the effective floor**, so `get_extents()` is only ever asked of 0.13's construction, which is what the `np.percentile` tests already cover.

But while I was in there I found the layout difference that *would* have mattered:

| | boxes drawn | collection order |
|---|---|---|
| 0.13 | `2k - 1`, tiled edge to edge | ladder, then fliers |
| 0.12 | `k`, nested inside one another | fliers, then ladder |

Reading nested boxes with the tiled rule gives **two rungs instead of four**, with `lo` and `hi` taken from different rungs and both mislabelled — plausible, not a crash. Recorded on #441 so whoever fixes the floor doesn't rediscover it.

## What I changed here, and what I deliberately didn't

**Changed** (`041eb1e`): the flier pairing. `_ladders()` took "the collection after this one", which is ownership by position — the pattern this PR replaced everywhere else — and it inverts between the two versions, pairing a ladder with the *next group's* fliers. Now matched by geometry, the same rule `_median_of` uses. That's a real improvement on 0.13 independently of 0.12, since it no longer depends on list order at all.

**Not changed**: `_levels()` still reads only the tiled layout. I wrote the nested branch, then removed it — it's unreachable while `import maidr` fails on 0.12, and shipping a branch for a layout nothing can reach is speculative rather than defensive. That's the same reasoning that removed the sweep fallback one commit earlier, and it seemed wrong to apply it in one direction only.

`uv run pytest` — **1877 passed**.

---

jooyoungseo:
All handled in `f1e43bf`. Thanks in particular for checking the `render()` ordering against the base class rather than taking the commit message's word for it — that one *was* load-bearing and I'd only asserted it.

## 1. `_ladders()` called twice

Right — the second call was pure re-scan for a number the first pass already had. The list is kept.

## 2. The nearest-tick assumption

This was the useful one, and you're right that "an assumption rather than something asserted" is the wrong shape for this PR specifically. It turns out to be a **bound**, not just something that holds in practice: seaborn dodges `n` levels into a slot `width` wide, so the outermost centre sits `width * (n - 1) / (2n)` from its tick — always under `width / 2`, against a categorical spacing of 1.

Measured, against a half-spacing of 0.5:

```
n=2 width=0.8  ->  0.2000
n=6 width=0.8  ->  0.3333
n=6 width=1.0  ->  0.4167
n=8 width=1.0  ->  0.4375
```

It tightens toward 0.5 as levels are added and reaches it only in the limit, at a width that would already have adjacent categories touching. The arithmetic is now at the comparison, and the tightest reachable case (8 levels, `width=1.0`) is a test.

## 3. `_median_of`'s O(ladders × lines)

Agreed and left alone.

## 4. The fully-empty extraction path

You're right that only the partial-failure path was covered. Added — and reaching it by constructing the layer without the patch's handover pins a second thing: that the handover is *required* rather than an optimisation, which is what removing the sweep fallback made true two commits ago.

`uv run pytest` — **1879 passed**.

---

jooyoungseo:
Blocking finding confirmed and fixed in `85d1635`. This was mine, and the second half of it means I told you something untrue earlier.

## Nine methods, ~250 dead lines

Verified with `ast` rather than by eye:

```
_cloud_within   x2  identical=True    236-274, 576-614
_box_bounds     x2  identical=True    276-290, 616-630
_median_of      x2  identical=True    292-366, 632-706
_is_vertical    x2  identical=True    368-379, 708-719
_hue_levels     x2  identical=True    381-399, 721-739
_tick_labels    x2  identical=True    401-407, 741-747
_dodge_offsets  x2  identical=True    409-456, 749-796
_category_of    x2  identical=True    458-516, 798-856
_levels         x2  identical=False   518-574, 858-927
```

Your diagnosis of the cause is exactly right — a later commit appended new versions instead of replacing them. 388 lines removed; the class now has 15 methods and no duplicates.

## The `_levels` pair — a correction

You spotted that the `_levels` duplicate "legitimately differs" and reads like an edit that wasn't applied in place. That is precisely what it was, and the consequence is worse than dead code: **the version that actually ran was the nested-aware one**, i.e. the one the edit meant to remove.

So `041eb1e`'s commit message and my comment above both said the seaborn 0.12 branch had been removed. **It hadn't been.** It is removed now, along with `_is_nested`, which is what those messages described and what their reasoning argued for.

No behaviour change on 0.13, which is why nothing caught it: the nested branch tests `_is_nested()` first, false for a tiled ladder, so both versions took the same path. Your point about tests passing throughout is the whole lesson here.

That also moots your note about the nested path being untested — there is no nested path now.

## A guard, since green tests can't see this

`tests/core/test_no_shadowed_methods.py` walks every class in `maidr/` and fails on a repeated method name. The package is clean today; run against the **committed** `boxenplot.py` it reports all nine.

It has two tests of its own, because the check asserts an empty list and a detector that never finds anything produces one too:
- a doubled class is reported;
- a name reused in a different scope (a closure inside a method, two classes each with `read`) is not.

## Remaining notes

- **Positional `ax`** — agreed, and agreed it's pre-existing: `histogram.py` and `pointplot.py` use the same `kwargs.get("ax")` idiom. Better hardened across all three together than in this PR.
- **`1e-6` clustering tolerance** — the categorical unit spacing is always 1 and distinct dodge positions are separated by a fraction of it, while the values being clustered are the same lattice position differing in the last bits. Comment warranted; I'll fold it in rather than leave it as an unstated scale assumption.
- **`CLAUDE.md`** — fair that it's outside the stated scope. It was stale in a way that would mislead the next reader of that file, so I'd rather carry it than leave it.

`uv run pytest` — **1882 passed**.

---

jooyoungseo:
Both handled in `0b3f24f`.

## The unnameable-ladder branch

You framed this exactly right — it was the one place in the file that failed *quietly*, which is out of step with everything else in it. A ladder losing its level reads as a chart that never had a hue split rather than one whose split went missing.

It warns now, naming the offset, the rank and both counts. And I drove the branch in a test rather than leaving it unexercised: a class written around "say so when you cannot read something" should hold its own unreachable branches to the same rule. Removing the warning fails that test and nothing else.

## `LineExtractorMixin` reach-across

Agreed, and I took your first suggestion — `ScatterPlot` mixes it in and calls it as its own. It worked as a static method, but reaching through an underscore-prefixed name on a class outside the MRO is a surprise for no gain.

## `_median_of`'s O(ladders × lines)

Still there, still deliberate. Worth noting the bound in case it comes up again: the loop is over `ax.get_lines()`, so it grows with *manually added* lines, not with data — seaborn's own medians are one per ladder. A chart would need dozens of hand-drawn reference lines before it registered.

## On the clustering tolerance

Folded the reasoning into the code in `5367129` rather than leaving it as an unstated scale assumption: a category axis puts its ticks one unit apart whatever the data is, so an offset is a fraction of 1, and two distinct lattice positions are separated by `width / n` — 0.1 at eight levels in a full-width slot, five orders of magnitude above the tolerance.

`uv run pytest` — **1883 passed**.

---

Issue #441, opened 2026-08-16
`import maidr` raises on seaborn 0.12.x, which pyproject.toml declares as supported


jooyoungseo:
## The defect

`pyproject.toml` declares `seaborn>=0.12` in both dependency groups. On seaborn 0.12.x the package cannot be imported at all:

```
$ python -c "import seaborn; print(seaborn.__version__); import maidr"
0.12.2
AttributeError: type object '_CategoricalPlotter' has no attribute 'plot_boxes'
```

`maidr/patch/boxplot.py`:

```python
sns_version = seaborn.__version__
min_version = "0.12"

if version.parse(sns_version) < version.parse(min_version):
    wrapt.wrap_function_wrapper("seaborn.categorical", "_BoxPlotter.plot", sns_version)
else:
    wrapt.wrap_function_wrapper(
        "seaborn.categorical", "_CategoricalPlotter.plot_boxes", sns_infer_new_orient
    )
```

Two things are wrong with it.

**The threshold names the wrong release.** `_CategoricalPlotter.plot_boxes` arrived with the categorical rewrite in seaborn **0.13**, not 0.12. `0.12.2 >= 0.12`, so 0.12 takes the 0.13 branch and `wrapt` raises while resolving the attribute — at import time, so nothing the user writes gets a chance to run.

**The legacy branch would not work either.** It passes `sns_version` — a *string* — where `wrapt.wrap_function_wrapper` expects the wrapper function. That branch is unreachable today (nothing is `< 0.12` and also installable), so the mistake has never surfaced, but it means there is no working path for pre-0.13 seaborn in either direction.

## Reproduce

```bash
uv venv sb12 --python 3.11
uv pip install --python sb12/bin/python "seaborn==0.12.2" "pandas<2.2" "matplotlib<3.10" wrapt scipy lxml htmltools jinja2
uv pip install --python sb12/bin/python --no-deps -e .
sb12/bin/python -c "import maidr"
```

(`pandas<2.2` is needed separately: seaborn 0.12 uses `mode.use_inf_as_na`, which pandas 2.2 removed.)

## What to decide

Either is defensible; the current state is the one that isn't.

1. **Raise the floor to `seaborn>=0.13`** in both dependency groups. Honest about what is tested — CI installs from the lockfile, so 0.13.2 is the only version any test has ever run against — and the smaller change.
2. **Make 0.12 genuinely work**: compare against `"0.13"`, and give the legacy branch a real wrapper function instead of a string.

[...]

PR #442, opened 2026-08-16
fix: describe the lines a layer's own calls drew, and declare a seaborn floor that imports

### LINKED ISSUE #440: A line layer sweeps every line on the axes, so a reference line over a box plot announces the box's geometry as data
## The defect

`MultiLinePlot._series()` sweeps every data-space line on the axes rather than the ones its own call drew:

```python
return [line for line in self.ax.get_lines() if self._is_in_data_space(line)]
```

That is correct while nothing else draws lines. Box plots, violins and boxen plots all do — matplotlib renders whiskers, caps and medians as `Line2D` objects in data space — so drawing one reference line over any of them makes the line layer describe the *box's internal geometry* as a chart.

Measured, one `ax.plot([0, 1], [0.5, 0.5])` followed by each:

| chart | series in the emitted `line` layer | correct |
|---|---|---|
| `ax.plot` alone | 1 | 1 ✓ |
| `ax.plot` + `sns.boxplot` | **11** | 1 |
| `ax.plot` + `sns.violinplot` | **7** | 1 |
| `ax.plot` + `sns.boxenplot` | **3** | 1 |

Every extra series is two points long, because a whisker or a cap is a two-point segment. A reader switching to the line layer of that chart is walked through ten two-sample "series" whose values are whisker endpoints and cap positions, announced exactly as data would be. Nothing says otherwise.

Drawing a threshold on a box plot is ordinary — a target, a control limit, a prior year's median.

## Reproduce

```python

df = pd.DataFrame({"g": np.repeat(list("ab"), 50),
                   "v": np.random.default_rng(0).normal(size=100)})
fig, ax = plt.subplots()
ax.plot([0, 1], [0.5, 0.5])          # one reference line
sns.boxplot(df, x="g", y="v", ax=ax)

for p in FigureManager.get_maidr(fig)._plots:
    if p.type.value == "line":
        print(len(p.schema["data"]), "series")   # 11
```

## Why this is not the reference-line issue

#434 was about `axhline`/`axvline`, whose blended transform puts them outside data space; `_is_in_data_space()` already excludes those. This is the reverse direction — lines that *are* in data space and belong to a different layer. The sweep cannot tell them apart because it never asks which call drew them.

## The shape of the fix

The mechanism already exists twice in this codebase. #380 introduced `drawn_as` so a layer receives the artists its own call produced, and #426 extended it to scatter as `DRAWN_POINTS`; py-maidr#438 uses the same pattern for boxen's collections. `MultiLinePlot` takes a `lines=` argument already — `maidr/patch/pointplot.py` passes it — so the patch has somewhere to hand them.

- [ ] Have `maidr/patch/lineplot.py` pass the `Line2D` objects its own call returned
- [ ] Keep the axes sweep as the fallback for a call that returns no artists, and say in the docstring which case that is for
- [ ] A test per companion chart (box, violin, boxen), since each draws a different number of lines and only the count distinguishes a correct layer from a swept one

## Scope

Pre-existing and independent of py-maidr#438 — found while reproducing a different defect there, and reproduces on `sns.boxplot`, which has been supported far longer. Filed separately rather than folded in.

## Related

- #380, #426 — the ownership-by-identity pattern this needs
- #434 — the blended-transform exclusion that is already in place and does not cover this
- #438 — where it surfaced

jooyoungseo:
Closes #440, #441. Two independent fixes on one branch (repository branch policy), each falsified separately.

---

## 1. `fix(line)` — a line layer described lines its own calls never drew (#440)

`MultiLinePlot._series()` swept every data-space line on the axes. That's right while nothing else draws lines — and box plots, violins and boxen plots all do, since matplotlib renders whiskers, caps and medians as `Line2D` objects in data space. So **one reference line** over any of them made the line layer describe the *box's own geometry* as a chart:

| chart | series in the `line` layer | correct |
|---|---|---|
| `ax.plot` alone | 1 | 1 ✓ |
| `ax.plot` + `sns.boxplot` | **11** | 1 |
| `ax.plot` + `sns.violinplot` | **7** | 1 |
| `ax.plot` + `sns.boxenplot` | **3** | 1 |

Every extra series is two points long, because a whisker is a segment. A reader switching to that layer was walked through ten two-sample "series" whose values are whisker endpoints and cap positions, announced exactly as data would be.

**The fix turns on something already there.** The internal context already separates the two, so nothing new has to: a companion chart draws its lines inside its own patch's context and `line()` declines them at the top, while a user's `ax.plot` arrives with the context clear. Measured:

```
user ax.plot ->  [(internal=False, 1 line)]
sns.boxplot  ->  [(internal=True, 1), ... ×12]
```

**The sweep was doing one thing worth keeping** — letting several `ax.plot()` calls form one multi-series layer. So lines accumulate on the axes and the list is handed over by reference; extraction is lazy, so later additions still reach it. That's the half a naive "pass this call's lines" fix breaks, trading a wrong reading for a missing one.

### Review round 1 — the implicit-axes case

The first version fixed only the spelling with `ax=` passed. `wrap_seaborn` wraps `lineplot` as a module-level function, so `instance` is always None; with no `ax=` either, the before-snapshot was empty and every line already on the axes looked newly drawn:

```python
sns.boxplot(data=df, x="g", y="v")          # no ax=
sns.lineplot(data=summary, x="g", y="v")    # no ax=
-> line layer: 11 series, values are whisker endpoints
```

That is #440 intact, in the *more common* spelling of the pairing the issue was filed about. The target is now resolved the way seaborn resolves it, `plt.gca()` included.

### Review round 2 — the same sweep, one level down

Narrowing `_series()` fixed which lines a layer *describes*. It left two reads of `ax.get_lines()` behind in `maidr/util/step_utils.py`, and both decide things about a **step** chart:

| ordering | was | should be |
|---|---|---|
| `sns.boxplot(...)` then `ax.step(...)` | type `line` | type `step` |
| `sns.violinplot(...)` then `ax.step(...)` | type `line` | type `step` |
| `ax.step(...)` then `sns.boxplot(...)` | `stepDirection` **absent** | `vh` |

The first is worse than a mislabel. `line` builds `MultiLinePlot` rather than `StepPlot`, so `_attach_level_labels` never runs — a hypnogram announces the stage code `3` where it should say `REM`.

The second needs no unusual ordering at all, because `resolve_step_direction` is read at **render** time: `ax.step()` first classifies correctly, and a box plot drawn afterwards is enough to make the drawstyles "disagree". The step line itself was never ambiguous.

Both helpers now take the lines a layer owns — the patch passes `series`, `StepPlot.render()` passes `self._series()`. `is_step_axes` is renamed `is_step_layer`, because "axes" was the scope that invited this.

### Performance

Two reviews flagged the unconditional before-snapshot, one asking whether it shows in a profile. It does. On an axes carrying 2,000 lines:

```
list(ax.get_lines())            369 us/call
whole patched ax.plot call     ~600 us/call
```

A third of the cost, on the path that never reads it — `Axes.plot` hands its lines straight back. `instance` tells the two apart before the call: the Axes for the bound method, `None` for the module-level seaborn function. **2,000 `ax.plot()` calls: 1213 ms → 749 ms.**

The snapshot is also `id()`-keyed now, matching the boxen patch, which turns the diff from quadratic in the lines on the axes into linear. The `series` accumulator keeps its list scan deliberately: `drawn` is one line in the ordinary case, so building a local set costs the same `O(len(series))` the scan does, and a *persistent* set would mean two pieces of state that have to stay in step. Measured at 17 µs/call.

### Falsification

|---|---|
| sweep the axes again in `_series()` | 3 |
| pass only the registering call's lines | 3 (different) |
| drop the `plt.gca()` fallback | 6 |
| `is_step_layer(series)` → `(ax.get_lines())` | 12 |
| `resolve_step_direction(self._series())` → `(self.ax.get_lines())` | 6 |

---

## 2. `fix(deps)` — the declared seaborn floor could not import (#441)

`pyproject.toml` said `seaborn>=0.12` in both dependency groups, while `patch/boxplot.py` reached for `_CategoricalPlotter.plot_boxes` — which arrived in **0.13**. On a 0.12.2 environment built for the purpose:

```
AttributeError: type object '_CategoricalPlotter' has no attribute 'plot_boxes'
```

At import time, so nothing the user writes gets to run. The branch meant to cover older seaborn couldn't have helped either — it passes `sns_version`, a *string*, where the wrapper function goes; nothing was ever `< 0.12` and installable, so it has never run.

Takes the smaller of the two ways out recorded on the issue: declare `seaborn>=0.13`, which is what CI installs and the only version any test has run against, and drop the branch rather than repair a path nothing reaches. **It breaks no working installation, because 0.12 could not import to begin with.** Making 0.12 genuinely work stays open on #441 — it needs the boxen extraction to read that release's nested ladder too.

`tests/core/test_seaborn_floor.py` guards both halves, since either drifts alone: the attribute must exist on the installed seaborn, and every declared floor must be at least the release that introduced it. A third case pins that the version under test satisfies what's declared. Putting the floor back to `0.12` fails that second case.

---

## Verification

- `uv run pytest` — **1923 passed** (1908 at the first push, +15 from the step class)
- `uv lock --check` — clean
- `ruff check` — clean on all touched files

## One note on a test expectation

I first asserted the reference line's x as `0` and `1`; it arrives as `a` and `b`, because the companion chart makes the axis categorical and #353's label recovery names the positions the line was drawn at. That's the correct reading, so the tests assert it rather than sidestepping it — the step cases inherit the same behaviour and pin it too.


jooyoungseo:
Confirmed and fixed in `dd05bea`. This was a real bug in the fix, and worse than a gap — it left #440 intact for the *more common* spelling of the very pairing the issue was filed about.

```python
sns.boxplot(data=df, x="g", y="v")          # no ax=
sns.lineplot(data=summary, x="g", y="v")    # no ax=

line layer series: 11
    [('a', -0.542), ('a', -2.325)]     <- whisker endpoints
    [('a',  0.783), ('a',  1.960)]
```

Your causal chain is right end to end: `wrap_seaborn` wraps `lineplot` as a module-level function so `instance` is always None, no `ax=` leaves `target` None, `before` becomes `[]`, and every line already on the axes then looks newly drawn.

**The fix is your suggestion** — resolve the target the way seaborn will, `plt.gca()` included. Creating an axes where there is none isn't a side effect worth avoiding: seaborn is about to call `plt.gca()` and create the same one a moment later. Same idiom the boxen patch already uses.

After: `line layer series: 1`, values `[('a', 0.5), ('b', 0.4)]` — the summary line.

**On the test gap.** You're right that the suite only exercised the explicit-`ax=` form. The new cases are parametrized across all three companions rather than just the box plot, because what failed is the *snapshot*, and it has no idea which chart drew what — testing one companion would have left the same hole one level down. Dropping the `plt.gca()` fallback fails all six.

**On `dict(kwargs, lines=series)`** — agreed it's unreachable today (`ax.plot` and `sns.lineplot` take no `lines` parameter) and agreed it's worth knowing about. Leaving it rather than adding a guard for a collision that can't occur; if a future caller does pass `lines`, the clobber is the least of what would go wrong.

`uv run pytest` — **1905 passed**.

---

jooyoungseo:
Both reproduced and both fixed in `5af3a39`. The step finding is a real one and I'd missed it: I narrowed `_series()` and left two reads of the axes behind that decide things about a step chart.

## The step sweep

Measured, exactly the two orderings you separated:

```
sns.boxplot(...); ax.step(...)      types=['box', 'line']    <- should be step
sns.violinplot(...); ax.step(...)   types=['violin_kde', 'line']
ax.step(...); sns.boxplot(...)      types=['step', 'box']    stepDirection=<absent>
```

Your causal chain holds end to end for both, including the part that makes the first one worse than a mislabel: `line` builds `MultiLinePlot` rather than `StepPlot`, so `_attach_level_labels` never runs and the ordinal names go with it. A hypnogram announces `3` where it should say `REM`.

The second is the quieter one, and your reading of *why* is right — it is a **render-time** loss, so it needs no unusual ordering at all: `ax.step()` first classifies correctly, and a box plot drawn afterwards is enough for `resolve_step_direction` to find the drawstyles "disagreeing". The step line itself was never ambiguous.

**Fixed rather than filed.** Both helpers now take the lines a layer owns: the patch passes `series`, `StepPlot.render()` passes `self._series()`. `is_step_axes` is renamed `is_step_layer`, because "axes" was the scope that invited this — leaving the name would invite it back.

After, in all three orderings and all three companions: `types=[..., 'step']`, `stepDirection='vh'`, level labels `['N3', 'N2', 'REM']`.

### Falsification

15 new tests in `TestAStepChartSharingItsAxes`, parametrized over box/violin/boxen and both orderings, reverted one half at a time:

|---|---|
| `is_step_layer(series)` → `is_step_layer(ax.get_lines())` | **12** |
| `resolve_step_direction(self._series())` → `(self.ax.get_lines())` | **6** |

Disjoint apart from the three that genuinely need both — classification has to reach `StepPlot` before the field can exist to be resolved. The level-names test is in there deliberately, since the count-and-type assertions alone would pass on a chart that had quietly stopped naming its stages.

## The unconditional `before` snapshot

Both reviews raised this, one asking whether it shows up in a profile. It does, so I measured rather than guessed. On an axes carrying 2,000 lines:

```
list(ax.get_lines())                 369 us/call
whole patched ax.plot call          ~600 us/call
```

A third of the cost, on the path that never reads the result. `Axes.plot` hands its lines straight back, and `instance` tells the two apart *before* the call — the Axes for the bound method, `None` for the module-level seaborn function. 2,000 `ax.plot()` calls in a loop: **1213 ms → 749 ms**.

One detail worth naming: the branch is now chosen by `before is None` rather than by `isinstance(plot, list)`. If a future matplotlib returned something other than a list, the shape test would have fallen through to the sweep this PR removes; keyed on the snapshot, it draws nothing instead. Empty is the safe answer there, describing the axes is not.

## The `not in series` scan

Same measurement, same axes: **17 us/call** — 20× smaller than the snapshot, and the accumulation is the *rare* path (several deliberate `ax.plot()` calls forming one layer), where the snapshot was on every call. Leaving it. De-duplicating by a parallel `set` of `id()`s would mean a second piece of state on the axes that has to stay in step with the list, which is worth more than 17 us at a scale nobody reaches.

## On the other two notes

- **`kwargs.get("ax")` vs `_argument()`** — agreed it's an inconsistency. Leaving it: `ax` is keyword-only in every seaborn release this package now supports (≥0.13 as of this PR), and the positional-resolution helper would be answering a question that cannot be asked.
- **`target` vs the axes seaborn finally draws on** — right that it is an assumption. It is now documented at the point it is made, and it is also narrower than before: the snapshot is only taken on the seaborn path, so `ax.plot` no longer depends on it at all.

`uv run pytest` — **1923 passed** (1908 before, +15 from the new class). `ruff check` clean.

---

jooyoungseo:
Taken for `before`, in `e18ecfc`. You're right that it should match the boxen idiom, and it is more than polish there: `before` is looked up once per line already on the axes, so a list makes the diff quadratic in that count where the set makes it linear. `Line2D` defines no `__eq__`, so the two match exactly the same artists.

**Not taken for `series`, and the reason is that it would not actually help.** `drawn` is one line in the ordinary case, so `item not in series` costs `O(len(series))` per call — and building `{id(x) for x in series}` locally costs `O(len(series))` too. Identical work, one more line of code. It only becomes O(1) if the set is *persistent* on the axes, and then there are two pieces of state that have to stay in step with each other for the layer to describe the right lines. Measured at 2,000 lines it is **17 µs/call**, against 369 µs for the snapshot this PR just removed from the `ax.plot` path. Not worth a second source of truth.

That asymmetry is why the two reads differ: `before` is a local computed once and discarded, `series` is the layer's own record handed over by reference.

Thanks for tracing the mechanism against `boxenplot.py`/`violinplot.py` and confirming the internal-context split holds for both — that is the assumption the whole fix rests on, and it is the one I would most want a second reading of. Your check that no call site is left on the old `is_step_axes(ax)` signature matches what I found; the rename was deliberately breaking so a missed one could not compile away quietly.

`uv run pytest` — **1923 passed**. `ruff check` clean.

---

PR #444, opened 2026-08-16
fix(api): fall back to a static image instead of raising a KeyError

### LINKED ISSUE #443: An unsupported chart crashes maidr.render() with a KeyError, but falls back gracefully through plt.show()
## The defect

The same figure, drawn with a chart type maidr does not patch, behaves two completely different ways depending on which door the user goes through:

```python
import matplotlib.pyplot as plt, numpy as np, seaborn as sns
import maidr

sns.rugplot(x=np.random.randn(20), height=0.1)
```

```
plt.show()            ok, warns "This figure contains plot type(s) not yet supported
                      by maidr. Falling back to static image. Supported types: ..."
maidr.render()        KeyError: 'No MAIDR found for figure: Figure(640x480).'
maidr.show()          KeyError: 'No MAIDR found for figure: Figure(640x480).'
maidr.save_html(...)  KeyError: 'No MAIDR found for figure: Figure(640x480).'
```

The graceful path exists and works — `maidr/backend.py::_show_fallback()` warns, names every supported plot type, and renders a static image. It is only wired into the matplotlib backend, so it is reached by `plt.show()` and by nothing else. The three functions in `maidr/api.py` that a user is actually told to call go straight to `FigureManager.get_maidr()`, which raises.

## Not specific to rugplot

Measured through `maidr.render()`, on a fresh figure each time:

| drawn | result |
|---|---|
| `sns.rugplot(x=data)` | `KeyError` |
| `ax.quiver(...)` | `KeyError` |
| `ax.pcolorfast(...)` | `KeyError` |
| `plt.subplots()` and nothing else | `KeyError` |
| `ax.imshow(...)` | ok — patched as a heatmap |

Anything maidr did not register reaches the same line. An empty axes does too, which is worth noting separately: a user who calls `maidr.render()` a moment too early gets the same opaque message as one whose chart type is unsupported.

## Why `KeyError` is the wrong shape

```python
# maidr/core/figure_manager.py
@classmethod
def get_maidr(cls, fig: Figure) -> Maidr:
    if fig not in cls.figs.keys():
        raise KeyError(f"No MAIDR found for figure: {fig}.")
    return cls.figs[fig]
```

`KeyError` is the exception Python raises when you index a dict wrong. Surfacing it from the library's documented entry point tells a user their *own* call did something illegal with a mapping, when what actually happened is that their chart type is not supported yet. The message repeats that framing — "No MAIDR found for figure" is a statement about maidr's internal bookkeeping, not about anything the user can act on. It names no chart type, no supported list, and no next step, even though `_show_fallback()` already computes exactly that string a few modules away.

It is also a *crash* where the same condition through `plt.show()` is a *warning plus a working static image*. For an accessibility library that is the wrong way round: the user who explicitly asked for accessible output is the one who gets nothing.

## R binding comparison

`r-maidr` treats this as a first-class path rather than an error: `should_fallback()` renders the plot as a static image and warns, and it does so through `save_html()`, `show()`, and the print method alike. There is no spelling of an unsupported chart in r-maidr that produces a traceback.

## What to decide

1. **Reuse the existing fallback from the API path.** `render()`/`show()`/`save_html()` fall back to the static image with the same warning `_show_fallback()` already emits, matching `plt.show()` and matching r-maidr. Most consistent; means `render()` returns something renderable for every figure.
2. **Raise a real, actionable exception.** A dedicated error carrying the same "not yet supported, here is the supported list" message, so the failure is loud but explicable. Keeps `render()` honest about having produced no accessible output.

These are not exclusive — `show()`/`save_html()` are display verbs and suit (1); `render()` returns a value and could reasonably take either. What should not stay is the current state, where the answer depends on which function was called and the user is shown a `KeyError` either way.

Whichever is chosen, the empty-axes case deserves its own message: "this figure has no plots yet" is a different problem from "this chart type is not supported", and telling the two apart is cheap.

## Related

- #250 — `sns.rugplot` support, which is how I hit this; the crash is independent of whether rug plots are ever supported
- `maidr/backend.py::_show_fallback` — the graceful path that already exists
- #421, #434 — the same principle applied one level in: a layer that cannot be read honestly should say so rather than produce a confident wrong reading

jooyoungseo:
Closes #443.

## What was wrong

The same figure, drawn with a chart type maidr does not patch, behaved two completely different ways depending on which door the user went through:

```python
sns.rugplot(x=np.random.randn(20), height=0.1)
```

| call | result |
|---|---|
| `plt.show()` | ok — warned, drew a static image |
| `maidr.render()` | `KeyError: 'No MAIDR found for figure: Figure(640x480).'` |
| `maidr.show()` | `KeyError` |
| `maidr.save_html(...)` | `KeyError` |

The graceful path existed and worked — `backend.py::_show_fallback()` warns, names every supported plot type, and renders a static image. It was wired into the matplotlib backend and nothing else, so the three functions a user is actually told to call were the ones that crashed.

Not specific to rugplot. Measured through `maidr.render()`:

| drawn | result |
|---|---|
| `sns.rugplot(x=data)` | `KeyError` |
| `ax.quiver(...)` | `KeyError` |
| `ax.pcolorfast(...)` | `KeyError` |
| `plt.subplots()` and nothing else | `KeyError` |
| `ax.imshow(...)` | ok — patched as a heatmap |

**`KeyError` was the wrong shape as well as the wrong outcome.** It is what Python raises when you index a dict wrong, so surfacing it from a documented entry point told a user their own call did something illegal with a mapping. The message repeated that framing: "No MAIDR found for figure" describes maidr's internal bookkeeping. No chart type, no supported list, no next step — even though `_show_fallback` already computed exactly that sentence a few modules away.

For an accessibility library the asymmetry ran the wrong way round: the user who explicitly asked for accessible output was the one who got nothing.

## The fix

All four paths now fall back, which is what **r-maidr** does through every one of its entry points and what `plt.show()` here already did through one.

**`maidr/util/fallback.py`** holds the two pieces both paths need, so they cannot drift — the warning and the image. The previous message drifted precisely because it was hand-listed in one place: it named "kde" and "violin", neither of which is a `PlotType`, while omitting smooth and the violin variants.

The HTML carries the **reason** as well as the image. A warning is seen by whoever ran the code; the page is what reaches everyone afterwards, and a reader who meets a bare image has no way to know an accessible version was expected and did not happen. `alt` repeats the note, since the image is exactly the thing a screen reader cannot read.

**An empty figure gets its own message.** "Your chart type is unsupported" is misleading advice for someone who called `maidr.render()` a moment too early, and telling the two apart is one pass over the axes' artist lists:

```
empty axes           []
empty + title/labels []                 <- still empty; the labels describe a chart never drawn
rugplot              collections=1
quiver               collections=1
pcolorfast           images=1
```

**`UnsupportedPlotError` subclasses `KeyError` deliberately.** The backend already catches `KeyError` around `get_maidr` to decide whether to fall back, and so may anyone else's code — narrowing the type without keeping the base would break the working half while fixing the broken half.

## Measured after

```
rugplot  maidr.render()      ok   warn=This figure contains plot type(s) not yet supported…
rugplot  maidr.save_html()   ok   warn=This figure contains plot type(s) not yet supported…
rugplot  plt.show()          ok   warn=This figure contains plot type(s) not yet supported…
empty    maidr.render()      ok   warn=This figure has no plots on it yet…
quiver   maidr.render()      ok   warn=This figure contains plot type(s) not yet supported…
barplot  maidr.render()      ok   warn=-
```

## Tests

`tests/core/test_unsupported_fallback.py` — 21 tests.

"Does not raise" is satisfiable by returning nothing at all, so two further things are asserted: the fallback carries the **image** (`data:image/png;base64,`), and it carries the **reason** in the page. `save_html` is asserted to leave a file on disk — raising there would leave a build step that expected an artefact with nothing but a traceback.

The guards: a supported chart still renders accessibly and emits **no** warning and **no** embedded image; a drawn chart is not reported as empty; a title and axis labels do not make an empty axes non-empty.

### Falsification

|---|---|
| the API fallback (re-raise instead) | **15** |
| the empty-vs-unsupported distinction | **3** |
| nothing | **0 / 21 pass** |

Disjoint.

## Verification

- `uv run pytest` — **1944 passed** (1923 before, +21)
- `ruff check` — clean on every touched file (the 7 `F401` findings in `__init__.py` files are pre-existing; confirmed identical on a clean tree)

## One note on the test for `plt.show()`

I first drove the backend path through `plt.show()` and it did not warn. The cause is the test environment, not the code: this module selects Agg and other test modules switch the global backend around, so `plt.show()` was reaching whichever backend happened to be installed by then. The test calls `_show_fallback` directly instead, which is the function actually under test.

---


jooyoungseo:
All four fixed in `3338a0d`. The `KeyError.__str__` point is the sharp one, and I'd rather fix it than acknowledge it — you're right that it reproduces the "wrong shape" complaint this PR exists to remove.

## The quoting

```
str(e)     -> 'This figure has no plots on it yet, so there is nothing for maidr…
e.message  -> This figure has no plots on it yet, so there is nothing for maidr t…
traceback  -> UnsupportedPlotError: 'This figure has no plots on…
```

A stray leading quote in the traceback, which is precisely how a dict-lookup failure reads — so an uncaught one would undo the fix for exactly the user who most needs it, in the place they most often meet an exception.

**`.message` as the documented contract is the weaker answer**, because the failure mode is the case where nobody remembered to read it. `__str__` now returns `self.message`, so the traceback reads as a sentence:

```
maidr.exception.unsupported_plot_error.UnsupportedPlotError: This figure has no plots on…
```

The base class is kept for what catches it, not for how it reads — and that split is now written on the method rather than left implicit. Two tests pin it, and reverting `__str__` to `super().__str__()` fails those two and nothing else.

## The three nits

- **Not re-exported** — fixed. `maidr.exception` now exports both, with an `__all__`, which also clears the `F401` that file was contributing.
- **Local import in `get_maidr`** — checked, and it was dodging nothing: `unsupported_plot_error` imports only `maidr.core.enum`, which has no path back to `figure_manager`. Moved to module scope with the rest; `import maidr` is clean.
- **`warn_unsupported` rebuilding the error** — fixed. It now takes a `Figure` *or* an `UnsupportedPlotError`, because the two callers genuinely arrive differently: the backend holds only a figure, the API entry points are inside an `except` and already hold the error. The API paths pass the error, so the artist lists are walked once rather than twice.

## On the two you traced rather than flagged

Thanks for checking both independently — the stacklevel arithmetic and the `has_drawn_artists` scoping are the two places I'd have most expected a quiet mistake, and having them read back is worth more than the nits.

`uv run pytest` — **1946 passed** (1944 before, +2 from the `__str__` tests). `ruff check` clean on every touched file; the remaining 6 `F401` are pre-existing in `__init__.py` files this PR does not touch.

---

Issue #439, opened 2026-08-16
A strip or swarm plot announces its jitter as the value of the category axis


jooyoungseo:
## The defect

`sns.stripplot` scatters each point sideways by a random offset so overlapping observations stay separable. `sns.swarmplot` runs a packing algorithm to the same end. Neither offset is an observation — both are chosen by the renderer to stop marks landing on top of each other.

But the offset is what `PathCollection.get_offsets()` returns, and that is what `ScatterPlot` announced. Measured on a three-category strip plot whose x axis is labelled `g` and whose ticks read `a`, `b`, `c`:

```json
{"x": -0.03994198890364789, "y": 0.1257302210933933}
{"x":  0.06297755824455262, "y": -0.1321048632913019}
{"x": -0.07398610980531882, "y": 0.6404226504432821}
```

A reader is given a precise number for a quantity that does not exist, in the place where the chart says a name. For a strip plot the number is *literally random* — run the same script twice and it changes.

This is not a missing feature. The chart loads, announces a complete-looking reading, and is false about it.

## The navigation half

`ScatterTrace` groups points into columns by exact `x` equality:

```ts
if (!currentX || currentX.x !== point.x) { … }
```

With jittered x every point gets its own column, so a 90-point three-category strip plot builds **90 columns of one point** instead of 3 columns of 30. Column navigation steps through individual observations and never through categories — which is the axis a strip plot exists to compare across.

## Reproduce

```python
import matplotlib; matplotlib.use("Agg")
import matplotlib.pyplot as plt, numpy as np, pandas as pd, seaborn as sns, maidr
from maidr.core.figure_manager import FigureManager

df = pd.DataFrame({"g": np.repeat(list("abc"), 8),
                   "v": np.random.default_rng(0).normal(size=24)})
fig, ax = plt.subplots()
sns.stripplot(df, x="g", y="v", ax=ax)
for p in FigureManager.get_maidr(fig)._plots:
    print(p.schema["data"][:3])
```

## What can be fixed here

Report the point at its nearest tick when — and only when — the axis really is a category axis. The test for "really is" is the one #353 established: matplotlib leaves a `UnitData` on an axis it mapped strings onto and nothing else. A numeric axis must stay untouched, because there the drawn position *is* the value and snapping it would round every measurement onto a gridline while the payload still looked like a scatter plot.

Both axes need asking, since categories sit on y when the chart is turned on its side — asking about x alone was itself the #353 defect.

## What cannot be fixed here

[...]

PR #447, opened 2026-08-16
fix(seaborn): read displot as the distribution it draws

### LINKED ISSUE #446: sns.displot reads as a dodged bar chart, loses its bin edges, and announces "_container0" as a group name
## The defect

`sns.displot` is seaborn's figure-level interface for distributions, and its default `kind="hist"` draws a histogram. maidr reads it as a **dodged bar chart**.

The same data, through the two interfaces:

```python
sns.displot(df, x="v", bins=3)
  type: dodged_bar
  {'x': '-1.61082',  'z': '_container0', 'y': 9.0}
  {'x': '-0.182386', 'z': '_container0', 'y': 33.0}

sns.histplot(df, x="v", bins=3)
  type: hist
  {'y': 9.0,  'x': -1.6108159261240345, 'xMin': -2.3250307746388343, 'xMax': -0.8966010776092346, ...}
  {'y': 33.0, 'x': -0.1823862290944347, 'xMin': -0.8966010776092346, 'xMax': 0.5318286194203652,  ...}
```

Three things go wrong at once, and each is worse than the last.

**1. The type is wrong.** `dodged_bar` is the chart that compares groups side by side within a category. A distribution is not that, so the reader is oriented to a chart that isn't on screen.

**2. The bin edges are gone.** A histogram bar means "9 observations fell between −2.325 and −0.897". Without `xMin`/`xMax` the reader gets `x: '-1.61082'` — the bin **centre**, formatted as a *string* and presented where a bar chart puts its category name. That is a precise-looking number that is neither an observation nor a boundary, and nothing marks it as a midpoint. It is the same class as announcing a jitter (#439): a rendering artefact reported as a value.

**3. `_container0` is announced as the group name.** `z` carries the series name a reader hears to tell dodged groups apart. `_container0` is maidr's own internal identifier for a matplotlib `BarContainer`. It has leaked into the announcement.

## Also: `kind="kde"` disagrees with `kdeplot`

```
sns.kdeplot(df, x="v")            -> smooth
sns.displot(df, x="v", kind="kde") -> line
```

A KDE is a fitted curve, not a series of observations. `smooth` is the type that says so, and the axes-level function gets it right while the figure-level one does not.

`kind="ecdf"` is unaffected — both give `step`.

## The cause is already documented

From `maidr/patch/common.py::wrap_seaborn`:

> `pairplot`, `jointplot`, `relplot` and `lmplot` therefore ran the *unpatched* function, and the panel was seen only by the matplotlib-level patches. (`catplot` and `displot` reach neither binding: they drive seaborn's plotter classes directly, and are pinned as unreached in `tests/core/test_seaborn_patch_reach.py`.)

and, on what that costs:

> A `histplot` panel arrived as bars, because `Axes.bar` cannot know it is drawing a histogram and the seaborn-level patch that would have known never ran.

That is exactly what is still happening for `displot`. #344 fixed the four functions that take the defining-module binding; `displot` and `catplot` take neither, so the reading is still wrong for them. The limitation is recorded — the *consequence* is not, and it is a wrong reading rather than a missing one.

## Proposed fix

Patch the plotter method both interfaces drive, the way `maidr/patch/boxplot.py` already patches `_CategoricalPlotter.plot_boxes`:

```python
from seaborn.distributions import _DistributionPlotter
_DistributionPlotter.plot_univariate_histogram   # histplot AND displot(kind="hist")
_DistributionPlotter.plot_univariate_density     # kdeplot  AND displot(kind="kde")
```

`histplot` and `kdeplot` are already patched at the seaborn level and set the internal context before calling through, so the inner patch declines for them and no panel registers twice — the same recursion guard the boxen and box patches rely on.

`tests/core/test_seaborn_patch_reach.py` pins `displot` and `catplot` as unreached today, so that test is the one to update rather than to work around.

`catplot` has the same shape and should be measured before deciding whether it belongs in the same change — it drives `_CategoricalPlotter`, which is already partly patched.

## Reproduce

```python

df = pd.DataFrame({"v": np.random.default_rng(0).normal(size=60)})
sns.displot(df, x="v", bins=3)
for p in FigureManager.get_maidr(plt.gcf())._plots:
    print(p.type.value, p.schema["data"][0][:2])
```

## Related

- #344 — the defining-module patching that fixed the same defect for `pairplot`, `jointplot`, `relplot`, `lmplot`
- #402 — bins the payload claims that the chart never drew, the other half of getting a histogram's intervals right
- #439 — a rendering artefact announced as a value, the same class as the bin centre here

jooyoungseo:
Closes #446.

## What was wrong

`sns.displot` is seaborn's figure-level interface for distributions, and its default `kind="hist"` draws a histogram. maidr read it as a **dodged bar chart**.

The same data, through the two interfaces:

```
sns.displot(df, x="v", bins=3)
  dodged_bar   {'x': '-1.61082', 'z': '_container0', 'y': 9.0}

sns.histplot(df, x="v", bins=3)
  hist         {'y': 9.0, 'x': -1.6108159, 'xMin': -2.3250307, 'xMax': -0.8966010, ...}
```

Three losses at once, each worse than the last:

1. **The type is wrong.** `dodged_bar` is the chart that compares groups side by side. A distribution is not that, so the reader is oriented to a chart that isn't on screen.
2. **The bin edges are gone.** A histogram bar means "9 observations fell between −2.325 and −0.897". Without `xMin`/`xMax` the reader gets `x: '-1.61082'` — the bin **centre**, as a *string*, in the slot where a bar chart puts its category name. A precise-looking number that is neither an observation nor a boundary, with nothing marking it as a midpoint. Same class as announcing a jitter (#439).
3. **`_container0` was announced as the group name.** `z` is what a reader hears to tell dodged series apart. `_container0` is maidr's own internal identifier for a matplotlib `BarContainer`.

`kind="kde"` had the smaller version of the same problem — `line` where the axes-level `kdeplot` gives `smooth`. A fitted curve is not a series of observations, and only one of those types says so.

## Why it survived

The gap is documented in `wrap_seaborn`, and there is a test pinning it:

> `catplot` and `displot` reach neither binding: they drive seaborn's plotter classes directly, and are pinned as unreached in `tests/core/test_seaborn_patch_reach.py`.

#344 fixed the four grids that take the defining-module binding. `displot` takes neither, so `Axes.bar` was the only patch that ever saw its panels — and the docstring already said what that costs: *"a `histplot` panel arrived as bars, because `Axes.bar` cannot know it is drawing a histogram"*. The limitation was recorded; the consequence for `displot` was left as a wrong reading rather than a missing one.

## The fix

Patch the plotter methods both interfaces drive, which is the idiom `maidr/patch/boxplot.py` already uses for `_CategoricalPlotter.plot_boxes`:

```python
_DistributionPlotter.plot_univariate_histogram   # histplot AND displot(kind="hist")
_DistributionPlotter.plot_univariate_density     # kdeplot  AND displot(kind="kde")
```

`histplot` and `kdeplot` set the internal context before calling through, so the inner patch declines for them and no panel registers twice — the recursion guard the box and boxen patches already rely on.

**One call covers the whole grid.** `plot_univariate_histogram` is reached *once* for a faceted `displot` and draws every panel — and `plotter.ax` is `None` in exactly that case, so reading `ax` alone would have registered **nothing at all** there, not merely the first panel. The panels come off the `FacetGrid` instead.

**Each panel is asked separately whether it drew bars**, so the two existing guards survive: a bivariate histogram keeps declining for the reason #388 gives (a `QuadMesh` of joint counts has neither one bin per bar nor a count), and an axes that already holds someone else's bars is not claimed.

## Measured after

```
displot kind=hist        hist    keys=[x, xmax, xmin, y, ymax, ymin]
displot col=g            hist, hist
displot kind=kde         smooth
displot kind=kde col=g   smooth, smooth
displot kind=ecdf        step          (unchanged)
histplot                 hist          (registers once, not twice)
histplot kde=True        hist, smooth  (unchanged)
histplot bivariate       declined      (unchanged)
jointplot                point, hist, hist (unchanged)
```

## Tests

`tests/core/test_displot_distribution.py` — 14 tests.

The sharpest one asserts the two interfaces produce **identical payloads** for the same data, which is what makes the defect legible: nothing about `displot`'s reading looked wrong on its own, only next to `histplot`'s.

Guards: `histplot`/`kdeplot` still register exactly once; the `kde=True` overlay is unchanged; a bivariate histogram still declines; a histogram does not claim a `barplot`'s bars; `jointplot` (which takes the defining-module path) does not shift.

`tests/core/test_seaborn_patch_reach.py` is **rewritten rather than worked around** — it asked to be:

> Pinned here so that the day either is fixed, this test fails and has to be rewritten, the way `test_hist2d.py` pinned `hexbin` before #368.

What it pinned as wrong is now pinned as right, and `catplot` — still unreached, still `[DODGED, LINE]` — keeps its own test so the remaining boundary stays visible.

### Falsification

|---|---|
| the `plot_univariate_histogram` patch | **7** |
| the `plot_univariate_density` patch | **3** |
| nothing | **0** |

Disjoint apart from the one reach test that covers both kinds.

## Verification

- `uv run pytest` — **1975 passed** (1959 before, +16)
- `ruff check` — clean on every touched file

## Still open

`catplot` has the same shape — it drives `_CategoricalPlotter`, which is already partly patched — and is worth measuring on its own rather than folding in here.

---


jooyoungseo:
Both taken in `5ecfebd`. Finding 1 is the one I should have caught myself — I wrote `test_seaborn_floor.py` earlier in this branch's history for exactly this failure and then added two new reaches without extending it.

## The floor guard

You're right that this is the same risk class as #441, and right that it is worse than a normal regression: an attribute that moves breaks `import maidr` for **every** user, not only users of that chart type, and with no diagnostic pointing at the cause.

Rather than adding two more constants, the guard now runs over a table of every private seaborn attribute a patch reaches at import time:

```python
PATCHED_INTERNALS = [
    ("seaborn.categorical",   "_CategoricalPlotter", "plot_boxes",                Version("0.13")),
    ("seaborn.distributions", "_DistributionPlotter", "plot_univariate_histogram", Version("0.11")),
    ("seaborn.distributions", "_DistributionPlotter", "plot_univariate_density",   Version("0.11")),
]
```

Parametrized, so a rename fails the case that **names** the attribute rather than whichever patch happens to import first, and the floor check runs against every entry with `REQUIRED_SINCE` as the max.

Verified it bites: renaming `plot_univariate_histogram` in the table fails exactly `[_DistributionPlotter.plot_univariate_histogram_renamed]` and nothing else.

Both new attributes date from the 0.11 distributions rewrite, comfortably under the declared 0.13 floor — so nothing changes about what the package supports, only about what happens if seaborn moves them.

The list is the part that can still rot, so that is written down in the constant: *"A patch added without an entry here is exactly the case this file cannot catch."*

## The local import

Correct — `maidr/patch/__init__.py` imports `histogram` before `kdeplot`, so there was no circularity to guard. Hoisted to module scope; `import maidr` is clean.

On moving `_plotter_axes` to `common.py`: I've left it in `histogram.py`. `common.py` holds the patch machinery every module uses (`_draw_quietly`, `common`, `wrap_seaborn`), and this is a seaborn-distributions detail that two of the twelve patch modules share. Moving it would put a `_DistributionPlotter`-shaped helper next to functions that know nothing about seaborn's plotter classes. Happy to move it if a third caller appears, which is the bar `CLAUDE.md` sets for abstraction.

`uv run pytest` — **1977 passed** (1975 before, +2 from the new parametrized cases). `ruff check` clean.

---

Issue #449, opened 2026-08-16
A horizontal seaborn violin raises TypeError, and a positional `data` silently drops the box summary


jooyoungseo:
Two failures in `maidr/patch/violinplot.py`, both from the same cause: the patch re-parses the caller's arguments to work out what was drawn, instead of asking seaborn's plotter what it resolved.

## 1. A horizontal violin takes the whole figure down

seaborn documents the orientation as *inferred*: assign the categorical variable to `y` and you get a horizontal violin, no `orient=` needed. That call raises:

```python
import matplotlib; matplotlib.use("Agg")
import matplotlib.pyplot as plt, numpy as np, pandas as pd, seaborn as sns, maidr

rng = np.random.default_rng(0)
df = pd.DataFrame({"v": rng.normal(size=60), "g": list("ab") * 30})

_, ax = plt.subplots()
sns.violinplot(data=df, y="g", x="v", ax=ax)   # horizontal, drawn correctly
maidr.render(ax.get_figure())
```

```
TypeError: ufunc 'isnan' not supported for the input types, and the inputs
could not be safely coerced to any supported types according to the casting rule 'safe'
  maidr/core/plot/violinplot.py:183  in ViolinBoxStatsCalculator.compute
```

No HTML at all — for the figure, not just the layer. It is not caught by the #443 fallback either, which only handles `UnsupportedPlotError`.

The mechanism is in `ViolinDataExtractor.extract`:

```python
orient = kwargs.get("orient", None)
is_horizontal = orient in ("h", "horizontal", "y")
```

`orient` is `None` when the caller let seaborn infer it, so `is_horizontal` is False, so the extractor takes `cat_col, val_col = x, y` — which for this call is `("v", "g")`. It then groups by the *numeric* column and reads the *category names* as the values, and `np.isnan` is handed an array of strings.

Adding `orient="h"` avoids it, so this is invisible to anyone who spells the orientation out.

## 2. A positional `data` drops the VIOLIN_BOX layer

`sns.violinplot(df, x=..., y=...)` is seaborn's first-positional signature. Measured on the same frame:

| call | layers |
|---|---|
| `sns.violinplot(data=df, x="g", y="v", ax=ax)` | `violin_box`, `violin_kde` |
| `sns.violinplot(df, x="g", y="v", ax=ax)` | `violin_kde` |

[...]

PR #450, opened 2026-08-16
fix(seaborn): read every catplot kind as the chart it draws

### LINKED ISSUE #448: sns.catplot reads a violin as a two-point line, and disagrees with its axes-level equivalent for six of eight kinds
## The defect

`sns.catplot` is seaborn's figure-level interface for categorical charts. It drives `_CategoricalPlotter` directly rather than importing the axes-level functions, so none of maidr's seaborn-level patches ever run and the panel is read only by the matplotlib-level ones.

Measured across every `kind`, against the axes-level function that draws the same chart:

| `kind` | `catplot` reads | axes-level reads | |
|---|---|---|---|
| `strip` | `point×3` | `point×3` | ✓ |
| `swarm` | `point×3` | `point×3` | ✓ |
| `box` | `box(3)` | `box(3)` | ✓ |
| **`violin`** | **`line(2)`** | `violin_kde(30)` | ✗ |
| **`boxen`** | **`line(2)`, `point×3`** | `boxen(3)` | ✗ |
| **`point`** | **`line(3)`** | `error_bar(3)` | ✗ |
| **`bar`** | **`dodged_bar(3)`, `line(2)`** | `bar(3)` | ✗ |
| **`count`** | **`dodged_bar(3)`** | `bar(3)` | ✗ |

Six of eight disagree, and they fail in three distinct ways.

### 1. A distribution announced as a two-point line

`kind="violin"` emits a **single `line` layer with two samples**. A violin plot is a density estimate over 30 observations per group; the axes-level `violinplot` reads it as `violin_kde` with 30. What the reader gets instead is a line chart of two points — and those two points are the whisker geometry, not data.

`kind="boxen"` is the same shape: a `line(2)` where the letter-value ladder should be, plus three `point` layers of the fliers. The ladder — the entire content of a boxen plot — is absent.

This is the #440 defect class: a `line` layer describing geometry that another chart drew. Fixed for the axes-level path, still live here because the patch that would suppress it never runs.

### 2. Uncertainty silently dropped

`kind="point"` emits `line(3)` where `pointplot` emits `error_bar(3)`. The estimates survive; the confidence intervals #246 added do not. A reader is given three means with no indication that the chart draws intervals around them, which is the thing a point plot exists to show.

### 3. Wrong type, plus a phantom layer

`kind="bar"` and `kind="count"` emit `dodged_bar` where the axes-level functions emit `bar` — a chart that compares groups side by side, announced for one that does not. `kind="bar"` additionally emits a `line(2)` of the error-bar caps.

## The cause is documented

From `maidr/patch/common.py::wrap_seaborn`:

> `catplot` and `displot` reach neither binding: they drive seaborn's plotter classes directly, and are pinned as unreached in `tests/core/test_seaborn_patch_reach.py`.

#344 fixed the four grids that take the defining-module binding. #446 reached `displot` by patching the plotter method it drives. `catplot` is the last one, and it is the one with the most kinds behind it.

## Proposed fix

The same shape as #446: patch the `_CategoricalPlotter` methods `catplot` drives, so the axes-level function and the figure-level one share a reading.

`_CategoricalPlotter.plot_boxes` is **already patched** for `boxplot` — which is exactly why `kind="box"` is the one categorical kind that already agrees. That is both the precedent and the evidence that the approach works.

The remaining methods to survey:

```python
from seaborn.categorical import _CategoricalPlotter
[m for m in dir(_CategoricalPlotter) if m.startswith("plot_")]
```

The recursion guard is already in place: each axes-level patch sets the internal context before calling through, so an inner plotter-level patch declines and nothing registers twice — the mechanism #446 relies on.

Worth doing per kind rather than in one change: `violin` and `boxen` are the severe ones (a distribution read as a two-point line), `point` is the uncertainty loss, and `bar`/`count` are a type correction. They can land separately and be falsified separately.

## Reproduce

```python

df = pd.DataFrame({"g": np.repeat(list("abc"), 20),
                   "v": np.random.default_rng(0).normal(size=60)})
sns.catplot(df, x="g", y="v", kind="violin")
print([p.type.value for p in FigureManager.get_maidr(plt.gcf())._plots])
# ['line']
```

## Related

- #446 — the same gap for `displot`, and the plotter-level idiom that closed it
- #344 — the defining-module patching that fixed the four grids taking that path
- #440 — a `line` layer describing another chart's geometry, the failure `kind="violin"` and `kind="boxen"` show
- #246 — the pointplot confidence intervals `kind="point"` drops
- `tests/core/test_seaborn_patch_reach.py` — where `catplot` is pinned as unreached

jooyoungseo:
Closes #448 and #449. Four commits, one per plotter method, each falsifiable on its own.

`sns.catplot` drives `_CategoricalPlotter` directly and imports nothing, so it reaches neither binding `wrap_seaborn` patches — its panels were read by the matplotlib-level patches alone. #448 measured six of eight kinds disagreeing with the axes-level function that draws the same chart:

| `kind` | `catplot` read | axes-level read | |
|---|---|---|---|
| `strip` | `point×2` | `point×2` | ✓ |
| `swarm` | `point×2` | `point×2` | ✓ |
| `box` | `box` | `box` | ✓ |
| **`violin`** | **`line`** | `violin_box`, `violin_kde` | ✗ |
| **`boxen`** | **`line`, `point×2`** | `boxen` | ✗ |
| **`point`** | **`line`** | `error_bar` | ✗ |
| **`bar`** | **`dodged_bar`, `line`** | `bar` | ✗ |
| **`count`** | **`dodged_bar`** | `bar` | ✗ |

All six now agree. `box` already did — because `_CategoricalPlotter.plot_boxes` was already patched, for `seaborn.boxplot`, which is both the precedent and the evidence the approach works. Each of the other five went the same way, to the plotter method the grid and the axes-level function share.

## The three ways they failed

**A distribution announced as a line chart** — `violin` and `boxen`. The violin's line is its inner box; the boxen's is its median segments, each announced as a two-sample series, with the point layers holding the outliers alone at numeric slots rather than at the category names. For the boxen that is precisely the reading #253 was filed to replace: the fix landed for `sns.boxenplot`, and `catplot` never got it.

**Uncertainty dropped** — `point`. The estimates read correctly as a line; what a `line` layer has no room for is `yMin`/`yMax`, the confidence intervals #246 added. A reading with no intervals in it sounds exactly like a correct reading of a chart that draws none, so the loss is inaudible.

**Wrong type plus a phantom layer** — `bar` and `count`. `dodged_bar` names a chart that compares groups side by side; a chart with no hue is not one, so the reader is oriented to a chart that is not there. Seaborn does not forward `hue` or `dodge` to `Axes.bar`, so the matplotlib patch infers grouping from bar widths and positions, and on seaborn's output it inferred wrong. The extra `line` is the error-bar geometry travelling as a series of its own — the #440 shape.

---

## Two defects found on the way, both worse than the one I went looking for

Both in the violin patch, both from the same cause as #448 itself: the patch worked out what had been drawn by re-reading the caller's keywords instead of asking seaborn what it resolved. Filed as **#449**.

**An inferred-horizontal violin raised, and the figure emitted no HTML.** Assigning the categorical variable to `y` is how seaborn documents a horizontal violin; no `orient=` is needed:

```python
sns.violinplot(data=df, y="g", x="v", ax=ax)
maidr.render(ax.get_figure())
```
```
TypeError: ufunc 'isnan' not supported for the input types, and the inputs
could not be safely coerced to any supported types according to the casting rule 'safe'
```

`ViolinDataExtractor` read `orient = kwargs.get("orient", None)`, and None is not in `("h", "horizontal", "y")` — so it took `cat_col, val_col = x, y`, grouped by the numeric column and handed `np.isnan` an array of category names. Writing `orient="h"` avoided it, which is why this survived. Not caught by #443's fallback either, which only handles `UnsupportedPlotError`.

**A positional frame silently lost the box summary.**

| call | layers |
|---|---|
| `sns.violinplot(data=df, x="g", y="v", ax=ax)` | `violin_box`, `violin_kde` |
| `sns.violinplot(df, x="g", y="v", ax=ax)` | `violin_kde` |

`extract` looked for `data` in the keywords only, so every `isinstance(df, pd.DataFrame)` branch fell through and it returned `[], []`. The chart loaded, the density curve read correctly, and the five summary statistics a violin exists to carry were not there.

Both are answered by reading `plotter.orient` (`"x"`/`"y"`, inference included), `plotter.plot_data` (resolved roles, however the data was passed) and `plotter.var_levels` (the order actually drawn, `order=` and `Categorical` categories already applied).

---

## A regression the fourth commit found and fixed

A `row`/`col` grid allocates an axes for **every** combination of the two, including ones the data does not hold. Seaborn draws nothing into those, and registering them promises a layer whose extraction has nothing to read — which raises and takes the whole figure's HTML down, not just the empty panel's:

```
catplot(kind="bar", col=..., row=...)  with one combination missing
  ExtractionError: Error extracting data for bar plot type from <class 'list'>
```

The violin path avoided this by accident: an empty panel yields no groups and no collections, so both its layers decline on their own. The boxen and point commits did not, and would have shipped this. The panel list now comes from seaborn's own `iter_data(allow_empty=False)` for all four patches, hoisted out of `violinplot.py` into `maidr/patch/common.py` as `plotter_panels`.

On the same data `kind="bar"` also stops raising `StopIteration` during the draw itself, which it did before this branch.

---

## Shared

Each `seaborn.*` function keeps its wrapper for the two jobs that still belong there: `_draw_quietly` across the whole call, so a deprecation warning from seaborn's own argument shims does not reach a screen-reader user; and `wrap_seaborn`, so both bindings of each name stay wrapped. None of them sets the internal context any more — that is what would silence the plotter patch — and the colour probe they used to suppress has been handled in `seaborn_probe.py` since #373.

`_plotter_axes` moves from `maidr/patch/histogram.py` to `common.py` as `plotter_axes`, now that five patches need it and it is no longer specific to `_DistributionPlotter`.

`tests/core/test_seaborn_floor.py` gains `plot_violins`, `plot_boxens`, `plot_points` and `plot_bars` (seaborn 0.13) — the guard added in #441 for exactly this: a new import-time reach into a seaborn private breaks `import maidr` for *everyone* if the attribute moves.

`tests/core/test_seaborn_patch_reach.py` replaces its "kinds not reached" pin with the whole of #448 as a table — all eight kinds asserted equal to their axes-level equivalents — plus a check that the table keeps up with `catplot`'s own `kind == "..."` dispatch, so a ninth kind fails a test rather than going silently unread.

## Tests

Four new files (21 + 11 + 12 + 17), two tests appended to `test_boxenplot.py`, and the reach/floor updates. Mostly agreement tests: two interfaces, or two spellings of one call, describing the same chart the same way. That shape is what makes these defects legible, since none of the wrong readings looked wrong on its own.

Suite: **2057 passed**. `ruff check` clean on every file touched (six pre-existing `F401` re-export findings in `maidr/core/plot/__init__.py` are unchanged).

## Falsification

Each mechanism reverted in turn, against that commit's own test files.

| commit | reverted | failures |
|---|---|---|
| violin | whole change | 12 |
| | plotter patch registers nothing | 18 |
| | orientation from `kwargs["orient"]` again | 5 |
| | every panel handed the whole figure's data | 1 |
| | no per-panel line snapshot | 1 |
| | no placeholder recovery for an invented category axis | 1 |
| | `var_levels` ignored, categories in value order | 1 |
| boxen | whole change | 6 |
| | plotter patch registers nothing | 33 |
| | no per-panel collection snapshot | 1 |
| | only the first panel registered | 2 |
| point | whole change | 7 |
| | plotter patch registers nothing | 27 |
| | no per-panel line snapshot | 2 |
| | only the first panel registered | 2 |
| bar | whole change | 13 |
| | plotter patch registers nothing | 14+ |
| | only the first panel registered | 3 |
| | bar type hard-coded to `bar` | 1 |
| shared | empty panels registered too | 4 |

Two of these rows are the reason the falsification pass is worth running rather than assumed. The boxen **snapshot** revert passed all 86 tests at first: every overlay case in `test_boxenplot.py` draws the strip plot *after* the boxen, so its clouds are not on the axes when the ladders are read. Drawn *before* it, each ladder takes seven of them at each end — fourteen outliers per category on a chart written with `showfliers=False`. That ordering now has two tests. And the **empty-panel** revert is what surfaced the regression above; it passed everything until an unbalanced grid was in the suite.


jooyoungseo:
Thanks — the unnamed-`hue` finding is a real defect, and it reproduces. Pushed as `4f1d9ed`.

## The bug

Reproduced first, on a fully populated cross rather than the example in the report (there, `a` only occurs with `10` and `b` only with `20`, so only two of four pairs are non-empty and the collision does not show):

```python
sns.violinplot(x=["a"]*12 + ["b"]*12, y=range(1, 25), hue=(["p"]*6 + ["q"]*6)*2)
  groups: ['a', 'a', 'b', 'b']
# the same data as named columns
  groups: ['a_p', 'a_q', 'b_p', 'b_q']
```

Four violins, two pairs sharing a name, nothing telling them apart. That is the same shape of defect this branch is about — a reading that sounds complete and is not — so it is worth closing, exactly as you say. `None` on either side now answers False, with `test_two_unnamed_variables_are_not_the_same_variable` failing alone when that guard is removed.

## The rest, in order of how much they mattered

**Dead code.** Confirmed: `_lines_before` has no callers left, and it was the only user of that module's `pyplot` import. Both deleted.

**Stale `ViolinDataExtractor` docstrings.** Confirmed and fixed on all three — class, method, and the module docstring above them, which said the same thing. The doctest example stays, since those keyword forms are spellings `Axes.violinplot` shares with seaborn's signature and a caller may still write one; the class now says it is reached from the matplotlib path alone and why the seaborn path left.

**`_levels` recomputed per category.** Confirmed — `pd.unique` was running once per outer iteration. Hoisted.

**Line lengths.** Confirmed, and worth saying that ruff does not select `E501` in this project's config, so `ruff check` passed on all of them. Six wrapped (one more than listed — `test_seaborn_patch_reach.py`'s new `parametrize` line). One 89-column line remains, in `test_a_function_defined_at_the_package_root_is_not_a_gap`, which predates this branch.

**`_argument`'s docstring.** Fixed rather than left as optional — it is a small thing, but the reason the reuse works is that `wrapped` is a bound method either way, and that is worth a sentence where a reader will hit it.

## The two test gaps

Both are the kind that would have let something through, so both are now pinned and both fail without their mechanism.

**Facets and a hue together.** You were right that it works and right that nothing pinned it. Writing the test found something worth noting: my first attempt derived the frame from the file's existing fixture, whose `g` and a two-level hue run in lockstep — so half the cross simply did not exist and the assertion I had written was wrong, not the code. The test now builds its own frame where all four combinations are populated in each panel, and checks each group's median against the rows it should have come from. Removing the hue split inside a panel fails it.

**Runtime attributes in the floor test.** Agreed, and I think the framing matters: it is a *different* risk rather than a smaller one. The existing table is about `import maidr` raising for every user; losing `iter_data` or `_get_axes` breaks nothing at import and nothing on a plain call, and surfaces on the first faceted categorical grid someone draws — narrow enough that it could ship. So they get their own table, `RUNTIME_INTERNALS`, with that written down next to it. Three of the six (`plot_data`, `orient`, `variables`) are set in `__init__` rather than declared, so a class-level `hasattr` answers False for a perfectly healthy seaborn; the check falls back to constructing a real plotter for those, which is also the only way to catch one that stops being set at all. Verified by renaming one entry: it fails, and names the attribute.

## Not changed

The `O(categories · hue_levels · n)` masking. It is real, and I left it: the alternative is a groupby whose ordering I would then have to reconcile with `var_levels` by hand, and the whole point of reading from the plotter is that seaborn has already decided that order. At the group counts a violin plot is legible at, the cross product is the cheaper thing to be sure about.

Suite: **2066 passed**, CI green across 3.9–3.12.

---

jooyoungseo:
Both nits addressed, one of them by fixing it. Pushed as `1f3fe05`.

## `_register_kde_layer` (nit 2) — live, not latent

You called this pre-existing and out of scope. It reproduces, and it is worse than "would still pick up all of them" — it **replaces** the violin's own density:

```python
ax.fill_between([0, 1], [0, 0], [1, 1])
sns.violinplot(data=df, x="g", y="v", ax=ax)
  violin_box(2), violin_kde(4)      # 30 samples per curve become 4
```

Four samples where the density has thirty. The band's four vertices are announced under the violin's name, and the curve the violin actually drew is not in the reading at all. A `PolyCollection` is not a violin-specific artist — `fill_between`, `stackplot` and a filled `kdeplot` all produce one — so the layer took whichever came first on the axes.

Your two-violins-on-one-axes case does not show it, incidentally: both violins hold identical data there, so the second layer describing the first's polygons is invisible. `fill_between` is what makes it legible.

Fixed rather than commented, since it is the same class the rest of the branch is about. The layer is now built from the collections the call added — the seaborn path from a per-panel snapshot alongside the one already taken for the inner-box lines, the matplotlib path from `bodies`, which is matplotlib's own name for the polygons `Axes.violinplot` returns. Filtered out of `ax.collections` rather than iterated directly, so the order stays the axes' own: `x_levels` is paired with the curves positionally, and a violin named for its neighbour's category would be the same defect one column along.

`test_the_density_is_the_one_this_call_drew` fails alone when the sweep is restored. Suite: **2067 passed**.

## Docstring length (nit 1) — not changed, and here is the reasoning

I read this one carefully, because "trim to what a reader needs to use the function" is usually right and I did not want to wave it away.

What decided it is the surrounding code rather than my preference. This is the existing convention in `maidr/patch/`, not a departure from it — `boxenplot.py::boxen`, `pointplot.py::point`, `histogram.py::_drew_bars`, `common.py::wrap_seaborn` and `seaborn_probe.py` all carry the same shape, including the measured before/after output and the issue number, and several of them predate this branch. A file where four functions explain their history and the fifth does not is harder to read than one where all five do, so matching them is the lower-cost choice.

On issue numbers going stale: the risk is real for a number *standing in place of* an explanation. These sit after one — the measured wrong reading is always in the text, so `(#448)` is a pointer to more, not the only record. Someone reading `maidr/patch/violinplot.py` outside this tracker still learns that an inferred-horizontal violin used to raise and why, without resolving anything.

The specific thing these docstrings are for is the class of bug this branch is full of: every one of them was a patch doing something that looked arbitrary until you knew what it was avoiding — the internal context deliberately *not* set, the placeholder recovery for an invented category axis, the per-panel snapshot. Each is a line someone would reasonably delete as redundant. The comment above it is what stops that, and it has to be at the code rather than in a commit message, because the person about to delete it is reading the code.

Happy to trim if a maintainer prefers the shorter form — it is a house-style call, and I would rather be consistent with whatever the answer is than argue for mine.

---

PR #452, opened 2026-08-17
feat(shiny): rebuild render_maidr on Shiny's Renderer contract

jooyoungseo:
## Description

Brings py-maidr's Shiny integration onto the extension point Shiny for Python documents, and fixes two defects found while doing it — one of which made an offline Shiny chart completely inaccessible with no error to say so.

`maidr/widget/shiny.py` was 29 lines that subclassed `shiny.render.ui` and overrode `render()` with a copy of the base implementation. Shiny's documented extension point is `Renderer`. Subclassing a concrete renderer inherits its value type, so the class was typed `Renderer[TagChild]` and a type checker rejects this repository's own example, whose render function returns an `Axes`. There was no UI function to pair it with, no options, no declared dependency on `shiny`, and no test had ever imported the file.

### Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Related Issues

None; found by auditing the integration against the py-shiny docs.

## Changes Made

**`fix(env)` — `Environment.is_shiny()` detects a session, not a package**

It asked whether `import shiny` succeeds, and the comparison it then made (`shiny.__name__ == "shiny"`) is true for any successful import. Six call sites gate on it to decide whether a render emits an `HTMLDependency` and whether it is wrapped in an iframe, so **every process that merely had Shiny installed took the Shiny path**. A plain script rendering a bar chart returned a 28 KB `srcdoc` iframe carrying no dependency where it should have returned a 15 KB div carrying one; a Streamlit app got a maidr iframe nested inside Streamlit's own.

`is_flask()` next door already asks the runtime question (`has_app_context()`). `get_current_session()` is Shiny's equivalent, and is non-`None` exactly where `render_maidr` runs.

**`fix(offline)` — an iframed `use_cdn=False` render shipped no maidr.js at all**

The chart arrived as a picture: no sonification, no braille, no keyboard navigation, and nothing in the console. Two mechanisms both failed outside a notebook — `Tag.get_html_string()` (which builds the `srcdoc`) silently drops `HTMLDependency` children, and the other branch reads the bundle from `window.parent.__maidrJsSource`, which only `init_notebook()` populates and which returns early outside a notebook. **Reachable from both Shiny and Flask.** `inline_bundle_tags()` now carries the source itself, with the `data-maidr-math` marker so maidr.js does not report the maths rules missing while they are on the page. An unreadable bundle warns and falls back to the CDN rather than raising; the sources are read once per process rather than once per ~2 MB render.

`use_cdn="auto"` and `True` are untouched.

**`feat(shiny)` — the renderer itself**

- `class render_maidr(Renderer[Any])` with `output_maidr()` as its container, returned from `auto_output_ui()` so Shiny Express places it.
- Options: `width`, `height`, `use_cdn`. `use_cdn` matters here — the only previous lever was `maidr.set_use_cdn()`, process-wide state shared by every concurrent session, which `docs/index.qmd` already warns against.
- `render()` rather than `transform()` is overridden, matching `shiny.render.plot`, which does the same because it too needs to bracket the value function.
- Figures a render opens are closed. A render function runs once per reactive flush, so `plt.subplots()` inside one opened a figure every time: 25 flushes left 25 open and matplotlib warning.
- An unrenderable return value raises `TypeError` naming the function and the type, instead of `AttributeError: 'NoneType' object has no attribute 'get_figure'` three frames down.
- `shiny` is now an optional extra (`pip install "maidr[shiny]"`), so a stock install fails with an actionable message and CI exercises this file for the first time.

### Backwards compatibility

`from maidr.widget.shiny import render_maidr` is unchanged, bare and parenthesised decoration both work, and `ui.output_ui("id")` still pairs with it — `output_maidr` is that call with a style attribute. Existing apps need no edit.

The one visible change: `@render_maidr` now closes pyplot figures the render function opened. An app that builds a figure at module scope and returns it each flush is unaffected — only figures that were not open beforehand are closed.

### A deliberate non-change

The chart stays in an iframe under Shiny. Serving it inline would let Shiny's own dependency machinery carry maidr.js, but it also puts N maidr instances with document-level keybindings into one Bootstrap page, and I have no way to check the keyboard and focus behaviour of that in a browser here. Changing it on an untested assumption is the wrong trade in an accessibility library. Filed as a follow-up instead.

## Checklist

## Additional Notes

**Verification.** `uv run pytest` — 1982 passed (from 1980; 22 new tests). `ruff check` clean on every changed file. `uv lock --check` clean. The renderer was driven end to end under a session context against **shiny 1.0.0 and 1.6.1**, so the declared `shiny>=1.0` floor is tested rather than assumed; the resolver picks 1.5.0 on Python 3.9, where htmltools 0.7 is unavailable.

**What has not been verified: anything in a browser.** No Playwright or manual pass was run. Every claim above is about emitted HTML and Python behaviour.

**Reviewed adversarially before opening.** That pass caught a regression in an earlier revision of this branch: the figure cleanup also dropped maidr's `FigureManager` record, which is correct for a figure the render throws away but wrong for one built lazily on the first flush and cached (`@reactive.calc`, any memoised helper) — those degraded to a static image from the second render on, silently. Now only `plt.close` is called, and `test_a_figure_built_lazily_and_cached_stays_accessible` fails if that returns.

`FigureManager.figs` still retains an entry per figure. That is pre-existing and not fixable from the renderer — the entry's value holds the figure, so even a weakly-keyed map keeps it reachable. Follow-up issue to come.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,3 +49,6 @@ See [Offline Use and the JavaScript Bundle](https://xability.github.io/py-maidr/
 ## Example Code
 
 We provide [some example code](https://github.com/xability/py-maidr/blob/main/example) for using py-maidr with matplotlib, seaborn, Jupyter Notebook, Quarto, Shiny, and Streamlit.
+
+Shiny support requires the optional extra `pip install "maidr[shiny]"`, which
+provides `output_maidr()` and `@render_maidr` in `maidr.widget.shiny`.
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -1125,6 +1125,57 @@ plt.show() #<<
 
 ### Shiny
 
+Shiny support ships as an optional extra:
+
+```bash
+pip install "maidr[shiny]"
+```
+
+Pair `output_maidr()` in the UI with `@render_maidr` in the server, the same
+way you would pair `ui.output_plot()` with `@render.plot`:
+
+```python
+import matplotlib.pyplot as plt
+from shiny import App, ui
+
+from maidr.widget.shiny import output_maidr, render_maidr
+
+app_ui = ui.page_fluid(
+    ui.input_slider("n", "Bars", min=2, max=10, value=4),
+    output_maidr("bars"),
+)
+
+
+def server(input, output, session):
+    @render_maidr
+    def bars():
+        fig, ax = plt.subplots()
+        ax.bar(range(input.n()), range(1, input.n() + 1))
+        return ax
+
+
+app = App(app_ui, server)
+```
+
+The decorated function may return a matplotlib or seaborn artist, a Plotly
+`Figure`, or an Altair chart. Returning `None` leaves the output blank.
+
+`@render_maidr` accepts `width`, `height`, and `use_cdn`. Prefer `use_cdn`
+here over `maidr.set_use_cdn()`: the setter is process-wide state shared by
+every concurrent session, while the argument is scoped to one output.
+
+```python
+@render_maidr(width="600px", use_cdn=False)
+def bars():
+    ...
+```
+
+`use_cdn=False` is the setting for an air-gapped deployment. The chart is
+rendered in an iframe, and an iframe's `srcdoc` cannot reach assets the app
+serves, so the bundled `maidr.js` travels inline in the document -- correct
+offline, but roughly 2 MB per chart. The default (`"auto"`) loads from the
+CDN and stays around 30 KB.
+
 Check out [a reactive Shiny dashboard example with maidr](https://xabilitylab.shinyapps.io/a11y_dashboard/) and its source code is available on [GitHub](https://github.com/xability/a11y_dashboard).
 
 ### Streamlit
--- a/example/shiny/example_shiny_scatter.py
+++ b/example/shiny/example_shiny_scatter.py
@@ -2,7 +2,7 @@ import matplotlib.pyplot as plt
 import seaborn as sns
 from shiny import App, ui
 
-from maidr.widget.shiny import render_maidr
+from maidr.widget.shiny import output_maidr, render_maidr
 
 # Load the dataset
 iris = sns.load_dataset("iris")
@@ -25,7 +25,7 @@ app_ui = ui.page_fluid(
                 selected="sepal_width",
             ),
         ),
-        ui.column(9, ui.output_ui("create_barplot")),
+        ui.column(9, output_maidr("create_scatterplot")),
     )
 )
 
@@ -33,7 +33,7 @@ app_ui = ui.page_fluid(
 # Define the server
 def server(inp, _, __):
     @render_maidr
-    def create_barplot():
+    def create_scatterplot():
         fig, ax = plt.subplots(figsize=(10, 6))
         s_plot = sns.scatterplot(
             data=iris, x=inp.x_var(), y=inp.y_var(), hue="species", ax=ax
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -26,12 +26,12 @@ from maidr.core.plot.barplot import BarPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
 from maidr.util.dependencies import (
     MAIDR_JS_FILENAME,
+    bundled_cdn_url,
+    inline_bundle_tags,
     maidr_bundled_files_dependency,
     maidr_bundled_relative_dir,
     maidr_html_dependency,
     maidr_js_cdn_url,
-    bundled_cdn_url,
-    inline_bundle_tags,
     schema_trace_types,
     warn_if_bundle_cannot_render,
     warn_if_bundle_is_stale,
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -30,6 +30,8 @@ from maidr.util.dependencies import (
     maidr_bundled_relative_dir,
     maidr_html_dependency,
     maidr_js_cdn_url,
+    bundled_cdn_url,
+    inline_bundle_tags,
     schema_trace_types,
     warn_if_bundle_cannot_render,
     warn_if_bundle_is_stale,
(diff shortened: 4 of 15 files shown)


jooyoungseo:
Both potential bugs reproduce. Fixed in b40afd6, with tests that fail without the fix.

**1. `UnicodeDecodeError` escaping the fallback — correct.**

```
read_text raises: UnicodeDecodeError | is OSError: False | is ValueError: True
```

The except clause is now `(OSError, ValueError)` — `FileNotFoundError` is an `OSError`, and `ValueError` is there for the decode error specifically, with a comment saying why it is not redundant. `test_inline_bundle_tags_survive_a_corrupted_bundle` covers it.

**2. `_check_supported` letting `StopIteration` through — correct, and worse than described.**

The bare `StopIteration` does not merely replace the intended `TypeError`; because `render()` is a coroutine, Python converts it on the way out:

```
[]     -> RuntimeError : coroutine raised StopIteration
{}     -> RuntimeError : coroutine raised StopIteration
[1, 2] -> AttributeError : 'int' object has no attribute 'axes'
```

Returning an empty list is an ordinary thing for a render function that filtered its data down to nothing to do, so this is reachable in normal use, not just by misuse.

You are right that it is latent in `FigureManager.get_axes` rather than newly introduced — the point is that it is a *resolver*, not a validator, so it raises whatever the traversal happens to hit. Rather than enumerate exception types at the call site, `_check_supported` now treats any failure to resolve as "not renderable", since the answer to the question it is asking is the same either way. The parametrized test covers `str`, `int`, `[]`, `{}` and `[1, 2]`.

**3. The `uv.lock` diff — checked, and it is clean.**

Your hypothesis was right: it is marker splitting, and no pinned version moved. Comparing every `name`/`version` pair between `64a1ef9` and this branch:

```
packages added:   14        (shiny + its transitive deps)
packages removed: 0
version sets changed for pre-existing packages: 0
```

So nothing already in the lock changed version, in either direction. The volume is the resolver re-expressing existing markers now that the graph has more Python-version-conditional edges in it.

**On the `FakeSession._process_ui = AppSession._process_ui` coupling** — agreed that it is a trade rather than a free win. It is deliberate in this direction because the failure modes are asymmetric: coupling to a private method means the test suite breaks loudly on a Shiny upgrade, whereas a hand-written copy of the payload shape means it passes quietly while production breaks. The earlier version of this file did have the hand-written copy, and a review pass demonstrated that deleting `AppSession._process_ui` entirely left all 17 tests green. Worth revisiting if it becomes noisy, but a red test is the outcome I want there.

---

jooyoungseo:
Thanks — both points are fair. One acted on, one already tracked.

**The `use_cdn="auto"` routing observation is correct, and your reading of the net effect is exactly right.** It was non-functional in a live Shiny app before this PR (the parent-source stash is only ever populated by `init_notebook()`, which returns early outside a notebook) and it is non-functional after. Not a regression, and the primary CDN path is untouched.

Filed as **#455** rather than folded in here, because the fix you suggest — routing `"auto"` through `iframe_inline_bundle` — would inline ~1.9 MB on the *default* setting, which is the one thing `"auto"` exists to avoid. The honest options are to serve the bundle from the host so a `srcdoc` document can reach it by absolute URL, or to keep `"auto"` CDN-only in an iframe and make the dead `onerror` say so instead of failing silently. Both are more than a predicate change, so they get their own issue.

You are right that `will_iframe` and the `"auto"` branch can still disagree about what an iframe render needs. What survives is dead weight rather than a wrong answer — the `HTMLDependency` is dropped and the relative fallback never fires — but it is the kind of thing that reads as intentional to the next person, so #455 says so explicitly.

**The `except ImportError` nit is a real one, and it bit me during this work.** Fixed in 341dc8f.

The two failures need opposite advice: `shiny` genuinely absent, versus installed with a broken import chain. The second is not hypothetical — `ImportError: cannot import name 'TagifiedTag' from 'htmltools'` is what this repo's own environment produced when shiny 1.7 met the locked htmltools, and it is an `ImportError`, so it was already being caught and answered with "install `maidr[shiny]`" — advice that would have sent someone to reinstall a package they already had.

The absent case is now recognised by `ModuleNotFoundError.name` and keeps the message naming the extra. Anything else says the package is installed but its imports failed, names the likely cause, and quotes the original error rather than swallowing it. Parametrized test over both shapes, including the exact `TagifiedTag` message.

I left non-`ImportError` failures propagating as themselves. `is_shiny()` widens to `Exception` because it is a *probe* — nobody asked for Shiny, so it must never fail a render. A module import is the opposite: the caller asked for this module explicitly, and relabelling an arbitrary exception as an `ImportError` would hide what actually went wrong.

CI is green on 3.9–3.12 (2007 tests).

---

jooyoungseo:
Two of the four applied in 909968b; two declined with reasons.

**Applied — `_close_new_figures` swallowing `Exception` silently.** Fair. It runs in a `finally`, so it must not raise, but swallowing without a word means a bug in the cleanup would present as figures quietly accumulating — the exact symptom it exists to prevent, with no hint where to look. It now warns and carries on.

**Applied — import ordering.** Sorted into the block's alphabetical order in both files.

**Declined — PEP 604 for `UseCdn`.** This one would break Python 3.9, which the package supports. `from __future__ import annotations` defers *annotations*; `UseCdn = ...` is an assignment, so it is evaluated eagerly:

```
--- python 3.11 ---
alias built OK
--- python 3.9 ---
    UseCdn = bool | Literal["auto"] | None
TypeError: unsupported operand type(s) for |: 'type' and '_LiteralGenericAlias'
```

The `use_cdn` parameter in `maidr/api.py` uses `|` legitimately, because those *are* annotations and the future import does defer them. The distinction is easy to lose, so it is now a comment on the alias rather than something for the next person to rediscover.

**Declined as an issue — `session._process_ui`.** Real coupling, but there is no public equivalent to migrate to: `shiny.render.ui.transform` is itself `session._process_ui(value)` (`shiny/render/_render.py`), so every custom renderer that returns UI is on the same footing. Filing "migrate when a public API appears" would be an issue nobody can act on.

What that risk needs is a detector, and it has one: `tests/widget/conftest.py` sets `FakeSession._process_ui = AppSession._process_ui`, so if the method is renamed, moved, or has its dependencies changed, the suite fails loudly instead of drifting. Verified during an earlier pass: deleting `AppSession._process_ui` leaves the tests red rather than green.

Everything green on 3.9–3.12 (2007 tests). Merging.

---

PR #458, opened 2026-08-17
fix(seaborn): read a binned regplot's intervals as uncertainty, not as fits

### LINKED ISSUE #451: A binned regplot announces each confidence interval as its own fitted curve, one layer per bin
## The defect

`sns.regplot(x_estimator=...)` collapses each x value to an estimate and draws a confidence interval around it. Those intervals are drawn as ordinary lines, so each one registers as a **`smooth` layer of its own**.

```python

rng = np.random.default_rng(7)
df = pd.DataFrame({"dose": np.repeat([1.0, 2.0, 3.0, 4.0], 15),
                   "resp": rng.normal(size=60) + np.repeat([1, 3, 5, 8], 15)})

_, ax = plt.subplots()
sns.regplot(data=df, x="dose", y="resp", x_estimator=np.mean, ax=ax)
print([p.type.value for p in FigureManager.get_maidr(ax.get_figure()).plots])
# ['point', 'smooth', 'smooth', 'smooth', 'smooth', 'smooth']
```

Six layers for four estimates and one regression line. The middle four are the intervals:

```
[0] point    n= 4   {'x': 1.0, 'y': 0.865}, {'x': 2.0, 'y': 2.307}, ...
[1] smooth   n= 2   {'x': 1.0, 'y': 0.573}, {'x': 1.0, 'y': 1.199}
[2] smooth   n= 2   {'x': 2.0, 'y': 1.802}, {'x': 2.0, 'y': 2.737}
[3] smooth   n= 2   {'x': 3.0, 'y': 4.409}, {'x': 3.0, 'y': 5.204}
[4] smooth   n= 2   {'x': 4.0, 'y': 7.693}, {'x': 4.0, 'y': 8.659}
[5] smooth   n=30   the regression line, with its band on yMin/yMax
```

## Three losses, and they compound

**The type is wrong.** `smooth` means *a computed fit*. A vertical confidence bar is not a fit — it is an uncertainty bound on one estimate. A reader who navigates into layer 1 is told they are on a curve, and hears two points at the same x with different y.

**The layer count scales with the data.** `x_estimator` bins by unique x when `x_bins` is not given, so a continuous x produces one layer per distinct value. On 60 distinct x values:

```
sns.regplot(data=df, x="a", y="b", x_estimator=np.mean)
  ['point', 'smooth' × 60]      # 61 layers
```

Sixty-one layers to navigate, sixty of which are two-point vertical bars.

**The link between an estimate and its interval is severed.** The estimates are in layer 0; each interval is in a layer of its own. Nothing connects `{'x': 1.0, 'y': 0.865}` to the bar that runs from `0.573` to `1.199`, so the uncertainty is unreachable from the value it belongs to.

## What the same numbers look like read correctly

`sns.pointplot` on the same frame draws the same thing — an estimate per level with a confidence interval — and reads as **one** layer:

```
[0] error_bar n=4  {'x': '1.0', 'y': 0.865, 'yMin': 0.593, 'yMax': 1.168}, ...
```

One sample per estimate, carrying its own bounds. That is what #246 added `PlotType.ERRORBAR` for, and `maidr/patch/pointplot.py::_register_point_layer` already contains the split that produces it: it separates the estimate lines from the interval polylines, verifies they pair one-to-one, and falls back to describing every drawn line when they do not.

The same shape of defect as #448's `catplot(kind="point")`, which dropped the intervals entirely — this one keeps them and mistypes them, which is arguably worse, since sixty phantom curves are audible while a missing bound is not.

## Scope

* `x_estimator=<callable>` and `x_bins=<int>` both take this path (measured: identical layer shapes).
* `ci=None` is already correct — seaborn draws no intervals, so the reading is `['point', 'smooth']`, two layers.
* `sns.lmplot` inherits it, since #344 made it run the patched `regplot`. Measured identical.
* The regression line and its band (layer 5 above) are correct and must stay: a binned regplot legitimately has *both* an uncertainty layer and a fit.

Target reading: `error_bar(4)` for the estimates and their intervals, plus `smooth(30)` for the fitted curve — two layers rather than six, and sixty-one becomes two.

## Related

- #246 — the `pointplot` intervals, and `_register_point_layer`'s estimate/interval split
- #448 — `catplot(kind="point")` dropping the same intervals through a different route
- #440 — a `line` layer describing another chart's geometry, which is this failure with a different type name

Measured on seaborn 0.13.2 / matplotlib 3.10.

jooyoungseo:
Closes #451.

`sns.regplot(x_estimator=...)` collapses each x to an estimate and draws a confidence interval around it. Those intervals are ordinary lines, and the patch asked the **axes** which of its lines were fits rather than knowing which lines its own call drew:

```python
sns.regplot(data=df, x="dose", y="resp", x_estimator=np.mean)
```

| | before | after |
|---|---|---|
| 4 dose levels | `point(4)`, `smooth(2)`×4, `smooth(30)` | `error_bar(4)`, `smooth(30)` |
| 60 distinct x | `point(60)`, `smooth(2)`×60, `smooth(30)` | `error_bar(60)`, `smooth(30)` |

## Three losses, and they compound

**The type is wrong.** `smooth` means *a computed fit*. A vertical confidence bar is not one, so a reader who navigates into layer 1 is told they are on a curve and hears two points at the same x.

**The count scales with the data.** `x_estimator` bins by unique x when `x_bins` is absent, so a continuous x gave one layer per distinct value — sixty-one layers, sixty of them two-point bars.

**The link is severed.** The estimates were in layer 0 and each interval in a layer of its own, so the uncertainty was unreachable from the value it bounds. `sns.pointplot` draws the same quantity and has read as one `error_bar` layer with `yMin`/`yMax` on each sample since #246; the two now agree.

## A second symptom, same root, no `x_estimator` needed

`find_smooth_lines_by_label` matched any label starting with `_child` — which is what matplotlib names *any* unlabelled artist, not something a regression line is distinguished by:

```python
ax.plot(...)                 # unlabelled -> matplotlib names it _child0
sns.regplot(data=df, ...)
```

| drawn | before | after |
|---|---|---|
| `regplot` alone | `point`, `smooth` | unchanged |
| a plain line, **then** `regplot` | `line`, `point`, `smooth`, **`smooth`** | `line`, `point`, `smooth` |
| `sns.lineplot`, **then** `regplot` | `line`, `point`, `smooth`, **`smooth`** | `line`, `point`, `smooth` |
| `regplot`, **then** a plain line | `point`, `smooth`, `line` | unchanged |

The caller's own series announced twice — once correctly, once as a model of itself. Ordering decided it: a line added *after* the regplot was safe, and "plot the series, overlay a fit" is the failing order.

## The change

A before/after snapshot of the lines — the idiom `lineplot.py` (`DRAWN_SERIES`) and `boxenplot.py` (`DRAWN_LADDERS`) already use — plus a geometric split of what it returns:

* an **interval bar** stands at one x (`np.ptp(finite_x) == 0`) with few enough vertices to be one bar;
* the **fitted curve** spans the axis.

Asked of the geometry rather than the label, because "is this a vertical bar or a curve" is the question the split actually needs answered, and `_child0` was never an answer to it.

The estimates and their intervals then travel as one `ERRORBAR` layer through the same `PointPlot` that `sns.pointplot` uses, so nothing about the estimate/interval reading is reimplemented.

Collections are snapshotted too, so the scatter layer is handed its own points instead of sweeping the axes — #426's rule, which this path had not got. A `regplot` drawn over an existing scatter was describing that one's points.

## Verified rather than assumed

The pairing follows `pointplot`'s discipline: the counts must agree and every bar must stand at an estimate, and when they do not the layer falls back to the scatter-and-curves reading — incomplete rather than wrong. `PointPlot` zips estimates against intervals positionally, so a pairing off by one would put every bound on the wrong estimate, which reads as complete and is wrong about every interval in it.

Neither guard is reachable from anything seaborn draws, so both are driven directly — the same rule `test_boxenplot.py` applies to `BoxenPlot`'s unreachable branch.

## Tests

`tests/core/plot/test_regplot_binned_estimates.py`, 20 tests: the layer shape and its independence from the bin count, the bounds on each sample checked against the estimator the caller passed, agreement with `pointplot` on the same estimates, the fit surviving alongside the intervals, `ci=None`/`scatter=False`/`order=2`/`lowess` unchanged, both drawing orders of the `_child` case, and the pairing guards.

Suite: **2087 passed**. `ruff check` clean on every file touched (six pre-existing `F401` re-export findings in `maidr/core/plot/__init__.py` unchanged).

## Falsification

| reverted | failures |
|---|---|
| the whole change | 9 |
| no interval/curve split | 6 |
| no before/after line snapshot | 2 |
| no collection snapshot | 1 |
| pairing position check removed | 1 |
| pairing count check removed | 1 |

The last four fail alone and disjointly. The two pairing rows are the reason the direct tests exist: both branches passed the whole file before they were written, because nothing seaborn draws reaches them.

---


jooyoungseo:
Thanks — the dead-code ask is right and it's done, in `745960f`.

## `find_smooth_lines_by_label` — deleted

Confirmed: after this PR its only reference is its own docstring example. `maidr/core/plot/regplot.py` uses the neighbouring `find_regression_line`, which stays.

Worth saying *why* it deserved deleting rather than keeping: it existed to answer "which lines on these axes are fits", and it answered by label — `_child*`, which is what matplotlib names any unlabelled artist. That heuristic **is** both symptoms of this issue. The snapshot replaces the question, not just the implementation: it is now "which lines did *this call* draw", and the split among those is geometric. Leaving the old answer exported would leave the next caller a way back into the bug. `SMOOTH_KEYWORDS` and `List` were imported for it alone and go too.

## `_prospective_axes` — noted

Correct, and worth writing down as you say. Added:

> On a figure that exists but holds no axes, `gca()` does create one, a beat before `regplot` would have. That is deliberate and costs nothing: seaborn's own `plt.gca()` a moment later returns the same axes, so the snapshot is taken on the axes actually drawn on — and an axes with nothing on it snapshots as empty either way.

## The fallback integration test — I tried, and it does not reproduce

This is the one I'd push back on, having measured it. Your suggested route was mixed group sizes with a single-observation bin. Seaborn draws a bar per estimate regardless:

```python
# dose 3.0 has exactly one observation
df = pd.DataFrame({"dose": [1.0]*15 + [2.0]*15 + [3.0] + [4.0]*15, ...})
sns.regplot(data=df, x="dose", y="resp", x_estimator=np.mean)

lines drawn: [(2, 1), (2, 1), (2, 1), (2, 1), (100, 100)]   # four bars, four estimates
layers: ['error_bar', 'smooth']
```

The n=1 bin gets a degenerate zero-height bar rather than none, so the counts still agree and the pairing succeeds. I could not find a spelling that reaches the fallback from seaborn at all — which is the same conclusion `_paired_estimates`' docstring reaches, and the reason both guards are driven directly instead. Writing a test that *forced* the fallback through some contrived monkeypatching would pin the mechanics of the mock rather than the behaviour, and the direct tests already do the real job (each fails alone when its check is removed).

If you know a real seaborn call that produces a count or position mismatch, I'll add the end-to-end test for it — I'd rather have one, I just could not construct one that was honest.

Suite: **2087 passed**, CI green across 3.9–3.12.

---

jooyoungseo:
All three addressed in `9210357`.

## The dropped-layer arm — restructured, not commented

You read it correctly: `_paired_estimates` declining *and* `points is None` left the bars in neither branch, and it needs `scatter=False` together with binning, which seaborn cannot produce today.

I took the restructure rather than the comment, because of what the failure would be. Mistyping a bar as a curve is the reading this PR replaces and is bad; **dropping** it is worse — a layer that is not there cannot be navigated to, and nothing announces its absence, so the chart reads as complete with a piece simply missing. And "unreachable today" is the property most likely to stop holding on a seaborn release, which is the whole reason this patch exists.

So the two arms now converge on one `described` list, and no path can drop an artist:

```python
if paired is not None:
    ... ERRORBAR ...
    described = curves
else:
    if points is not None:
        ... SCATTER ...
    described = intervals + curves
_register_curves(axes, described, ...)
```

Two tests drive it by making the pairing decline, since seaborn will not — one with the scatter present, one without. The first fails when the dropping arm is restored.

## The O(n) snapshots — right, and it is the same precedent

Confirmed and fixed: `before_lines`/`before_collections` are now sets. Worth noting the one thing I kept from the current form — they hold the **artists**, not their `id()`s. An id is only unique while its object is alive, so an id-keyed snapshot can be matched by an unrelated artist allocated at a freed address; holding the objects keeps every one alive for the comparison. That costs nothing here because an `Artist` hashes by identity, so the set is O(1) either way. `histogram.py::_containers_of` documents the same reasoning.

## The duplicated constant

Moved to `maidr/patch/common.py` as `MAX_INTERVAL_VERTICES`, imported by both. It is a fact about seaborn's interval geometry rather than about either reader of it, and your point stands that two copies would need to move together.

Suite: **2089 passed**.

---

Separately, on your earlier note asking for an end-to-end test of the pairing fallback: I measured the scenario you suggested (a bin with a single observation) and it does not reach it — seaborn draws a degenerate zero-height bar rather than none, so the counts still agree. Details in the comment above. The monkeypatched tests added here are the closest honest thing: they pin the *dispatch* when pairing declines, which is the part that could silently drop a layer, rather than the mechanics of a mock.

---

jooyoungseo:
Both addressed in `d922175`.

## The three blank lines — fixed

Confirmed and removed. It was the hole `_MAX_INTERVAL_VERTICES` left when it moved to `common.py` in the previous round.

## The `capsize` caution — measured, and written down

You're right that this is the assumption to name, so I checked it rather than assert it:

```
regplot has capsize:   False
pointplot has capsize: True
seaborn 0.13.2
```

So the two functions genuinely differ, and `_is_interval_bar`'s single-x test is the thing resting on it. Added to the docstring, including which check would give way first — a useful detail, because it isn't the obvious one:

> The single-x test is the part that depends on `regplot` having no `capsize`, which it does not — `pointplot` takes one and `regplot`'s signature has no equivalent. A capped bar draws its caps at two further x values, so were one ever added here the width test would fail first and the bar would be described as a curve; the vertex ceiling is only an upper bound and would not be what gave way.

The vertex ceiling reads like the capsize-aware part of the check (8 is the capped count `pointplot` needs), which makes it the natural place to look — and it would still pass. Worth a reader not having to derive that under time pressure.

## The synthetic estimate `Line2D` — agreed, and it holds today

Your reading is right: it is never added to the axes, and that is safe only because `PointPlot` calls `.get_data()` on it and puts the real interval lines alone into `self._elements`. I re-checked that both are still true on this branch. Nothing to change now — if `PointPlot` ever highlights the estimate too, this would have to become a real artist, and it would fail loudly at selector-generation time rather than silently, since a line with no axes has no SVG element to match.

## The comment density

Noted, and I've left it. The rationale here is mostly *why not the obvious thing* — why geometry rather than the label, why the snapshot is per-call, why the unpairable case describes rather than drops — and each of those is a wrong turn this patch has already been down once. I'd rather a future maintainer read a paragraph than re-derive the label heuristic.

Suite: **2089 passed**.

---

PR #459, opened 2026-08-17
feat(streamlit): add a first-class Streamlit integration

jooyoungseo:
## Description

py-maidr had no Streamlit code at all — `grep -rn streamlit maidr/ tests/` returned nothing. The only guidance was `example/streamlit/example_streamlit_app.py`, which hand-rolled:

```python
components.html(maidr.render(plot).get_html_string(),
                scrolling=True, height=fig_height * 100, width=fig_width * 100)
```

Every part of that line is now wrong, and one part was always wrong.

Follows #452, which made the equivalent fixes on the Shiny path and whose `inline_bundle_tags()` this reuses.

### Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Documentation update

## Related Issues

None; found by auditing the integration against the Streamlit components docs.

## Changes Made

**`maidr/widget/streamlit.py`** — new module.

`render_maidr(plot, *, height, width, tab_index, use_cdn)` draws the chart. It prefers `st.iframe` and falls back to `components.v1.html`, which is deprecated with a removal date that has already passed and warns on every call.

Four things the hand-rolled line got wrong:

| | Before | Now |
|---|---|---|
| **Runtime** | `get_html_string()` drops `HTMLDependency`, so `use_cdn=False` produced a chart with **no maidr.js at all** — the same silent failure #452 fixed for Shiny | the bundle travels inline |
| **Height** | guessed as `fig_height * 100`; the API default renders as 150 px and crops | `height="content"`, so Streamlit measures the chart and maidr's braille and text panels stay visible when they open |
| **API** | `components.v1.html`, past its removal date | `st.iframe`, with the old call kept as a fallback |
| **Size units** | `width=fig_width * 100` | `"stretch"` / `"content"` / int |

`maidr_html(plot, *, use_cdn)` returns the same chart as a string. It exists separately because the useful lever against Streamlit's rerun-everything model is caching the *string*, and that needs a string-returning entry point.

`_warn_if_no_runtime` checks the emitted HTML actually carries a source for `maidr.js`. A chart with none still looks right — it is the SVG, unchanged — while being silently unusable. That failure is invisible to a sighted developer testing their own app, which is exactly why it gets an assertion rather than trust.

**The iframe is deliberate, and the module says so.** Streamlit's app binds `r`, `c` and `esc` at the document level, exempting only form fields; maidr binds all three. Two listeners on one document cannot both win and `preventDefault` in one does not stop the other, so rendering the chart directly into the Streamlit page would take those keys away from maidr. This is also why Streamlit's components v2 is not the target.

Also: `streamlit` optional extra, imported lazily so `maidr_html` works without it; example rewritten; docs, README and CLAUDE.md updated.

### Backwards compatibility

Purely additive. No existing module or behaviour changes, and the old hand-rolled `components.html(...)` call keeps working exactly as before — including keeping its silent `use_cdn=False` failure, which is why the example is updated in the same PR.

## Checklist

## Additional Notes

**Verification.** `uv run pytest` — 2118 passed (from 2112; 21 new tests). `ruff check` clean on every changed file. `uv lock --check` clean. `st.iframe` was driven for real against the installed Streamlit 1.61.1, not only through stubs.

**What has not been verified: anything in a browser.** No Playwright or manual pass was run. Every claim above is about emitted HTML and Python behaviour.

**Reviewed adversarially before opening, and it caught five defects in the first draft** — all fixed in 4fbf2af, each with a test that fails without the fix:

1. **Altair got 1.9 MB it never loads.** `maidr_html` decided whether to inline from a mode it re-derived itself, but `maidr.render` hands an Altair chart to the Vega-Lite adapter *before* `use_cdn` is consulted. Under `use_cdn=False` an Altair chart went from 3,960 to 1,871,570 characters, on every rerun, while still requiring the network. The decision now reads the emitted HTML and says plainly that `use_cdn=False` cannot be honoured for such a chart.
2. **`tab_index` silently dropped on Streamlit 1.45–1.55.** The fallback is gated on `hasattr(st, "iframe")` and treated that as "too old for `tab_index`" — but `components.v1.html` accepted it from 1.45, eleven releases before `st.iframe` landed in 1.56. It now asks the installed function what it takes.
3. **`tab_index` defaulted to `0` on a false premise.** The docstring claimed the browser default leaves the frame unreachable from the keyboard. It does not: an iframe's contents take part in sequential focus navigation, and maidr gives the chart its own tab stop. Forcing `0` only added a stop on the frame — announced as "st.iframe" identically for every chart, since Streamlit hardcodes that name. Default is now the browser's; the argument remains for layouts that want a deterministic landing point.
4. **`_bundle_marker` vouched for a chart with no runtime.** It returned `""` when the bundle could not be read, and `""` is a substring of every string. A zero-byte `maidr.js` produced 382,504 characters, an empty `<script>`, and no warning.
5. **The documented caching recipe returned the first chart forever.** With every argument underscore-prefixed, Streamlit hashes nothing and the cache key is constant — changing the selectbox left the chart pinned to the first selection, silently. The recipe now carries an explicit key and says why it is not optional.

`streamlit>=1.30` is deliberate rather than raised to 1.56: py-maidr supports Python 3.9, and no Streamlit at or above 1.55 does, so a 1.56 floor would make `maidr[streamlit]` uninstallable on a supported interpreter. Both embed APIs are handled instead.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -37,7 +37,8 @@ The `maidr/patch/` modules use `wrapt` to intercept matplotlib/seaborn plot call
 - **`maidr/core/plot/`** — Factory pattern: `MaidrPlotFactory` dispatches to concrete `MaidrPlot` subclasses (BarPlot, BoxPlot, HeatPlot, etc.) based on `PlotType` enum. Each subclass implements `_extract_plot_data()`
 - **`maidr/patch/`** — One module per plot type (barplot.py, boxplot.py, etc.) plus `highlight.py` (injects maidr attributes into SVG elements) and `clear.py` (cleanup on `plt.clf`/`plt.cla`)
 - **`maidr/util/mixin/`** — Reusable extraction logic: `ContainerExtractorMixin`, `LevelExtractorMixin`, `LineExtractorMixin`, `CollectionExtractorMixin`, `FormatExtractorMixin`
-- **`maidr/widget/shiny.py`** — Shiny framework integration
+- **`maidr/widget/shiny.py`** — Shiny framework integration (`output_maidr`, `@render_maidr`)
+- **`maidr/widget/streamlit.py`** — Streamlit integration (`render_maidr`, `maidr_html`)
 
 ### Supported Plot Types
 
--- a/README.md
+++ b/README.md
@@ -52,3 +52,5 @@ We provide [some example code](https://github.com/xability/py-maidr/blob/main/ex
 
 Shiny support requires the optional extra `pip install "maidr[shiny]"`, which
 provides `output_maidr()` and `@render_maidr` in `maidr.widget.shiny`.
+Streamlit support requires `pip install "maidr[streamlit]"`, which provides
+`render_maidr()` and `maidr_html()` in `maidr.widget.streamlit`.
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -1223,14 +1223,20 @@ open panel, and the audio state. Caching the *HTML* rather than the render
 call is the lever that helps:
 
 ```python
-from maidr.widget.streamlit import maidr_html, render_maidr
+from maidr.widget.streamlit import maidr_html
 
 @st.cache_data
 def chart_html(_fig, key):     # `_fig` is not hashed; `key` is the cache key
     return maidr_html(_fig)
 
-render_maidr(fig)              # or place chart_html(fig, key=day) yourself
+st.iframe(chart_html(fig, key=day), height="content", width="stretch")
 ```
+
+`st.iframe` arrived in Streamlit 1.56. On older versions place the same
+cached string with `st.components.v1.html(chart_html(fig, key=day),
+height=600, scrolling=True)` — it cannot size itself, so give it a height.
+`render_maidr()` handles that difference for you, which is why it is the
+better default when you are not caching.
 
 Both arguments matter. The leading underscore tells Streamlit not to hash the
 figure, which a matplotlib `Figure` does not support -- which leaves `key` as
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -1180,6 +1180,65 @@ Check out [a reactive Shiny dashboard example with maidr](https://xabilitylab.sh
 
 ### Streamlit
 
+Streamlit support ships as an optional extra:
+
+```bash
+pip install "maidr[streamlit]"
+```
+
+```python
+import matplotlib.pyplot as plt
+import streamlit as st
+
+from maidr.widget.streamlit import render_maidr
+
+day = st.selectbox("Day", ["Thu", "Fri", "Sat", "Sun"])
+
+fig, ax = plt.subplots()
+ax.bar(["lunch", "dinner"], [12, 30])
+ax.set_title(f"Tips on {day}")
+
+render_maidr(ax)
+```
+
+`render_maidr()` accepts a matplotlib or seaborn artist, a Plotly `Figure`, or
+an Altair chart, plus `height`, `width`, `tab_index`, and `use_cdn`.
+
+The defaults are chosen for accessibility rather than for layout:
+
+- `height="content"` lets Streamlit measure the chart, so maidr's braille and
+  text panels stay visible when they open. A fixed height crops them.
+- `tab_index=0` puts the chart in the page's tab order. The browser default
+  leaves the frame unreachable from the keyboard, which for a library whose
+  entire interface is keyboard-driven means unreachable without a mouse.
+
+::: {.callout-note}
+## Streamlit reruns the whole script on every interaction
+
+Each rerun rebuilds the chart from scratch, so maidr loses its position, any
+open panel, and the audio state. Caching the *HTML* rather than the render
+call is the lever that helps:
+
+```python
+from maidr.widget.streamlit import maidr_html
+
+@st.cache_data
+def chart_html(_fig):          # the underscore skips hashing the figure
+    return maidr_html(_fig)
+
+st.iframe(chart_html(fig), height="content", tab_index=0)
+```
+:::
+
+::: {.callout-note}
+## Air-gapped deployments
+
+`use_cdn=False` embeds the ~1.9 MB `maidr.js` bundle in the page. That is
+required rather than wasteful: the chart is rendered inside an iframe, and an
+iframe built from an HTML string cannot fetch anything the app serves. The
+default loads from the CDN and stays around 15 KB.
+:::
+
 Check out [this Streamlit dashboard](https://0193dbe9-1a63-183e-ba7f-b62075dd1d71.share.connect.posit.cloud/) with maidr, and its source code is available on [GitHub](https://github.com/xability/maidr_streamlit).
 
     * Note: `Streamlit` framework has some "Unlabeled 0 Button" which does not have to do with our maidr package. This issue needs to be addressed by the `Streamlit` team.
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -1208,9 +1208,12 @@ The defaults are chosen for accessibility rather than for layout:
 
 - `height="content"` lets Streamlit measure the chart, so maidr's braille and
   text panels stay visible when they open. A fixed height crops them.
-- `tab_index=0` puts the chart in the page's tab order. The browser default
-  leaves the frame unreachable from the keyboard, which for a library whose
-  entire interface is keyboard-driven means unreachable without a mouse.
+- `tab_index=None` leaves the frame's tab order to the browser. The chart is
+  still reached by tabbing: an iframe's contents take part in sequential focus
+  navigation, and maidr gives the chart inside its own tab stop. Passing `0`
+  makes the frame a stop as well, which is one extra Tab before the chart, and
+  Streamlit hardcodes the frame's accessible name as `st.iframe`, so that stop
+  announces identically on every chart on the page.
 
 ::: {.callout-note}
 ## Streamlit reruns the whole script on every interaction
@@ -1220,14 +1223,20 @@ open panel, and the audio state. Caching the *HTML* rather than the render
 call is the lever that helps:
 
 ```python
-from maidr.widget.streamlit import maidr_html
+from maidr.widget.streamlit import maidr_html, render_maidr
 
 @st.cache_data
-def chart_html(_fig):          # the underscore skips hashing the figure
+def chart_html(_fig, key):     # `_fig` is not hashed; `key` is the cache key
     return maidr_html(_fig)
 
-st.iframe(chart_html(fig), height="content", tab_index=0)
+render_maidr(fig)              # or place chart_html(fig, key=day) yourself
 ```
+
+Both arguments matter. The leading underscore tells Streamlit not to hash the
+figure, which a matplotlib `Figure` does not support -- which leaves `key` as
+the only thing it *can* hash. Drop it and every argument is skipped, the cache
+key is constant, and the first chart is handed back for the rest of the
+session: a chart that quietly stops matching its own controls.
 :::
 
 ::: {.callout-note}
--- a/example/streamlit/example_streamlit_app.py
+++ b/example/streamlit/example_streamlit_app.py
@@ -1,9 +1,8 @@
 import matplotlib.pyplot as plt
 import seaborn as sns
 import streamlit as st
-import streamlit.components.v1 as components
 
-import maidr
+from maidr.widget.streamlit import render_maidr
 
 # Streamlit app setup
 st.title("Maidr Streamlit Example")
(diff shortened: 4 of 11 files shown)


jooyoungseo:
All three acted on in 908bc5f. The `uv.lock` observation was the valuable one — it was a real regression, and not for the reason I'd have guessed.

**1. `stacklevel` off by one — correct, and reproduced.**

```
maidr_html     blamed stk.py:25  (want stk.py:17)
render_maidr   blamed stk.py:25  (want stk.py:20)
```

Both overshot to the same line — the frame above the actual caller — exactly as you described. `_warn_if_no_runtime` warns from inside its own body, a frame deeper than the Altair branch, so sharing one number cannot be right for both. That branch now takes `stacklevel - 1`:

```
maidr_html     blamed stk.py:17  (want stk.py:17)
render_maidr   blamed stk.py:20  (want stk.py:20)
```

And `test_a_chart_that_only_loads_maidr_remotely_is_not_inlined` now asserts *where* the warning lands, from both entry points — the guard the sibling warning already had and this one lacked.

**2. `_stacklevel` visibility — noted in the docstring.** It now says it is internal, and why the underscore alone is not enough.

**3. The `altair` downgrade — real, and worth the dig.**

Not an artifact of my environment. Root cause:

```
streamlit 1.50.0  ->  ['altair!=5.4.0,!=5.4.1,<6,>=4.0']
```

1.50 is the newest Streamlit that supports Python 3.9, and it caps `altair<6`. So `uv sync --all-extras` on 3.9 resolves altair to 5.5.0 instead of 6.0.0. Python 3.10+ is unaffected (6.1.0). The other three lock changes are benign: `packaging` gained a marker-split variant, and `starlette`/`websockets` lost variants without any version moving.

I tried scoping the dependency with `python_version >= '3.10'` first. It does nothing — uv locks all extras into one graph, so the user-facing `streamlit` extra drives the resolution regardless of what `test` says. The only mechanism that would actually separate them is `[tool.uv] conflicts`, and that breaks `uv sync --locked --all-extras --dev`, which is this repo's documented setup command and what CI runs.

So I did the two things that are honest:

- **Removed `streamlit` from the `test` extra.** It bought nothing — every Streamlit test stubs the module rather than importing it, so they run on 3.9 with or without the real package. That leaves one place asking for it instead of two.
- **Recorded the constraint next to the extra**, so the next person reading a lock diff does not have to re-derive it.

The downgrade itself stays, and I think that is right: it touches only the locked dev environment, three of the four supported interpreters still exercise the Altair adapter against altair 6, and on Python 3.9 anyone installing both `maidr[altair]` and `maidr[streamlit]` gets this same constraint from pip. It is a property of the ecosystem, not of this PR.

`uv run pytest` — 2118 passed. `ruff check` and `uv lock --check` clean.

---

jooyoungseo:
Real bug, and a good one — fixed in 7344924. The diagnosis is exactly right; the suggested fix would have broken the common path, which turned out to be the interesting part.

**The bug reproduces.** A Plotly chart always carries `cdn.plot.ly`, so `"<script" in html and "src=" in html` is satisfied by a script belonging to a different library, and the warning is suppressed for a chart with no maidr runtime. Same shape as the empty-`_bundle_marker` trap, via a different substring, in the one function written to catch exactly this. There's now a test that fails against the old gate:

```
=== with the OLD loose gate ===
FAILED tests/widget/test_streamlit.py::test_an_unrelated_cdn_script_does_not_vouch_for_maidr
```

**But `if _loads_maidr_remotely(html): return` would have made it worse.** That helper matches a literal `<script src=...maidr...>`, and maidr only arrives that way for Altair. The matplotlib and Plotly paths construct the element in JavaScript (`var s = document.createElement('script'); s.src = '...'`), so the markup holds the URL and no tag. Measured across every combination:

```
matplotlib  True  literal_script_src_maidr=False  maidr_url_anywhere=True
matplotlib  auto  literal_script_src_maidr=False  maidr_url_anywhere=True
matplotlib  False literal_script_src_maidr=False  maidr_url_anywhere=False
plotly      True  literal_script_src_maidr=False  maidr_url_anywhere=True   plotly_cdn=True
plotly      auto  literal_script_src_maidr=False  maidr_url_anywhere=True   plotly_cdn=True
plotly      False literal_script_src_maidr=False  maidr_url_anywhere=False  plotly_cdn=True
```

`literal_script_src_maidr` is `False` in all six. A tag-shaped gate would have warned "this chart has no runtime" on every ordinary render — a louder bug than the quiet one it fixed, and one no existing test would have caught either, since none asserted the *absence* of that warning.

So the check matches a **URL** naming the `maidr` package rather than a tag. That column is correct in all six rows: `True` wherever a runtime is fetched, `False` for `use_cdn=False` where the bundle is inline and the `_bundle_marker` gate takes over, and never confused by `cdn.plot.ly`.

You were also right that the two helpers were asking the same question — `_loads_maidr_remotely` is now `_references_maidr_runtime` and serves both callers, including the Altair branch.

Two tests, one per direction:
- `test_an_unrelated_cdn_script_does_not_vouch_for_maidr` — fails on the old gate.
- `test_a_normal_render_is_recognised_as_having_a_runtime[True|auto]` — fails on the tag-shaped one.

`uv run pytest` — 2121 passed. `ruff check` and `uv lock --check` clean.

---

jooyoungseo:
Checked, and the `src`/`srcdoc` concern is refuted — but the second half of your suggestion was right, and it is now in the suite (69cc2c1).

**`st.iframe` has no `srcdoc` parameter.** The proposed call would be a `TypeError`:

```
signature: (src: 'str | Path', *, width, height, tab_index) -> 'DeltaGenerator'
has srcdoc param: False
```

It classifies the argument instead. From its own docstring: *"**HTML string**: If `src` doesn't match any of the above patterns, it's treated as raw HTML and embedded directly in the iframe."* Absolute URLs, relative URLs and `Path` objects are the patterns; a string beginning `<div>` is none of them.

Driven through the real Streamlit, the routing is what we want:

```
exceptions: []
proto fields set: ['srcdoc', 'scrolling']
src set? False | srcdoc len: 14958
srcdoc starts: <div>   <script type="text/javascript">
maidr schema in srcdoc: True
```

`src` unset, the document in `srcdoc`, schema intact.

**Your other point stands, though, and it is the more useful one:** nothing in the suite could have told you that. The dispatch tests stub `st.iframe` with a recorder, so they assert what we *call* and are blind to what Streamlit *does*. That is exactly the gap that lets a `src`/`srcdoc` mistake stay green in Python and show up only as a blank embed.

So there is now `test_a_real_streamlit_app_embeds_the_chart_as_srcdoc`, which runs a one-chart app through `AppTest` and asserts on the element Streamlit produced: `src` unset, `srcdoc` populated, and both the MAIDR schema and the runtime URL present inside it. It skips where `st.iframe` does not exist — that is the range the `components.v1.html` fallback covers, and where the stubbed tests are the coverage. A URL/document mix-up now fails the suite instead of a browser.

**The stale `# pragma: no cover` is gone** — you are right that it was copied from `shiny.py`, where the branch really is unreachable in CI because shiny is always installed. Here `test_render_maidr_without_streamlit_names_the_extra` exercises it.

`uv run pytest` — 2122 passed. CI green on 3.9–3.12.

---

jooyoungseo:
Both addressed in a6be796. The docs one was a real regression I introduced while fixing an earlier review point.

**The caching recipe didn't cache — correct, and it's my fault twice over.** The previous round flagged that the snippet called `st.iframe` directly, which needs Streamlit 1.56 while the extra declares `>=1.30`. I removed the call — and took the caching with it, leaving `chart_html` defined and never used. The prose underneath then explained a `key` argument that nothing passed.

Fixed by placing the cached string again and stating the version difference instead of dodging it:

```python
@st.cache_data
def chart_html(_fig, key):     # `_fig` is not hashed; `key` is the cache key
    return maidr_html(_fig)

st.iframe(chart_html(fig, key=day), height="content", width="stretch")
```

followed by the `components.v1.html(..., height=600, scrolling=True)` form for older Streamlit, and a line saying `render_maidr()` is the better default precisely because it handles that difference for you. Now the recipe caches, the prose matches the code, and neither half is version-silent.

**`_accepts_tab_index` against a decorated function — checked, and it holds:**

```
streamlit 1.61.1
components.html is wrapped: True
signature: (self, html, width=None, height=None, scrolling=False, *, tab_index: 'int | None' = None)
_accepts_tab_index(real components.html): True
```

It *is* wrapped, as you suspected — Streamlit's `deprecate_func_name` — but the wrapper carries `__wrapped__`, which `inspect.signature` follows by default, so it reports the true signature rather than the wrapper's `(*args, **kwargs)`.

That was the right thing to be uneasy about, though, and you correctly identified that nothing in the suite would have caught it: every existing test probes a hand-written stub. There's now `test_tab_index_support_is_detected_on_the_real_streamlit`, which asserts the probe's answer matches `inspect.signature` on the installed function *and* that `__wrapped__` is still present — so if Streamlit ever swaps in a decorator that doesn't preserve it, the test says so instead of the library silently warning "too old" at users on a version that supports the argument.

Thanks for re-deriving the `stacklevel` arithmetic and the regex-vs-CDN-URL-shape reasoning independently rather than taking the PR description's word for it — that is the check that matters most on those two, since both are the kind of thing that looks right in prose.

`uv run pytest` — 2123 passed. CI green on 3.9–3.12.

---

jooyoungseo:
Took both of the first two in 8165a2c — the TOCTOU one turned out to have a fix cheaper than the doc note you suggested.

**1. TOCTOU on `use_cdn=None`.** Rather than document the window, it is closed: `resolved` is now passed to `render` instead of letting it read the shared default a second time.

```python
resolved = maidr.get_use_cdn() if use_cdn is None else use_cdn
rendered = maidr.render(plot, use_cdn=resolved)   # was: use_cdn=use_cdn
```

`_resolve_use_cdn` returns a non-`None` value unchanged, so this is behaviour-identical on the happy path and simply removes the second read. `test_the_resolved_mode_is_the_one_the_chart_was_built_with` asserts `render` never receives `None`.

The docstring caution went in as well — you are right that it belongs here for the same reason it is in `shiny.py`, and more so, since Streamlit's threading model makes it easier to hit.

**2. Regex false-positive.** Confirmed: `"notmaidr@1.js"` matched. Anchored on the path segment every real URL carries:

```
false-positive check notmaidr@1.js -> False
true-positive check npm/maidr@1.js -> True
```

Still `True` for `{matplotlib, plotly} × {True, "auto"}` and `False` for `use_cdn=False` across both, so nothing regressed. The near-misses — a package ending in the same letters, and `cdn.plot.ly` — are now rows in a parametrized table rather than assumptions.

**3. `_stacklevel` as a public kwarg.** Leaving it, and agreed it is a tradeoff rather than a clean win. The alternative — a module-level or thread-local depth — trades a visible parameter for hidden state, which is worse to reason about for the same problem. The underscore plus the docstring is the weakest signal Python offers, but it is the conventional one.

That is every round's findings addressed. `uv run pytest` — 2128 passed; `ruff check` and `uv lock --check` clean; CI green on 3.9–3.12. Merging once this run finishes.

---

jooyoungseo:
Good question, and the honest answer is that it was an oversight rather than a deliberate simplification. Fixed in ae95799.

Your reasoning for why it might have been intentional — "its import chain is shallower" — is the opposite of true. Streamlit reaches pyarrow, tornado, protobuf, altair and pydeck; Shiny reaches htmltools and shinychat. Streamlit is the *likelier* of the two to hit the case the distinction was written for, so if either module were going to skip it, it should have been the other one.

Rather than copy the eight lines across, both now call `missing_extra_error` in `maidr/widget/_extras.py`. This PR is what makes that the second caller, which is the bar `CLAUDE.md` sets for adding an abstraction. Tested for both failure shapes on the Streamlit side — a genuine `ModuleNotFoundError('streamlit')` gets the extra, and `ImportError: cannot import name 'Foo' from 'pyarrow'` gets the skew message with the original error quoted — matching the pair that already existed for Shiny.

That leaves only `_stacklevel`, which we have both now landed on as an accepted tradeoff.

`uv run pytest` — 2130 passed. Merging once CI confirms.

---

jooyoungseo:
Both notes checked. Neither needs a change, but the second was worth verifying rather than asserting — and your "every workflow uses `--all-extras`" is not quite true.

**The extras-less job exists.** `release.yml:40` runs `uv sync --locked --extra test --extra visualization` — no `streamlit` extra, so the real package is absent there. So I ran the suite in exactly that state:

```
$ .venv/bin/python -c "import streamlit"
ModuleNotFoundError: No module named 'streamlit'

$ pytest tests/widget
53 passed, 2 skipped
```

The two skips are precisely the two real-Streamlit tests (`..._embeds_the_chart_as_srcdoc` and `..._detected_on_the_real_streamlit`), both behind `importorskip`. Everything else stubs the module and runs regardless. The invariant holds, and now by measurement rather than by reading the workflows.

**On `htmltools` dropping out of the `absent` check** — your analysis is right, and it is unreachable. `maidr/widget/shiny.py` is only importable through the `maidr` package, whose `__init__` imports `htmltools` unconditionally as a hard runtime dependency, so a missing `htmltools` fails long before that except-block is reached. I left the helper keyed on the single package rather than threading a list through for a branch that cannot execute — a parameter that exists only for an unreachable case is a worse trade than the slightly-off message it would produce there.

That is every round resolved. Final state: 2130 tests passing locally, all 16 checks green on 3.9–3.12, and the eight commits each carry the defect they fix and its regression test.

Merging.

---

PR #463, opened 2026-08-17
fix(a11y): name the iframe every chart is rendered into

### LINKED ISSUE #453: The maidr iframe has no accessible name
Every iframe-wrapped render — notebook, Shiny, Flask — emits an `<iframe>` with no `title` attribute.

`maidr/util/iframe_utils.py`, both `wrap_in_iframe_matplotlib` and `wrap_in_iframe_plotly`:

```python
return tags.iframe(
    id=unique_id,
    width="100%",
    height="100%",
    scrolling="no",
    style="background-color: #fff; position: relative; border: none",
    frameBorder=0,
    onload=resizing_script,
)
```

A frame with no accessible name is a WCAG 2.2 **4.1.2 Name, Role, Value** failure. Screen readers announce it as "frame" with no further information, and a notebook or dashboard with several charts gives a reader several identical, unlabelled frames to tell apart — which is exactly the navigation problem this library exists to solve.

The chart's own title is already known at render time (`ax.get_title()`, the Plotly `layout.title`, or the MAIDR schema), so there is a good name available rather than a generic one.

**Suggested shape**

- Give the wrappers a `title` parameter and pass the chart's title through from `Maidr._create_html_tag` / `PlotlyMaidr._create_html_tag`.
- Fall back to something specific but generic (`"MAIDR accessible chart"`) when the chart is untitled — better than nothing, and distinguishable from unrelated frames on the page.
- A test asserting the emitted iframe carries a non-empty `title`.

Pre-existing rather than a regression; noticed while auditing the Shiny integration in #452, which does not touch these wrappers.

jooyoungseo:
Closes #453.

Every iframed render — notebook, Shiny, Flask — emitted an `<iframe>` with no `title`, so a screen reader announced an unnamed frame. A reader arriving at a chart was told a frame was there and nothing else: not that it held a chart, and on a page carrying several, not which one.

| | before | after |
|---|---|---|
| `ax.set_title("Body mass by species")` | *(no title attribute)* | `Body mass by species, accessible chart` |
| `fig.suptitle("Two panels")` | *(no title attribute)* | `Two panels, accessible chart` |
| no title anywhere | *(no title attribute)* | `Accessible chart` |

## Why the chart's own title, and not just a label

A generic name fixes the WCAG failure and leaves the usable one in place. A page carrying three charts all named "Accessible chart" gives a reader tabbing between them no way to tell which they have reached — so the chart's title leads, because that is the part that distinguishes one frame from the next.

The qualifier follows so the name still says what *kind* of thing the frame is: an interactive chart to enter rather than a picture to skip. It is the shape `Description.tsx` already uses in the core for its own table — a name for the chart when there is one, a bare label when there is not.

## Where the title comes from

Read back off the **emitted schema** rather than off the figure object, for two reasons:

* it is the one shape matplotlib and Plotly both produce, so the rule lives in `iframe_utils.chart_title_of` once instead of drifting between renderers — the `MAX_INTERVAL_VERTICES` lesson from #458;
* it names the frame the same thing the chart announces itself as. Reading the figure separately could let the two come apart, and a frame whose name no chart on the page says is a frame a reader cannot find.

A figure-level title wins where there is one. Failing that, **a single title shared by every layer is the figure's name too** — a one-axes figure carries its title on the layer, which is where `ax.set_title()` puts it and much the commoner spelling.

A multi-panel figure whose panels are titled differently has no one name. Naming the frame after the first panel would name it after a part of what it holds, so those fall back to the bare label.

**Altair** builds no MAIDR schema on the Python side — the JS adapter reads the Vega-Lite spec directly — so its name comes from the spec, in all three spellings Vega-Lite allows: a bare string, `{"text": …}`, and a list of lines for a multi-line title. The lines are joined rather than truncated, since the first alone may not be the name.

## Measured, not assumed

The Plotly schema keys its top level with `MaidrKey.TITLE`, the enum, not the plain string:

```
top-level keys: ['id', 'subplots', <MaidrKey.TITLE: 'title'>]
chart_title_of -> 'Revenue by quarter'
```

That reads correctly only because `MaidrKey` subclasses `str` and so hashes equal to its value. Pinned by a test, because the failure would be silent: an enum that stopped subclassing `str` would leave every Plotly frame named "Accessible chart" with nothing raising.

## Tests

23 new. `tests/util/test_iframe_accessible_name.py` (17) asserts the attribute on the **rendered HTML** rather than on the helper alone — the helper returning the right string is not the claim, the claim is that it reaches the tag — and covers both wrappers, the whitespace-is-absent rule, each schema shape, and the enum-keyed schema. `tests/altair/test_altair_maidr.py` (+6) covers the three Vega-Lite title spellings and both ends of the Altair path.

## Falsification

| reverted | failures |
|---|---|
| the `title=` attribute on both wrappers | **8** |
| the shared-layer-title fallback in `chart_title_of` | **3** |
| Altair reading its spec title | **1** |

Disjoint, and each fails only the tests written for it.

## Verification

`ruff check` clean on every file touched. Full suite:

```
2142 passed in 93.56s
```

Note for anyone reproducing: `tests/widget/conftest.py` (new on `main`) does a module-level `importorskip("shiny")`, and without the extra installed that aborts collection for the **whole** run — `collected 0 items / 1 skipped`, which reads as a passing run at a glance. `uv sync --locked --all-extras --dev`, as `CLAUDE.md` prescribes, is what makes the suite collect.

---


jooyoungseo:
CI caught a real break in this PR, fixed in `9b35f1b`.

## What broke

The Python 3.9 job failed, and not as a test failure — as a hard import break:

```
maidr/util/iframe_utils.py:21: in <module>
    def iframe_title(chart_title: str | None = None) -> str:
E   TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

`pyproject.toml` sets `requires-python = ">=3.9"`, and 3.9 evaluates annotations at definition time. I wrote `str | None` in a module with no `from __future__ import annotations`, and `maidr/core/maidr.py` imports it — so **`import maidr` raised on 3.9**. Every version from 3.10 up was unaffected, which is why the local suite was green.

Fixed by adding the future import, with a comment saying which interpreter it is there for.

Verified on a real 3.9, not inferred — executing the module directly, which is the step that used to raise:

```
python 3.9.23
module executed OK
'Body mass, accessible chart'
'Accessible chart'
'Sales'
```

and then the whole suite on that interpreter: **2249 passed, 4 skipped**.

## The guard

I added `tests/test_python_floor.py` rather than only fixing the line, because of how this failure presents. It is invisible on every version above the floor, so it reads as one red job among fifteen rather than as "the package does not import on a version we ship for" — and the next `str | None` in a fresh module lands exactly the same way.

The test walks each module's AST and requires the future import wherever an annotation writes a PEP 604 union. It covers the cases the naive version would miss:

- nested unions — `list[str | None]`, `dict[str, int | None]` — which raise just the same;
- return annotations and `x: int | None = None` variable annotations, not only parameters;
- positional-only, keyword-only, `*args` and `**kwargs`, since any one of them is enough;
- quoted annotations (`"str | None"`) are *not* reported, because a string is not evaluated at definition time.

It runs clean across all 110 modules, so this was the only instance.

The detector is itself tested against the shape that broke and against the safe spellings (`Optional[str]`, quoted, future-guarded). A check for an absent condition otherwise passes by never detecting anything, which is the failure mode that would have let this through a second time.

**Deliberately out of scope**, and noted in the module docstring: a PEP 604 union used at *runtime* rather than in an annotation — `isinstance(x, int | str)` — is equally broken on 3.9 and the future import does not save it. Nothing in the package does that today, and catching it would mean knowing which call arguments are evaluated as types.

## One practical note

`uv run --python 3.9 …` re-creates the project `.venv` at that version rather than using a throwaway one. Worth knowing before reaching for it mid-session — a following `uv sync` keeps 3.9, so the suite silently changes interpreter. I left mine on 3.9 and re-ran there, which is the stricter check anyway.

---

jooyoungseo:
All five addressed in `4e48bb0`. The ruff one was right and cost me 150 lines, which is the good outcome.

## 1. Ruff already ships this rule — walker deleted

You're right, and I checked rather than took it on trust. FA102 flags exactly the fault:

```
FA102 Missing `from __future__ import annotations`, but uses PEP 604 union
  --> maidr/util/iframe_utils.py:21:31
```

`tests/test_python_floor.py` is gone; `extend-select = ["FA102"]` is in `[tool.ruff.lint]`.

**FA102 only, not `FA`.** Selecting the set enables FA100 too, which is a different thing wearing a similar name — a style opinion about `typing.Union[str, None]`, which runs perfectly well on 3.9. It accounts for **all 26** of the other findings, every one of them working code:

```
--select FA102 .  → All checks passed!
--select FA100 .  → Found 26 errors
```

`extend-select` rather than `select`, which would have replaced ruff's defaults (`E4`, `E7`, `E9`, `F`) instead of adding to them.

Worth recording that the rule agrees with the walker: FA102 finds nothing across the package once the one real break is fixed, which is the verdict the walker reached too.

### The rule alone gates nothing, though

This is the part that would have quietly undone the protection. FA102's only fix is an *unsafe* one, and the existing step runs `check --diff`, which reports fixable diffs — so with the fault present **and the rule enabled**:

```
$ ruff check --diff maidr/util/iframe_utils.py
No errors would be fixed (3 fixes available with `--unsafe-fixes`).
exit=0
```

Green. So I added a second step running a plain `check --select FA102`, scoped to that one rule rather than switching the existing step to a plain `check` — which would fail on the 51 findings that predate it (19 `E402`, 19 `F401`, 13 `F811`).

Verified against the pinned **ruff 0.3.4**, not just my local 0.15.8: it supports FA102, accepts the `[tool.ruff.lint]` section, reports the three breaks when present, and passes clean otherwise.

## 2. Docstring parameter order — fixed

`chart_title` now sits last, after `use_cdn`, matching the signature.

## 3. Candlestick fixed titles — real, and narrower than it looks

Measured both spellings. `mpf.plot(title=…)` sets the figure **suptitle**, which lands in the figure-level title and is never overwritten, so that path is fine:

```
figure-level title: 'AAPL 2026 Q1'   →  'AAPL 2026 Q1, accessible chart'
```

The axes-level spelling is where it bites:

```python
axlist[0].set_title("AAPL 2026 Q1")
→ figure-level title: None
  layer titles      : ['Candlestick Chart']
  iframe name       : 'Candlestick Chart, accessible chart'
```

Filed as #464 rather than fixed here. This PR names the frame from whatever title the chart *already reports*; changing which title a candlestick reports alters the chart's own announcement too, for three plot types, which is a behaviour change that deserves its own change and its own tests. Your instinct that it's inherited rather than introduced is right — but it is now inherited somewhere more visible, and the issue says so.

## 4. Type inconsistency — aligned

`_inject_plot` takes `chart_title: str | None = None`, matching the wrappers it feeds.

## 5. End-to-end Plotly test — added

Agreed, and it is the case a synthetic dict genuinely cannot cover: that renderer builds its own schema, so nothing else would notice it drifting from the keys the title is read from. Two tests now build a real `plotly.express` figure and assert the resolved name, titled and untitled, mirroring `TestTheRenderedChartCarriesItsName`.

## Verification

```
2140 passed, 4 skipped
ruff check --select FA102 .  → All checks passed!
uv lock --check              → in sync
```

2249 → 2140 is the 111-test walker leaving and the 2 Plotly tests arriving.

---

PR #465, opened 2026-08-17
fix(mplfinance): keep the title the caller gave the chart

### LINKED ISSUE #464: fix: a candlestick discards the title the caller set on the axes
## Description

`CandlestickPlot.render()` overwrites whatever title the layer had with a fixed string:

```python
base_schema = super().render()
base_schema[MaidrKey.TITLE] = "Candlestick Chart"     # maidr/core/plot/candlestick.py:224
```

`mplfinance_barplot.py:163` and `mplfinance_lineplot.py:201` do the same with `"Volume Bar Plot"` and `"Moving Average Line Plot"`.

So a caller who names their chart on the axes loses that name. Measured:

```python
fig, axlist = mpf.plot(df, type="candle", returnfig=True)
axlist[0].set_title("AAPL 2026 Q1")
```

```
figure-level title: None
layer titles      : ['Candlestick Chart']
```

## The spelling that works

`mpf.plot(title=…)` sets the figure **suptitle**, which lands in the figure-level title and is never overwritten:

```python
fig, axlist = mpf.plot(df, type="candle", returnfig=True, title="AAPL 2026 Q1")
```

```
figure-level title: 'AAPL 2026 Q1'
```

So the loss is confined to the axes-level spelling — but that is the ordinary matplotlib way to title a chart, and nothing tells a caller it will be ignored here.

## Why it is worth fixing now

It has become more visible than it was. #463 names the iframe every chart renders into after the chart's own title, so this fixed string is now a chart's **accessible name** as well as its announced title:

```
'Candlestick Chart, accessible chart'
```

Three candlesticks on one page are announced identically, and a reader tabbing between them has no way to tell which they have reached — which is the failure #453 was filed about, reappearing for this one plot type through a different route.

## Suggested fix

Prefer the authored title and keep the fixed string as the fallback, rather than overwriting unconditionally:

```python
base_schema[MaidrKey.TITLE] = base_schema.get(MaidrKey.TITLE) or "Candlestick Chart"
```

Deliberately **not** done in #463: that PR names the frame from whatever title the chart already reports, and changing which title a candlestick reports alters the chart's own announcement too. That is a behaviour change for existing users of these three plot types and belongs in its own change, with its own tests.

## Scope

Three call sites, one line each, plus a test per plot type that an axes-level title survives and that an untitled chart still gets the descriptive fallback.

## Difficulty

Small.

jooyoungseo:
Closes #464. Follow-up to #463, which found this and deliberately left it.

`CandlestickPlot`, `MplfinanceBarPlot` and `MplfinanceLinePlot` each overwrote the layer's title with a fixed description of the chart type, discarding whatever the caller had set:

```python
fig, axlist = mpf.plot(prices, type="candle", returnfig=True)
axlist[0].set_title("AAPL 2026 Q1")
```

| | before | after |
|---|---|---|
| layer title | `Candlestick Chart` | `AAPL 2026 Q1` |
| accessible name | `Candlestick Chart, accessible chart` | `AAPL 2026 Q1, accessible chart` |

## Why now

It became more visible than it was. #463 names the iframe every chart renders into after the chart's own title, so this fixed string is now a chart's **accessible name** as well as its announced title. Three candlesticks on one page were announced identically, and a reader tabbing between them had no way to tell which they had reached — which is the failure #453 was filed about, reappearing for these three plot types by a different route.

## The spelling that was never broken

`mpf.plot(title=…)` sets the figure **suptitle**, which lands in the figure-level title and was never the thing being overwritten:

```
mpf.plot(..., title="AAPL 2026 Q1")   →  figure-level title: 'AAPL 2026 Q1'
```

So the loss was confined to the axes-level spelling — the ordinary matplotlib way to title a chart, with nothing telling a caller it would be ignored here. Both spellings are covered by tests now, so they are pinned as agreeing rather than assumed to.

## The change

Three call sites, one expression each: the caller's title wins, the description stays as the fallback.

```python
base_schema[MaidrKey.TITLE] = (
    str(base_schema.get(MaidrKey.TITLE, "") or "").strip()
    or "Candlestick Chart"
)
```

The fallback is kept deliberately rather than dropped. It tells a reader who was given no name what kind of chart they are on, which is better than the empty string the base render leaves for an untitled layer. Whitespace counts as absent, matching the trimmed "authored" check used everywhere else — otherwise `set_title("   ")` would name the frame `"   , accessible chart"`.

## Tests

`tests/core/plot/test_mplfinance_titles.py`, 10 tests in three groups:

- **the caller's title survives** — on the candlestick, the volume panel and the moving-average line; that the frame is named after it; and that two charts titled differently get **different** names, which is the property the fix exists for;
- **the descriptive label is still the fallback** — for each of the three plot types, and for a whitespace-only title;
- **the figure-level spelling is unchanged** — `mpf.plot(title=…)` still resolves as it did.

The moving-average case also asserts `"Moving Average Line Plot"` is *absent* once the axes is titled, since both layers on that axes take the axes' title and neither should be left announcing the fixed string beside a named sibling.

## Falsification

Restoring the unconditional overwrite at all three sites fails **5** tests — every case in `TestTheCallersTitleSurvives` — and leaves the five fallback and figure-level cases passing, which is what says the fallback was preserved rather than replaced.

## Verification

```
2181 passed, 6 skipped
```

`ruff check` clean on every file touched (the six repo-wide findings are the pre-existing `F401` re-exports in `maidr/core/*/__init__.py`).

---


jooyoungseo:
Two acted on in `a3e1649`, two left with reasons.

## The `axlist[2]` comment — you were right to query it, and the comment was wrong

The index is correct; my comment describing it was not. Measured rather than reasoned:

```
len(axlist) = 4
  axlist[0]: ylabel='Price'
  axlist[1]: ylabel=''
  axlist[2]: ylabel='Volume'
  axlist[3]: ylabel=''
```

So the layout is exactly the `[main, main_twin, volume, volume_twin]` you guessed, and index 2 is the second *panel* — not "the second axes mplfinance returns", which is what I wrote and is plainly false of a four-element list. Corrected, with the measurement recorded so the next reader does not have to re-derive it.

Worth saying the test was not trivially passing either way: with the fix reverted it fails, and the falsification run in the PR body shows it among the five.

## The dead `set_title` — removed

Confirmed dead: `self._title` is only ever assigned, and nothing in the package calls the setter.

I took it out rather than leaving it to a follow-up, for the reason you gave — it sits in the file this change is about, and a dead title setter there reads as the mechanism a title fix should have used. Three lines, provably unreachable, on the same subject as the change.

## The duplication — left, deliberately

Three call sites, one expression each, and `CLAUDE.md` asks for no abstraction before a second caller genuinely needs it. A `MaidrPlot._authored_title(base_schema, fallback)` helper would be right the moment a fourth of these appears; today it would move the logic away from the only three places that explain why it exists. If the helper does arrive, the three comments are what should collapse into it.

## The defensive `str(...)` — left, and here is the case for it

You're right that `ax.get_title()` returns `str` today, so both guards are unreachable through the current base class. I've kept them because the value is read out of `base_schema`, not off the axes — it is whatever `MaidrPlot.render()` last put there, and the failure if that ever changes is silent: a non-string title would flow into the accessible name and be announced as whatever `f"{...}"` makes of it. The cost is one call per render.

## Verification

```
2181 passed, 6 skipped
```

`ruff check` clean on both files touched.

---

Issue #461, opened 2026-08-17
Report two upstream Streamlit accessibility issues


jooyoungseo:
Two Streamlit behaviours block accessible embedding, both concretely evidenced while adding the Streamlit integration in #459. py-maidr has standing to report them.

**1. `st.iframe` hardcodes the frame's accessible name.**

`streamlit/static/static/js/IFrame.*.js` renders the element with a literal `title: "st.iframe"`, and `inspect.signature(st.iframe)` is `(src, *, width, height, tab_index)` — no way to override it. Every embedded chart on a page therefore announces with the same meaningless name, so a screen-reader user tabbing a dashboard cannot tell one chart from another. That is a WCAG 2.2 **4.1.2 Name, Role, Value** problem, and it is a regression against the older `declare_component` path, which used the component's name as the title.

Ask: an overridable `title` on `st.iframe`.

Until it exists, the workaround is to render an `st.caption` or a visually-hidden heading next to each chart so there is *something* distinguishing adjacent to it. Note this is also why `render_maidr(tab_index=...)` defaults to `None`: forcing `tab_index=0` adds a tab stop on the frame itself, and that stop announces as "st.iframe" on every chart.

Related: py-maidr's own iframes have the same gap, tracked separately as #453.

**2. Streamlit binds `r`, `c` and `esc` at the document level with no opt-out.**

The root app installs document-level handlers for those keys, with a filter that exempts only `INPUT` / `SELECT` / `TEXTAREA` / `contentEditable`. maidr binds all three (replay, braille, dismiss). Two listeners on one document cannot both win, and `preventDefault` in one does not stop the other's independent listener.

This blocks any keyboard-driven visualization library from adopting Streamlit's *recommended* component API: components v2 renders with no iframe, which is precisely what would expose maidr's bindings to the collision. py-maidr therefore stays on the iframe path deliberately — see the module docstring in `maidr/widget/streamlit.py` — which in turn causes the state loss in #460.

Ask: a documented opt-out from the app-level key handling for custom components.

---

This issue tracks *filing* these upstream and linking the reports back here. Both need reduced reproductions against a current Streamlit before submission.

Follow-up from #459.

jooyoungseo:
This issue says both claims "need reduced reproductions against a current Streamlit before submission." Both now have one, run in Chromium against **Streamlit 1.61.1**. One claim holds exactly as written. The other is **partly wrong**, and the part that is wrong is the part that decides #460.

The app is four lines:

```python
[...]

PR #466, opened 2026-08-17
fix(a11y): give every chart iframe an accessible name

### LINKED ISSUE #453: The maidr iframe has no accessible name

jooyoungseo:
## Description

Closes #453.

Every iframe maidr emits — notebook, Shiny, Flask — was built without a `title`:

```python
return tags.iframe(
    id=unique_id,
    width="100%", height="100%", scrolling="no",
    style="...", frameBorder=0, onload=resizing_script,
)
```

A frame with no accessible name is announced as "frame" and nothing more. A notebook or dashboard holding several charts therefore gave a screen-reader user several identical, indistinguishable landmarks to tell apart — a WCAG 2.2 **4.1.2 Name, Role, Value** failure, and the navigation problem this library exists to solve, reproduced in its own output.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues

Closes #453. Found while auditing the Shiny integration in #452; filed then rather than folded in, since it affects every renderer and not just Shiny.

## Changes Made

**`maidr/util/iframe_utils.py`** — both wrappers take an optional `title` and always emit one. `iframe_title()` returns the chart's own title, or `"MAIDR accessible chart"` when there isn't one. Whitespace-only counts as unauthored, matching how the schema already decides the same question.

**The name was available all along.** Each renderer already reads a figure title for its schema, so the same value now names the frame:

| Renderer | Source |
|---|---|
| matplotlib | `fig.get_suptitle()`, else the axes title when there is exactly one axes |
| Plotly | `layout.title.text`, read directly rather than through `to_dict()` |
| Altair | the chart title — read defensively, since Altair accepts a bare string *or* a `TitleParams`, and stringifying the latter would announce a repr |

**An axes title is used only when the figure has one axes.** On a multi-panel figure, naming the frame after one panel would announce a part as though it were the whole, so a generic name is the more honest answer there.

Measured across every renderer:

```
mpl titled       : 'Tips by day'
mpl untitled     : 'MAIDR accessible chart'
mpl suptitle wins: 'Whole figure'
plotly titled    : 'Plotly titled'
altair titled    : 'Altair titled'
altair untitled  : 'MAIDR accessible chart'
```

### Backwards compatibility

Additive. `title` is optional on both wrappers, so any existing caller keeps working and simply gets the generic name. The only change to emitted HTML is the new attribute.

## Checklist

- [ ] I have updated the documentation, if applicable — no user-facing API changed.

## Additional Notes

**Verification.** `uv run pytest` — 2168 passed (from 2152; 16 new). `ruff check` and `uv lock --check` clean.

The tests were checked against the defect rather than only against the fix: removing `title=` from the two wrappers fails **12 of the 16**, including the parametrized guarantee that no `use_cdn` mode may emit an unnamed frame.

**What this does not do.** Streamlit embeds are unaffected — Streamlit builds its own iframe and hardcodes its accessible name as `st.iframe` with no override, which is tracked separately as an upstream report in #461. This PR covers the frames maidr builds itself.

**Not verified in a browser.** No screen-reader pass was run. The claim here is about the emitted attribute, not about how any particular AT announces it.

---

### DIFF
--- a/maidr/altair/altair_maidr.py
+++ b/maidr/altair/altair_maidr.py
@@ -38,7 +38,7 @@ import tempfile
 import uuid
 import webbrowser
 from pathlib import Path
-from typing import Any, Literal, cast
+from typing import Any, Literal, Optional, cast
 
 from htmltools import HTML, HTMLDocument, Tag, tags
 
@@ -278,7 +278,26 @@ class AltairMaidr:
         with delayed retries (500/1500/3000 ms) plus a ResizeObserver
         (MutationObserver fallback) on ``document.body``.
         """
-        return wrap_in_iframe_plotly(inner)
+        return wrap_in_iframe_plotly(inner, self._chart_title())
+
+    def _chart_title(self) -> Optional[str]:
+        """Return a name for this chart, for the iframe's accessible name.
+
+        Altair allows a title to be a plain string or a ``TitleParams``
+        carrying one, and ``Undefined`` when unset -- so the text is reached
+        defensively rather than assumed.
+
+        Returns
+        -------
+        str or None
+            The title, or ``None`` when the chart has no usable one.
+        """
+        title = getattr(self._chart, "title", None)
+        if not isinstance(title, str):
+            title = getattr(title, "text", None)
+        if not isinstance(title, str):
+            return None
+        return title.strip() or None
 
     def _open_in_browser(self) -> None:
         """Save to a temp HTML file and open it in the default browser."""
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -390,7 +390,9 @@ class Maidr:
             maidr = f"\nvar maidr = {json.dumps(schema, indent=2)}\n"
 
         # Inject plot's svg and MAIDR structure into html tag.
-        return Maidr._inject_plot(svg, maidr, self.maidr_id, use_iframe, use_cdn)
+        return Maidr._inject_plot(
+            svg, maidr, self.maidr_id, use_iframe, use_cdn, self._chart_title()
+        )
 
     def _create_html_doc(
         self,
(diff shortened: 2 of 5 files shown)


jooyoungseo:
Correct on all counts — closing this rather than resolving the conflicts. #463 landed the same fix an hour before I opened this, and I verified it is a strict superset rather than taking that on faith.

**The multi-line Altair gap is real.** Confirmed against my own branch:

```
title obj: TitleParams
title.text: ['Line one', 'Line two']  (list)
my branch yields: None
```

`title.text` is a `list`, my `isinstance(title, str)` check rejects it, and a chart with a real title silently gets the generic name. Merging this would have been a regression against what is already on `main`.

**Measured `main`'s output across every case I had tested here:**

```
mpl titled       : 'Tips by day, accessible chart'
mpl untitled     : 'Accessible chart'
mpl suptitle     : 'Whole figure, accessible chart'
plotly titled    : 'Plotly titled, accessible chart'
altair titled    : 'Altair titled, accessible chart'
altair multiline : 'Line one Line two, accessible chart'
altair untitled  : 'Accessible chart'
```

Seven for seven, including the one mine drops.

**And the schema-based approach is the better design.** Reading the title back off the emitted schema means one implementation instead of three, and — more to the point — it names the frame with whatever the *schema* says the chart is called. My per-renderer version read `fig.get_suptitle()` and `layout.title.text` off the objects directly, which is a second source of truth that can drift from the first. #465, merged just after, is exactly that class of bug in the schema's own title extraction; a renderer reading around the schema would not have picked up that fix.

No cherry-pick needed. The naming convention on `main` (`"..., accessible chart"`) is also the better of the two: it names the chart *and* says what kind of thing the frame is, which my bare-title version left to the reader to infer.

My mistake was not re-checking `main` between filing #453 and starting work on it. Closing.

---

Issue #455, opened 2026-08-17
use_cdn="auto" has no working offline fallback inside a srcdoc iframe


jooyoungseo:
`use_cdn="auto"` is documented as "try the CDN, fall back to the bundled copy client-side if it is unreachable". Inside a `srcdoc` iframe — which is what a notebook, Shiny, or Flask render produces — the fallback cannot resolve, so `"auto"` silently degrades to CDN-only.

Two variants, both dead outside their intended context:

- **Notebook branch** (`maidr/core/maidr.py`, `iframe_in_notebook`): `onerror` calls `fallbackFromParent`, reading `window.parent.__maidrJsSource`. Only `init_notebook()` populates that, and it returns early outside a notebook — so this works in a notebook and nowhere else.
- **Non-iframe branch**: `onerror` sets `fb.src = 'lib/maidr-<version>/maidr.js'`. The `HTMLDependency` that would have put that file there is dropped by `Tag.get_html_string()` when the iframe wrapper serialises the tag, so the file is never served and the fallback 404s.

#452 fixed the equivalent hole for `use_cdn=False` by inlining the bundle, and deliberately left `"auto"` alone: inlining unconditionally would defeat the point of `"auto"`, which is to stay small when the network is available. So `"auto"` is correct whenever the CDN is reachable and has no offline path when it is not.

**Consequence.** An air-gapped Shiny or Flask deployment on the default setting gets a chart with no MAIDR runtime — silently. `use_cdn=False` is the setting that works, and that is currently only discoverable by reading the code.

**Options**

1. Document it, and have the `onerror` handler log something actionable instead of failing silently. (Cheapest; worth doing regardless.)
2. Serve the bundle from the host and reference it by absolute URL, so a `srcdoc` document can reach it. Under Shiny this means registering an `HTMLDependency` on the *container* rather than the rendered value, or a `session.dynamic_route`.
3. Detect reachability once per process and pick a concrete mode, rather than deferring to the browser.

Also worth checking as part of this: the `lib/maidr-<version>/maidr.js` path is relative, so even where the dependency *is* served it may not resolve under a sub-path deployment (Posit Connect, shinyapps.io behind a prefix).

Follow-up from #452.

PR #467, opened 2026-08-17
fix(plotly): report when a plotly chart runs out of maidr.js sources

jooyoungseo:
## Description

Rebased onto `main` and cut down to the half that #468 did not cover.

#468 gave the **matplotlib** loader a `reportNoRuntime` for the case where neither the CDN nor the bundled fallback loads. The **Plotly** loader has the same shape and the same failure, and is still uncovered on `main`:

```js
s.onerror = function() {
    var fb = document.createElement('script');
    fb.src = '{bundled_js_rel}';
    document.head.appendChild(fb);      // no onerror at all
};
```

So a Plotly chart in a Shiny or Flask iframe with no network becomes an image with no MAIDR runtime — no sonification, no braille, no keyboard navigation — and says nothing. Exactly what #468 fixed, one file over.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues

Partially addresses #455, alongside #468.

## Changes Made

- **`maidr/util/dependencies.py`** — `_OFFLINE_FALLBACK_REPORT` moves here from `maidr/core/maidr.py` as `OFFLINE_FALLBACK_REPORT`. Both renderers already import from this module; leaving it beside one of them would mean the Plotly path importing from `maidr.core` for a JS string, or a second wording.
- **`maidr/plotly/plotly_maidr.py`** — defines the reporter in its loader and calls it from a new `fb.onerror`.
- **`maidr/core/maidr.py`** — import site only; the emitted script is byte-identical.

Verified both renderers now define the helper, call it, and say the same thing.

### What changed since the first version of this PR

The original also patched `maidr/core/maidr.py` with a separately-worded `NO_RUNTIME_CONSOLE_ERROR`. #468 landed the same fix there ~5 hours before this PR was opened, and I had not re-checked `main` between filing #455 and starting. Merging that would have put **two differently-worded diagnostics** on one failure — the drift the second commit claimed to be preventing, reintroduced under another name. Both the matplotlib hunk and the second constant are gone.

## Checklist

- [ ] I have updated the documentation, if applicable — no user-facing API changed.

## Additional Notes

**Verification.** `uv run pytest` — 2198 passed (from 2194; 4 new). `ruff check` and `uv lock --check` clean.

Tests extend `tests/core/test_offline_fallback_report.py` rather than adding a parallel file, so the two renderers' coverage sits together. Removing the new `fb.onerror` fails `test_the_plotly_fallback_script_reports_on_error`, and `test_both_renderers_say_the_same_thing` asserts the shared constant's wording reaches *both* emitted documents — the drift check neither file could make on its own.

**Not verified in a browser.** The assertions are about emitted script text. That a 404 on a `<script>` fires `onerror` is the one behavioural assumption underneath this that nothing here proves.

### DIFF
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -26,13 +26,13 @@ from maidr.core.plot.barplot import BarPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
 from maidr.util.dependencies import (
     MAIDR_JS_FILENAME,
+    OFFLINE_FALLBACK_REPORT,
     bundled_cdn_url,
     inline_bundle_tags,
     maidr_bundled_files_dependency,
     maidr_bundled_relative_dir,
     maidr_html_dependency,
     maidr_js_cdn_url,
-    OFFLINE_FALLBACK_REPORT,
     schema_trace_types,
     warn_if_bundle_cannot_render,
     warn_if_bundle_is_stale,
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -32,6 +32,7 @@ from maidr.util.dependencies import (
     maidr_bundled_relative_dir,
     maidr_html_dependency,
     maidr_js_cdn_url,
+    OFFLINE_FALLBACK_REPORT,
     schema_trace_types,
     warn_if_bundle_cannot_render,
     warn_if_bundle_is_stale,
@@ -67,23 +68,6 @@ _SEGMENTED_BAR_PLOTS = (GroupedBarPlot,)
 #: never set ``onerror`` at all. What the reader gets either way is a chart
 #: with no MAIDR runtime -- a picture, with nothing saying why (#455).
 #:
-#: Names the setting that works rather than describing the failure, because
-#: the person who hits this is on an air-gapped deployment and the answer
-#: (``use_cdn=False``) is otherwise only discoverable by reading the source.
-#:
-#: A plain string rather than an f-string: it is interpolated *into* f-strings,
-#: so its braces must not be doubled.
-_OFFLINE_FALLBACK_REPORT = """
-                        function reportNoRuntime(why) {
-                            console.error(
-                                '[maidr] The chart loaded but its runtime did not: '
-                                + why + '. The CDN was unreachable and the bundled '
-                                + 'copy could not be resolved from inside this frame. '
-                                + 'Re-render with use_cdn=False to inline the bundle, '
-                                + 'which works without network access.'
-                            );
-                        }
-"""
 
 
 class Maidr:
@@ -1012,7 +996,7 @@ class Maidr:
                                 if (window.main) window.main();
                             }}
                         }}
-{_OFFLINE_FALLBACK_REPORT}
+{OFFLINE_FALLBACK_REPORT}
                         function fallbackFromParent() {{
                             try {{
                                 var jsSrc = window.parent && window.parent.__maidrJsSource;
@@ -1064,7 +1048,7 @@ class Maidr:
                 rel_dir = maidr_bundled_relative_dir()
                 bundled_js_rel = f"{rel_dir}/{MAIDR_JS_FILENAME}"
                 fallback_script = f"""
-                    (function() {{{_OFFLINE_FALLBACK_REPORT}
+                    (function() {{{OFFLINE_FALLBACK_REPORT}
                         function bootstrap() {{
                             if (document.readyState === 'loading') {{
                                 document.addEventListener('DOMContentLoaded', function() {{ if (window.main) window.main(); }});
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -61,6 +61,7 @@ _AXES_WIDE_BAR_PLOTS = (BarPlot, GroupedBarPlot)
 #: the layer to keep when one axes holds more than one of the family.
 _SEGMENTED_BAR_PLOTS = (GroupedBarPlot,)
 
+
 class Maidr:
     """
     A class to handle the rendering and interaction
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -61,15 +61,6 @@ _AXES_WIDE_BAR_PLOTS = (BarPlot, GroupedBarPlot)
 #: the layer to keep when one axes holds more than one of the family.
 _SEGMENTED_BAR_PLOTS = (GroupedBarPlot,)
 
-#: What the browser says when ``use_cdn="auto"`` runs out of sources.
-#:
-#: Both fallbacks can fail to resolve, and both used to do it in silence: the
-#: notebook one only acts ``if (jsSrc)`` and swallows the miss, and the other
-#: never set ``onerror`` at all. What the reader gets either way is a chart
-#: with no MAIDR runtime -- a picture, with nothing saying why (#455).
-#:
-
-
 class Maidr:
     """
     A class to handle the rendering and interaction
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -1111,6 +1111,12 @@ class PlotlyMaidr:
             CDN was simply unreachable and the fix is ``use_cdn=False``.
             Sharing one message would send half the callers somewhere useless.
 
+            Built by placeholder replacement rather than as an f-string,
+            unlike everything else in this file: the JS body is full of
+            literal braces, and an f-string would need every one of them
+            doubled -- which is how a template like this acquires a
+            mismatched brace that only shows up as broken JS in a browser.
+
             Parameters
             ----------
             on_missing : str
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -1100,7 +1100,7 @@ class PlotlyMaidr:
         # KaTeX travels as a string because ``maidr.js`` resolves
         # ``maidr-math.css`` against the URL it was loaded from, and an
         # inline script inside a srcdoc iframe has no URL to offer it.
-        def parent_source(on_missing: str) -> str:
+        def parent_source(on_missing: str, on_unreachable: str = "") -> str:
             """Return the parent-window loader, reporting failure as told.
 
             The two ``use_cdn`` modes reach this for different reasons and so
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Both taken in 7ca1686.

**De-duplicated.** You're right, and it's the second time this pattern has come up — the Shiny/Streamlit `ImportError` messages had the same shape and got the same treatment in #459. One wording in two files is one wording that can drift, and here the message *is* the change, so drift would be the whole defect coming back halfway.

`NO_RUNTIME_CONSOLE_ERROR` now lives in `maidr/util/dependencies.py`, beside the other bundle and CDN concerns both renderers already import from — rather than `environment.py`, which answers "where am I running", not "what do we ship". `test_both_renderers_use_one_diagnostic` asserts the constant's wording reaches both emitted documents, so a future copy-paste divergence fails rather than passes.

**The coupled assertion is gone.** Fair catch — `"+ fb.src +"` tested the concatenation that builds the message, not the message. It now slices from `"could not be loaded from"` and checks the URL is interpolated within it, so rewriting the snippet as a template literal changes nothing about the test. The docstring says why, since the obvious "simplification" is to put the substring assertion back.

`uv run pytest` — 2194 passed (one more than before; the new shared-wording test). `ruff check` and `uv lock --check` clean.

On the residual gap you flagged: agreed, and it's the same limit as the rest of this series — the tests assert on emitted script text, not on `onerror` firing in a real DOM. Worth being plain that "a 404 on a `<script>` fires `onerror`" is the one behavioural assumption underneath this change that nothing here proves.

---

jooyoungseo:
Right on every point. Rebased onto `main` and cut to the half that is actually missing — 51be121.

**Verified #468 independently before acting.** `main` has `_OFFLINE_FALLBACK_REPORT` / `reportNoRuntime` wired into both matplotlib branches, and:

```
$ git show origin/main:maidr/plotly/plotly_maidr.py | grep -n "fb.src\|fb.onerror\|reportNoRuntime"
1179:                            fb.src = '{bundled_js_rel}';
```

One line, no handler. So your read is exactly right: the matplotlib hunk was redundant, the Plotly gap is real.

**Your suggested resolution is what landed.** The helper moved to `maidr/util/dependencies.py` — the home you named and the one I'd already picked for the constant — and Plotly calls the same `reportNoRuntime` rather than carrying its own wording. `maidr/core/maidr.py` changes at the import site only; its emitted script is byte-identical. `NO_RUNTIME_CONSOLE_ERROR` is gone.

The alternative you flagged is worth spelling out because it was the trap: resolving the conflict naively would have left two `fb.onerror` assignments on one script object, the second silently winning, and #468's wording replaced by mine on that path. A green merge, a lost fix.

**Tests extend `tests/core/test_offline_fallback_report.py`** instead of adding a parallel file, so both renderers' coverage sits together and `test_auto_fallback_diagnostics.py` is gone. `test_both_renderers_say_the_same_thing` asserts the shared constant's wording reaches both emitted documents — the drift check that neither file can make alone, and the one that would have caught the two-wordings outcome.

On the import-ordering nit: the constant is now grouped with the other names from that module rather than sorted into the lowercase run.

Worth noting for anyone reading later: this is the second PR in this series superseded mid-flight (#466 by #463, this one's matplotlib half by #468). The lesson I'm taking is to re-check `main` immediately before starting each piece, not just when filing the issue.

`uv run pytest` — 2198 passed.

---

jooyoungseo:
All three correct, and the first one is the good catch — it falsified this PR's own headline claim. Fixed in 2ba4158.

**1. The notebook branch was untouched.** Right: Plotly has the two loader paths matplotlib has, and I wired only the relative-path one. A Plotly chart in a real Jupyter session with no network stayed a picture with nothing saying why — the exact failure the PR says it removes, on the branch I did not look at.

**2. And the tests could not have caught it,** for the reason you give: forcing `is_shiny` leaves `is_notebook` false, so all three ran the already-fixed branch. They now parametrize over both, mirroring `TestEveryAutoPathCanReport` beside them. Reverting just the notebook wiring fails exactly the two `[True]` cases:

```
FAILED ...::test_every_plotly_auto_path_defines_the_reporter[True]
FAILED ...::test_every_plotly_auto_path_reports[True]
```

**A trap worth recording, since the obvious fix has one.** Wiring the notebook branch first required splitting the two modes apart. Plotly built one `parent_source_snippet` and shared it between `use_cdn=False` and `"auto"` — but the same miss means different things:

- `use_cdn=False`: the caller asked for the bundle. The fix is `init_notebook()`.
- `"auto"`: the CDN was merely unreachable. The fix is `use_cdn=False`.

My first attempt replaced the shared snippet's warning outright, which would have told everyone who had *already* passed `use_cdn=False` to pass `use_cdn=False`. Checking `main` showed matplotlib deliberately keeps the `init_notebook()` wording on that path and uses `reportNoRuntime` only under `"auto"` — so Plotly's snippet is now a builder taking the failure text, and both modes keep their own advice. `test_use_cdn_false_still_advises_init_notebook` pins it. The full matrix:

```
notebook=True  use_cdn=auto  reports=True   init_notebook_advice=False
notebook=True  use_cdn=False reports=False  init_notebook_advice=True
notebook=False use_cdn=auto  reports=True   init_notebook_advice=False
notebook=False use_cdn=False reports=False  init_notebook_advice=False
```

**3. The orphaned comment was my bad slice** when moving the constant — I cut from the assignment rather than from the top of its comment block. Removed, and its content, including the `#455` context you flagged as lost, now sits with the constant in its new home rather than being dropped.

The import grouping nit is addressed too.

`uv run pytest` — 2202 passed.

---

jooyoungseo:
The `NameError` doesn't occur — but you were right that the code is wrong, just not in the way reported. Fixed in dbaf78e.

**It passes:**

```
$ pytest "...::TestThePlotlyRenderReportsToo::test_both_renderers_say_the_same_thing"
1 passed
```

There is a module-level `_rendered` at line 43, as well as the staticmethod at 126. A bare `_rendered(...)` inside a method resolves to the module-level one — class scope is not in the name lookup chain for code inside methods — so the two lines are deliberately different: `self._rendered(...)` renders **Plotly**, the bare `_rendered(...)` renders **matplotlib**. That is the assertion. Deleting the last line as a duplicate would have removed the only half of the test that touches the other renderer, and the test would still have passed while checking nothing it was written for.

The "2202 passed" claim holds; I re-ran it on the current head.

**But the readability point underneath is right, and it's the more important one.** You read the code carefully and drew the wrong conclusion, which is the strongest possible evidence that two same-named helpers are indefensible here — the next person will "fix" it into a duplicate of the line above, and the suite will stay green. Renamed to `_plotly_rendered`, so the two renderers are told apart by their names rather than by a scoping rule, with a comment on the assertion saying which is which.

That is a better outcome than the reported bug would have been, so thank you for flagging it even though the diagnosis missed.

**Import grouping** done as well — `OFFLINE_FALLBACK_REPORT` now sits with `MAIDR_JS_FILENAME` at the top of both blocks rather than sorted among the lowercase names.

`uv run pytest` — 2202 passed. `ruff check` and `uv lock --check` clean.

---

jooyoungseo:
Both taken in 5da55fc.

**The vestigial indentation is gone.** It was ~28 spaces carried from when the constant lived nested inside `_inject_plot` — meaningless once two files read it at two different depths. Re-verified all four paths after the trim, since "only whitespace" is exactly the kind of claim that turns out to matter in a language where a stray edit can close a string early:

```
notebook=True  mpl    defines=True calls=True
notebook=True  plotly defines=True calls=True
notebook=False mpl    defines=True calls=True
notebook=False plotly defines=True calls=True
```

**The `srcdoc` fallback was the better catch, and I'd call it more than a nit.** You're right that it degrades silently: both environments `_plotly_rendered` drives produce an iframe, so a regex miss can only mean the attribute order moved — and the fallback would then have asserted against the *outer* HTML while still passing. A test that quietly starts checking a different document is worse than one that fails, because nothing tells you the coverage left. It asserts now, with a comment on why falling back is not an option here rather than a convenience.

Thanks for tracing the JS scoping by hand. That was the thing I was least able to prove from Python — `reportNoRuntime` is declared in the `loader` string and called from two closures nested inside it, and the tests can only show both substrings are present, not that one is in the other's scope. An independent read of it is worth more here than another assertion.

Nothing outstanding from any round. CI green on 3.9–3.12; `uv run pytest` — 2202 passed. Merging.

---

jooyoungseo:
Both taken in d011cc7 — the second one deserved more than "nit", since it was the last place this PR's parity claim wasn't actually true.

**Cross-origin vs. empty stash.** You're right: `parent_source` swallowed the exception and fell through to the same `on_missing` string, so under `"auto"` an unreachable parent reported "the notebook page has no stashed copy". Both end the same way, but they send the reader somewhere different — one says "call `init_notebook()`", the other says "your frame can't see its parent" — and naming the wrong one is worse than a generic message, because it's confidently wrong.

`parent_source` now takes `on_unreachable`, defaulting to `on_missing`. The default is right exactly where the two share an answer: under `use_cdn=False` the fix is `init_notebook()` whichever happened, so that path keeps one message. Under `"auto"` they diverge, matching how `fallbackFromParent` has always split them:

```
use_cdn=auto  no_stash=True   unreachable=True   init_advice=False
use_cdn=False no_stash=False  unreachable=False  init_advice=True
```

`test_the_two_notebook_failures_are_told_apart` pins it.

**Blank line restored** — good catch, that was collateral from slicing the constant out from directly above `class Maidr`.

On the `srcdoc` regex coupling: agreed it's a trade rather than a solved problem. Asserting on a miss means a change in attribute order fails loudly and gets a one-line fix, instead of the tests quietly checking the outer HTML — which is the version I'd rather be wrong in.

`uv run pytest` — 2203 passed. Merging once this run is green; that's five review rounds with nothing outstanding.

---

jooyoungseo:
The sentinel point is a good one and is fixed; the blank line was already done a commit before this review.

**Blank lines** — landed in d011cc7, one commit ahead of what this round read:

```
$ git show HEAD:maidr/core/maidr.py | grep -n -B3 "^class Maidr:"
62-_SEGMENTED_BAR_PLOTS = (GroupedBarPlot,)
63-
64-
65:class Maidr:
```

Two blank lines. Flagged in the previous round and fixed then.

**`on_unreachable=""` → `None`.** Agreed, and worth doing precisely because nothing exercises it yet. `""` with `on_unreachable or on_missing` made "not given" and "deliberately empty" indistinguishable, so a future caller wanting the unreachable branch to stay quiet would have silently got the missing branch's message — the same class of confusion as the two-wordings problem this PR exists to remove, one layer down. `None` keeps them separable, and the docstring now says why it is `None` rather than `""` so the "simplification" back to a falsy check doesn't look free.

Thanks for tracing `reportNoRuntime`'s hoisting through all three branches. That the declaration is emitted before the `onerror` that closes over it is the one property nothing on the Python side can assert, so an independent read of it is the only check there is.

`uv run pytest` — 2203 passed. Nothing outstanding across six rounds; merging on green.

---

Issue #454, opened 2026-08-17
maidr.render() blocks the asyncio event loop in a Shiny app


jooyoungseo:
`render_maidr.render()` calls `maidr.render(value)` synchronously, on the event loop. SVG export (the matplotlib writer plus lxml) is CPU-bound, so while it runs, every other session on that worker is stalled.

Measured on a trivial three-bar chart: an `asyncio.sleep(0.001)` ticker recorded **0 ticks during 600 ms of continuous rendering**, and steady-state render time was ~43 ms per flush (~35 ms with `use_cdn=False`, which also serialises the bundle). A large scatter or heatmap will be substantially worse, and the cost is paid per session per input change.

`docs/index.qmd:304` already has a "Shiny and other async servers" callout, but it is only about the CDN version lookup — it reads as though that was the only thing that ever blocked the loop. It should say plainly that the render itself does.

**Why this was not fixed in #452**

The obvious fix — `await asyncio.to_thread(maidr.render, value)` — touches a matplotlib `Figure` from a worker thread, and matplotlib/pyplot are not thread-safe. This is exactly why `shiny.render.plot` branches on `self.fn.is_async()`. It needs a design decision and benchmarks, not a one-line change.

**Options worth evaluating**

1. `asyncio.to_thread` with a process-wide render lock, after establishing which parts of the pipeline actually touch the figure.
2. `session.dynamic_route` to serve the SVG out of band, instead of pushing ~25 KB (or ~2 MB offline) through the websocket on every flush.
3. Accept it and document it accurately.

Whichever lands, update the callout at `docs/index.qmd:304`.

Follow-up from #452.

jooyoungseo:
Correcting a number in the issue body: **"0 ticks during 600 ms" is wrong.** I re-measured before writing it into the docs, and it does not reproduce in steady state.

```
idle control:          594 ticks in 0.6s
11 renders in 0.63s -> 57 ms each
ticks while rendering:  86
```

So the loop is **starved, not frozen** — it runs between renders, just at roughly a seventh of its idle rate. The original figure was almost certainly taken across a cold first render, which costs ~1.8 s on its own (imports, font cache, CDN version resolution) and swamps everything after it. My warm-up loop is what separates the two.

[...]

PR #471, opened 2026-08-17
docs: say that the render blocks the loop, and that "auto" has no offline path

jooyoungseo:
## Description

Two pieces of guidance in `docs/index.qmd` that did not match what the code does, plus an opt-in benchmark so the number in the first one stays honest.

### 1. The async callout implies nothing blocks the loop

It is headed **"No lookup happens on an event loop"** — true of the version *lookup*, and easy to read as "the loop is fine". `maidr.render()` is synchronous and CPU-bound (matplotlib's SVG writer, then lxml), and it never awaits, so **nothing preempts it: for as long as one render runs, the loop does not run at all.**

```
longest loop gap: 1.4 ms idle, 62.1 ms around one render
```

One render on a trivial two-bar chart holds the loop for its full duration, against ~1.5 ms of ordinary scheduler jitter. Every other session on that worker waits the whole time, per render, per input change.

### 2. The offline section doesn't mention that `"auto"` can't fall back in an iframe

Notebook, Shiny and Flask renders are wrapped in a `srcdoc` iframe, whose fallback is a relative path that resolves only where the host serves the copied bundle — which nobody does for a `srcdoc` document. So on the **default** setting an air-gapped deployment gets the CDN or nothing.

The paragraph states this as intended behaviour rather than a pending fix: inlining unconditionally is what `use_cdn=False` is for, and doing it in `"auto"` would cost every online reader ~1.9 MB to serve the offline one. Since #467/#468 the failure at least reports itself in the console.

### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Related Issues

Documents #454. #455 is already closed (by #468, which did its reporting half); the offline paragraph deliberately links no issue, since the behaviour it describes is by design.

## Changes Made

- `docs/index.qmd` — two paragraphs.
- `tests/core/test_render_is_synchronous.py` — new, skipped unless `--run-benchmark`.
- `tests/conftest.py` — the `--run-benchmark` flag and `benchmark` marker.

No production code changed.

## Checklist

## Additional Notes

`uv run pytest` — 2203 passed, 1 skipped (the benchmark). `ruff check` and `uv lock --check` clean.

**The benchmark was wrong twice before it was right, and that is the argument for having it.**

Its first version used a busy-wait as the "idle" control, which starved the loop exactly as rendering does — a control that could not fail. Caught by its own assertion, which had been written first.

Its second version started the ticker *before* the settling sleep and counted those wakeups as though they had occurred during rendering. It reported a healthy-looking "86 wakeups, starved not frozen", and would have reported the same number had rendering been a hard freeze — which it is. Caught in review here.

That second bug means an earlier revision of this PR put a **wrong** claim into the docs, and that I had "corrected" #454's original zero-ticks measurement with my own artefact. The original was right; [my correction is retracted](https://github.com/xability/py-maidr/issues/454#issuecomment-5312983703). Counting wakeups over a window cannot distinguish a blocked loop from a slow one, so the benchmark now measures the longest gap between wakeups around a single render instead — one number, and the one a deployment actually wants.

Verified it fails when the render is removed, so it measures the render rather than the harness, and passes with the proxy pointed at a dead port, so it does not depend on the network it documents the absence of.


jooyoungseo:
Both points taken in 48db350 — and the second one changed what the paragraph says, not just its link.

**#455 is closed, and you're right that "Tracked in" then reads as "already fixed".** I checked what closed it: PR #468, *"say something when the offline fallback runs out of sources"* — which is option 1 from that issue, the reporting half. Options 2 and 3, the ones that would make `"auto"` actually work offline in an iframe, were never done.

So the behaviour I described is still real, but calling it "tracked" was wrong in the other direction: it implies a pending fix, and I don't think there should be one. Inlining the bundle unconditionally is what `use_cdn=False` *is*; doing it in `"auto"` would cost every online reader ~1.9 MB to serve the offline one. The paragraph now says that plainly — this is how `"auto"` behaves and why, use `use_cdn=False` when air-gapped — with no issue link and no implication that something is coming.

**Took the benchmark offer.** `tests/core/test_render_is_synchronous.py`, skipped unless `--run-benchmark`, since gating CI on timing buys flakes rather than information.

Writing it immediately caught a bug in its own first draft, which is the argument for having it:

```
>       assert idle_ticks > busy_ticks
E       rendering did not slow the loop at all (592 vs 592)
```

My idle control busy-waited instead of awaiting — so it starved the loop exactly as rendering does, the two windows agreed, and the comparison measured nothing. A control that cannot fail is worse than no control. It failed loudly only because the assertion was written before the harness. Fixed, and the guard stays.

It now reproduces the figures the docs quote:

```
idle 592 ticks / rendering 87 ticks (12 renders, 51 ms each)
```

The docs cite the ratio and the command rather than exact counts, so the prose doesn't drift from the machine it was measured on. Assertions are a wide band — "much slower, still running" — for the same reason.

`uv run pytest` — 2203 passed, 1 skipped (the benchmark).

---

jooyoungseo:
All three taken in 253e9d5. The first one you flagged as ironic was the one that actually mattered.

**The CDN lookup.** You're right, and it's worse than ironic — a benchmark shipped alongside documentation about air-gapped deployments would have been the thing that hung in one. Pinned to `MAIDR_CDN_VERSION=latest`, which skips the version resolution and changes nothing about the CPU-bound work being measured. Verified by running it with the proxy pointed at a dead port:

```
idle 988 wakeups/s / rendering 136/s (11 renders, 57 ms each)
1 passed in 3.27s
```

No stall, no `MAIDR_CDN_TIMEOUT`. I kept `use_cdn=True` rather than switching to `False` for exactly the reason you suspected: `False` inlines ~1.9 MB per render, which would measure something else.

**Unequal windows.** Good catch — the rendering loop only checks the clock *between* renders, so it overshoots by up to one render's duration while the idle window is exactly `_WINDOW`. It compares rates now, which makes the overshoot irrelevant rather than merely tolerable under wide bands, and means the assertion survives someone tightening it later.

**`try/finally` around the figure** — added.

Thanks for verifying both doc claims against the source independently. That's the check that matters most on a "the docs were wrong" PR, since the failure mode is replacing one confident wrong statement with another. Your note that the corrected text now matches what the code's own comments have said all along is a useful signal that the *code* was right and only the prose had drifted.

`uv run pytest` — 2203 passed, 1 skipped. Benchmark on demand:

```
idle 980 wakeups/s / rendering 140/s (10 renders, 61 ms each)
```

---

jooyoungseo:
You're right, and this is the most important catch of the series — the benchmark was measuring its own settling period, and the docs paragraph it was written to support was **wrong**. Fixed in 463c570.

**Confirmed by moving the ticker after the settle, exactly as you suggested:**

```
ticks DURING rendering only: 0  (13 renders, 49 ms each)
```

Zero. Your reasoning was exact: `_render_once()` never awaits, asyncio is cooperative, so the loop cannot run at all while the render loop runs. Every tick I reported came from the 0.1 s free-run before `start` was captured — which is why "86" sat suspiciously near 100 regardless of window length, as you spotted.

**So the correction I made was the error.** #454's original "0 ticks" was right; I "corrected" a correct number with my own artefact, and then wrote *"Starved, not frozen"* into the docs on the strength of it — arguing for **less** urgency than the truth supports. I've retracted [that comment on #454](https://github.com/xability/py-maidr/issues/454#issuecomment-5312983703).

**The benchmark now measures the loop being held rather than counting wakeups**, because as you note, a wakeup count over a window cannot tell a blocked loop from a slow one — it would report the same healthy-looking number for either, which is precisely how this went wrong:

```
longest loop gap: 1.4 ms idle, 62.1 ms around one render
```

One render, one number, and the thing a deployment actually wants to know. Verified it fails when the render is removed, so it is measuring the render and not the harness.

The docs now say the loop does not run at all for the duration of a render, with the held-time and the jitter baseline.

**On the redundant `MAIDR_CDN_VERSION` pin** — you're right that `tests/conftest.py`'s autouse fixture already pins it and stubs `urlopen`. I'll keep the explicit `monkeypatch.setenv` anyway: this benchmark's whole point is to be run deliberately, sometimes in a sandbox, and I'd rather it not depend on a global fixture continuing to exist for its network-independence. But it's belt-and-braces, not protection, and worth knowing that.

`uv run pytest` — 2203 passed, 1 skipped.

---

jooyoungseo:
Both done in c0f595f, and thank you for checking the 1.9 MB figure against the actual bytes rather than taking it — 1,502,512 + 367,928 + 406 = 1,870,846. That the maths CSS is included is exactly what makes the number right, and it's the sort of thing that quietly rots.

**The `use_cdn=True` comment.** You're right that the file should stand on its own; a reader shouldn't have to find `conftest.py` to learn a benchmark never leaves the machine.

Worth flagging that your note also caught me overstating: I told you in the last round I'd *keep* an explicit `MAIDR_CDN_VERSION` pin, but the rewrite that fixed the tick-counting bug had silently dropped it, so at that moment the test relied on the autouse fixture alone. It's back now, with the reason written down — the suite is safe without it, and it repeats it because a benchmark meant to be run deliberately, often in a sandbox, shouldn't be one global fixture away from stalling on `MAIDR_CDN_TIMEOUT`.

Verified rather than asserted this time, with the proxy pointed at a dead port:

```
longest loop gap: 1.4 ms idle, 39.5 ms around one render
1 passed in 2.17s
```

**PR description updated.** It was a narrative of three iterations rather than a description of the diff, including numbers from a measurement approach that no longer exists — the exact confusion you predicted for a future reader. It now describes the final state, and keeps the two benchmark bugs as a note about why the benchmark is worth having, rather than as stale claims in the body.

`uv run pytest` — 2203 passed, 1 skipped. Merging on green.

---

PR #472, opened 2026-08-17
fix(shiny): let @output_args size a chart in Shiny Express

jooyoungseo:
## Description

`shiny.express.output_args()` is Shiny's documented way to pass UI arguments to a renderer that places its own container. It works by splatting its keywords into the renderer's `auto_output_ui()`:

```python
# shiny/render/renderer/_renderer.py
def _render_auto_output_ui(self):
    with session_context(self._session):
        return self.auto_output_ui(
            # Pass the `@output_args(foo="bar")` kwargs through to the auto_output_ui function.
            **self._auto_output_ui_kwargs,
        )
```

`render_maidr.auto_output_ui()` took no keywords, so in Shiny Express:

```python
from shiny.express import output_args

@output_args(width="50%")
@render_maidr
def bars():
    ...
```

```
TypeError: render_maidr.auto_output_ui() got an unexpected keyword argument 'width'
```

`shiny.render.plot` accepts `**kwargs` there for exactly this reason — it is the precedent this integration was built against, and this is the piece of that contract it had missed.

### Precedence

`@output_args` overrides the decorator; it does not reset it. Given `@output_args(width="50%")` over `@render_maidr(width="42px", height="7em")`, the container is `width: 50%; height: 7em`.

Implemented with `kwargs.setdefault` rather than Shiny's `set_kwargs_value` helper. That helper exists to skip `MISSING`/`None` so an unset argument does not override the UI function's own default, and `render_maidr`'s `width`/`height` are plain strings that can be neither — the two behave identically here, and the simpler one does not reach into a private module.

### Unknown keywords

`output_maidr` keeps its narrow signature, so a keyword it does not know raises there:

```
TypeError: output_maidr() got an unexpected keyword argument 'bogus'
```

An `**kwargs`-absorbing container would have dropped a typo on the floor instead, which for a sizing argument means silently getting the wrong layout.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Changes Made

- `maidr/widget/shiny.py` — `auto_output_ui` accepts and forwards `**kwargs`.
- `tests/widget/test_shiny.py` — two tests, driven through `_render_auto_output_ui` since the splat is the step under test. Both fail on `main`.
- `docs/examples.qmd` — the Shiny section only showed Core mode; adds the Express spelling.

## Checklist

## Additional Notes

`uv run pytest` — 2205 passed, 1 skipped. `ruff check` clean.

Verified against py-shiny 1.6.1 by driving the real `_render_auto_output_ui` rather than a stand-in, so the test fails if upstream stops splatting.

Nothing here was checked in a browser: the assertions are on the emitted container markup, not on how it lays out.

---

### DIFF
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -1176,6 +1176,19 @@ serves, so the bundled `maidr.js` travels inline in the document -- correct
 offline, but roughly 2 MB per chart. The default (`"auto"`) loads from the
 CDN and stays around 30 KB.
 
+In Shiny Express there is no separate UI call: the decorated function places
+its own container, and `@output_args()` sizes it, exactly as it does for
+`@render.plot`.
+
+```python
+from shiny.express import output_args
+
+@output_args(width="600px")
+@render_maidr
+def bars():
+    ...
+```
+
 Check out [a reactive Shiny dashboard example with maidr](https://xabilitylab.shinyapps.io/a11y_dashboard/) and its source code is available on [GitHub](https://github.com/xability/a11y_dashboard).
 
 ### Streamlit
--- a/maidr/widget/shiny.py
+++ b/maidr/widget/shiny.py
@@ -273,9 +273,32 @@ class render_maidr(Renderer[Any]):
         self.use_cdn = use_cdn
         super().__init__(_fn)
 
-    def auto_output_ui(self) -> Tag:
-        """Return the container Shiny Express places for this output."""
-        return output_maidr(self.output_id, width=self.width, height=self.height)
+    def auto_output_ui(self, **kwargs: Any) -> Tag:
+        """Return the container Shiny Express places for this output.
+
+        Parameters
+        ----------
+        **kwargs : Any
+            Arguments for :func:`output_maidr`, supplied by
+            :func:`shiny.express.output_args` and taking precedence over
+            the ones given to the decorator.  Shiny splats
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Took the first nit, skipping the second.

**`**kwargs: Any`** — done (`96b3d78`). I had followed `shiny.render.plot`, which spells it `**kwargs: object` and suppresses the resulting complaint with `# pyright: ignore[reportArgumentType]`. The reason that trade is worth making upstream does not hold here: this repo runs no type checker in CI, only `ruff`, so the suppression can never be checked against a real diagnostic — including whether `arg-type` was even the right code. An unverifiable suppression is worse than none, and `Any` says "opaque passthrough" without one.

**`@output_args(id=...)`** — leaving it. It would raise `TypeError: output_maidr() got multiple values for argument 'id'` rather than the "unexpected keyword" wording, which is less friendly but still an immediate, named failure at UI-construction time. Guarding it would mean either intercepting `id` to re-raise a nicer message, or passing the output id as a keyword — and the second only moves the collision, since a caller-supplied `id` would then silently *replace* the renderer's own, wiring the container to an output that does not exist. That is a worse failure than an ugly `TypeError`, so the corner stays as it is.

CI was green on the pre-nit commit (all 16 checks); re-running on the amend. `uv run pytest` — 2205 passed, 1 skipped. Merging on green.

---

PR #474, opened 2026-08-17
fix(deps): declare the packages `import maidr` needs

### LINKED ISSUE #473: `pip install maidr` produces a package that cannot be imported
The install both the README and `docs/index.qmd` give is `pip install -U maidr`. On a clean environment that install cannot be imported:

```console
$ python -m venv bare && ./bare/bin/pip install maidr
$ ./bare/bin/python -c "import maidr"
Traceback (most recent call last):
  ...
  File ".../maidr/core/plot/violin_kde_plot.py", line 9, in <module>
    from scipy import interpolate
ModuleNotFoundError: No module named 'scipy'
```

Install `scipy` and the next one appears:

```console
  File ".../maidr/patch/barplot.py", line 377, in <module>
    wrap_seaborn("barplot", sns_bar)
  File ".../maidr/patch/common.py", line 357, in wrap_seaborn
    import seaborn
ModuleNotFoundError: No module named 'seaborn'
```

Install `seaborn` and `import maidr` succeeds.

## Why it is not caught

Three things line up:

- **`scipy` is declared nowhere.** It reaches a working environment only as a `statsmodels` dependency, and `statsmodels` sits in the `visualization` extra.
- **`seaborn` is declared only in the `visualization` and `test` extras**, but `wrap_seaborn` imports it at import time rather than lazily, so it is a hard requirement.
- **`matplotlib` is declared only in those same extras** and is imported throughout the model layer. It does not fail today *only* because `mplfinance` — a base dependency — happens to pull it in. That is an accident, not a promise.

Every CI job installs `uv sync --locked --all-extras --dev`, so no job in the matrix ever sees the environment a user gets.

## Affected

`v1.19.0`, `v1.19.1` and `v1.20.0` all ship `maidr/core/plot/violin_kde_plot.py`, and PyPI's metadata for the current release lists no `scipy`:

```
requires_dist: ['htmltools>=0.5', 'lxml>=5.1.0', 'mplfinance>=0.12.10b0',
                'numpy>=1.26', 'wrapt<2,>=1.16.0', ...extras only...]
```

Reproduced on Python 3.9 and 3.11.

## Fix

Declare what is imported: move `matplotlib` and `seaborn` into the base dependencies and add `scipy`. The `visualization` extra keeps `statsmodels` and `flask`, which really are optional.

Making `seaborn` and `scipy` genuinely optional — guarding the imports so a matplotlib-only install works without them — is the larger alternative. It is a feature rather than a fix, and it changes what a `maidr[visualization]` install means, so it is not what this issue asks for.

The guard that would have caught it is a CI job that installs the package with no extras and imports it. Everything else in the matrix installs `--all-extras` by design, so nothing else can.

Found while verifying the Streamlit integration against a real Streamlit install.

jooyoungseo:
## Description

`pip install -U maidr` — the install the README and `docs/index.qmd` both give — produces a package that cannot be imported. On a clean venv, against the current tree:

```console
$ ./bare/bin/python -c "import maidr"
  File ".../maidr/core/plot/violin_kde_plot.py", line 9, in <module>
    from scipy import interpolate
ModuleNotFoundError: No module named 'scipy'
```

Install `scipy` and the next one appears:

```console
  File ".../maidr/patch/barplot.py", line 377, in <module>
    wrap_seaborn("barplot", sns_bar)
  File ".../maidr/patch/common.py", line 357, in wrap_seaborn
    import seaborn
ModuleNotFoundError: No module named 'seaborn'
```

Install `seaborn` and it imports. `v1.19.0`, `v1.19.1` and `v1.20.0` are all affected — PyPI's metadata for the current release lists no `scipy`. Details and the full reproduction are in #473.

## Changes Made

**`pyproject.toml`** — `matplotlib`, `seaborn` and `scipy` move into the base dependencies, because all three are imported while `import maidr` runs:

| package | imported by | was declared |
|---|---|---|
| `scipy` | `maidr/core/plot/violin_kde_plot.py` | nowhere |
| `seaborn` | `wrap_seaborn`, `maidr/patch/barplot.py:377` | `visualization`, `test` |
| `matplotlib` | throughout the model layer | `visualization`, `test` |

`matplotlib` does not fail today, but only because `mplfinance` — a base dependency — happens to pull it in. Being right by way of somebody else's dependency graph is what this PR is about.

The `visualization` extra is left as it is. It still adds `statsmodels` and `flask`, which really are optional, and trimming the two now-redundant names would change what `maidr[visualization]` means for no functional gain.

**`.github/workflows/ci.yml`** — a `bare-install` job. Every existing job installs `uv sync --locked --all-extras --dev`, so none of them can ever see the environment a user gets; that is the whole reason this shipped three times. The new job installs the package alone, on the 3.9 floor, and imports it — no extras, no dev group, no test dependencies, no test run.

There is precedent for this shape a few lines below it: the `FA102` step guards the same failure mode from the other direction — "anything reachable from `maidr/__init__.py` therefore takes `import maidr` down with it, and every interpreter from 3.10 up sees nothing."

**`uv.lock`** — regenerated; `uv lock --check` passes.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues

Closes #473.

## Checklist

## Additional Notes

The new CI job was run locally as CI will run it, on both ends of the supported range:

```console
$ uv venv bare39 -p 3.9 && uv pip install --python bare39/bin/python .
$ ./bare39/bin/python -c "import maidr, sys; print('py', sys.version.split()[0], '| import maidr OK')"
py 3.9.23 | import maidr OK
```

and it fails on `main` at `scipy`, then at `seaborn`, as above — so it tests the fix rather than the harness.

`uv run pytest` — 2205 passed, 1 skipped. `ruff check` reports the same six pre-existing `F401`s as `main` and no new ones. `actionlint` clean on the edited workflow.

No unit test accompanies this: an undeclared dependency is invisible from inside a test suite whose environment has every extra installed, which is exactly how it got here. The CI job is the assertion.

Found while verifying the Streamlit integration from #459 against a real Streamlit install, which needed a venv the project's own `--all-extras` sync does not produce.

---


jooyoungseo:
Took nit 2, and it turned out to be more than cosmetic. Nits 1 and 3 answered below.

**`scipy>=1.10` (`8b972db`)** — you were right to ask why 1.10, and the answer is that it was a number I picked rather than derived. Checking PyPI's metadata, 1.10 is the wrong one:

| scipy | `requires-python` |
|---|---|
| 1.10.x | `>=3.8,<3.12` |
| **1.11.0** | **`>=3.9,<3.13`** |
| 1.14.0 | `>=3.10` |

1.10 caps itself below 3.12, which `requires-python` promises and the test matrix runs. It would not have broken anything — the resolver picks a version with wheels for the interpreter in front of it — but a stated floor that cannot be installed on a supported interpreter is a claim that is simply false. **1.11** is the oldest release installable across the whole 3.9–3.12 range, so that is the floor now, with a comment saying the constraint is about which wheels exist rather than about what the code needs (`interpolate.interp1d` long predates all of these). Re-verified on the 3.9 floor: resolves scipy 1.13.1, `import maidr` OK.

**Nit 1, the dead `try/except ImportError` in `_activate_backend`** — confirmed dead, and worse than dead. `maidr/__init__.py:194` runs `from .api import (...)`, `maidr/api.py:8` imports `matplotlib.axes` at module scope, and `_activate_backend()` is not called until line 248. Nothing can reach line 73 with matplotlib missing. It is misleading rather than merely redundant: it tells a reader maidr is designed to run without matplotlib, which is the belief this PR exists to correct. Not bundling it — this commit is `fix(deps)` and deleting dead code is `refactor`, and semantic-release reads those differently. Filing it as a follow-up once this lands.

**Nit 3, `uv pip install .` rather than `uv sync --locked`** — deliberate, and worth stating outright since it is the one thing that makes the job meaningful. Installing from the lockfile would install the environment *this repo* resolved, which is the environment every other job already has. The bug is in the metadata a user resolves against, so the job has to resolve the way a user does. It is the only job in the file that should skip the lock.

`uv run pytest` — 2205 passed, 1 skipped. `uv lock --check` clean. Merging on green.

---

jooyoungseo:
Took nit 1 (`f608a81`), and the run that raised it proves the point: `Import with no extras` appears **twice** in the checks for the previous commit — once under run `32008455428` (push) and once under `32008458970` (pull_request). That is the duplication `test`'s `if:` exists to prevent.

The job now carries the same guard, verbatim. It belongs in that category by cost rather than by kind: it resolves and installs matplotlib, scipy and seaborn from the index, which is a test-sized job wearing a lint-sized name — unlike `lint`, `commit-lint` and `workflow-lint`, which are cheap enough that nobody bothered. Left those alone; making the file uniform is a separate change from making this job correct.

Nit 3, whether `scipy>=1.11` over-constrains: it excludes 1.9 and 1.10 only, both of which cap themselves at `<3.12` and so cannot be installed on two of the four interpreters in the matrix. On the 3.9 floor the resolver still has 1.11 through 1.13.1 to choose from and picks 1.13.1. Nothing installable is ruled out.

`actionlint` clean. All 18 checks were green on the previous commit, including the new job on both of its accidental runs; re-running on the amend. Merging on green.

---

jooyoungseo:
The guard landed and behaves: on this commit `Import with no extras` is `success` in run `32008591375` and `skipped` in `32008595406`, where the previous commit ran it in both. Same shape as `test`.

**Declining the checkout-ref nit**, though it is a fair thing to raise.

Copying `ref: ${{ github.event.pull_request.head.sha }}` from `test` would make this job build the PR head. Without it, a `pull_request` run builds the merge commit — the tree that will exist on `main`. For a job whose entire question is *"can the thing we are about to publish be installed and imported"*, the merge result is the tree worth asking about; the PR head is a tree that will never be released. That the plain checkout is also what `lint`, `commit-lint` and `workflow-lint` use makes it the majority pattern here rather than an oversight.

Worth noting the difference is narrower than it looks: with the `if:` guard, an internal PR reaches this job only through the `push` event, where `github.event.pull_request.head.sha` is empty and `ref:` resolves to the pushed branch either way. The two spellings can only diverge on a fork PR.

`fetch-depth: 0` I have also left off — `test` needs history, and this job needs one tree and an index.

18 checks, `Import with no extras` green and no longer doubled; the 3.9–3.12 matrix is still running. Merging on green.

---

PR #476, opened 2026-08-17
refactor: drop the useless matplotlib guard in `_activate_backend`

### LINKED ISSUE #475: `_activate_backend`'s matplotlib guard is unreachable, and says the wrong thing
`maidr/__init__.py:72` opens with:

```python
try:
    import matplotlib
except ImportError:
    # matplotlib is not installed — nothing to activate.
    return
```

Nothing can reach that `except`. `maidr/__init__.py:194` runs `from .api import (...)`, `maidr/api.py:8` imports `matplotlib.axes` at module scope, and `_activate_backend()` is not called until line 248. A missing matplotlib takes the package down 54 lines earlier, with a traceback pointing at `api.py`.

Since #474 matplotlib is a declared base dependency, so the branch is not merely unreachable in practice — it contradicts the package metadata.

The reason to remove it is not the dead line. It is the comment, which tells the next reader that maidr is built to run without matplotlib and degrade gracefully. That is not true of any part of this codebase: the model layer, `FigureManager` and nearly every module under `maidr/patch/` import it unconditionally. Anyone acting on that belief -- adding a second guard elsewhere, or wondering which code path serves the matplotlib-less user -- is working from a premise the guard invents.

Deleting the `try`/`except` and importing matplotlib plainly is the whole change.

Raised in review of #474 and deliberately not bundled there: that commit is a `fix` and this is a `refactor`, and semantic-release reads the two differently.

jooyoungseo:
## Description

`_activate_backend` opened with:

```python
try:
    import matplotlib
except ImportError:
    # matplotlib is not installed — nothing to activate.
    return
```

The branch is reachable, and it cannot do what it says.

`_activate_backend()` is called twice. The first call is `maidr/__init__.py:192`, two lines before `from .api import (...)` at 194 — so with matplotlib missing it really does catch and return. Then line 194 runs, `maidr/api.py:8` imports `matplotlib.axes` at module scope, and `import maidr` fails anyway. All the guard buys is a traceback pointing at `api.py` rather than at the line that actually wanted matplotlib. By the second call (line 248) matplotlib is loaded and the branch is dead.

Since #474 matplotlib is a declared base dependency, so the configuration it defends against is not a supported one either.

The reason to remove it is the comment rather than the branch. It tells the next reader that maidr is built to run without matplotlib and degrade gracefully. Nothing in this codebase does: the model layer, `FigureManager` and nearly every module under `maidr/patch/` import it unconditionally. Anyone acting on that belief — adding a matching guard elsewhere, or looking for the code path that serves the matplotlib-less user — is working from a premise the guard invents.

### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor

## Related Issues

Closes #475. Raised in review of #474 and deliberately not bundled there: that commit is a `fix` and this is a `refactor`, and semantic-release reads the two differently.

## Changes Made

- `maidr/__init__.py` — five lines become one `import matplotlib`.

## Checklist

## Additional Notes

`uv run pytest` — 2205 passed, 1 skipped, including the four `tests/test_backend.py` tests that drive `_activate_backend` directly. `ruff check` clean on the edited file.

No test accompanies this. On any supported install nothing observable changes; the one difference is in an install that #474 made impossible, where the `ModuleNotFoundError` now names `maidr/__init__.py` instead of `maidr/api.py`. That is the point of the change rather than a side effect of it, and it is not a state a test can construct from inside a suite that has matplotlib installed.

The first revision of this PR claimed the branch was unreachable, having read only the second call site. It is reachable; it is merely useless. Corrected in `5268d5b` — the change is the same, the reason for it is not.

### DIFF
--- a/maidr/__init__.py
+++ b/maidr/__init__.py
@@ -69,11 +69,7 @@ def _activate_backend() -> None:
     """
     global _original_backend, _backend_message_shown
 
-    try:
-        import matplotlib
-    except ImportError:
-        # matplotlib is not installed — nothing to activate.
-        return
+    import matplotlib
 
     # Save the original backend once, before we override it.
     # Guard against saving our own backend (could happen if _activate_backend


jooyoungseo:
You are right and I was wrong. **The branch is reachable.**

`_activate_backend()` is called twice, and I had read only the second site. On `main`:

```
192:_activate_backend()          <- first call
194:from .api import (  # noqa: E402
248:_activate_backend()          <- the one I looked at
```

With matplotlib missing, line 192 really does reach the `except` and return. The import then fails two lines later at 194, from `maidr/api.py:8`. So the guard is not dead code that never runs — it is code that runs and accomplishes nothing except relocating the traceback away from the line that wanted matplotlib.

Commit message and PR body corrected in `5268d5b`. The change is unaltered; only the justification was wrong, and it was wrong in a way a future reader would have inherited from the commit log.

One consequence I had also stated too broadly: "no observable behaviour change" is not quite true. In an install with matplotlib absent, the `ModuleNotFoundError` now names `maidr/__init__.py` rather than `maidr/api.py`. That install is not reachable since #474 made matplotlib a declared dependency, and the relocation is the improvement rather than a side effect — but it is a difference, and the PR body now says so instead of claiming none.

Nothing else in the diff moves. Merging on green.

---

jooyoungseo:
Filed the `set_backend()` guard as #477 rather than pulling it in here — agreed it is the same remnant, and agreed it is worth thinking about separately because it is public API.

Checked it before writing it down, since I got the reachability wrong once already on this PR. It is dead for a different reason than the one just removed: `set_backend()` can only be called after `import maidr` succeeded, and `import maidr` loads pyplot on the way through — the module docstring at the top of this very file exists to work around that.

```console
$ python -c "import sys; import maidr; print('matplotlib.pyplot' in sys.modules)"
True
```

So `import matplotlib.pyplot` there is a `sys.modules` lookup and cannot raise.

Worth noting it fails worse than the one this PR removes, which is why it earned its own issue rather than a line in this one: that guard was internal and returned from something with no contract. `set_backend()` is public, and taking the branch would return `None` — its ordinary return — having changed no backend, reporting it only at `DEBUG`. The caller gets silence.

---

PR #478, opened 2026-08-17
refactor: drop the unreachable pyplot guard in `set_backend`

### LINKED ISSUE #477: `set_backend()` silently does nothing on an ImportError it cannot get
`maidr/__init__.py:158`, in the public `set_backend()`:

```python
try:
except ImportError:
    _logger.debug("matplotlib is not installed; cannot set backend.")
    return
```

`set_backend()` can only be called after `import maidr` has succeeded, and `import maidr` loads pyplot on the way through (`maidr.core.maidr` imports it — the module docstring at the top of the same file describes working around exactly that). Verified:

```console
$ python -c "import sys; import maidr; print('matplotlib.pyplot' in sys.modules)"
True
```

So by the time this line can run, `import matplotlib.pyplot` is a `sys.modules` lookup. It cannot raise `ImportError`.

This is the same remnant as #475, but it fails worse. #475's guard was in an internal helper and returned from something with no contract. This one is public API: were the branch ever taken, `maidr.set_backend(use_maidr=False)` would return `None` — its normal return — having changed nothing, and say so only at `DEBUG`, which is off by default. A caller gets silence and a backend that did not change.

Since #474 matplotlib is a declared base dependency, so the configuration it defends against is not a supported one either.

Deleting the `try`/`except` and importing pyplot plainly is the whole change. Kept out of #476 because that PR is scoped to the internal helper and this one touches public API, which is worth its own diff.

jooyoungseo:
## Description

The public `set_backend()` opened with:

```python
try:
except ImportError:
    _logger.debug("matplotlib is not installed; cannot set backend.")
    return
```

`set_backend()` is reachable only after `import maidr` has succeeded, and `import maidr` loads pyplot on the way through — the "Matplotlib backend activation" block at the top of this same file exists to work around exactly that (`maidr.core.maidr` imports pyplot, which is why `_activate_backend()` has to be called twice). Verified:

```console
$ python -c "import sys; import maidr; print('matplotlib.pyplot' in sys.modules)"
True
```

So the `import` there is a `sys.modules` lookup and cannot raise `ImportError`.

Unlike the sibling guard removed in #476, this one is on public API, and the way it would fail is the reason it is worth removing rather than leaving alone: `return` is `set_backend`'s ordinary return, so a caller would get `None` back — indistinguishable from success — having changed no backend, told only at `DEBUG`, which is off by default. Silence is the worst available answer for a function whose entire job is a side effect.

`_activate_backend` already imports pyplot unguarded a few lines above (`maidr/__init__.py:91`), so this also makes the two agree.

Left alone: the `try`/`except ImportError` further down in `_activate_backend`, around the backend switch. That one guards a missing `maidr.backend` entry point, which is a different and still-reachable condition.

### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor

## Related Issues

Closes #477. Raised in review of #476 and kept out of it: that PR was scoped to an internal helper, and this touches public API.

## Changes Made

- `maidr/__init__.py` — five lines become one `import matplotlib.pyplot as plt`. `_logger` is still used elsewhere in the module, so the import stays.

## Checklist

## Additional Notes

`uv run pytest` — 2205 passed, 1 skipped. `ruff check` clean.

`tests/test_backend.py::TestSetBackend` covers the changed function directly — eight tests including both directions, a round trip, the idempotent repeat, and the Jupyter `%matplotlib inline` path.

No test is added. The removed branch cannot be entered, and a suite that can `import maidr` at all is a suite where pyplot is loaded — constructing the case would mean faking the very condition the change says is impossible.

*(An earlier revision of this section claimed `tests/test_backend.py` drives `_activate_backend` rather than `set_backend`, and offered a manual smoke test in place of coverage that already exists. It covers both; the smoke test was redundant, not a substitute.)*

### DIFF
--- a/maidr/__init__.py
+++ b/maidr/__init__.py
@@ -155,11 +155,7 @@ def set_backend(use_maidr: bool = True) -> None:
     active backend.  ``set_backend`` only controls the behavior of
     ``plt.show()``.
     """
-    try:
-        import matplotlib.pyplot as plt
-    except ImportError:
-        _logger.debug("matplotlib is not installed; cannot set backend.")
-        return
+    import matplotlib.pyplot as plt
 
     if use_maidr:
         plt.switch_backend("module://maidr.backend")


jooyoungseo:
Correcting my own PR body on the strength of your `TestSetBackend` note: I had written that `tests/test_backend.py` drives `_activate_backend` rather than `set_backend`, and offered a manual smoke test in its place. That was wrong. `TestSetBackend` has eight tests on the changed function — both directions, a round trip, the idempotent repeat, and the Jupyter `%matplotlib inline` path. The smoke test was redundant rather than a substitute, and the body now says so.

Also folded in your observation about the *other* `except ImportError` in `_activate_backend`: it is untouched deliberately, because it guards a missing `maidr.backend` entry point — a different and still-reachable condition — and a reader seeing two similar guards and one of them deleted deserves to know which is which. Said so explicitly rather than leaving it to inference.

No change to the diff. 14 checks green, the 3.9–3.12 matrix still running. Merging on green.

---

PR #479, opened 2026-08-17
test(streamlit): cover the `components.v1.html` fallback end to end

jooyoungseo:
## Description

`test_a_real_streamlit_app_embeds_the_chart_as_srcdoc` — the only test that drives a real Streamlit rather than a stub — opened with:

```python
if not hasattr(st, "iframe"):
    pytest.skip("this Streamlit predates st.iframe; the fallback covers it")
```

Nothing covers it. The fallback has stub coverage only, and the stubs assert *which function is called with which arguments*, which is the one thing that cannot go wrong silently. What can go wrong silently is what Streamlit then does with the HTML — and that was checked on exactly one of the two paths.

It matters more than "an old Streamlit" suggests: the `streamlit` extra allows `>=1.30`, `st.iframe` arrived in 1.56, and **CI's own Python 3.9 job resolves streamlit 1.50** (`uv.lock`). So the job most likely to catch a version-sensitive break was the one job where this test skipped itself.

Both APIs marshal into the same `IFrame` proto, so the skip buys nothing. Removing it makes the 3.9 job exercise the fallback for real while 3.10–3.12 keep exercising `st.iframe`, from one set of assertions.

Verified on both, locally:

| Python | Streamlit | path taken | before | after |
|---|---|---|---|---|
| 3.9 | 1.50.0 | `components.v1.html` | `SKIPPED` | `passed` |
| 3.11 | 1.61.1 | `st.iframe` | passed | passed |

### A second assertion, on `tab_index`

`render_maidr`'s `tab_index` defaults to `None`, documented as the browser default, on the reasoning that an iframe's contents already take part in sequential focus navigation and maidr gives the chart its own tab stop — so `0` would put one extra Tab in front of every chart, announced with Streamlit's hardcoded frame name.

That default only holds if `None` reaches the frame as *absent*. A protobuf scalar cannot express absence on its own, and `IFrame.tab_index` reads as `0` when unset — the very value the default exists to avoid. It survives because the field is declared with explicit presence, so `HasField` can tell them apart. The new test asserts both: that presence is still tracked, and that nothing set the field.

### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Test coverage

## Changes Made

- `tests/widget/test_streamlit.py` — the skip becomes coverage; the app run moves into a `_run_smoke_app` helper, which also asserts exactly one embed is emitted; one new test for the `tab_index` default.

No production code changed.

## Checklist

## Additional Notes

`uv run pytest` — 2206 passed, 1 skipped. `ruff check` clean.

Confirmed the change is load-bearing rather than cosmetic: on Python 3.9 with streamlit 1.50, `main` reports

```
SKIPPED [1] tests/widget/test_streamlit.py:421: this Streamlit predates st.iframe; the fallback covers it
```

and this branch reports `3 passed`.

One thing a test cannot reach, so as not to imply it is covered: the fallback's `_LEGACY_FALLBACK_HEIGHT`. Streamlit carries width and height in the element's `LayoutConfig` rather than on the `IFrame` proto, and `AppTest` exposes only the proto — so whether the 600 px lands is still unverified here, as is anything about how the embed actually lays out.

Found while driving the #459 integration against a real Streamlit install for the first time, which also confirmed the pieces it rests on: `st.iframe` routes a raw HTML string to `srcdoc` rather than `src`, and `height="content"` measurement is available only on that srcdoc path.

---

### DIFF
--- a/tests/widget/test_streamlit.py
+++ b/tests/widget/test_streamlit.py
@@ -406,20 +406,15 @@ def test_a_normal_render_is_recognised_as_having_a_runtime(bar_axes, use_cdn):
     assert not [w for w in caught if "no source for maidr.js" in str(w.message)]
 
 
-def test_a_real_streamlit_app_embeds_the_chart_as_srcdoc(monkeypatch):
-    """Drive the real Streamlit, not a stub, once end to end.
+def _run_smoke_app(monkeypatch):
+    """Run the smoke app under Streamlit's own script runner.
 
-    The stubbed dispatch tests assert which function is called with which
-    arguments; they cannot see what Streamlit then *does* with them. This
-    is the check that ``st.iframe``'s positional argument really carries a
-    self-contained document into the frame's ``srcdoc`` rather than being
-    treated as a URL -- a mistake that would raise nothing in Python and
-    show up only as a blank embed in a browser.
+    Returns the emitted ``IFrame`` proto.  Which of the two embedding APIs
+    produced it depends on the installed Streamlit, and deliberately is not
+    asserted here -- both marshal into the same proto, which is what lets
+    one set of assertions cover both.
     """
-    st = pytest.importorskip("streamlit")
-    if not hasattr(st, "iframe"):
-        pytest.skip("this Streamlit predates st.iframe; the fallback covers it")
-
+    pytest.importorskip("streamlit")
     from streamlit.testing.v1 import AppTest
 
     monkeypatch.setenv("MAIDR_CDN_VERSION", "latest")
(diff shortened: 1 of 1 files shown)


jooyoungseo:
No changes from this round — all three observations land where you put them, but the protobuf one was worth checking rather than assuming, since it is the one that could fail on the job this PR exists to reach.

**`has_presence` on the 3.9 job.** Confirmed on both ends of the matrix, in the environments themselves rather than from the lock alone:

```
py 3.9.23  | protobuf 6.33.6 | has_presence attr: True
py 3.11.15 | protobuf 7.35.1 | has_presence attr: True
```

The 3.9 pin arrives through streamlit 1.50's own protobuf floor, so loosening the `streamlit` extra cannot drop protobuf below it without changing that. And the failure mode is benign either way: `has_presence` missing would be an `AttributeError` on the assertion line, naming the attribute — noisy, but not a false pass.

**`_run_smoke_app` running twice.** Agreed, leaving it. Two `AppTest` runs cost ~0.5 s together, and a shared fixture would hand both tests the same proto object — cheaper, but it would couple them, and a mutation in one would surface as a failure in the other. Not a trade worth making for half a second.

**What is not covered.** Kept explicit in the PR body on purpose. `_LEGACY_FALLBACK_HEIGHT` is the one I would most like to assert and cannot: Streamlit carries width and height in the element's `LayoutConfig`, and `AppTest` exposes only the proto. Better to name the hole than to let "the fallback is covered end to end" imply the layout is.

All checks green except the 3.9–3.11 test jobs, still running. Merging on green — and the 3.9 result is the one worth watching, since it is the job that used to skip.

---

Issue #457, opened 2026-08-17
Evaluate rendering Shiny charts inline instead of inside an iframe


jooyoungseo:
Under Shiny, a maidr chart is wrapped in a `srcdoc` iframe. That choice is what forces two of the awkward things about the Shiny path:

- `Tag.get_html_string()` drops `HTMLDependency` children, so Shiny's own asset machinery cannot deliver `maidr.js` and the bundle has to travel inline for `use_cdn=False` (~2 MB per chart — see #452) with no offline path at all for `use_cdn="auto"` (#455).
- The iframe has no accessible name (#453).

Rendering inline instead would let `session._process_ui` register a normal `HTMLDependency`, which Shiny serves under `lib/`. Verified experimentally: with the iframe suppressed, `_process_ui` picks up and registers the `maidr` dependency correctly (`deps: ['maidr']`), and the payload drops from ~28 KB to ~15 KB — or from ~2 MB to ~15 KB offline.

**Why #452 did not do it.** Every existing non-iframe path (`save_html`) renders a single chart. Notebooks and Flask get one iframe per plot. Dropping the iframe puts N maidr instances into one Bootstrap document with document-level keybindings, and it exposes maidr's runtime styles to the page and vice versa. Neither can be settled by reading emitted HTML — they need a browser.

**What this needs before it can land**

Manual or Playwright QA of a Shiny app with **two** charts on one page:

- Tab order reaches each chart, in a sensible order.
- `b` / `t` / `s` / `r` act on the focused chart only, and do not fire while a `ui.input_text` has focus.
- Braille panel and audio work for both charts independently.
- No visual regression from Bootstrap styles bleeding into maidr's UI, or the reverse.

If the keyboard behaviour does not hold, the iframe stays and this issue closes as won't-fix — in which case #453 and #455 become the things to fix instead.

Follow-up from #452.

jooyoungseo:
## Correcting this issue's premise: browser QA was available all along

I filed this issue saying the inline-versus-iframe question could not be settled without browser QA that was unavailable, and I repeated "nothing here was verified in a browser" in every PR in this series. **That was wrong.** The environment has Chromium preinstalled; Playwright drives it once pointed at the binary explicitly:

```python
p.chromium.launch(executable_path="/opt/pw-browsers/chromium-1194/chrome-linux/chrome",
                  args=["--no-sandbox"])
```

I never checked. Worse, I used the unchecked assumption as the reason to file this issue rather than do the work.

## What now runs

[...]

PR #485, opened 2026-08-17
fix(shiny): keep the reader on the chart across a re-render, and test it in a browser

### LINKED ISSUE #484: A reactive re-render drops the reader out of the chart, silently
In a Shiny app, a reader navigating a chart who changes any input is returned to the top of the document. No announcement, no way to tell it happened other than pressing a key and hearing nothing.

Measured in Chromium against a live `shiny run` app, with the shipped `@render_maidr` integration (`use_cdn=False`, so the runtime loads with no network):

```
BEFORE re-render
  outer focus : IFRAME#iframe_b2394527-bb
  in-frame    : DIV[application]
  position    : b, 2            <- two data points in, text mode on

AFTER re-render
  outer focus : BODY
  in-frame    : BODY
  charts      : 1               <- the chart is fine; the reader is not
```

The chart itself survives: it re-attaches, and Tabbing back in works (`a, 1`). What is lost is the reader's place, their focus, and any signal that either went.

For a sonification-first, keyboard-driven interface this is the difference between a dashboard being usable and not. A sighted user changing a slider keeps their eyes on the chart; a blind user changing the same slider is silently ejected to the document body and must Tab back through every preceding control.

## It is the swap, not the trigger

Focus is sampled every 200 ms from the moment the input changes, alongside whether the output container's HTML has been replaced:

```
t=0.0s  focus=IFRAME  output_swapped=False
t=0.4s  focus=BODY    output_swapped=True
```

Focus survives the input change and dies in the same sample the output is replaced. `output_ui` replaces the container wholesale, taking the focused element with it, and nothing restores it.

## Not caused by the iframe

The same measurement against an inline render — no iframe — is identical:

```
t=0.0s  focus=DIV     output_swapped=False
t=0.2s  focus=BODY    output_swapped=True
```

| | inline | iframe (shipped) |
|---|---|---|
| before | `DIV[application]`, at `b, 2` | `DIV[application]`, at `b, 2` |
| after | `BODY` | `BODY` |

So this is independent of #457 and should not wait on it. Whichever embedding wins, this bug is in both.

## Notes toward a fix

`render_maidr` returns UI that Shiny swaps into the container, so the renderer knows when a swap is about to happen and could restore focus after it. Restoring *focus* looks tractable from this repo. Restoring the reader's **position within the chart** likely needs something from the maidr runtime, since the position is its state, not ours — that part may be an upstream (`xability/maidr`) conversation.

Worth deciding explicitly whether restoring focus without the position is an improvement or a new confusion: landing back on the chart at point 1 with no announcement may read as "nothing happened". An announcement on re-render ("chart updated, 5 bars") is probably the missing piece either way.

Found with browser testing, which — contrary to what I wrote in #457 and in every PR in this series — this environment supports. Harness and correction in https://github.com/xability/py-maidr/issues/457#issuecomment-5317214856.

jooyoungseo:
## Description

Two things, in that order: an accessibility bug, and the test infrastructure that proves it fixed — because the second turned out to be the only way to establish the first.

### The bug

A Shiny output is replaced wholesale on every reactive flush, taking the focused element with it. A reader partway through navigating a chart who changed any input was returned to the top of the document, silently. Measured on `main`, in Chromium against a live `shiny run` app:

```
BEFORE re-render   outer focus: IFRAME#iframe_b2394527   in-frame: DIV[application]   at: b, 2
AFTER  re-render   outer focus: BODY                     in-frame: BODY
```

The chart itself was fine — it re-attached, and Tabbing back in worked. What was lost was the reader's place and any signal that it had gone. A sighted user changing a slider keeps their eyes on the chart; a blind user changing the same slider was ejected to the document body and had to Tab back through every preceding control.

Sampling focus every 200 ms against whether the container had been replaced shows it is the swap, not the input change:

```
t=0.0s  focus=IFRAME  output_swapped=False
t=0.4s  focus=BODY    output_swapped=True
```

### Why the implementation looks like this

Two findings, both established in a browser rather than reasoned about, rule out the obvious event-driven version:

1. **Removing the focused element does not reliably fire `blur`/`focusout`.** A first attempt hung the restore off `focusout` and never fired.
2. **Moving focus *into* an iframe fires no `focusin` in the parent document at all** — `document.activeElement` simply becomes the frame. Instrumented directly: focusing the chart inside the frame produced `events: []` while `activeElement` was `IFRAME`.

Since charts render into an iframe today, an event-driven hook sees nothing on the path that matters. So "the chart had focus" is established by sampling `activeElement`, and the re-render is detected with a `MutationObserver` on the output container.

The conditions are deliberately narrow, because being too eager is worse than the bug — pulling someone out of a control mid-task is more disruptive than dropping them from a chart. Focus is restored only when the element that held it has **left the document** *and* focus **landed nowhere**.

## Browser tests

Neither of the two findings above, nor the fix itself, is visible in emitted markup — which is all the other 2208 tests can see. The unit tests here can only assert the script is *present* in the payload, which is not the same as it working. Until now the only evidence was me driving a browser by hand, which does not survive the session.

`tests/browser/` runs a real Shiny app in Chromium. The two cases that matter pull in opposite directions:

| test | asserts |
|---|---|
| `test_the_chart_is_reachable_and_answers_the_keyboard` | the premise — a chart nobody can focus cannot lose focus |
| `test_focus_returns_to_the_chart_after_a_rerender` | the bug |
| `test_a_deliberate_move_away_is_not_undone` | the opposite failure, plus keystrokes still land in the input |
| `test_a_chart_nobody_focused_does_not_take_focus` | page load is a re-render too |

A suite covering only the second would pass on a version that grabs focus from whatever the reader was doing.

**Verified load-bearing.** With `_process_ui(payload)` reverted to `_process_ui(rendered)`:

```
FAILED tests/browser/test_focus_restore.py::test_focus_returns_to_the_chart_after_a_rerender
1 failed, 3 passed
```

The three over-reach guards still pass, which is what they should do when there is no fix to over-reach.

### A second use of the same infrastructure

`test_offline_report.py` guards the runtime report from #455/#467 — the message an air-gapped reader gets instead of a silently dead chart. `tests/core/test_offline_fallback_report.py` asserts that message is in the emitted script, which cannot tell whether it ever reaches a console: a `ReferenceError` earlier in the script, a fallback that never fires, or a report wired to a branch that is not the one taken would all pass it. It blocks the CDN with `page.route` rather than relying on the runner having no network, so the test creates the air-gapped condition itself.

Also load-bearing: with the report body emptied, `test_a_chart_with_no_runtime_says_so_in_the_console` fails.

It is here rather than in its own PR because it is the same infrastructure, and because one test is thin evidence that browser testing generalises beyond the thing it was built for.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Test infrastructure

## Related Issues

Closes #484.

Independent of #457: the same measurement inline, with no iframe, loses focus identically (`t=0.2s focus=BODY output_swapped=True`), so this is not caused by the iframe and should not wait on that decision.

## Changes Made

- `maidr/widget/_focus.py` — new; the script and the reasoning for its shape.
- `maidr/widget/shiny.py` — the payload becomes `TagList(rendered, tags.script(...))`. It rides with the chart rather than with the container because `output_maidr` is not always what places the output — Express goes through `auto_output_ui`, and an app may write its own `ui.output_ui` — but every chart comes through the renderer. It self-guards, so arriving per render costs nothing after the first.
- `tests/widget/test_shiny.py` — the payload carries the hook, and the hook is idempotent (without the guard, N flushes would leave N samplers running).
- `tests/browser/` — new; fixtures, two apps, six tests.
- `tests/conftest.py` — `--run-browser`, alongside the existing `--run-benchmark` but for a different reason, and the comment says which: the benchmark is skipped because timing is machine-dependent, this because it needs a browser binary and a Shiny server, and a missing browser should read as "not run here" rather than as a failure.
- `.github/workflows/ci.yml` — a `browser` job that installs Chromium and runs them. Kept out of the `test` matrix because it needs a browser as well as a wheel.
- `pyproject.toml` — `playwright` as a `browser` extra rather than a `test` one, so `uv sync --all-extras` does not imply the browser tests can run.
- `docs/examples.qmd` — two sentences, including the "only when the chart held focus" condition, since that is the part a reader might otherwise find surprising.

## Checklist

## Additional Notes

`uv run pytest` — 2208 passed, 5 skipped (1 benchmark, 4 browser… 6 browser as of the last commit). `uv run pytest tests/browser --run-browser` — 6 passed in 117 s. `ruff check` shows the same six pre-existing `F401`s as `main`; `actionlint` and `uv lock --check` clean.

**What this does not fix.** Focus comes back to the chart; the reader's **position** in it does not — they land at the first data point. I checked whether that was fixable from here rather than assuming: the runtime's entire public surface is `maidrLive.{setData, appendData}`, there is no API to read or set a position, and no DOM signal of one. `setData` looked like a better route — update in place, never destroy the element — but it accepts the whole figure schema, a single layer, and a bare data array, returns `ok` for all three, and changes nothing. Filed upstream as xability/maidr#956; detail in #484.

**Limits of the verification.** Chromium only, one chart type, no screen reader — focus is read from `document.activeElement`, which is not the same as hearing an announcement. The 200 ms sampler is forced by the two findings above; if the runtime ever exposed a focus event crossing the frame boundary, this could become event-driven.

**CI has not validated this yet.** GitHub has been in a Partial System Outage throughout (`API Requests`, `Actions`, `Pull Requests` all major outage), and every red check on this PR failed before reaching any code — `429` downloading actions, `503` on permission checks, `Not Found` reading commits. Per-check diagnosis is in the comments. The new `browser` job in particular has never run on a runner. Not merging until it has.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,40 @@ jobs:
         id: tests
         run: uv run pytest -vvv
 
+  browser:
+    name: Accessibility tests in a browser
+
+    # Guarded like `test`: an internal PR fires both `push` and
+    # `pull_request`, and this job installs a browser.
+    if:
+      github.event_name == 'push' || github.event.pull_request.head.repo.full_name !=
+      github.repository
+
+    runs-on: ubuntu-latest
+
+    # The only tests that check what the library actually promises -- that
+    # a chart can be reached and driven from the keyboard. Everything else
+    # in the suite asserts on emitted markup, which cannot tell a working
+    # chart from a well-formed one.
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv and set the python version
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install the project
+        run: uv sync --locked --all-extras --dev
+
+      - name: Install Playwright and its browser
+        run: |
+          uv pip install playwright
+          uv run playwright install --with-deps chromium
+
+      - name: Run the browser tests
+        run: uv run pytest tests/browser -vv --run-browser
+
   bare-install:
     name: Import with no extras
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,10 +83,10 @@ jobs:
       - name: Install the project
         run: uv sync --locked --all-extras --dev
 
-      - name: Install Playwright and its browser
-        run: |
-          uv pip install playwright
-          uv run playwright install --with-deps chromium
+      # The wheel arrives with `--all-extras` above; only the browser
+      # binary is missing, and it is not a Python package.
+      - name: Install the browser Playwright drives
+        run: uv run playwright install --with-deps chromium
 
       - name: Run the browser tests
         run: uv run pytest tests/browser -vv --run-browser
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -1176,6 +1176,12 @@ serves, so the bundled `maidr.js` travels inline in the document -- correct
 offline, but roughly 2 MB per chart. The default (`"auto"`) loads from the
 CDN and stays around 30 KB.
 
+A reactive flush replaces the whole output, so a reader who was navigating the
+chart when some input changed would be dropped to the top of the page without
+being told. `@render_maidr` puts focus back on the updated chart when — and
+only when — the chart was what held focus, so a reader who had deliberately
+moved to another control keeps their place.
+
 In Shiny Express there is no separate UI call: the decorated function places
 its own container, and `@output_args()` sizes it, exactly as it does for
 `@render.plot`.
--- a/maidr/util/iframe_utils.py
+++ b/maidr/util/iframe_utils.py
@@ -247,6 +247,11 @@ def wrap_in_iframe_matplotlib(base_html: Tag, chart_title: str | None = None) ->
     return tags.iframe(
         id=unique_id,
         title=iframe_title(chart_title),
+        # Marks this frame as maidr's own. A host page may put other
+        # iframes in the same container, and anything keying off "an
+        # iframe is the chart" -- the Shiny focus restore does -- would
+        # otherwise act on one of those. See `maidr/widget/_focus.py`.
+        **{"data-maidr-chart": ""},
         srcdoc=str(base_html.get_html_string()),
         width="100%",
         height="100%",
@@ -383,6 +388,11 @@ def wrap_in_iframe_plotly(base_html: Tag, chart_title: str | None = None) -> Tag
     return tags.iframe(
         id=unique_id,
         title=iframe_title(chart_title),
+        # Marks this frame as maidr's own. A host page may put other
+        # iframes in the same container, and anything keying off "an
+        # iframe is the chart" -- the Shiny focus restore does -- would
+        # otherwise act on one of those. See `maidr/widget/_focus.py`.
+        **{"data-maidr-chart": ""},
         srcdoc=str(base_html.get_html_string()),
         width="100%",
         height="100%",
--- a/maidr/widget/_focus.py
+++ b/maidr/widget/_focus.py
@@ -25,6 +25,15 @@ from __future__ import annotations
 #: A reader who tabs away, clicks another control, or moves focus on
 #: purpose is never pulled back, and a chart nobody had focused never
 #: takes focus at all.
+#:
+#: The price of not being event-driven is a poll that never stops: a
+#: 200 ms timer, running for the life of the tab, re-scanning
+#: ``.shiny-html-output`` on every tick. On a dashboard with many charts
+#: that is a small continuous cost paid whether or not anyone ever
+#: touches a chart. It is deliberate rather than overlooked -- there is
+#: no signal to wait on instead -- and it is the first thing to revisit
+#: if the runtime ever exposes a focus event that crosses the frame
+#: boundary.
 FOCUS_RESTORE_JS = """
 (function () {
   if (window.__maidrShinyFocusRestore) return;
--- a/maidr/widget/_focus.py
+++ b/maidr/widget/_focus.py
@@ -1,0 +1,76 @@
+"""Keeping a reader's place when Shiny replaces a chart.
+
+A Shiny output is replaced wholesale on every reactive flush, taking the
+focused element with it.  For a chart that is navigated by keyboard, that
+ejects the reader to the top of the document mid-read, with nothing said
+(#484).  The script here puts focus back.
+"""
+
+from __future__ import annotations
+
+#: Installed once per page, alongside the first rendered chart.
+#:
+#: Two things rule out an event-driven version, and both were established
+#: in a browser rather than reasoned about.  Removing the focused element
+#: does not reliably fire ``blur``/``focusout``; and moving focus *into*
+#: an iframe fires no ``focusin`` in the parent document at all --
+#: ``document.activeElement`` simply becomes the frame.  Since a chart is
+#: rendered into an iframe today, an event-driven version sees nothing at
+#: all on the path that matters.  So "the chart had focus" is established
+#: by sampling, and the re-render is detected by observing the container.
+#:
+#: The conditions are deliberately narrow, because the failure mode of
+#: being too eager is worse than the bug: focus is restored only when the
+#: element that held it has left the document *and* focus landed nowhere.
+#: A reader who tabs away, clicks another control, or moves focus on
+#: purpose is never pulled back, and a chart nobody had focused never
+#: takes focus at all.
+FOCUS_RESTORE_JS = """
+(function () {
+  if (window.__maidrShinyFocusRestore) return;
+  window.__maidrShinyFocusRestore = true;
+
+  var CHART = 'iframe, [role="img"], [role="application"]';
+  var SAMPLE = 200;
+  var held = null;
+  var observed = new WeakSet();
+
+  function adrift() {
+    var a = document.activeElement;
+    return !a || a === document.body || a === document.documentElement;
+  }
+
+  function sample() {
+    if (adrift()) return;
+    var a = document.activeElement;
+    var c = a.closest && a.closest('.shiny-html-output');
+    held = c && c.id && c.querySelector(CHART) ? { id: c.id, el: a } : null;
+  }
+
+  function restore(container) {
+    if (!held || held.id !== container.id) return;
+    if (held.el.isConnected) return;
+    if (!adrift()) return;
+    var target = container.querySelector(CHART);
+    if (!target) return;
+    target.focus();
+    held = { id: container.id, el: target };
+  }
+
+  function watch() {
+    var containers = document.querySelectorAll('.shiny-html-output');
+    for (var i = 0; i < containers.length; i++) {
+      var c = containers[i];
+      if (observed.has(c) || !c.id) continue;
+      observed.add(c);
+      new MutationObserver(
+        (function (el) { return function () { restore(el); }; })(c)
+      ).observe(c, { childList: true, subtree: true });
+    }
+  }
+
+  setInterval(function () { sample(); watch(); }, SAMPLE);
+})();
+"""
+
+__all__ = ["FOCUS_RESTORE_JS"]
--- a/maidr/widget/_focus.py
+++ b/maidr/widget/_focus.py
@@ -39,10 +39,15 @@ FOCUS_RESTORE_JS = """
   if (window.__maidrShinyFocusRestore) return;
   window.__maidrShinyFocusRestore = true;
 
-  var CHART = 'iframe, [role="img"], [role="application"]';
+  // `[data-maidr-chart]` rather than a bare `iframe`: a host page may put
+  // its own iframe in the same output container, and treating that as the
+  // chart would force focus onto someone else's embed. The roles cover
+  // the non-iframe render, where the runtime owns the element and marks
+  // it itself.
+  var CHART = '[data-maidr-chart], [role="img"], [role="application"]';
   var SAMPLE = 200;
   var held = null;
-  var observed = new WeakSet();
+  var observed = [];
 
   function adrift() {
     var a = document.activeElement;
@@ -52,8 +57,14 @@ FOCUS_RESTORE_JS = """
   function sample() {
     if (adrift()) return;
     var a = document.activeElement;
-    var c = a.closest && a.closest('.shiny-html-output');
-    held = c && c.id && c.querySelector(CHART) ? { id: c.id, el: a } : null;
+    if (!a.closest) { held = null; return; }
+    // The chart itself, not merely something in a container that has one.
+    // A caption or a download link beside the chart is a place a reader
+    // may have gone on purpose, and losing it should not send them to the
+    // chart instead.
+    var chart = a.closest(CHART);
+    var c = a.closest('.shiny-html-output');
+    held = chart && c && c.id ? { id: c.id, el: chart } : null;
   }
 
   function restore(container) {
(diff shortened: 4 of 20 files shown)


jooyoungseo:
Holding this rather than merging: `claude-review` is red for a reason outside the diff, and the platform is mid-incident.

The job never reached the review. It dies in the permission check:

```
GET /repos/xability/py-maidr/collaborators/jooyoungseo/permission - 503
##[error]Failed to check permissions: HttpError: No server is currently available
          to service your request.
```

Re-ran it once — the legitimate case for a re-run, since it died before any review work began — and it failed identically at ~1.1 s. A second re-run would not be a diagnosis, so here is the actual one, from GitHub's status API:

```
status: Partial System Outage          (incident open since 2026-08-17T13:40Z)
 - API Requests  -> major_outage
 - Actions       -> major_outage
 - Pull Requests -> major_outage
 - Issues        -> major_outage
 - Webhooks      -> partial_outage
```

`CodeQL` also returned `neutral` on this run rather than a verdict, which fits the same cause.

Everything that did complete is green: `lint`, `commit-lint`, `Lint workflows and shell scripts`, and `Import with no extras`. The 3.9–3.12 test matrix was still running when the incident took the API down, so I am treating those results as unreliable regardless of what they report.

Locally, on the merge base: `uv run pytest` — 2208 passed, 1 skipped; `ruff check` shows the same six pre-existing `F401`s as `main` and no new ones.

Not merging until the incident clears and CI has run cleanly end to end. Re-checking shortly; nothing here needs a code change.

---

jooyoungseo:
Pushed `d244a44`, which closes the gap this PR admitted to: the fix had no test, only me driving a browser by hand.

The behaviour is a browser behaviour — an element losing focus because it left the document — and nothing about it is visible in emitted markup, which is all the other 2208 tests can see. The two tests already here only assert that the script is *present* in the payload, which is not the same as it working.

**`tests/browser/test_focus_restore.py`** drives a real Shiny app in Chromium. Four cases, and the two that matter pull in opposite directions:

| test | asserts |
|---|---|
| `test_the_chart_is_reachable_and_answers_the_keyboard` | the premise — a chart nobody can focus cannot lose focus |
| `test_focus_returns_to_the_chart_after_a_rerender` | the bug |
| `test_a_deliberate_move_away_is_not_undone` | the opposite failure, plus keystrokes still land in the input |
| `test_a_chart_nobody_focused_does_not_take_focus` | page load is a re-render too |

A suite covering only the second would pass on a version that grabs focus from whatever the reader was doing — which is a worse bug than the one being fixed.

**Verified load-bearing.** With `_process_ui(payload)` reverted to `_process_ui(rendered)`:

```
FAILED tests/browser/test_focus_restore.py::test_focus_returns_to_the_chart_after_a_rerender
1 failed, 3 passed
```

The three over-reach guards still pass, which is what they should do when there is no fix to over-reach.

**How it is wired.** Opt-in behind `--run-browser`, alongside the existing `--run-benchmark`, but for a different reason and the comment says so: the benchmark is skipped because timing is machine-dependent, this because it needs a browser binary and a Shiny server, and a missing browser should read as "not run here" rather than as a failure. `conftest.py` locates a Chromium via `PLAYWRIGHT_BROWSERS_PATH` or `PATH` and skips if it finds none — Playwright pins an exact build and refuses anything else, so an environment shipping its own browser needs `executable_path`.

A `browser` CI job installs Chromium and runs them. Kept out of the `test` matrix because it needs a browser as well as a wheel; `playwright` is a `browser` extra rather than a `test` one for the same reason, so `uv sync --all-extras` does not imply the browser tests can run.

`uv run pytest` — 2208 passed, 5 skipped (1 benchmark, 4 browser). `uv run pytest tests/browser --run-browser` — 4 passed in 92s. `actionlint` clean, `uv lock --check` clean.

Still blocked on the platform incident for CI; the new job has not run on a runner yet, only locally. Not merging until it has.

---

jooyoungseo:
Four checks are red on `d244a44`. All four failed before reaching any code in this PR — diagnosed individually rather than waved off as flakes:

| check | what actually happened |
|---|---|
| `lint` | could not **download the action**: `Failed to download archive 'codeload.github.com/astral-sh/setup-uv/...'` — `429 Too Many Requests`, 3 attempts. `ruff` never ran. |
| `Lint workflows and shell scripts` | `curl: (22) The requested URL returned error: 429` fetching actionlint's installer, then `./actionlint: No such file or directory`. `shellcheck` passed; actionlint never ran. |
| `commit-lint` | `HttpError: Not Found at async getPushEventCommits` — could not read the commits from the API, then crashed: `Cannot read properties of undefined (reading 'map')`. The `You have commit messages with errors` line is that crash, not a finding. |
| `claude-review` | `GET /collaborators/.../permission → 503`, as on the previous commit. |

Every one is a `429`/`503`/`Not Found` against github.com infrastructure, which matches the open incident:

```
status: Partial System Outage
API Requests = major_outage      Actions  = major_outage
Pull Requests = major_outage     Issues   = major_outage
Git Operations = degraded        Webhooks = degraded
```

Locally, on this exact commit: `ruff check .` clean apart from the six pre-existing `F401`s; `actionlint .github/workflows/ci.yml` clean; `uv lock --check` clean; `uv run pytest` 2208 passed, 5 skipped; `uv run pytest tests/browser --run-browser` 4 passed.

The commit subject `test(shiny): verify the focus fix in a real browser` is a valid conventional commit, so commit-lint has nothing to object to once it can read the commit.

Not re-running any of these while the incident is open — a re-run that fails on the next 429 is noise, and the two I already re-ran failed identically. Holding until the platform recovers, then a clean end-to-end run; the new `browser` job in particular has never executed on a runner and needs to before this merges.

---

jooyoungseo:
GitHub has now described the incident in terms that match the per-check diagnosis above exactly:

> We are experiencing high error rates around 20% for web experiences and api traffic. **Archive downloads and raw repository content downloads are experiencing an approximate 50% error rate.**

That is the two failure shapes on this PR, named by the platform:

- *archive and raw content downloads* — `lint` could not download the `setup-uv` action archive, and `Lint workflows and shell scripts` could not `curl` actionlint's installer from `raw.githubusercontent.com`. Both `429`.
- *api traffic* — `claude-review`'s `503` on the permission check, and `commit-lint`'s `Not Found` reading the push commits.

A ~50% failure rate on action downloads means any given re-run is a coin toss, which is why I am not re-running: a pass would say as little as the failures do.

`claude-review` and `commit-lint` have gone red again on `afe417d` for the same reasons. I will not keep posting a diagnosis per commit — this comment covers the class, and I will confirm each check individually once the incident closes.

Nothing in this PR has changed in response to any of it, and nothing should.

---

jooyoungseo:
All three points taken. The first one did not go the way I expected, and the honest result is worth more than the fix I set out to write.

## Multi-chart isolation — behaviour covered, that one line still not isolable

`tests/browser/test_focus_isolation.py` puts two charts on a page, re-renders both, and asserts the reader comes back to the chart they were in. It passes.

It does **not** prove the `held.id !== container.id` check, which is what you actually asked for. Two attempts, both of which stayed green with that line deleted:

1. **A re-rendering chart paired with a static one.** `restore` never reaches the id check — the held element is still connected, so it bails on `isConnected` a line earlier. The fixture's asymmetry, which I had picked precisely to make theft visible, is what made the test vacuous.
2. **Both charts re-rendering.** Now the held element really is destroyed and both observers fire — but they fire in an order that restores the right chart first, after which `adrift()` is false and the second call returns anyway.

That ordering is incidental rather than guaranteed, so the check is still doing real work; it just is not observable from outside. Both dead ends are written into the test's docstring so the next reader does not take a green run as proof of that line — or delete it on the strength of one.

I would rather say this than quietly ship the first version, which looked like it closed the gap and closed nothing.

## `.shiny-html-output` going silently inert

Fair, and the "silently" is the bad part. `test_the_container_class_the_focus_script_looks_for_still_exists` now asserts the script and `output_maidr` agree on the class, in the **default** suite — so a Shiny upgrade that renames it fails for a contributor running plain `pytest`, rather than waiting for someone to opt into `--run-browser`.

## Redundant install step

Correct — `--all-extras` already brings the wheel via the `browser` extra; only the binary is missing, and it is not a Python package. Dropped, with a comment saying why the remaining step is not also redundant.

---

`uv run pytest` — 2209 passed, 11 skipped. `uv run pytest tests/browser --run-browser` — 10 passed in 176 s. `ruff` shows the same six pre-existing `F401`s; `actionlint` clean.

On your last point — that the `browser` job has never run on a runner, and that `--no-sandbox`, `--with-deps` and the dynamic port could all behave differently there: agreed, and that is exactly why this is not merged. GitHub is still in a Partial System Outage with `Actions` at major outage, so the job still has not run. It will before this merges.

---

jooyoungseo:
The pipe finding is a good catch and is fixed. Taking the docstring point too; answering the other two rather than changing them.

## The subprocess pipe — fixed

You are right, and the failure mode is the part that makes it worth fixing: a wedged writer gives a **hang**, not a red test, so it would present as a CI job that times out with no useful output. Streamlit is the likeliest trigger, since its access logs are chattier than Shiny's and the newly added key-collision test drives a Streamlit app.

Output now goes to a `NamedTemporaryFile` rather than a pipe, read only on failure and truncated to the last 4 KB. Nobody has to drain it, so there is no buffer to fill and no draining thread to get wrong. Both failure paths — early exit and boot timeout — now include the log; the timeout path previously reported nothing at all, which would have been unhelpful in exactly the situation it fires.

## The permanent 200 ms poll — documented

Fair, and the docstring was lopsided: it explained at length why the fix is not event-driven, and never said what that costs. It now states plainly that the timer runs for the life of the tab, re-scans on every tick, is paid whether or not anyone touches a chart, and is the first thing to revisit if the runtime ever exposes a focus event that crosses the frame boundary.

## `_process_ui` on a `TagList` — leaving as is

Agreed it is an unguarded coupling to a private API. I am not adding a targeted assertion because I do not think one would earn its place: `test_payload_has_the_shape_shiny_expects` already drives the real `_process_ui` through a `TagList` and asserts the `{"deps", "html"}` shape, so a change in what it accepts fails there — with a slightly worse message than a dedicated test, but it fails. The `fake_session` fixture borrows Shiny's real `AppSession._process_ui` rather than stubbing it, which is what makes that true.

## An empty container after a failed render — leaving as is, with reasoning

`restore` ends with `if (!target) return;`, so an empty container is a no-op and focus stays wherever it landed. Testing it would mean building a render that fails *after* the container has been replaced, which is a state I could not construct without stubbing the renderer — at which point the test would be exercising the stub. The behaviour is also the conservative one: doing nothing is right when there is no chart to return to.

---

`uv run pytest` — 2209 passed, 13 skipped. `uv run pytest tests/browser --run-browser` — 12 passed in 198 s. `ruff` unchanged from `main`; `actionlint` clean.

Still not merged: `Actions` remains at major outage, so the `browser` job has still never run on a runner. Given your point that `--no-sandbox`, `--with-deps` and the dynamic port can all behave differently there, that run is the last thing this needs.

---

jooyoungseo:
All three correctness findings were real and are fixed. They share a property worth naming: none was reachable from the browser suite as scoped — one static output, one focusable element each — so they had to be found by reading. That is a fair criticism of the tests, not just of the code.

## 1. The selector matched any iframe

Correct, and the consequence is the bad kind: focus forced onto someone else's embed. The rendered frame now carries `data-maidr-chart` (both sites in `iframe_utils.py`), and the selector is `[data-maidr-chart], [role="img"], [role="application"]` — the roles still cover the non-iframe render, where the runtime owns the element and marks it itself.

Pinned in the default suite by `test_the_rendered_iframe_is_marked_as_maidrs_own`, which asserts both sides agree, the same shape as the `.shiny-html-output` guard.

## 2. `sample()` held whatever was focused, not the chart

The sharpest of the three, because the docstring already claimed the narrower behaviour ("the element that held it") while the code did not implement it. Now:

```js
var chart = a.closest(CHART);
var c = a.closest('.shiny-html-output');
held = chart && c && c.id ? { id: c.id, el: chart } : null;
```

Your example is exactly right about why it matters: a caption or download link beside a chart is somewhere a reader may have gone deliberately, and losing it should not redirect them onto the chart.

## 3. Observers never disconnected

`watch()` now prunes any whose container is no longer `isConnected`, before adding new ones. `removeUI` and an unmounting `conditionalPanel` are the reachable cases. This also meant swapping the `WeakSet` for a list — a `WeakSet` cannot be iterated, so it could dedupe but never clean up, which is presumably how it got missed.

## Not changed

- **Fixed sleeps rather than auto-retrying locators.** Agreed as a follow-up if the suite grows. Most of the waiting is for the ~1.5 MB inlined bundle to parse in the frame and for a reactive flush to land — neither is a locator appearing, so `expect`'s auto-retry would not replace them, though a poll-until-condition helper would trim the tail. Not doing it in this PR, which is large enough.
- **`_free_port()` TOCTOU.** Real, and I do not have a better option: both Shiny and Streamlit take a port number rather than a bound socket, so something has to pick one and let go. The window is microseconds and a collision fails loudly at bind.

---

`uv run pytest` — 2210 passed, 13 skipped. `uv run pytest tests/browser --run-browser` — 12 passed, still green after all three changes. `ruff` unchanged from `main`.

On holding for the real CI run: agreed, and I have narrowed what is unknown about it rather than only waiting. I reproduced the job's steps in a clean checkout — `uv sync --locked --all-extras --dev` on Python 3.12, then `playwright install chromium`, then the suite — and it passes end to end, 12 tests in 205 s. That confirms `--all-extras` really does supply the wheel (so dropping the extra `uv pip install` was right) and that a freshly downloaded Chromium launches with `--no-sandbox` here. What it cannot confirm is the `ubuntu-latest` image itself or `--with-deps`' apt step, so the job still has to run for real. `Actions` is still at major outage.

---

Issue #305, opened 2026-08-05
ci: OIDC to app-token exchange returns 401, so Claude reviews cannot post as claude[bot]


jooyoungseo:
Claude reviews on this repo are attributed to **github-actions[bot]**, while `maidr` and `r-maidr` show **claude[bot]**. The difference is that those repos let `anthropics/claude-code-action` obtain its own app token; here that exchange fails, so the workflows pass `github_token: ${{ secrets.GITHUB_TOKEN }}` and the action speaks as the workflow instead.

## The failure

Observed on runs [31040855269](https://github.com/xability/py-maidr/actions/runs/31040855269) and [31041322462](https://github.com/xability/py-maidr/actions/runs/31041322462), after #302 removed the input:

```
Exchanging OIDC token for app token...
Attempt 1 of 3...
App token exchange failed: 401 Unauthorized - Invalid OIDC token
Attempt 2 of 3...
App token exchange failed: 401 Unauthorized - Invalid OIDC token
Attempt 3 of 3...
App token exchange failed: 401 Unauthorized - Invalid OIDC token
Action failed with error: Invalid OIDC token
```

Reverted in #304. This issue is about the underlying exchange.

## What is known

- The Claude GitHub App **is** installed here — it posted as `claude[bot]` on #273 on 2026-03-04.
- The bypass went in the next day, `f22e3187` "ci: add github_token to bypass OIDC in claude review workflow", the same day `a35b5925` moved the review to `pull_request_target`. That adjacency is suggestive but not established: no error was recorded at the time, so whether `pull_request_target` caused it is a guess.
- Both jobs already grant `id-token: write`, so the OIDC permission itself is present.
- `maidr` uses plain `pull_request` and needs no `github_token`; `r-maidr` is worth checking as a second comparison.

## Worth checking

- Whether the exchange fails specifically under `pull_request_target`, or on any event here. A branch with a temporary `workflow_dispatch` job would answer this without touching the review workflows.
- Whether the app installation is scoped to selected repositories and whether `py-maidr` is among them with the permissions the exchange expects.
- Whether the OIDC subject or audience differs on `pull_request_target` in a way the exchange rejects.

## Why it matters beyond the label

[...]

PR #486, opened 2026-08-17
fix(patch): say which seaborn is installed when it is too old

jooyoungseo:
## Description

#441 asked for one of two things. The floor half is done — `pyproject.toml` declares `seaborn>=0.13` and `maidr/patch/boxplot.py`'s wrong version branch is gone. This addresses what the issue said mattered *beyond* tidiness:

> Someone whose resolver lands on 0.12 — an older lockfile, a conflicting pin elsewhere in their environment — gets an `AttributeError` from `import maidr` with no indication that their seaborn version is the reason.

That is still true today. Reproduced on a real seaborn 0.12.2, following the issue's own steps:

```
AttributeError: type object '_CategoricalPlotter' has no attribute 'plot_bars'
```

Nothing there names seaborn, names a version, or says what to do, and it arrives while `maidr.patch` is being imported — before any of the user's own code runs.

**Note the attribute.** The issue reports `plot_boxes`; a real 0.12 install raises on **`plot_bars`**. The boxplot site had already been fixed, and four others had not. Five `_CategoricalPlotter` methods and two `_DistributionPlotter` ones are wrapped by name and all arrived in 0.13, so fixing them one at a time as they surface would be chasing a list.

Now:

```
ImportError: maidr requires seaborn >= 0.13, but seaborn 0.12.2 is installed.
maidr patches seaborn internals that arrived in 0.13 (the categorical
rewrite), so importing it against an older seaborn fails while those patches
are applied. Upgrade with: pip install --upgrade 'seaborn>=0.13'
```

### What this is not

It does **not** make seaborn 0.12 work — that is option 2 in the issue, still open, and it would need the boxen extraction to read 0.12's nested ladder as well. This replaces an error that says nothing with one that says which version is installed, which is needed, and how to fix it.

### Deliberately permissive

An unreadable or absent `__version__` is allowed through rather than treated as unsupported. Refusing to import on a version string we merely failed to parse would break installs that work, and the check exists to improve a failure — not to create one.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues

Refs #441 — the diagnosis half. Leaving the issue open for option 2 (making 0.12 genuinely work), which is a separate decision with the boxen work attached to it.

## Changes Made

- `maidr/patch/_seaborn_version.py` — new; the check, and why it is tolerant.
- `maidr/patch/__init__.py` — runs it before the patch modules are imported, which is the only point where it can precede the failure.
- `tests/util/test_seaborn_version.py` — 14 tests: version parsing including `.dev0` and `rc1` suffixes, the raise, both permissive paths, and that the check's floor and `pyproject.toml`'s agree, so the two cannot drift into a resolver installing a version the code then refuses.

## Checklist

## Additional Notes

`uv run pytest` — 2256 passed, 13 skipped. `ruff check` shows the same six pre-existing `F401`s as `main`.

Verified both directions against real installs rather than only against the mocked check:

```
seaborn 0.12.2  ->  ImportError, message above
seaborn 0.13.2  ->  import maidr OK
```

The unit tests drive the check with a stub module, since the suite runs on a supported seaborn and cannot install an unsupported one to prove it — the real 0.12.2 environment is what established that the check fires on the right thing.

---

### DIFF
--- a/maidr/patch/__init__.py
+++ b/maidr/patch/__init__.py
@@ -1,5 +1,12 @@
+# Before anything else: the modules below wrap seaborn internals by name,
+# and on a seaborn too old to have them `wrapt` raises an AttributeError
+# that names neither seaborn nor a version (#441).
+from ._seaborn_version import check_seaborn_version
+
+check_seaborn_version()
+
 # Import all patches to ensure they are applied
-from . import (  # noqa: F401
+from . import (  # noqa: E402, F401
     areaplot,
     barplot,
     boxenplot,
--- a/maidr/patch/_seaborn_version.py
+++ b/maidr/patch/_seaborn_version.py
@@ -1,0 +1,78 @@
+"""Fail understandably when seaborn is older than the patches expect.
+
+Nearly every module in this package wraps a seaborn internal by name, and
+the ones under ``_CategoricalPlotter`` and ``_DistributionPlotter``
+arrived with the categorical rewrite in seaborn **0.13**.  On an older
+seaborn ``wrapt`` cannot resolve the attribute and raises while
+``maidr.patch`` is being imported, so ``import maidr`` does not survive::
+
+    AttributeError: type object '_CategoricalPlotter' has no attribute 'plot_bars'
+
+Nothing in that names seaborn, names a version, or suggests what to do,
+and it arrives before any of the user's own code runs.  ``pyproject.toml``
+declares ``seaborn>=0.13``, so a resolver will not choose 0.12 on its own
+-- but `--no-deps`, a conflicting pin elsewhere in the environment, or an
+old lockfile all still land there, which is exactly the case #441 was
+filed about.
+
+This check does not make an old seaborn work.  It replaces an error that
+says nothing with one that says which version is installed, which is
+needed, and that the two do not match.
+"""
+
+from __future__ import annotations
+
+#: The release that introduced the internals the patches wrap.
+MIN_SEABORN = (0, 13)
+
+
+def _parse(version: str) -> tuple[int, ...]:
+    """Return the leading numeric components of a version string.
+
+    Deliberately tolerant: a version this cannot read must not become a
+    second, more confusing import failure than the one being replaced.
+    """
+    parts: list[int] = []
+    for chunk in version.split(".")[:2]:
+        digits = ""
+        for char in chunk:
+            if not char.isdigit():
+                break
+            digits += char
+        if not digits:
+            break
+        parts.append(int(digits))
+    return tuple(parts)
+
+
+def check_seaborn_version() -> None:
+    """Raise a readable ``ImportError`` on a seaborn the patches cannot wrap.
+
+    Raises
+    ------
+    ImportError
+        If the installed seaborn predates :data:`MIN_SEABORN`.
+    """
+    try:
+        import seaborn
+    except ImportError:
+        # Not our problem to report: seaborn is a declared dependency, and
+        # the import that needs it will say so plainly enough.
+        return
+
+    installed = getattr(seaborn, "__version__", "")
+    parsed = _parse(installed)
+    if not parsed or parsed >= MIN_SEABORN:
+        return
+
+    needed = ".".join(str(part) for part in MIN_SEABORN)
+    raise ImportError(
+        f"maidr requires seaborn >= {needed}, but seaborn {installed} is "
+        f"installed. maidr patches seaborn internals that arrived in "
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Both nits taken; the third I agree is not worth the parser complexity.

**Test location** — you are right, and I had put it in `tests/util/` on the strength of "it is a small utility" rather than where the code lives. Moved to `tests/patch/test_seaborn_version.py`, which starts that package with the empty `__init__.py` its siblings (`tests/core/`, `tests/widget/`, `tests/util/`) all carry.

**`__import__("sys")`** — replaced with a plain `import sys`. It was a leftover from writing `_with_seaborn` before the imports existed, and there was no reason to keep it.

**`"0.13.0.dev0"` parsing to `(0, 13)`** — agreed, leaving it. The parser is deliberately coarse because its failure mode should be "let the import through", and a pre-release of the floor version is nearer to supported than to not. Tightening it would mean a full version comparator to serve a case nobody hits.

## On the CodeQL failures

`Analyze (python)` and `Analyze (javascript-typescript)` went red on the previous commit, `Analyze (actions)` passed, and the aggregate `CodeQL` check returned `neutral` rather than a verdict. I could not get the underlying error — the log endpoint returns only the cleanup tail for those jobs — and `rerun-failed-jobs` was refused with `403 This workflow run cannot be retried`.

The discriminator that does not need the log: **this diff is Python-only**, and a Python module cannot break the JavaScript/TypeScript analysis. Combined with `neutral` on the aggregate, that is the same shape as the failures seen across this repo during today's incident, which GitHub has since marked "mitigated ... monitoring to ensure stability".

The push above starts a fresh CodeQL run, which settles it either way. If it goes red again on a Python-only change I will treat it as real and dig further rather than attribute it to the incident twice.

`uv run pytest` — 2256 passed, 13 skipped. `ruff` unchanged from `main`.

---

jooyoungseo:
The second run settles the CodeQL question, and not by repetition — by which job failed.

| check | run 1 (`f0ca2f1`) | run 2 (`24a1f94`) |
|---|---|---|
| `Analyze (python)` | **failure** | success |
| `Analyze (javascript-typescript)` | **failure** | success |
| `Analyze (actions)` | success | **failure** |

The two commits differ by a file move and `__import__("sys")` → `import sys`. **Neither touches a workflow file**, so the workflow YAML that `Analyze (actions)` reads is byte-identical across the two runs — and it passed on one and failed on the other. Meanwhile the two languages that failed first now pass, on a diff that only added Python.

A code defect does not move between languages while the code stays the same. I said I would not attribute a second red to the incident without digging; this is not the same failure twice, it is a different one, and the pair together is the evidence rather than the excuse. `rerun-failed-jobs` is refused with `403 This workflow run cannot be retried` on both runs, so I could not isolate it further.

`claude-review` also failed, in 7 s, producing no review — having posted a complete review of this same branch four minutes earlier. That is the auth fragility already tracked in #305, not a verdict on this diff.

## Everything that tests the code is green on `24a1f94`

```
Python 3.9 / 3.10 / 3.11 / 3.12 tests   success
Accessibility tests in a browser        success
Import with no extras                   success
lint, commit-lint, Lint workflows       success
Analyze (python), Analyze (js-ts)       success
```

The browser job is worth calling out separately: it ran on a real runner, green, on a branch that has now exercised it four times across two PRs.

Merging on that basis, with the reasoning recorded here rather than in a squash message nobody will read. If `Analyze (actions)` keeps failing on `main` after this, it is a repository-level problem rather than a property of this change, and deserves its own issue — I will watch the next few runs and file one if the pattern holds.

---

jooyoungseo:
Closing the loop on the CodeQL prediction, since I said I would watch `main` and file an issue if the pattern held.

**It did not hold.** CodeQL runs on `main`, newest first:

```
19:14Z  success   (after #488)
17:38Z  success   (after this PR merged)
17:09Z  failure
16:13Z  success
15:13Z  success
14:49Z  failure
14:39Z  failure
13:57Z  success
```

Every failure falls inside the incident window that opened at 13:40Z, and both runs since this merged are green. So there is no repository-level problem to file — the reading that the failures were the incident rather than the code is what the subsequent runs show, and I am recording that rather than leaving a prediction unresolved.

Worth keeping the discriminator on record for the next time this happens, because it did not depend on the status page: **the failing language moved between two runs whose workflow files were byte-identical.** A status page tells you the platform is unwell; it cannot tell you whether your specific red check is yours. That one could.

---

Issue #487, opened 2026-08-17
A plotly heatmap navigates upside down, and its highlight breaks when maidr 4.4 lands


jooyoungseo:
The Python-side counterpart of maidr#971, which was just fixed for the
JavaScript plotly adapter in maidr#972. `maidr/plotly/heatmap.py` has the same
defect, and one consequence of the JS fix makes it worse rather than better.

## Summary

The MAIDR grammar's `HeatmapData` runs **top-first**: `y[0]` and `points[0]`
are the row drawn at the top. The core reverses both on construction so its own
row 0 is the bottom of the drawn grid, which is what makes <kbd>↑</kbd> — which
increments the row index — move visually upward.

`PlotlyHeatmapPlot._extract_plot_data` passes plotly's `z` and `y` through
untouched:

```python
z = as_list(self._trace.get("z"))
y = self._trace.get("y", None)
points = []
for row in z:
    points.append([self._to_native(v) for v in row])
```

Plotly numbers a heatmap's rows from the **bottom**, so the layer is emitted
bottom-first and the chart is navigated upside down: the cursor enters at the
top row and <kbd>↑</kbd> walks down.

No value is announced against the wrong label — both arrays are reversed
together, so the pairing survives. What is wrong is the reader's model of where
they are.

## Measured

The emission, through the real class:

```python
fig = go.Figure(go.Heatmap(x=['c1','c2'], y=['first','second','third'],
                           z=[[1,2],[3,4],[5,6]]))
p = PlotlyHeatmapPlot(fig.to_dict()['data'][0], {})
p._extract_plot_data()
# y      -> ['first', 'second', 'third']
# points -> [[1.0, 2.0], [3.0, 4.0], [5.0, 6.0]]
# selector -> '.subplot.xy .heatmaplayer image'
```

Which end plotly draws `y[0]` at, measured on plotly.js 3.7.0 in Chromium via
`_fullLayout.yaxis.c2p` (a smaller pixel is higher on screen):

```
c2p(0) = 266.67   ← 'first'
c2p(2) =  53.33   ← 'third'
```

So `y[0]` is the **bottom** row, and the emitted order is inverted.

## The part that gets worse

The selector is `.heatmaplayer image` — plotly rasterises the grid — so the
core synthesises its own overlay rects in `Heatmap.createOverlayRects`. That
function used to place its row 0 on the **top** band of the image, contradicting
the model's own row-0-is-the-bottom convention.

[...]

PR #491, opened 2026-08-17
fix(plotly): read a heatmap in the order plotly draws it

### LINKED ISSUE #489: A plotly heatmap with categoryorder is read in the data's order, not the drawn one
## What

`PlotlyHeatmapPlot._extract_plot_data` takes its labels straight off the trace:

```python
result[MaidrKey.X] = [self._to_native(v) for v in as_list(x)]
...
y_labels = [self._to_native(v) for v in as_list(y)]
```

Plotly does not necessarily draw them in that order. `xaxis.categoryorder` / `yaxis.categoryorder` re-sort a categorical axis and leave the trace's `x`, `y` and `z` exactly as the author wrote them, so the emitted payload describes a grid plotly is not drawing.

This is the Python counterpart of xability/maidr#985, and the same family as #487 one level up: #487 fixed the *direction* of the y axis and still assumes the trace's order is the axis' order. `categoryorder` breaks that assumption on **both** axes.

## Measured

plotly.js in Chromium, since that is what actually resolves the order. `x = ['charlie', 'alpha', 'bravo']`, columns summing to 3 / 30 / 300 respectively so a value-based sort has something to sort by. Reading `xaxis._categories`, which is the resolved drawn order:

| `categoryorder` | resolved order |
|---|---|
| `'trace'` (default) | `charlie, alpha, bravo` |
| `'category ascending'` | `alpha, bravo, charlie` |
| `'category descending'` | `charlie, bravo, alpha` |
| `'total ascending'` | `charlie, alpha, bravo` |
| `'total descending'` | `bravo, alpha, charlie` |
| `'sum descending'` | `bravo, alpha, charlie` |
| `'mean descending'` | `bravo, alpha, charlie` |

So **every** form applies to a heatmap, including the aggregate ones — `total descending` is the columns in descending sum order (300, 30, 3), and `total ascending` matches the trace here only because the sums happen to already ascend.

Separately, a reversed **x** axis flips the drawn direction without touching the category list, exactly as a reversed y axis does — and `_draws_first_row_at_top` currently asks that question of y only.

## What a reader gets

Every label still sits on its own value, which is why nothing looks broken. What is wrong is where the cells are: arrowing across walks one order while the chart shows another, the panning and braille row are laid out in that same false order, and the core's highlight — which places a cell purely by index over the rendered image — outlines a different cell from the one announced.

## Proposed

Resolve the drawn order from the **declared** layout, the way `_draws_first_row_at_top` already resolves the reversal, and permute both the labels and `points` to match. Then extend the reversal check to x so the columns turn over when the x axis is reversed.

Scope it to what a declared spec can answer exactly:

- **`categoryorder: 'array'` with `categoryarray`** — exact, no computation needed. This is what `plotly.express`'s `category_orders={...}` compiles to, which makes it both the most common path here and the one with no risk.
- **`'category ascending'` / `'category descending'`** — a plain sort of the labels.
- **the aggregate orders** (`total`/`sum`/`mean`/`min`/`max`/`median`, either direction) — **decline**, and leave the trace's order. They are resolvable in principle from `z`, but only by reimplementing plotly's own aggregation and tie-breaking offline; getting that subtly wrong would put the chart back in exactly the confidently-wrong state this is meant to remove, and silently. Better to leave the sort unapplied than to apply a different one.

Also decline when the resolved order is not a permutation of what the trace names. Measured in xability/maidr#985: a `categoryarray` naming categories the trace does not carry makes plotly draw **empty** columns, and `points` has no way to say "empty", so inventing or dropping a column would be worse than keeping the sort unapplied.

Needs a test that would have failed: a heatmap under `category_orders` whose emitted `x`/`y`/`points` match the drawn grid, plus tests pinning each decline path.

jooyoungseo:
Closes #489. Python counterpart of xability/maidr#986.

## The bug

`PlotlyHeatmapPlot._extract_plot_data` took its labels straight off the trace. `categoryorder` sorts a categorical axis and leaves the trace's own `x`, `y` and `z` exactly as the author wrote them, so the emitted payload described a grid plotly was not drawing.

Same family as #487, one level up: #487 fixed the *direction* of the y axis and still assumed the trace's order was the axis' order.

Every label still sat on its own value, which is why nothing looked broken. What was wrong was where the cells were: arrowing across walked one order while the chart showed another, the panning and braille row were laid out in that same false order, and the core's highlight — which places a cell purely by index over the rasterised image — outlined a third.

## Measured

plotly.js rendered in Chromium, since that is what actually resolves the order.

**Which orders re-sort a heatmap** — `x = ['charlie','alpha','bravo']`, column sums 3 / 30 / 300, reading `xaxis._categories`:

| `categoryorder` | resolved order |
|---|---|
| `'trace'` (default) | `charlie, alpha, bravo` |
| `'category ascending'` | `alpha, bravo, charlie` |
| `'category descending'` | `charlie, bravo, alpha` |
| `'total ascending'` | `charlie, alpha, bravo` |
| `'total descending'` | `bravo, alpha, charlie` |
| `'sum descending'` | `bravo, alpha, charlie` |
| `'mean descending'` | `bravo, alpha, charlie` |
| `'min ascending'` | `charlie, alpha, bravo` |
| `'max descending'` | `bravo, alpha, charlie` |
| `'median ascending'` | `charlie, alpha, bravo` |

**Every** form applies, the aggregate ones included.

**When an order applies at all.** Plotly resolves the axis *type* before any order, and a numeric-looking label gives it a **linear** axis, on which `categoryorder` and `categoryarray` are ignored outright:

| `x` | `xaxis` | resolved type | drawn left→right |
|---|---|---|---|
| `[1, 2, 3]` | `array [3, 1, 2]` | **linear** | `1, 2, 3` |
| `[1, 2, 3]` | `type: category`, `array [3, 1, 2]` | category | `3, 1, 2` |
| `['1', '2', '3']` | `array ['3','1','2']` | **linear** | `1, 2, 3` |
| `[1, 3, 'b']` | `array ['b', 3, 1]` | **linear** | `1, 2, 3` |
| `['a', 'c', 'b']` | `array ['c','a','b']` | category | `c, a, b` |

Numeric strings count as numeric, and one numeric label anywhere is enough.

**How an array is defaulted:** a non-empty `categoryarray` with no `categoryorder` resolves to `array`; an empty one stays `trace`; a declared order beats the array.

## The change

`maidr/plotly/heatmap.py` only.

- `_draws_first_row_at_top` → `_axis_runs_backwards(axis_name)`, same body, now asked of x as well. On y "backwards" means first-category-at-top; on x it means the columns run right to left, which was never handled.
- New `_drawn_category_order(axis_name, labels)` resolves the sort from the **declared** layout — there is no browser here — and returns indices into the trace's labels in drawn order, or `None` to decline.
- `_extract_plot_data` composes the two: take the origin order (bottom for y, left for x), turn the rows over unless y is reversed, turn the columns over only when x is, and permute `points` by the same mapping so no value leaves its label.

**Resolved only where a declared spec can answer exactly:**

- `'array'` with a `categoryarray` — what `plotly.express`'s `category_orders={...}` compiles to, and the common path;
- `'category ascending'` / `'category descending'` — a plain sort of the labels;
- **the aggregate orders are declined.** Rebuilding plotly's aggregation and tie-breaking offline would trade one silent wrongness for another.

**Declines** when the axis is not categorical (any numeric-looking label without a declared `type: "category"`), when the resolved order is not a permutation of what the trace names (a superset, a subset, an unknown name, a repeated array entry, a repeated trace label), and when the grid is ragged — a ragged `z` has no column to move a value to, so neither the sort nor the reversal touches its columns.

## Verification

```
uv run pytest tests/plotly -q   → 915 passed (38 new)
uv run pytest -q                → 2296 passed, 20 skipped
ruff check / format             → clean on both changed files
```

Falsified per fix rather than in aggregate, so each test is tied to the change it covers: reverting the emission fails 8 of the new cases; reverting the ragged-column guard, the implicit-array default, the permutation check, the numeric normalisation, or the axis-type guard each fails **exactly** its own cases and leaves the rest green.

## Review history worth keeping

Four rounds found four real defects, one of which I introduced mid-PR:

1. a **ragged `z` with a reversed x axis crashed** with `IndexError` — a regression this PR introduced, since the reversal was ungated where the sort was gated;
2. a `categoryarray` with no `categoryorder` beside it was declined, and plotly draws it;
3. a repeated `categoryarray` entry passed the length check and silently emitted one column twice while dropping another;
4. **numeric labels declined every sort**, because `_to_native` floats them and the array was compared raw — and chasing that turned up that the earlier "fix" for it asserted behaviour plotly does not have, so the axis-type guard above replaced it.

## Not verified

- The resolved orders come from plotly.js in a browser; the tests drive real `go.Figure` traces against hand-written layouts shaped like what was measured.
- `ruff check maidr tests` reports 6 pre-existing `F401` re-export warnings in `maidr/core/`. Untouched here; both changed files pass on their own.
- **The mixed-label case declines** rather than resolving. Plotly's own rule is a proportion of numeric values, and reproducing that offline is the same trap as the aggregate orders — so this errs toward leaving a sort unapplied.
- **A label list whose length does not match the grid** keeps the plain reversal it would have had, so malformed input behaves as it did before.


jooyoungseo:
All three were real. Fixed in `3c16b5b`, each with its own test and each falsified separately.

## 1. The ragged `z` crash — confirmed, and it was a regression I introduced

Traced it exactly as you described. `width` comes from `points[0]`, so a ragged grid gives `rectangular = False` and `cols` stays `[0, 1, 2]` — and then the ungated reversal flipped it to `[2, 1, 0]`, `columns_moved` went true, and `points[1][2]` raised on a two-wide row.

Worth naming what that means: before this PR a ragged `z` with a reversed x axis merely reversed the rows and carried on. My change turned that into an unhandled exception. Not a pre-existing gap — a regression, and the worst kind, since it only fires on the combination.

Fixed by moving the reversal inside the `rectangular` guard alongside the sort, which is what the comment beside them already claimed:

```python
if rectangular:
    if x_labels is not None and len(x_labels) == width:
        ...
    if x_backwards:
        cols.reverse()
```

Three tests, since you were right that the path had no coverage at all: reversed x, sorted x, and the plain case — the last to pin that a ragged grid still turns its *rows* over, which stays well defined.

## 2. Implicit `categoryorder: "array"` — you were right, and I measured it

I did not want to take this one on the strength of reading plotly's defaults, so I rendered it in Chromium and read the resolved `categoryorder` back:

| layout | resolved `categoryorder` | drawn order |
|---|---|---|
| `categoryarray` only | **`array`** | `bravo, charlie, alpha` |
| `categoryarray` + `categoryorder: 'array'` | `array` | `bravo, charlie, alpha` |
| neither | `trace` | `charlie, alpha, bravo` |
| `categoryarray: []` | `trace` | `charlie, alpha, bravo` |
| `categoryarray` + `categoryorder: 'category ascending'` | `category ascending` | `alpha, bravo, charlie` |

So the rule is exactly as you said, with two edges worth pinning: an **empty** array does not trigger it, and a **declared** order beats the array. The code now resolves `array` when one was not declared and the array is non-empty, and there are tests for all three rows.

You were also right that the cost was only a missed opportunity rather than corruption — but a hand-built `go.Figure` setting just `categoryarray` is an ordinary thing to write, so it was a real gap.

## 3. The repeated `categoryarray` entry — confirmed

`['charlie', 'alpha', 'alpha']` against `['charlie', 'alpha', 'bravo']` resolves to `[0, 1, 1]`: right length, every name known, not a permutation. It would have emitted `alpha`'s column twice and dropped `bravo` entirely — a heatmap missing a column and showing another twice, with nothing to indicate it. That is precisely the failure mode this whole change exists to remove, so contrived input or not, it is worth the one line:

```python
if len(set(resolved)) != len(labels):
    return None
```

## Verification

```
uv run pytest tests/plotly -q   → 908 passed (31 new, up from 24)
uv run pytest -q                → 2289 passed, 20 skipped
ruff check / format             → clean on both changed files
```

Falsified one at a time rather than together, so each test is tied to its own fix: reverting any single one of the three fails **exactly** its own new case and leaves the other 907 green.

---

jooyoungseo:
You flagged this as plausible-but-unverified and asked for a runtime check. I ran it. **The first half reproduces; the second does not.**

## Confirmed: numeric labels declined the sort silently

Exactly your reasoning, exactly your repro:

```python
fig = go.Figure(go.Heatmap(x=[1, 2, 3], y=["r1", "r2"], z=[[1, 2, 3], [4, 5, 6]]))
layout = {"xaxis": {"categoryorder": "array", "categoryarray": [3, 1, 2]}}
```

```
emitted x            : [1.0, 2.0, 3.0]
expected if sorted   : [3.0, 1.0, 2.0]
```

`_to_native` floats the labels, so `position` was keyed `"1.0"/"2.0"/"3.0"` while `drawn` was `["3","1","2"]`. Every lookup missed and the function returned `None`.

Worth being blunt about how bad a miss that is: `categoryorder: "array"` is what `plotly.express`'s `category_orders` compiles to, and the PR called it "the common path and the risk-free one" — so the change did not work on the case it was mainly written for, whenever the categories were numbers. Fixed by putting `declared` through `_to_native` too, so both sides normalise identically:

```python
drawn = [str(self._to_native(v)) for v in declared]
```

Three tests: an array of numbers sorting, a numeric array with no `categoryorder` beside it, and a numeric array naming a category the trace lacks still declining.

## Not confirmed: the typed-array export

I checked this rather than defending against it blind:

```python
fig.update_xaxes(categoryorder="array", categoryarray=[3, 1, 2])
fig.to_dict()["layout"]["xaxis"]
# → {'categoryorder': 'array', 'categoryarray': [3, 1, 2]}
```

A plain list, not `{dtype, bdata}`. So the compaction you were thinking of does not apply to this layout key, at least here — reasonable inference from how trace arrays behave, but it does not hold.

I routed it through `as_list` anyway, for a reason unrelated to the one you gave: `as_list` passes a list through untouched and turns `None` into `[]`, which lets both emptiness checks drop their `isinstance` guards and read as `if declared:`. Not insurance against a failure mode I could not reproduce.

## Verification

```
uv run pytest tests/plotly -q   → 911 passed (34 new)
uv run pytest -q                → 2292 passed, 20 skipped
ruff check / format             → clean
```

Falsified: comparing raw again fails exactly the two new cases that need the normalisation, and leaves the other 909 green.

Thanks for pushing on a finding you could not run — it was the one real gap left, and the suite's all-strings fixture would never have shown it.

---

jooyoungseo:
Your minor note turned out to be the most important thing on this PR. Chasing it found that **the numeric case I added two commits ago asserts behaviour plotly does not have** — that commit made this wrong rather than right, and I want to be plain about it since it briefly shipped past a review that had already approved.

## What I measured

You asked whether lexicographic sorting matches what plotly draws for numeric categories. The answer turned out to be upstream of that question: **plotly resolves the axis *type* before it resolves any order, and a numeric-looking label gives it a linear axis** — on which `categoryorder` and `categoryarray` are ignored outright.

| `x` | `xaxis` | resolved type | drawn left→right |
|---|---|---|---|
| `[1, 2, 3]` | `array [3, 1, 2]` | **linear** | `1, 2, 3` |
| `[1, 2, 3]` | `type: category`, `array [3, 1, 2]` | category | `3, 1, 2` |
| `['1', '2', '3']` | `array ['3','1','2']` | **linear** | `1, 2, 3` |
| `[1, 3, 'b']` | `array ['b', 3, 1]` | **linear** | `1, 2, 3` |
| `['a', 'c', 'b']` | `array ['c','a','b']` | category | `c, a, b` |

So numeric strings count as numeric, and a numeric label *anywhere* is enough — the mixed row surprised me too.

My `test_sorts_them_by_an_array_of_numbers` asserted `x == [3.0, 1.0, 2.0]`. Plotly draws `1, 2, 3`. The previous commit read the array in exactly the case plotly ignores it and **reordered a chart plotly had not reordered** — which is the failure this entire change exists to remove, arrived at from the other side. Worse than the gap it was fixing, because it fired on ordinary numeric data rather than on an unusual layout.

## The fix

Decline unless the axis is genuinely categorical: `type: "category"` declared, or no label that parses as a number.

```python
if axis.get("type") != "category" and any(
    self._looks_numeric(label) for label in labels
):
    return None
```

The mixed case declines too. That is conservative rather than exact — plotly's own rule is a proportion of numeric values, and reproducing that offline is the same trap as the aggregate orders — but it errs toward leaving a sort unapplied rather than inventing one, which is the direction the rest of this change already errs in.

Tests rewritten to assert the measured behaviour, plus one for the `type: "category"` case that *does* sort, and one for mixed labels.

## Worth noting for the core adapter

xability/maidr#986 does the same job by reading a **rendered** `_fullLayout._categories`, and it is naturally immune to this: a linear axis reports `_categories: []`, so its length check declines before anything else runs. The asymmetry is the whole difference between reading a chart plotly has drawn and reading one it has only been told about — the rendered side gets plotly's type resolution for free, and the declared side has to infer it. Nothing to change there; just the reason the same guard is not needed on both.

## Your two coverage gaps

Both added, and both were the right things to ask for:

1. **y sort composed with a reversed y** — the asymmetry runs opposite to x (the resolved order counts from the bottom, so a *reversed* y is the case left alone), which is exactly the kind of thing a later change flips without noticing.
2. **ragged grid with a sorted y** — pins that row moves stay well defined however ragged the rows are, which is why the `rectangular` guard covers columns only.

## Verification

```
uv run pytest tests/plotly -q   → 915 passed
uv run pytest -q                → 2296 passed, 20 skipped
ruff check / format             → clean
```

Falsified: dropping the guard fails exactly the three cases that need it.

---

jooyoungseo:
All three taken; the third was a real difference. `b3e12ea`.

## `nan` / `inf` — you were right to ask

I had encoded the numeric test from Python's parser rather than from plotly's, and they disagree. Measured:

| `x` | `xaxis` | resolved type | array honoured |
|---|---|---|---|
| `['nan', 'inf', 'zeta']` | `array ['zeta','nan','inf']` | **category** | **yes** |
| `['1e10','2e10','3e10']` | `array ['3e10','1e10','2e10']` | linear | no |
| `['aa','bb','cc']` | `array ['cc','aa','bb']` | category | yes |

So exponent form *is* numeric to plotly, and the two special names are not — the JavaScript coercion it tests with rejects them where `float()` accepts them. My guard therefore declined a sort plotly actually applies, on a heatmap whose categories happen to be called `nan` or `inf`.

Narrow, and the cost was a *missed* sort rather than a wrong one — the safe direction. But the rule was wrong for a reason worth fixing rather than documenting: it was derived from the wrong language's parser. `math.isfinite(float(value))` reproduces plotly's answer on all three rows, and both are now pinned by tests.

## The other two

- **Duplicate `self._layout.get(axis_name, {})`** — left as is. Passing the resolved axis dict in would spread knowledge of the layout's shape to the call site, and at one dict lookup per axis per heatmap the cost is not what is buying the coupling. Happy to change it if you would rather.
- **61 commits on the branch** — squash-merging, as you suggest and as the rest of this series has been.

## Verification

```
uv run pytest tests/plotly -q   → 917 passed (40 new)
uv run pytest -q                → 2298 passed, 20 skipped
ruff check / format             → clean
```

Falsified: parsing without the finiteness check fails exactly the `nan`/`inf` case and leaves the other 916 green.

That is four rounds of review on this PR and four real defects, one of which I introduced mid-PR and one of which I had to be pushed on twice. The pattern in all four is the same and worth naming: every one came from encoding a rule about plotly from something *other* than plotly — its docs, Python's semantics, or my own reasoning — instead of rendering the case and reading the answer.

---

PR #493, opened 2026-08-17
fix(plotly): decline a category order on a date axis too

jooyoungseo:
Follow-up to #491, which is merged. A reviewer there raised this as unmeasured and unguarded; it is real, so #491 as it stands applies a declared order to a chart plotly draws chronologically.

## The gap

#491 guards against plotly resolving a **linear** axis from numeric-looking labels, on which `categoryorder` and `categoryarray` are ignored. It does not guard against a **date** axis, which ignores them the same way.

## Measured

plotly.js in Chromium, `categoryarray` declared on every case:

| labels | resolved type | array honoured |
|---|---|---|
| `2024-03-01`, `2024-01-01`, `2024-02-01` | **date** | no — drawn chronologically |
| `2024-03`, `2024-01`, `2024-02` | **date** | no |
| `2024-03-01 12:00`, … | **date** | no |
| `2024/03/01`, … | category | yes |
| `03-01-2024`, … | category | yes |
| `Mar 2024`, `Jan 2024`, `Feb 2024` | category | yes |
| `Q1`, `Q2`, `Q3` | category | yes |
| `2024-13-45`, `2024-99-99`, … | category | yes |

Two things worth having in the record:

1. **Year-first and hyphen-separated is the whole rule.** Slashes, day-first, and month names all stay categories and keep their array — so the guard has to be narrow or it declines sorts that do apply.
2. **Plotly validates the parts, not the shape.** `2024-13-45` is still a category. So the pattern checks month `01`–`12` and day `01`–`31` rather than matching four digits and a dash, or it would decline a legitimate set of category names that merely look date-shaped.

## The change

`maidr/plotly/heatmap.py` only — a `_looks_like_date` companion to `_looks_numeric`, and one clause added to the existing guard:

```python
if axis.get("type") != "category" and any(
    self._looks_numeric(label) or self._looks_like_date(label)
    for label in labels
):
    return None
```

A declared `type: "category"` still brings the order back, exactly as it does for numbers, and there is a test for that direction too.

## Verification

```
uv run pytest tests/plotly -q   → 925 passed (8 new)
uv run pytest -q                → 2306 passed, 20 skipped
ruff check / format             → clean
```

Falsified: dropping the date half of the guard fails exactly the three ISO cases and leaves the other 922 green.

## Not verified

- The measurements are plotly.js in a browser; the tests drive real `go.Figure` traces against hand-written layouts shaped like what was measured.
- **The mixed case still declines conservatively** — one date-looking or number-looking label anywhere is enough. Plotly's own rule is a proportion of the values, and reproducing that offline is the same trap as the aggregate orders, so this errs toward leaving a sort unapplied.
- **This enumerates the axis types plotly infers rather than deriving them.** Numeric and date are the two that turned up; a third inferred type would need the same treatment, and the guard says so in a comment rather than pretending to be general.

---


jooyoungseo:
All three edge cases measured, and **two of them were wrong in the direction that matters**. `54604bd`.

| label | plotly | the pattern said | |
|---|---|---|---|
| `2024-3-1` | **date** | not a date | **missed** — applies an order plotly ignores |
| `' 2024-03-01'` | **date** | not a date | **missed** — same |
| `2024-02-30` | category | a date | over-declined — loses a sort that applies |
| `2024-04-31` | category | a date | same |
| `2024-03-01T12:00:00` | date | a date | already right |

You framed the single-digit case as "a false negative rather than a false positive", and that is exactly why it was the one to chase: a false negative here *applies* a reordering to a chart plotly draws chronologically. That is the confidently-wrong failure, not a missed opportunity.

The per-month point was right too. Plotly is **more lenient about the shape and stricter about the parts** than a single regex can express: it takes what a person would plausibly write, and rejects what a calendar would not. So the pattern now matches the shape loosely and hands the parts to `datetime.date`, which rejects February 30th for the same reason plotly does:

```python
match = _ISO_DATE.match(value.strip())
...
year, month, day = match.groups()
try:
    date(int(year), int(month), int(day) if day else 1)
except ValueError:
    return False
```

`.strip()` covers the whitespace case, and a year-month with no day defaults to the 1st, since `2024-03` is a date to plotly.

**On the duplication nit** — you are right that the table now appears in three places. I have cut the module-level copy down to a pointer at the method, so the table lives once next to the code that acts on it, with the test class keeping its own copy of what it exercises.

## Verification

```
uv run pytest tests/plotly -q   → 929 passed (12 new)
uv run pytest -q                → 2310 passed, 20 skipped
ruff check / format             → clean
```

Falsified: restoring the pattern-only check fails exactly those three cases — the two it missed and the one it over-declined — and leaves the other 926 green.

This is the fifth review finding on this pair of PRs that came down to *measure it rather than reason about it*, and the second where reasoning had produced something worse than the gap it was closing. Thanks for pushing on the corners.

---

Issue #293, opened 2026-07-29
Split maidr/util/dependencies.py into focused modules


jooyoungseo:
Follow-up from #291.

`maidr/util/dependencies.py` grew from ~170 lines to ~1,050 during that PR and now owns three separable concerns:

1. **Bundled asset access** — `bundled_js_path`, `read_bundled_js`, `maidr_html_dependency`, `maidr_js_version`
2. **CDN version resolution** — `get_cdn_version`, `set_cdn_version`, the resolver endpoints, the timeout budget, the semver guards
3. **Bundle staleness reporting** — `bundle_status`, `BundleStatus`, `warn_if_bundle_is_stale`, `STALE_MINOR_GAP`

## Suggested shape

`maidr/util/cdn.py` and `maidr/util/bundle_freshness.py`, leaving asset access in `dependencies.py`. That also makes the test files map 1:1 onto source modules — `tests/core/test_cdn_version.py` and `tests/core/test_bundle_freshness.py` already split along exactly these lines.

## Why it was deferred rather than done in #291

Six separate reviewers raised it, and I agreed each time — but it is pure code motion in a file that had accumulated many rounds of review. Doing it inside that PR would have meant re-reviewing the whole file for *movement* instead of *behaviour*, which is a poor trade at the end of a long review. It is much safer as its own change with its own diff.

## Notes for whoever picks it up

- `maidr/api.py` currently imports 8 names from `dependencies`; the split should reduce that.
- `tests/core/test_cdn_version.py::test_no_render_path_references_the_unresolved_constants` scans the package for `MAIDR_JS_CDN_URL` / `MAIDR_CSS_CDN_URL` outside their home module — it will need its exclusion updated to whichever module ends up owning them.
- `_warn_once` and its bookkeeping are shared by both the CDN and staleness sides, so it needs a deliberate home rather than being duplicated.

jooyoungseo:
First cut landed in #494 (`c68e197`): `bundle_capability.py`. Recording the state and the plan here, because a review of that PR made the fair point that the "Moved out of this module" section in `dependencies.py` will drift from the actual plan unless the plan is written down somewhere.

## Where it stands

| concern | state |
|---|---|
| Bundled asset access | stays in `dependencies.py` |
| **Bundled-copy capability** | **moved — `maidr/util/bundle_capability.py`** (263 lines) |
| Bundle staleness reporting | still in `dependencies.py` — **next** |
| CDN version resolution | still in `dependencies.py` — last |

[...]

PR #494, opened 2026-08-17
refactor(util): move the bundle capability check into its own module

jooyoungseo:
## Description

First cut of the split #293 asks for. `maidr/util/dependencies.py` had grown to **2248 lines** owning three separable concerns; this takes the one that answers *"can the bundled `maidr.js` draw what we are about to hand it"* — `bundle_trace_types`, `warn_if_bundle_cannot_render`, `schema_trace_types`, `MaidrBundleTraceWarning` and their private helpers.

The seam is one the tests already drew. `tests/core/test_bundle_capability.py` (16 tests) and `tests/core/test_bundle_freshness.py` (23 tests) split along exactly this line before the module did, which is the 1:1 source↔test mapping #293 wants.

Deliberately one cut rather than the whole three-way split: this is pure code motion, and a 2248-line file is where a motion mistake hides. One module at a time keeps each diff reviewable as movement.

## What stays behind, and why

The section boundary in the file is not quite the dependency boundary, so three things did **not** move with the block they sit in:

- **`_RELEASE_RE`** — defined inside the moved region, but `_version_key` uses it and `_version_key` stays. It is a version-parsing primitive, not a capability one. Moving it would have broken every version comparison in the package.
- **`_bundle_warning_enabled` / `BUNDLE_WARNING_ENV_VAR`** — the warning-suppression policy is shared with the freshness side, so it belongs to neither module alone. Imported rather than duplicated; it wants a deliberate home when the freshness half moves, the same question #293 raises about `_warn_once`.

## Keeping the old import path working

`maidr.util.dependencies` still resolves the moved names, through a lazy module `__getattr__` (PEP 562) rather than a re-export at the bottom of the file:

```python
def __getattr__(name):
    if name in _MOVED_TO_BUNDLE_CAPABILITY:
        from maidr.util import bundle_capability
        return getattr(bundle_capability, name)
    raise AttributeError(...)
```

`bundle_capability` imports **from** `dependencies`, so an eager re-export would make the cycle real: whichever module a program imported first would find the other half-initialised. Resolving on attribute access means the import only runs once both are loadable. Verified in both orders, including `import maidr.util.bundle_capability` before anything touches `dependencies`.

### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor (no behaviour change)

## Related Issues

Refs #293. Does not close it — the CDN-resolution and freshness halves remain, and the notes below are what the next cut has to deal with.

## Changes Made

- `maidr/util/bundle_capability.py` — new; 237 lines moved verbatim, with its own imports and logger.
- `maidr/util/dependencies.py` — 2248 → 2011 lines; gains the `__getattr__` shim; loses a `collections.abc.Iterable` import that became unused when its only user moved.
- `tests/core/test_bundle_capability.py` — repointed at the new module. It monkeypatches module-private state (`_bundle_trace_warned`, `_bundle_tokens`) and asserts on the logger name, all of which are tied to module identity.

## Checklist

## Additional Notes

`uv run pytest` — **2265 passed, 13 skipped**, identical to `main`. `ruff check` reports the same **6** pre-existing findings as `main` and no new ones.

No new tests: this is motion, and the 16 that already cover the moved code are the check that it still behaves. They were repointed, not rewritten.

**For whoever takes the next cut.** Two things cost me most of the time here, and neither is visible from the issue:

1. The extraction initially failed with 97 test failures from a single missing import — `warn_if_bundle_cannot_render` calls `_bundle_warning_enabled`, which lives in the *freshness* section. Every render path calls it, so one `NameError` looked like a catastrophe. Worth checking cross-section helpers before assuming a section is self-contained.
2. Only one test file couples to module identity, but it does so in three ways at once (private attributes, a `cache_clear`, and a `caplog` logger name). The freshness half will have the same shape.

---

### DIFF
--- a/maidr/__init__.py
+++ b/maidr/__init__.py
@@ -210,6 +210,7 @@ from .patch import (  # noqa: E402, F401
     mplfinance,
     violinplot,
 )
+from .util.bundle_capability import MaidrBundleTraceWarning  # noqa: E402
 from .util.dependencies import (  # noqa: E402
     BUNDLE_WARNING_ENV_VAR,
     BUNDLED_TAG,
@@ -220,7 +221,6 @@ from .util.dependencies import (  # noqa: E402
     BundleStatus,
     ResolverOutcome,
     MaidrBundleStaleWarning,
-    MaidrBundleTraceWarning,
     bundle_status,
     resolver_outcome,
     bundled_css_path,
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -24,6 +24,10 @@ from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.plot import MaidrPlot
 from maidr.core.plot.barplot import BarPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
+from maidr.util.bundle_capability import (
+    schema_trace_types,
+    warn_if_bundle_cannot_render,
+)
 from maidr.util.dependencies import (
     MAIDR_JS_FILENAME,
     OFFLINE_FALLBACK_REPORT,
@@ -33,8 +37,6 @@ from maidr.util.dependencies import (
     maidr_bundled_relative_dir,
     maidr_html_dependency,
     maidr_js_cdn_url,
-    schema_trace_types,
-    warn_if_bundle_cannot_render,
     warn_if_bundle_is_stale,
 )
 from maidr.util.environment import Environment
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -35,6 +35,10 @@ from maidr.plotly.step_shape import (
     is_step_trace,
     renders_through_webgl,
 )
+from maidr.util.bundle_capability import (
+    schema_trace_types,
+    warn_if_bundle_cannot_render,
+)
 from maidr.util.dependencies import (
     MAIDR_JS_FILENAME,
     OFFLINE_FALLBACK_REPORT,
@@ -44,8 +48,6 @@ from maidr.util.dependencies import (
     maidr_bundled_relative_dir,
     maidr_html_dependency,
     maidr_js_cdn_url,
-    schema_trace_types,
-    warn_if_bundle_cannot_render,
     warn_if_bundle_is_stale,
 )
 from maidr.util.environment import Environment
--- a/maidr/util/bundle_capability.py
+++ b/maidr/util/bundle_capability.py
@@ -1,0 +1,263 @@
+"""What the bundled ``maidr.js`` is able to draw.
+
+Split out of :mod:`maidr.util.dependencies` (#293), which had grown to own
+three separable concerns. This is the one that answers "can the copy in
+this wheel render the chart we are about to hand it" -- a question that
+distance in version numbers cannot answer, and that the reader who most
+needs the answer (``use_cdn=False``, offline) is the one a staleness
+warning cannot reach.
+
+Its tests live in ``tests/core/test_bundle_capability.py``, which already
+split along this line before the module did.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+import threading
+import warnings
+from collections.abc import Iterable
+
+from maidr.util.dependencies import (
+    BUNDLE_WARNING_ENV_VAR,
+    _bundle_warning_enabled,
+    bundled_js_path,
+    maidr_js_version,
+)
+
+_logger = logging.getLogger(__name__)
+
+#: An enum member assignment, which is how the bundle spells a trace type.
+#: TypeScript compiles ``TraceType`` to a run of them, so the 4.3.0 bundle
+#: carries ``e.AREA=`area`,e.ALLUVIAL=`alluvial`,e.BAR=`bar`,...``.
+#:
+#: All three quote characters, because the minifier picks whichever is
+#: shortest for the surrounding context -- the 4.1.0 bundle uses backticks
+#: throughout, and a check written for double quotes alone would have
+#: reported every type missing.
+#:
+#: Anchored on the ``.UPPER_SNAKE =`` rather than matching every quoted
+#: lower-snake string anywhere. The looser form was tenable while the bundle
+#: was small, and stopped being so at 4.3.0: nine adapters gained chart types
+#: and the vocabularies that came with them (MathML attribute names among
+#: others) took the scrape from a few hundred tokens to **1265**, including
+#: ordinary words like ``count``, ``above`` and ``accent``. A capability
+#: check whose vocabulary contains most of the dictionary answers "yes, your
+#: bundle knows that" to anything, so it can no longer raise the warning it
+#: exists for -- failing silent rather than merely failing safe (#436).
+#: Narrowed, 4.3.0 yields 60.
+_BUNDLE_TOKEN_RE = re.compile(
+    r"\.[A-Z][A-Z0-9_]*\s*=\s*(?P<quote>[`\"'])"
+    r"(?P<token>[a-z][a-z0-9_]{1,39})(?P=quote)"
+)
+
+#: Parsed once; the bundle does not change under a running process.
+_bundle_tokens: frozenset[str] | None = None
+_bundle_token_lock = threading.Lock()
+
+#: Which (severity, trace type) pairs have already been reported, so a
+#: chart rendered in a loop says it once per type rather than per render.
+#: Its own lock: it shares no invariant with the token cache, and the
+#: staleness warning next door already keeps these two concerns apart.
+_bundle_trace_warned: set[tuple[bool, str]] = set()
+_bundle_trace_lock = threading.Lock()
+
+
+class MaidrBundleTraceWarning(UserWarning):
+    """Raised when the bundled ``maidr.js`` cannot draw an emitted layer.
+
+    Distinct from :class:`MaidrBundleStaleWarning`, which is about age.
+    This one is about capability, and it is the question a reader actually
+    has: a bundle five minors behind may render everything, and one minor
+    behind may render none of a newly added type.
+
+    Re-exported as ``maidr.MaidrBundleTraceWarning``, so a consumer running
+    under ``-W error`` can silence this advisory alone::
+
+        warnings.filterwarnings("ignore", category=maidr.MaidrBundleTraceWarning)
+    """
+
+
+def bundle_trace_types() -> frozenset[str]:
+    """Return the trace types the bundled ``maidr.js`` mentions.
+
+    Read straight out of the shipped bundle rather than from a table kept
+    in step by hand.  A table costs one line per new plot type and forgetting
+    that line is silent in exactly the way this exists to prevent; the
+    bundle already knows the answer, because its factory switches on these
+    strings.
+
+    Returns
+    -------
+    frozenset of str
+        Every quoted lower-snake identifier in the bundle.  Empty when the
+        bundle cannot be read, which callers must treat as "cannot tell"
+        rather than as "renders nothing".
+
+    Notes
+    -----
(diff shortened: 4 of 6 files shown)


jooyoungseo:
Took the test-coverage point — it was the right thing to catch. Four tests added in `6b4f7f7`, covering exactly the three things you listed plus the import order:

| test | what breaks without it |
|---|---|
| `test_the_old_import_path_still_resolves` (×4, parametrized) | a name silently dropped from the shim's set |
| `test_the_shim_returns_the_object_rather_than_a_copy` | identity — `except MaidrBundleTraceWarning` and `pytest.warns` compare class objects, so a lookalike passes `is not None` and still fails to catch the warning it names |
| `test_an_unknown_attribute_still_raises_attribute_error` | your `hasattr` point — asserts both the raise and `not hasattr(...)` |
| `test_importing_the_new_module_first_does_not_deadlock_the_old_one` | the cycle the shim exists to avoid |

The last one runs in a **subprocess**, and that is not incidental: within the test session both modules are long since imported, so import order cannot be observed from inside it. A test written in-process would have passed against an eager re-export too — which is to say, it would have tested nothing.

You were right that the manual verification did not count. Since these tests are themselves the only thing standing behind the shim, I mutation-tested them rather than trusting that they pass:

```
mutation 1 — __getattr__ returns None instead of raising
  FAILED test_an_unknown_attribute_still_raises_attribute_error

mutation 2 — "schema_trace_types" removed from _MOVED_TO_BUNDLE_CAPABILITY
  ImportError: cannot import name 'schema_trace_types' from 'maidr.util.dependencies'
```

Mutation 2 is worth reading closely: it does not fail the new test, it fails at **collection**, because `maidr/core/maidr.py` imports that name at module scope. So the shim's completeness was already load-bearing before I wrote a line — the suite could not even start without it. The new tests add the cases that would *not* have announced themselves that loudly.

On the `_bundle_warning_enabled` wart — agreed it is a wart, and I would rather it stayed visible than be papered over by duplicating the policy on both sides. It gets a real home when the freshness half moves.

`uv run pytest` — **2272 passed, 13 skipped** (2265 + the 7 new). `ruff check` unchanged from `main` at 6 pre-existing findings.

---

jooyoungseo:
All three taken (`ec5ebf6`). The third was a correction rather than a suggestion, so I checked it before agreeing — and you were right.

## The docstring was overclaiming, and here is the proof

I claimed the subprocess test exercised "``bundle_capability`` before ``dependencies``, the order that would trip an eager re-export". It does not:

```console
$ python -c "import sys; import maidr.util.bundle_capability; print([m for m in
    ('maidr','maidr.util.dependencies','maidr.util.bundle_capability') if m in sys.modules])"
['maidr', 'maidr.util.dependencies', 'maidr.util.bundle_capability']

$ python -c "import sys; import maidr; print('maidr.util.bundle_capability' in sys.modules)"
True
```

Exactly your reasoning: the parent package initialises first, `maidr/__init__.py` imports `.util.dependencies`, and `maidr/core/maidr.py` pulls a moved name through the shim on the way — so `bundle_capability` is loaded before the line that names it, and is already loaded after a bare `import maidr`.

Worth stating the stronger conclusion: **that ordering is unobservable from inside this package at all.** Any entry point initialises `maidr` first. So there is no version of this test that checks what I said it checked, and rewording it is the only honest fix rather than a concession. The test is now named for what it does — a cold interpreter imports the package and reaches a moved name from its old home — and the docstring says plainly what it does *not* prove, and why the subprocess is still needed (within the session, nothing about import time is observable).

The laziness the shim provides is real and load-bearing regardless; it is just not what this test demonstrates.

## The other two

**Internal call sites** — agreed, and a good catch on the framing: the shim's docstring says "an import path *other code* already uses", which is a claim about external consumers. Leaving first-party code routed through it made that sentence false. `maidr/core/maidr.py` and `maidr/plotly/plotly_maidr.py` now import `schema_trace_types` and `warn_if_bundle_cannot_render` from `maidr.util.bundle_capability` directly, so the shim is purely the back-compat surface it was designed to be, and the decoupling is real for the package's own code rather than only for its tests.

**Blank lines** — collapsed 4 → 2. Confirmed with `cat -n` rather than by eye, since as you note `E3xx` is not enabled and nothing would have caught it.

`uv run pytest` — **2272 passed, 13 skipped**. `ruff check` unchanged from `main` at 6 pre-existing findings.

Merging on green.

---

jooyoungseo:
Both taken (`99f1e80`) — and one correction to the review itself, which I would rather raise than quietly accept.

**`maidr/__init__.py`** — you are right that this was the inconsistent one, and the sharpest version of the point is that it is the package's *own public re-export*: `maidr.MaidrBundleTraceWarning` is the name users actually reach for, and routing it through a back-compat shim made the shim load-bearing for the canonical path. Now `from .util.bundle_capability import MaidrBundleTraceWarning`. Verified the identity survives, since `except maidr.MaidrBundleTraceWarning` depends on it:

```
public re-export is the canonical class: True
still in __all__: True
```

**The stale docstring reference** — fixed, and qualified rather than just repointed. It now names `maidr.util.bundle_capability.warn_if_bundle_cannot_render` in full and says outright that this file no longer defines it but still resolves it. A bare rename would have removed the trap for someone reading the docstring while leaving it for someone grepping the file.

## One thing in the review I have to push back on

> a subprocess test that actually exercises cold-interpreter import order

It does not, and I retracted exactly that claim two comments ago. `import maidr.util.bundle_capability` initialises the parent package first, which imports `.util.dependencies`, and `maidr/core/maidr.py` pulls a moved name through the shim on the way — so both modules are loaded before the line that names either:

```console
$ python -c "import sys; import maidr; print('maidr.util.bundle_capability' in sys.modules)"
True
```

That ordering is unobservable from inside this package at all. The test's value is narrower and real — a cold interpreter imports the package and reaches a moved name from its old home — and its docstring now says so, including what it does not prove. Flagging it because a merged PR whose thread endorses a claim its own code comments retract is worse than either version alone.

The `__getattr__` design is correct regardless of ordering; that part of the review stands.

`uv run pytest` — **2272 passed, 13 skipped**. `ruff check` unchanged from `main` at 6 pre-existing findings. Merging on green.

---

jooyoungseo:
Your import-order trace is right, and it means **something I told the previous reviewer is now false.** Worth correcting on the record, because the change that made it false was made in this PR.

Two comments ago I said the ordering was unobservable from inside the package. That was true then. Repointing the internal call sites — the previous round's suggestion — inverted the graph:

```console
$ python -X importtime -c "import maidr" | grep -E "maidr\.util\.(dependencies|bundle_capability)$"
import time:  2102 | 15198 |           maidr.util.dependencies
import time:   428 | 15626 |         maidr.util.bundle_capability
```

The indentation is the point: `bundle_capability` is the outer import and `dependencies` completes inside it. `bundle_capability` now **starts** first, exactly as you traced.

So the cycle is live rather than theoretical, and I confirmed it by making the substitution instead of arguing about it:

```
ImportError: cannot import name 'MaidrBundleTraceWarning' from partially
initialized module 'maidr.util.bundle_capability'
(most likely due to a circular import)
```

## What I changed as a result

The subprocess test's docstring had been rewritten (last round) to say the ordering could not be observed. That is now wrong in the other direction, so it is rewritten again, and I added the assertion the situation actually deserves — `test_the_shim_must_stay_lazy`, which parses `dependencies.py` and fails if `bundle_capability` is imported at module scope.

Its value is not the case you and I just reproduced. While the cycle is live an eager re-export is impossible to miss — nothing imports at all; the suite dies at `conftest` before a single test runs. **The case it catches is the quiet one:** if the internal call sites are ever pointed back at `dependencies`, the cycle stops being live, an eager re-export starts working again, and the trap is rearmed for whoever repoints them next. The shape is the invariant; the ImportError is only today's symptom.

Mutation-checked both ways: with the eager re-export the suite fails at collection as described, and with the shim restored the guard passes.

## The three notes

1. **`_bundle_warning_enabled` crossing a private boundary** — agreed, and your framing is the one I want carried forward: the risk is it becoming *a pattern* rather than this one instance. Noted for the freshness cut.
2. **Subprocess cost** — fair. It is one spawn (~2 s of the suite's 99 s). If the freshness split wants another, they should share one subprocess rather than each paying for an interpreter.
3. **Design-note density in docstrings** — I take the point and am not going to pretend otherwise, though I would defend `_BUNDLE_TOKEN_RE`'s specifically: it records *why the obvious looser regex is wrong*, which is the kind of thing a reader is likely to "simplify" back. That is different from history for its own sake.

`uv run pytest` — **2273 passed, 13 skipped**. `ruff check` unchanged from `main` at 6.

---

jooyoungseo:
Point 3 was worth raising — there is a real gap there, though not quite the one described. Checked rather than reasoned (`dda63e4`).

## The star-import gap is real, and the direction is the other way round

`dependencies.py` has **no `__all__` at all**, so the concern is not that the moved names need dropping from one — it is that without one, `import *` reads `__dict__` and never consults `__getattr__`:

```console
$ python -c "ns={}; exec('from maidr.util.dependencies import *', ns); ..."
  schema_trace_types:           NOT reachable via *
  bundle_trace_types:           NOT reachable via *
  warn_if_bundle_cannot_render: NOT reachable via *
  MaidrBundleTraceWarning:      NOT reachable via *
```

So `import *` really did lose four names. **Deliberately not fixing it**, on evidence rather than taste:

- Nothing in the package, tests, or docs uses `from maidr.util.dependencies import *` — grepped for every spelling.
- `maidr.util.dependencies` appears nowhere in `docs/` or `README.md`. The documented surface is `maidr.X`, and `MaidrBundleTraceWarning` is still in `maidr.__all__`.
- Adding an `__all__` to a 2,000-line module to preserve an unused, undocumented access pattern buys a list that goes stale the first time anything is added — and this module is mid-split, so things will be added.

## What I did add: `__dir__`

PEP 562 pairs `__dir__` with `__getattr__` for exactly this reason — attribute lookup and introspection are answered by different machinery, so without it `dir()` and interactive tab-completion omit names that `getattr` resolves fine.

```
moved names in dir(): True
own names still there: True
```

Its docstring records that this does *not* restore `import *`, and why — so the next reader finds the decision rather than re-deriving the gap.

**Typo** — fixed, thanks. `` ``getattr(..., default)` `` → `` ``getattr(..., default)`` ``.

**`_bundle_warning_enabled`** — third mention across three rounds, which is itself the signal. Carrying it into the freshness cut as a named task rather than a note.

`uv run pytest` — **2273 passed, 13 skipped**. `ruff check` unchanged from `main` at 6.

---

jooyoungseo:
Took the optional one (`cae795b`). It looked cosmetic and turned out not to be.

The docstring already said `warn_if_bundle_cannot_render` had moved — but it said so mid-paragraph, inside prose about the freshness-versus-capability distinction. Someone scanning the file for *what it owns* would not read that sentence as a structural statement, which is exactly the reader you described. The module now ends its docstring with a named section:

```
What this module no longer owns
-------------------------------
The capability check above lives in :mod:`maidr.util.bundle_capability`
as of #293, which is splitting this file up one concern at a time.  The
names it took are still reachable from here -- see the "Moved out of this
module" section at the foot of the file, which resolves them lazily.  A
reader scanning this file for what it owns should read that section as
part of the answer.
```

Your point about it mattering more once the freshness half lands is the reason I did it now rather than deferring: a second shim entry appended to the bottom with no signpost at the top is how a file becomes one you have to read end-to-end to understand.

On the unverifiable numbers — fair to flag, and CI is the right arbiter. For the record from this environment: `uv run pytest` → **2273 passed, 13 skipped**; `ruff check` → **6**, identical to `main`. The new run on `cae795b` will confirm independently.

`_bundle_warning_enabled` has now been raised by four consecutive reviews. That is no longer a note in a PR description — it is the first item of the freshness cut.

---

PR #496, opened 2026-08-17
refactor(util): give the shared warning policy its own module

jooyoungseo:
## Description

Prerequisite for the next cut of #293, and a fix for the one thing four consecutive reviews of #494 flagged: `bundle_capability` imported the private `_bundle_warning_enabled` from `dependencies` across a module boundary, because the warning-suppression policy belongs to neither half.

`maidr/util/warn.py` now owns it — `warn_once` and its dedup bookkeeping, `bundle_warning_enabled`, and `BUNDLE_WARNING_ENV_VAR`.

## The property that makes this work

**It is a leaf: it imports nothing from the package.** That is why both halves can import it eagerly, with none of the lazy `__getattr__` machinery `bundle_capability` needed — `warn.py` cannot participate in a cycle it has no edge into.

`test_the_warn_module_imports_nothing_from_maidr` asserts it on the parse tree, and the reason it is not a runtime check is the same one from #494: importing any submodule initialises the `maidr` package first, so at runtime everything looks imported regardless. Mutation-checked by adding one import back:

```
maidr/util/dependencies.py:63: from maidr.util.warn import (
maidr/util/warn.py:20:        from maidr.util.dependencies import maidr_js_version
E   ImportError: cannot import name 'maidr_js_version' from partially
    initialized module 'maidr.util.dependencies' (circular import)
```

As with the shim guard, a live cycle announces itself at collection; the AST assertion is for the quiet case, and for naming the invariant.

## Not pure motion — one behaviour change

`warn_once` now logs through the **`maidr.util.warn`** logger rather than `maidr.util.dependencies`. Calling that out rather than burying it:

- Filtering or configuring on the `maidr` prefix — the common case — is unaffected.
- Only a filter naming `maidr.util.dependencies` exactly would stop seeing these lines.

I judged that acceptable for an internal module whose path appears nowhere in `docs/` or `README.md`, but it is a real difference and worth a maintainer's eye rather than my say-so alone. Keeping the old name would have meant a module lying about where its logging comes from.

The two names also lose their underscores — `_warn_once` → `warn_once`, `_bundle_warning_enabled` → `bundle_warning_enabled` — since crossing a module boundary on purpose is exactly what a leading underscore says you are not doing.

### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor

## Related Issues

Refs #293, which names this directly: *"`_warn_once` and its bookkeeping are shared by both the CDN and staleness sides, so it needs a deliberate home rather than being duplicated."*

## Changes Made

- `maidr/util/warn.py` — new leaf module.
- `maidr/util/dependencies.py` — 2090 → 2013 lines; imports the four names it still uses.
- `maidr/util/bundle_capability.py` — no longer reaches into `dependencies` for a private name.
- `tests/util/test_warn_module.py` — new: the leaf invariant, dedup-per-key, and the env-var switch.
- `tests/conftest.py` and three test files — repointed at the module that now owns the state.

## Checklist

## Additional Notes

`uv run pytest` — **2329 passed, 13 skipped** (2325 + the 4 new). `ruff check` unchanged from `main` at 6 pre-existing findings.

Two things this cost that are worth knowing before the freshness cut:

1. **`_MAX_WARNED_KEY_LEN` is used by callers, not just by `warn_once`** — ten sites truncate keys before calling. It is part of the shared contract rather than an implementation detail, so it moved too and is imported back.
2. **A single line in `tests/conftest.py` patches this state**, so getting it wrong fails the entire suite at collection rather than in one file. It looked like a 100-file blast radius and was one line.

---

### DIFF
--- a/maidr/util/bundle_capability.py
+++ b/maidr/util/bundle_capability.py
@@ -19,12 +19,8 @@ import threading
 import warnings
 from collections.abc import Iterable
 
-from maidr.util.dependencies import (
-    BUNDLE_WARNING_ENV_VAR,
-    _bundle_warning_enabled,
-    bundled_js_path,
-    maidr_js_version,
-)
+from maidr.util.dependencies import bundled_js_path, maidr_js_version
+from maidr.util.warn import BUNDLE_WARNING_ENV_VAR, bundle_warning_enabled
 
 _logger = logging.getLogger(__name__)
 
@@ -159,7 +155,7 @@ def warn_if_bundle_cannot_render(
     rendered in a loop says it once and a second unsupported type is not
     swallowed by the first.
     """
-    if not _bundle_warning_enabled():
+    if not bundle_warning_enabled():
         return
 
     known = bundle_trace_types()
--- a/maidr/util/dependencies.py
+++ b/maidr/util/dependencies.py
@@ -60,6 +60,13 @@ from typing import NamedTuple
 from urllib.error import HTTPError
 from urllib.request import Request, urlopen
 
+from maidr.util.warn import (
+    BUNDLE_WARNING_ENV_VAR,
+    _MAX_WARNED_KEY_LEN,
+    bundle_warning_enabled,
+    warn_once,
+)
+
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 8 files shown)


jooyoungseo:
Took the stale reference and **one** of the two blank-line spots (`cf03d9b`) — checked both against `main` rather than accepting them, and they turned out to differ.

**Stale cross-reference** — good catch, and it was outside every hunk so nothing would have flagged it. Now `:func:`maidr.util.warn.warn_once``, fully qualified since the reader is in a different module from the target. Rewrapped the paragraph while I was there; my first edit dropped "reasoning here" and left a sentence reading *"…, the / does not depend on the GIL"*, which I caught on re-read rather than at the diff.

**Blank lines** — one real, one not:

| spot | `main` | this branch | verdict |
|---|---|---|---|
| before `_logger` | 2 blank lines | 2 blank lines | **not a regression** — matches the file already |
| after `STALE_MINOR_GAP` | 1 | 2 | **mine**, from removing the `BUNDLE_WARNING_ENV_VAR` block — fixed |

Checked with `git show origin/main:maidr/util/dependencies.py`. I did not want to "fix" spacing into something the file did not have before, since as you say nothing here lints for it — the only reference that means anything is what was there.

**On `tests/core/test_bundle_capability.py` / `test_bundle_freshness.py`** — thanks for checking those explicitly rather than assuming. Your read is right and worth stating for the record, because it is load-bearing for the next cut: `dependencies` imports `BUNDLE_WARNING_ENV_VAR` eagerly from `warn`, which makes it a real attribute of `dependencies` rather than something resolved through the `__getattr__` shim. So `dependencies.BUNDLE_WARNING_ENV_VAR` keeps working with no shim entry, and that is a property of the leaf design rather than luck.

`uv run pytest` — **2329 passed, 13 skipped**. `ruff check` unchanged from `main` at 6.

---

PR #497, opened 2026-08-18
refactor(util): move bundle staleness reporting into its own module

jooyoungseo:
## Description

`maidr/util/bundle_freshness.py` takes staleness reporting — `bundle_status`, `BundleStatus`, `warn_if_bundle_is_stale`, `warn_bundle_unreadable`, `resolver_outcome`, `STALE_MINOR_GAP`, `MaidrBundleStaleWarning`.

`dependencies.py` goes **2033 → 1781 lines**.

| #293's concern | module | done |
|---|---|---|
| Bundled asset access | `dependencies.py` | ✅ stays put, as the issue intended |
| **Bundled-copy freshness** | **`bundle_freshness.py`** | ✅ **this PR** |
| CDN version resolution | `dependencies.py` | ❌ **not cut** — see below |
| *(not in the issue)* Bundled-copy capability | `bundle_capability.py` | ✅ #494 |
| *(the issue's third note)* Shared warning policy | `warn.py` | ✅ #496 |

**This does not finish #293, and an earlier revision of this description wrongly said it did.** The issue's suggested shape names *two* new modules, `cdn.py` and `bundle_freshness.py`. Only the second exists. `dependencies.py` is still 1781 lines because it holds asset access **and** the whole CDN resolver — the endpoints, the timeout budget, the semver guards — which is the concern the `cdn.py` half was for. Splitting that out is a separate PR, and I have not written it. #293 should stay open on those grounds specifically, not on the "four modules vs. two" bookkeeping I originally offered as the reason.

Two of the issue's three "notes for whoever picks it up" are settled: `api.py` is down from 8 imports to 4, and `_warn_once` got the deliberate home it asked for (`warn.py`, #496). The third — `test_no_render_path_references_the_unresolved_constants` needing its exclusion updated — needed no change, because `MAIDR_JS_CDN_URL` / `MAIDR_CSS_CDN_URL` never moved. It becomes live again if `cdn.py` is ever cut.

## Three things this cut needed that the first one did not

**1. `_RELEASE_RE` stays behind — again.** It sits inside the moved section, but `_version_key` uses it and it is a version-parsing primitive rather than a freshness one. I recorded this on #293 after the first cut as the thing to check before assuming a section is self-contained, and it caught itself here.

**2. There is an edge back into `dependencies`.** Unlike the capability cut, which was one-way, `inline_bundle_tags` calls `warn_bundle_unreadable`. Every other apparent reference turned out to be a docstring — the only real one is that single call, so it is imported inside the function and the module-scope dependency stays one-way.

**3. Reads that tests patch go through the module object.** `_deps.maidr_js_version()` and `_deps._published_version()` rather than from-imports:

```python
# tests/core/test_bundle_freshness.py
monkeypatch.setattr(dependencies, "maidr_js_version", lambda: version)
```

A from-import binds the function object at import time, so that patch would have silently stopped reaching this module — the fixture would set a version nothing read. **29 tests failed and said so**, which is the only reason I found it rather than shipping it.

## One observable change, not covered by "no behaviour change"

Moving the functions moves their `_logger`, so the log **records they emit now carry a different logger name**:

| record | before | after |
|---|---|---|
| `use_cdn="auto"` staleness drift | `maidr.util.dependencies` | `maidr.util.bundle_freshness` |
| bundle-unreadable (since #496) | `maidr.util.dependencies` | `maidr.util.warn` |

Message text, level, and the conditions that produce them are unchanged. This only matters to a caller that filters by logger name — e.g. `logging.getLogger("maidr.util.dependencies").setLevel(...)`, which would now silence nothing. `maidr.util` still catches everything, as before.

It also broke six tests without failing any of them — the same class of trap as point 3 above, applied to logger identity rather than a function reference. `test_bundle_freshness.py` had six `caplog.at_level(logging.WARNING, logger=dependencies.__name__)` calls left naming a logger that no longer emits. They kept passing because `caplog.records` collects by handler, not by name — the `logger=` argument only raises a level, and these records were already at WARNING. Idle today, and silently useless the first time a sub-WARNING record needs asserting on.

Which of the six belongs to which module was settled by spying on `logging.Logger.handle` while they ran, not by reading what each one calls:

```
test_use_cdn_auto_render_reports_to_the_logger              -> maidr.util.bundle_freshness
test_auto_render_first_does_not_swallow_a_later_offline_…   -> maidr.util.bundle_freshness
test_offline_render_first_does_not_swallow_a_later_auto_…   -> maidr.util.bundle_freshness
test_plotly_auto_render_reports_to_the_logger               -> maidr.util.bundle_freshness
test_init_notebook_use_cdn_false_never_resolves_when_…      -> maidr.util.warn
test_auto_path_reports_drift_to_the_logger_not_as_a_warning -> maidr.util.bundle_freshness
```

Worth doing that way: a first pass sorted them by call site and put the last one in `warn`, because `warn_if_bundle_is_stale` reaches `warn.py` on its *other* branch. The suite could not tell me, for the reason above.

### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor (behaviour unchanged except the logger names above)

## Related Issues

Refs #293 — advances it, does not close it. The `cdn.py` half is untouched; see the table at the top.

## Changes Made

- `maidr/util/bundle_freshness.py` — new, 314 lines.
- `maidr/util/dependencies.py` — the `__getattr__` shim gains a second set, `__dir__` covers both, one function-local import. `BUNDLE_WARNING_ENV_VAR` is imported straight from `warn` rather than forwarded through the shim, so the public path does not depend on what `bundle_freshness` happens to import (review point).
- `maidr/__init__.py`, `api.py`, `core/maidr.py`, `plotly/plotly_maidr.py` — repointed at the canonical module, so the shim stays the external back-compat surface rather than something first-party code leans on (the #494 review's point).
- `tests/conftest.py` and three test files — repointed at the module that owns the state.
- `tests/core/test_bundle_freshness.py` — the six `caplog` sites above.
- `tests/core/test_bundle_capability.py` — the laziness guard now covers **both** halves; they share the shim, so a module-scope import of either closes the same loop.

## Checklist

## Additional Notes

`uv run pytest` — **2329 passed, 13 skipped**, identical to `main`. `ruff check` unchanged at 6 pre-existing findings.

Verified rather than assumed:

- `_RELEASE_RE` occurs 3× in `dependencies.py` and 0× in the new module.
- Old import paths resolve, with identity preserved — `maidr.bundle_status is bundle_freshness.bundle_status` → `True`, which matters because `except MaidrBundleStaleWarning` compares class objects.
- Mutation-tested the widened laziness guard both ways: an eager `from … import` breaks the cycle live (`ImportError: … partially initialized module`), and a plain `import maidr.util.bundle_freshness` — which would *not* break — is caught by the AST assertion, which is the quiet case the guard exists for.
- Mutation-tested the `BUNDLE_WARNING_ENV_VAR` decoupling: with the name removed from `bundle_freshness` entirely, `dependencies.BUNDLE_WARNING_ENV_VAR` still resolves.
- The emitting logger for each repointed `caplog` site, per the table above.

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -362,6 +362,12 @@ It is only raised as a warning when the bundle genuinely runs
 (`use_cdn=False`). Under `use_cdn="auto"` the CDN copy normally loads
 instead, so the drift is reported to the logger rather than as a warning
 that could redden a test suite over code that never executed.
+
+That record comes from `maidr.util.bundle_freshness`, and the separate
+"bundle could not be read" record from `maidr.util.warn`. Both were
+emitted from `maidr.util.dependencies` in earlier releases, so filter on
+`maidr.util` — which has always caught both, and still will if they move
+again — rather than on a leaf module name.
 :::
 
 ::: {.callout-important}
--- a/maidr/__init__.py
+++ b/maidr/__init__.py
@@ -211,18 +211,20 @@ from .patch import (  # noqa: E402, F401
     violinplot,
 )
 from .util.bundle_capability import MaidrBundleTraceWarning  # noqa: E402
+from .util.bundle_freshness import (  # noqa: E402
+    STALE_MINOR_GAP,
+    BundleStatus,
+    MaidrBundleStaleWarning,
+    bundle_status,
+    resolver_outcome,
+)
+from .util.warn import BUNDLE_WARNING_ENV_VAR  # noqa: E402
 from .util.dependencies import (  # noqa: E402
-    BUNDLE_WARNING_ENV_VAR,
     BUNDLED_TAG,
     CDN_TIMEOUT_ENV_VAR,
     CDN_VERSION_ENV_VAR,
     LATEST_TAG,
-    STALE_MINOR_GAP,
-    BundleStatus,
     ResolverOutcome,
-    MaidrBundleStaleWarning,
-    bundle_status,
-    resolver_outcome,
     bundled_css_path,
     bundled_js_path,
     bundled_math_css_path,
--- a/maidr/__init__.py
+++ b/maidr/__init__.py
@@ -218,7 +218,6 @@ from .util.bundle_freshness import (  # noqa: E402
     bundle_status,
     resolver_outcome,
 )
-from .util.warn import BUNDLE_WARNING_ENV_VAR  # noqa: E402
 from .util.dependencies import (  # noqa: E402
     BUNDLED_TAG,
     CDN_TIMEOUT_ENV_VAR,
@@ -235,6 +234,7 @@ from .util.dependencies import (  # noqa: E402
     reset_cdn_version_cache,
     set_cdn_version,
 )
+from .util.warn import BUNDLE_WARNING_ENV_VAR  # noqa: E402
 
 # Second call: reclaim the backend after maidr's own imports.
 # NOTE: See the "Matplotlib backend activation" block at the top of this
--- a/maidr/api.py
+++ b/maidr/api.py
@@ -240,13 +240,15 @@ def init_notebook(
         # IPython not importable — nothing to display into.
         return
 
+    from maidr.util.bundle_freshness import (
+        warn_bundle_unreadable,
+        warn_if_bundle_is_stale,
+    )
     from maidr.util.dependencies import (
         MAIDR_JS_FILENAME,
         read_bundled_js,
         read_bundled_math_css,
         bundled_cdn_url,
-        warn_bundle_unreadable,
-        warn_if_bundle_is_stale,
     )
 
     mode = _resolve_use_cdn(use_cdn)
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -24,11 +24,11 @@ from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.plot import MaidrPlot
 from maidr.core.plot.barplot import BarPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
-from maidr.util.bundle_freshness import warn_if_bundle_is_stale
 from maidr.util.bundle_capability import (
     schema_trace_types,
     warn_if_bundle_cannot_render,
 )
+from maidr.util.bundle_freshness import warn_if_bundle_is_stale
 from maidr.util.dependencies import (
     MAIDR_JS_FILENAME,
     OFFLINE_FALLBACK_REPORT,
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -24,6 +24,7 @@ from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.plot import MaidrPlot
 from maidr.core.plot.barplot import BarPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
+from maidr.util.bundle_freshness import warn_if_bundle_is_stale
 from maidr.util.bundle_capability import (
     schema_trace_types,
     warn_if_bundle_cannot_render,
@@ -37,7 +38,6 @@ from maidr.util.dependencies import (
     maidr_bundled_relative_dir,
     maidr_html_dependency,
     maidr_js_cdn_url,
-    warn_if_bundle_is_stale,
 )
 from maidr.util.environment import Environment
 from maidr.util.iframe_utils import chart_title_of, wrap_in_iframe_matplotlib
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -35,11 +35,11 @@ from maidr.plotly.step_shape import (
     is_step_trace,
     renders_through_webgl,
 )
-from maidr.util.bundle_freshness import warn_if_bundle_is_stale
 from maidr.util.bundle_capability import (
     schema_trace_types,
     warn_if_bundle_cannot_render,
 )
+from maidr.util.bundle_freshness import warn_if_bundle_is_stale
 from maidr.util.dependencies import (
     MAIDR_JS_FILENAME,
     OFFLINE_FALLBACK_REPORT,
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -35,6 +35,7 @@ from maidr.plotly.step_shape import (
     is_step_trace,
     renders_through_webgl,
 )
+from maidr.util.bundle_freshness import warn_if_bundle_is_stale
 from maidr.util.bundle_capability import (
     schema_trace_types,
     warn_if_bundle_cannot_render,
@@ -48,7 +49,6 @@ from maidr.util.dependencies import (
     maidr_bundled_relative_dir,
     maidr_html_dependency,
     maidr_js_cdn_url,
-    warn_if_bundle_is_stale,
 )
 from maidr.util.environment import Environment
 from maidr.util.iframe_utils import chart_title_of, wrap_in_iframe_plotly
--- a/maidr/util/bundle_freshness.py
+++ b/maidr/util/bundle_freshness.py
@@ -1,0 +1,314 @@
+"""How far the bundled ``maidr.js`` has drifted from what is published.
+
+Split out of :mod:`maidr.util.dependencies` (#293), which owned three
+separable concerns.  This is the one that answers "how old is the copy in
+this wheel" -- the coarse signal.  Whether that copy can actually draw a
+given chart is a different question, answered by
+:mod:`maidr.util.bundle_capability`, and the two are deliberately kept
+apart: "you are drifting" and "this chart will not draw" are different
+claims, and the second is the one worth acting on.
+
+Its tests live in ``tests/core/test_bundle_freshness.py``, which split
+along this line before the module did.
+"""
+
+from __future__ import annotations
+
+import logging
+import threading
+import warnings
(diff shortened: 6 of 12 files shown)


jooyoungseo:
Both points addressed in cec7758, and the first one was right in a way I'd have missed.

**The `caplog` logger names.** Confirmed, and worse than "worth swapping": all six were already broken and the suite could not say so. `caplog.records` collects by handler rather than by logger name, so `logger=` only raises a level — and these records are at WARNING, which the root logger passes anyway. Six assertions naming a logger that emits nothing, all green.

You were also right that only the `init_notebook` site belongs to `warn`, and I nearly got that wrong. My first pass sorted the six by what each test calls and sent *two* to `warn_module`, on the reasoning that `warn_if_bundle_is_stale` reaches `warn.py`. So I spied on `logging.Logger.handle` while the six ran instead:

```
test_use_cdn_auto_render_reports_to_the_logger              -> maidr.util.bundle_freshness
test_auto_render_first_does_not_swallow_a_later_offline_…   -> maidr.util.bundle_freshness
test_offline_render_first_does_not_swallow_a_later_auto_…   -> maidr.util.bundle_freshness
test_plotly_auto_render_reports_to_the_logger               -> maidr.util.bundle_freshness
test_init_notebook_use_cdn_false_never_resolves_when_…      -> maidr.util.warn
test_auto_path_reports_drift_to_the_logger_not_as_a_warning -> maidr.util.bundle_freshness
```

`test_auto_path_reports_drift_…` was the one I'd misfiled — `warn_if_bundle_is_stale(bundle_is_primary=False)` logs from `bundle_freshness`; it's the *other* branch that reaches `warn.py`. Tests pass either way, so nothing but the probe would have caught it. Five now name `freshness`, the `init_notebook` one names `warn`.

Taking the aside in scope rather than deferring it: it's one line in the same file, in the same class of bug, and leaving a knowingly-wrong logger name behind to file an issue about costs more than fixing it.

**Import ordering.** Fixed — `bundle_capability` before `bundle_freshness` in `core/maidr.py` and `plotly/plotly_maidr.py`, `util.warn` after `util.dependencies` in `__init__.py`. No load-graph significance either way (`warn` is a leaf, and `bundle_capability` imports both regardless), so this is purely the alphabetical convention the rest of those blocks follow. Verified by import after the move rather than assumed.

**The "no behaviour change" claim.** Withdrawn as stated and replaced with the specific exception, now a table in the PR body: the `use_cdn="auto"` drift record moves to `maidr.util.bundle_freshness`, and the bundle-unreadable record moved to `maidr.util.warn` back in #496. Message, level, and trigger conditions unchanged; this only reaches someone filtering on `logging.getLogger("maidr.util.dependencies")`, which would now silence nothing. `maidr.util` still catches both.

Local: **2329 passed, 13 skipped**; `ruff check` at the same 6 pre-existing findings as `main`.

---

jooyoungseo:
Both observations taken in 05d0c13 — the first structurally rather than as a comment, because the fix turned out to be smaller than the caveat.

**1. `BUNDLE_WARNING_ENV_VAR`.** You're right that the coupling is incidental, and rather than document it I removed it: the name is now imported straight from `maidr.util.warn` at module scope in `dependencies.py` and dropped from `_MOVED_TO_BUNDLE_FRESHNESS`. That works precisely because of the property you identified — `warn` is a leaf, so it raises none of the cycle questions the lazy shim exists for, and `dependencies` already imported `warn_once` and `_MAX_WARNED_KEY_LEN` from it exactly this way. The shim now forwards only names whose owner genuinely can't be imported eagerly.

Confirmed by making your hypothetical cleanup: `bundle_freshness` mutated so it no longer carries the name at all —

```
bundle_freshness has the name : False
dependencies still resolves it: MAIDR_BUNDLE_STALE_WARNING
```

Under the previous arrangement that is the `AttributeError` you predicted.

One deliberate asymmetry: `bundle_warning_enabled` gets no forwarding. It was `_bundle_warning_enabled` before the split (checked at `2e43684~1:maidr/util/dependencies.py:1622`), so no caller could have been depending on the old path — the only name in that module that was public before #496 is the env var.

**2. `ast.unparse` called twice.** Hoisted, though into the comprehension rather than a bare local: the module-scope imports are now collected once with their unparsed text, and the offender filter reads the strings. Slightly clearer about what the two steps are, and it drops the double call.

Local after both: **2329 passed, 13 skipped**; `ruff check` at the same 6 pre-existing findings as `main`.

Also worth flagging in the other direction — your note that the six `caplog` assertions were "previously passing for the wrong reason" is the part of this I'd have shipped wrong without the review. My first attempt at repointing them sorted the six by what each test *calls* and put `test_auto_path_reports_drift_to_the_logger_not_as_a_warning` in `warn`, on the reasoning that `warn_if_bundle_is_stale` reaches `warn.py`. It does, on its other branch. Nothing in the suite could tell me, for exactly the reason you gave, so I settled it by spying on `logging.Logger.handle` while the six ran.

---

jooyoungseo:
Coverage gap taken in f1c6c86. You're right that the previous commit's verification was throwaway — I proved the fix by hand and left nothing behind to hold it, which is the same "verified once, unprotected thereafter" shape the PR is otherwise about.

`test_the_env_var_resolves_to_one_object_from_every_path` asserts your two lines plus the part that makes them durable:

```python
assert dependencies.BUNDLE_WARNING_ENV_VAR is warn.BUNDLE_WARNING_ENV_VAR
assert maidr.BUNDLE_WARNING_ENV_VAR is warn.BUNDLE_WARNING_ENV_VAR

assert "BUNDLE_WARNING_ENV_VAR" not in dependencies._MOVED_TO_BUNDLE_FRESHNESS
assert "BUNDLE_WARNING_ENV_VAR" not in dependencies._MOVED_TO_BUNDLE_CAPABILITY
assert bundle_freshness.BUNDLE_WARNING_ENV_VAR is warn.BUNDLE_WARNING_ENV_VAR
```

The identity pair alone would still pass if someone reinstated the shim path, because the shim also returns the right object *while `bundle_freshness` happens to hold the name* — which is precisely the coincidence that made this fragile. The set assertions are what fail on the reintroduction rather than on its eventual consequence. Kept as identity rather than equality for the reason your snippet implies: both pass on a plain string, and `is` is what says these are one constant rather than three that agree.

Falsified rather than assumed to work — direct import removed, name put back in `_MOVED_TO_BUNDLE_FRESHNESS`:

```
FAILED tests/core/test_bundle_capability.py::test_the_env_var_resolves_to_one_object_from_every_path
1 failed
```

Only that test fails, which is what makes it worth having.

Local: **2330 passed, 13 skipped** (+1); `ruff check` at the same 6 pre-existing findings as `main`.

Two notes on the rest of your review, since both touch things I'd stated:

- You read `dependencies.py` as now being "asset access + CDN version resolution". That's right, and it's also why this PR **does not close #293** — the issue's suggested shape asks for a `cdn.py` too, and that half is untouched. I'd claimed completion in an earlier revision of the description and have since corrected it; the table at the top now marks CDN resolution as not cut.
- Confirming your `_MOVED_TO_BUNDLE_FRESHNESS` check independently: it matched `bundle_freshness.py` when you looked, and this commit removes `BUNDLE_WARNING_ENV_VAR` from it — that entry is now deliberately *not* in the set, with the assertion above pinning that.

---

jooyoungseo:
Took the logger-name discoverability point in 718025c, but not as a `CHANGELOG.md` edit — semantic-release owns that file and the project's git-workflow rules say not to hand-edit it. Two things I checked before deciding where it should go instead:

**`refactor` is already in the changelog allowlist**, so this lands in the release notes on its own:

```toml
exclude_commit_patterns = [
  "^(?!(feat|fix|docs|perf|refactor|test|build|ci|style)(\(.*\))?!?: ).*"
]
```

But only the *subject line* renders there, and the subject is about moving a module, not about logger names. So the release-notes path alone would not tell anyone what you're worried about.

**The docs were the thing actually saying something incomplete.** `docs/index.qmd`'s staleness callout already said the `use_cdn="auto"` drift "is reported to the logger" — without naming which logger, so a reader wanting to filter it had to read the source anyway. That was true before this PR and would have stayed true after. It now names both records and, more usefully, points at the filter that survives this kind of move:

> That record comes from `maidr.util.bundle_freshness`, and the separate "bundle could not be read" record from `maidr.util.warn`. Both were emitted from `maidr.util.dependencies` in earlier releases, so filter on `maidr.util` — which has always caught both, and still will if they move again — rather than on a leaf module name.

That reaches the reader on an upgrade *and* the one arriving fresh six months later, which a changelog line only does for the first.

Two things I deliberately did **not** write:

- **A version number.** My first draft said "moved in 1.22.0". That's a guess and probably wrong — `minor_tags = ["feat"]`, `patch_tags = ["fix", "perf"]`, and this branch is `refactor` + `test` + `docs`, so it bumps nothing on its own and the release these arrive in depends on what else lands first.
- **The names read off the source.** Confirmed by watching `logging.Logger.handle` while each record was actually emitted:
  ```
  maidr.util.warn  | maidr: use_cdn=False was requested but the b…
  ```
  and `maidr.util.bundle_freshness` for the drift record, from the probe over `test_auto_path_reports_drift_to_the_logger_not_as_a_warning`. Getting a logger name wrong *in the documentation telling people how to filter it* would be a worse version of the bug this PR already contained once.

Local: **2330 passed, 13 skipped**; `ruff check` unchanged at the 6 pre-existing findings.

---

Issue #498, opened 2026-08-18
MaidrPlot holds live matplotlib artists, so no lifetime fix in FigureManager can work


jooyoungseo:
Split out of #456, where I said this "deserves its own issue rather than being folded in". #456 is about `FigureManager.figs` never dropping entries. This is about why the obvious fixes for that cannot work from inside `FigureManager` at all.

> **Both open questions below are now measured — see [this comment](https://github.com/xability/py-maidr/issues/498#issuecomment-5322145248).** One answer reverses what the "what a real fix has to do" section originally guessed; that section has been rewritten accordingly.

## The defect

`MaidrPlot` keeps two strong references into the live matplotlib object graph, and both reach the `Figure`:

```python
# maidr/core/plot/maidr_plot.py:45
self.ax = ax
self._elements = []          # filled with the artists a layer highlights
```

Measured on a three-bar chart after one render:

```
MaidrPlot.ax is an Axes        : Axes
plot.ax.figure is the dict key : True
len(plot._elements)            : 3
element type                   : Rectangle
element.axes is that same Axes : True
```

So `figs[fig] → Maidr._plots[i] → MaidrPlot.ax → Axes.figure → fig` closes a cycle through the dictionary's own key, and `_elements` closes a second one independently. Weakening either alone frees nothing.

## Why this blocks #456's options

- **Option 1 (`WeakKeyDictionary` + `weakref.ref` on `Maidr._fig`)** — I built it and reported the result on #456: `figs` stayed at 25 entries, nothing collected. A `WeakKeyDictionary` whose values reach their own keys collects nothing, and does so *silently* — the diff reads like a fix.
- **Any `id()`-keyed map with a `weakref.finalize` hook** has the same problem for the same reason: the registry's value keeps the figure alive, so the finaliser never runs.
- **Option 3 (drop on `close_event`)** is wrong independently, per #452 — a figure built lazily and cached (`@reactive.calc`, any memoised helper) is closed by the render that opened it but is still the live chart, and dropping its record makes every later render fall back to a static image.

That leaves #456's **option 2 (bounded LRU)** as the only one implementable without touching this, which is why #456 needs a decision rather than a patch. Worth being explicit that option 2 carries the same hazard as option 3 in miniature: evicting an entry for a figure the app still re-renders strips its data. A generous bound makes that rare, not impossible.

## What it costs

25 renders, each building its own figure and calling `plt.close()`, versus the identical workload with the `figs` entry also dropped:

[...]

Issue #456, opened 2026-08-17
FigureManager.figs retains every figure for the life of the process


jooyoungseo:
`FigureManager.figs` is a plain dict keyed by `Figure` (`maidr/core/figure_manager.py:41`). Nothing removes an entry except an explicit `maidr.close()` / `FigureManager.destroy()`, so in any long-running process that renders repeatedly, every figure ever plotted stays reachable — along with the `Maidr` object holding its extracted plot data.

`maidr/backend.py:91` already names this ("Without this, `FigureManager.figs` grows unboundedly…") and works around it for the `plt.show()` path, but the general case remains.

Reproduction — 25 renders, each building its own figure, with no explicit close:

```
open figs = 25   FigureManager.figs = 25
```

plus matplotlib's `More than 20 figures have been opened` warning.

A Shiny app is the sharpest case, since it renders once per reactive flush. #452 makes `@render_maidr` call `plt.close()` on figures the render opened, which frees the canvas and stops matplotlib's warning — but it deliberately does **not** drop the `FigureManager` entry, because "this figure was not open before the render" is also true of a figure built lazily on the first flush and cached (`@reactive.calc`, any memoised helper), and dropping the record there strips the chart of its data so every later render falls back to a static image.

**Why the obvious fix does not work.** Making `figs` a `weakref.WeakKeyDictionary` looks right but is not: `Maidr.__init__` stores `self._fig = fig` (`maidr/core/maidr.py:101`), so the dictionary's value holds a strong reference to its own key. The entry keeps itself alive and nothing is ever collected.

**Options**

1. Have `Maidr` hold a `weakref.ref` to its figure and expose `fig` through it, then make `figs` a `WeakKeyDictionary`. Correct, and the widest blast radius — `Maidr.fig` is public.
2. Bound the map (LRU by insertion, evicting the least recently rendered).
3. Register a `Figure` finaliser at creation time that removes the entry.

Option 1 is the real fix; it needs its own change and its own tests.

Follow-up from #452.

jooyoungseo:
I tried option 1 and it does not work. Reporting rather than shipping it, because the change is invisible — the leak is unchanged and the diff looks like a fix.

## What I did

Exactly what the issue prescribes: `Maidr` holds `weakref.ref(fig)` and exposes `_fig` through it, and `FigureManager.figs` becomes a `WeakKeyDictionary`.

```
open figs          = 0
FigureManager.figs = 25
```

Unchanged. Nothing was collected.

## Why

[...]

PR #500, opened 2026-08-18
fix(patch): forget a cleared axes' layers instead of appending beside them

### LINKED ISSUE #499: ax.clear() leaves a phantom layer, so a redrawn chart is read as two charts
`Figure.clear` is patched to drop the figure's registered layers (`maidr/patch/clear.py`). `Axes.clear` / `Axes.cla` is not. Re-plotting into a cleared axes therefore *appends* a layer instead of replacing one, and the reader is offered a layer describing bars that are no longer drawn.

## Measured

Same sequence — plot, render, clear, re-plot — through each of the three clear paths:

```
ax.clear()               layers=2  y per layer=[[1.0, 2.0, 3.0], [9.0, 9.0, 9.0]]
plt.cla()                layers=2  y per layer=[[1.0, 2.0, 3.0], [9.0, 9.0, 9.0]]
fig.clear() [patched]    layers=1  y per layer=[[9.0, 9.0, 9.0]]
```

The first layer is the stale one. Its artists are gone from the axes:

```
layer 0: 3 elements, still on axes: [False, False, False]
layer 1: 3 elements, still on axes: [True, True, True]
```

And it accumulates — nothing ever drops it:

```
layers after 5 ax.clear() cycles: 6
```

## Why this matters

The emitted schema carries both layers, so the frontend offers both. A reader switching layers gets one that announces `1, 2, 3` for a chart displaying `9, 9, 9`, and whose highlight resolves to nothing, because the artists it names are no longer drawn and so never reach `HighlightContextManager`.

That is the failure mode this project treats as the worst kind: not a crash, not a missing chart, but **a complete, confident, wrong answer**, delivered only in the non-visual modality. A sighted reviewer looking at the figure sees one correct chart.

`ax.clear()` is ordinary matplotlib — it is the normal way to redraw into a reused axes, which is exactly what an animation loop or a long-lived app does. `fig.clear()` being the only patched path means the two spellings of the same intent behave differently, and the one that behaves correctly is the less common one for a single-axes figure.

## Reproduce

```python
import matplotlib.pyplot as plt, maidr

fig, ax = plt.subplots()
ax.bar(["a", "b", "c"], [1, 2, 3])
m = FigureManager.get_maidr(fig)
m.render(use_cdn=True)

ax.clear()
ax.bar(["a", "b", "c"], [9, 9, 9])

print(len(m._flatten_maidr()["subplots"][0][0]["layers"]))   # 2
```

## The fix is not quite the `Figure.clear` one

`Maidr.clear()` is `self._plots = []` (`maidr/core/maidr.py:276`), which is right for `Figure.clear` because every layer on the figure is going. For `Axes.clear` it is too broad: on a multi-axes figure, clearing one axes must drop only the layers whose `ax` is that axes and leave the others registered.

So this wants a narrower operation — drop the plots belonging to one `Axes` — rather than reusing `clear()`. Worth checking against `maidr/patch/clear.py`'s existing `KeyError` guard, which handles a figure maidr never registered.

Two cases to cover in tests beyond the single-axes one:

- A `subplots(1, 2)` figure where one axes is cleared and the other keeps its layer.
- The `plt.cla()` spelling, which goes through the same `Axes` method and is measured above as equally affected.

## Found via

Investigating #498 — I was checking what happens to retained artists when an axes is cleared while its figure survives, which is a case the weakref proposal there has to handle. The retention question turned out to be the smaller half of what that sequence does.

## Related

- #498 — artist retention; this is the same sequence seen from the schema side
- #354 — the earlier case of a layer's element list going wrong and the outline landing on the wrong mark

jooyoungseo:
## Description

`Figure.clear` was patched to drop a figure's registered layers. `Axes.clear` was not. Re-plotting into a cleared axes therefore **appended** a layer rather than replacing one, and the reader was offered a layer describing artists that are no longer drawn.

Same sequence — plot, render, clear, re-plot — through each spelling:

```
ax.clear()               layers=2  y per layer=[[1.0, 2.0, 3.0], [9.0, 9.0, 9.0]]
plt.cla()                layers=2  y per layer=[[1.0, 2.0, 3.0], [9.0, 9.0, 9.0]]
fig.clear() [patched]    layers=1  y per layer=[[9.0, 9.0, 9.0]]
```

The stale layer's artists are gone from the axes:

```
layer 0: 3 elements, still on axes: [False, False, False]
layer 1: 3 elements, still on axes: [True, True, True]
```

and nothing ever drops it — **five clear cycles left six layers**.

### Why this is the bad kind of bug

Not a crash and not a missing chart: a **complete, confident, wrong answer delivered only in the non-visual modality**. The phantom layer announces `1, 2, 3` for a chart displaying `9, 9, 9`, and its highlight resolves to nothing because artists that are no longer drawn never reach `HighlightContextManager`. A sighted reviewer looking at the figure sees one correct chart.

`ax.clear()` is ordinary matplotlib — the normal way to redraw into a reused axes, which is what an animation loop or a long-lived app does. So the two spellings of one intent behaved differently, and the correct one was the less common for a single-axes figure.

## A second defect in the same area

`Maidr.clear()` emptied `_plots` and left `selector_ids` behind:

```
before clear:      plots=1 selector_ids=1
after fig.clear(): plots=0 selector_ids=1
after re-plot:     plots=1 selector_ids=2   <- not aligned
```

The two are paired by index in both directions, and `_drop_superseded_layers` already documents exactly what that costs:

> dropping a layer without dropping its id hands every surviving layer its neighbour's, and the highlight lands on the wrong mark with nothing raised.

Measured consequence: the re-plotted layer's artists were tagged with `selector_ids[0]` — the id minted for the discarded layer — while the id minted for the new layer was never used. Nothing raised.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues

Closes #499. Found while investigating #498 (artist retention), checking what happens to retained artists when an axes is cleared while its figure survives.

## Changes Made

- `maidr/core/maidr.py` — new `clear_axes(ax)`, and `clear()` now drops `selector_ids` too.
  - `clear_axes` rather than reusing `clear()`: on a figure with several axes, clearing one must leave the others registered. Keyed by `_layer_axes_key`, so a `twinx` twin — a second axes at the same `(row, col)` — does not take the axes it was twinned from with it, which is the trap that helper already records for the supersede rules.
- `maidr/patch/clear.py` — patches **both** `Axes.clear` and `Axes.cla`. They delegate to each other depending on `Axes._subclass_uses_cla`, so patching one leaves the other reachable on a subclass that inverts the delegation. That can fire the hook twice for a single call, which is why `clear_axes` is idempotent. Runs after `wrapped`, matching the existing `Figure.clear` patch.
- `tests/patch/test_clear.py` — new; the clear patch had **no** coverage before this.

## Checklist

## Additional Notes

`uv run pytest` — **2338 passed, 13 skipped** (+8). `ruff check` unchanged at the 6 pre-existing findings; the new files are clean.

Every test was falsified against the code it pins, rather than trusted because it passed:

| mutation | tests that fail |
|---|---|
| `Axes.clear` / `cla` not patched | **5** |
| `clear_axes` drops every axes instead of one | **3** |
| `clear()` leaves `selector_ids` behind | **1** |

The last one is worth flagging, because **it initially passed under mutation** — a test that would have shipped asserting nothing. Going through `fig.clear()` now reaches `clear_axes` first (`Figure.clear` clears its axes, which is patched), and that drops both lists in step, so the figure route cannot fail even with the line removed. The test now calls the public `Maidr.clear()` directly, which is what that line actually guards. I only caught it because the mutation run reported `8 passed` where it should have reported a failure, and I checked that the mutation had really applied rather than assuming the test was fine.

Behaviour deliberately **not** changed: a figure whose layers are all cleared keeps its `FigureManager.figs` entry, exactly as the `Figure.clear` path already did. Dropping the entry is #456, and it has an open hazard (#452's cached-figure case) that this change should not quietly decide.

---

### DIFF
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -299,9 +299,12 @@ class Maidr:
         same ``(row, col)``, and clearing one of a twinned pair must not
         take the other with it.
 
-        Idempotent: clearing an axes that holds no layers is a no-op, which
-        matters because ``Axes.clear`` and ``Axes.cla`` delegate to each
-        other and both are patched.
+        Idempotent: clearing an axes that holds no layers is a no-op. That
+        is not only about being called twice by a caller -- it is what makes
+        patching both ``Axes.clear`` and ``Axes.cla`` safe. The two delegate
+        to each other, and the inner call resolves to the *patched* method,
+        so a single ``ax.cla()`` fires the hook twice by construction. The
+        second firing has to find nothing left to do.
 
         Parameters
         ----------
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -274,7 +274,47 @@ class Maidr:
         return html.show(_renderer)
 
     def clear(self):
+        """Forget every layer on this figure.
+
+        ``selector_ids`` goes with them. It used to be left behind, and the
+        two lists are paired by index -- see ``_drop_superseded_layers`` for
+        the invariant -- so a figure cleared and then re-plotted tagged its
+        new layer's artists with the id minted for a layer that no longer
+        existed, and the new layer's own id was never used. Nothing raised
+        (#499).
+        """
         self._plots = []
+        self.selector_ids = []
+
+    def clear_axes(self, ax: Axes) -> None:
+        """Forget the layers drawn on one axes, keeping the rest.
+
+        ``clear()`` is right for ``Figure.clear``, where every layer is
+        going. It is too broad for ``Axes.clear``: on a figure with more
+        than one axes, clearing one must leave the others registered, or
+        redrawing a single panel silently unregisters its neighbours.
+
+        Keyed by ``_layer_axes_key`` rather than by grid position, for the
+        reason recorded there -- ``ax.twinx()`` puts a second axes at the
+        same ``(row, col)``, and clearing one of a twinned pair must not
+        take the other with it.
+
+        Idempotent: clearing an axes that holds no layers is a no-op, which
+        matters because ``Axes.clear`` and ``Axes.cla`` delegate to each
+        other and both are patched.
+
+        Parameters
+        ----------
+        ax : Axes
+            The axes whose layers should be forgotten.
+        """
+        kept = [
+            (plot, selector_id)
+            for plot, selector_id in zip(self._plots, self.selector_ids)
+            if self._layer_axes_key(plot) is not ax
+        ]
+        self._plots = [plot for plot, _ in kept]
+        self.selector_ids = [selector_id for _, selector_id in kept]
 
     def destroy(self) -> None:
         del self._plots
--- a/maidr/patch/clear.py
+++ b/maidr/patch/clear.py
@@ -6,6 +6,7 @@ from matplotlib.axes import Axes
 from matplotlib.figure import Figure
 
 from maidr.core.figure_manager import FigureManager
+from maidr.patch.lineplot import forget_axes_state
 
 
 @wrapt.patch_function_wrapper(Figure, "clear")
@@ -35,11 +36,25 @@ def _clear_axes(wrapped, instance, args, kwargs) -> None:
     Narrower than ``Figure.clear``'s ``maidr.clear()`` on purpose: on a
     figure with several axes, clearing one must leave the others registered.
 
+    Dropping the layers is not enough on its own. ``lineplot`` keeps its
+    own "already registered" latch on the axes, which matplotlib does not
+    reset because it does not own it -- so clearing the layers while
+    leaving the latch set makes the next ``ax.plot()`` register *nothing*,
+    and the chart goes from mis-described to undescribed. See
+    ``forget_axes_state``.
+
     Runs after ``wrapped``, matching the ``Figure.clear`` patch above -- the
     layers are dropped once matplotlib has actually removed the artists, not
     before.
     """
     wrapped(*args, **kwargs)
+
+    # Before the registration lookup, not after: this state lives on the
+    # axes and outlives any maidr entry. A figure closed with `maidr.close()`
+    # and then cleared would otherwise keep the latch set, and nothing drawn
+    # on it afterwards would register.
+    forget_axes_state(instance)
+
     figure = instance.get_figure()
     if figure is None:
         return
--- a/maidr/patch/clear.py
+++ b/maidr/patch/clear.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import wrapt
 
+from matplotlib.axes import Axes
 from matplotlib.figure import Figure
 
 from maidr.core.figure_manager import FigureManager
@@ -15,3 +16,44 @@ def clear(wrapped, instance, args, kwargs) -> None:
     except KeyError:
         return
     maidr.clear()
+
+
+def _clear_axes(wrapped, instance, args, kwargs) -> None:
+    """Drop the layers drawn on an axes when matplotlib clears it.
+
+    Only ``Figure.clear`` was patched, so re-plotting into a cleared axes
+    *appended* a layer rather than replacing one, and the reader was offered
+    a layer describing artists no longer drawn -- announced with confident
+    values, and with a highlight resolving to nothing because those artists
+    never reach ``HighlightContextManager``. It accumulated: five clear
+    cycles left six layers (#499).
+
+    ``ax.clear()`` is the ordinary way to redraw into a reused axes, so the
+    two spellings of the same intent behaved differently and the correct one
+    was the less common.
+
+    Narrower than ``Figure.clear``'s ``maidr.clear()`` on purpose: on a
+    figure with several axes, clearing one must leave the others registered.
+
+    Runs after ``wrapped``, matching the ``Figure.clear`` patch above -- the
+    layers are dropped once matplotlib has actually removed the artists, not
+    before.
+    """
+    wrapped(*args, **kwargs)
+    figure = instance.get_figure()
+    if figure is None:
+        return
+    try:
+        maidr = FigureManager.get_maidr(figure)
+    except KeyError:
+        return
+    maidr.clear_axes(instance)
+
+
+# Both spellings, because they delegate to each other depending on
+# ``Axes._subclass_uses_cla``: ``clear()`` may call ``cla()`` or vice versa.
+# Patching only one leaves the other reachable on a subclass that inverts
+# the delegation. When both fire for a single call the second finds nothing
+# left to drop, which is why ``clear_axes`` is idempotent.
+wrapt.wrap_function_wrapper(Axes, "clear", _clear_axes)
+wrapt.wrap_function_wrapper(Axes, "cla", _clear_axes)
--- a/maidr/patch/lineplot.py
+++ b/maidr/patch/lineplot.py
@@ -15,6 +15,39 @@ from maidr.util.step_utils import is_step_layer
 #: its own calls drew; see `line()` for what sweeping the axes collected.
 DRAWN_SERIES = "_maidr_line_series"
 
+#: The "this axes already has a layer" latch. Kept beside ``DRAWN_SERIES``
+#: because the two are reset together and nothing else may set either.
+PLOT_CREATED = "_maidr_plot_created"
+
+
+def forget_axes_state(ax: Axes) -> None:
+    """Drop the per-axes registration state this module stashes.
+
+    ``line()`` records two things directly on the ``Axes`` object: the
+    accumulated series, and a latch saying a layer already exists. Both are
+    plain instance attributes, so ``ax.clear()`` leaves them untouched --
+    matplotlib resets the state *it* owns, and knows nothing about these.
+
+    That makes them the one thing a clear has to be told about. Without
+    this, clearing an axes and drawing on it again finds the latch still
+    set, registers no layer at all, and leaves the chart **completely
+    undescribed** -- worse than the stale layer that clearing exists to
+    remove. It also leaves the series list holding ``Line2D`` objects that
+    were detached by the clear.
+
+    ``ax.containers``, which ``barplot`` gates on, needs none of this: it
+    belongs to matplotlib, which empties it on clear. This module is the
+    only one that keeps its own state on an axes.
+
+    Parameters
+    ----------
+    ax : Axes
+        The axes being cleared.
+    """
+    for attribute in (DRAWN_SERIES, PLOT_CREATED):
+        if hasattr(ax, attribute):
+            delattr(ax, attribute)
+
 
 def line(wrapped, instance, args, kwargs) -> Axes | list[Line2D]:
     """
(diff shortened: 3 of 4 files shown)


jooyoungseo:
This was a genuine catch and the PR was wrong to ship without it. Fixed in 8ee232a.

## The line/step regression — confirmed exactly as described

Reproduced before touching anything:

```
after first plot : layers=1   _maidr_plot_created set: True
after ax.clear() : layers=0   _maidr_plot_created STILL set: True
after re-plot    : layers=0
  *** chart is now COMPLETELY undescribed ***
  _flatten_maidr: {'id': '…', 'subplots': [[{}]]}
```

So the PR as written **turned one failure into a worse one** for this plot family: before it, a cleared-and-redrawn line chart was mis-described (stale layer beside the live one); after it, undescribed. Your framing of that is right, and it is the more dangerous direction — a phantom layer at least announces something a tester might notice, while an empty `subplots` announces nothing at all.

The fix is `forget_axes_state(ax)`, placed in `lineplot.py` beside the attributes it clears rather than in `clear.py`, so what this module stashes stays described in one place. It runs **before** the registration lookup, not after — the state is on the axes and outlives any maidr entry, so a figure closed with `maidr.close()` and then cleared would otherwise keep the latch set forever. After:

```
after ax.clear() : _maidr_plot_created STILL set: False   series len: 0
after re-plot    : layers=1
```

I also audited your "only patch module that stashes state on the Axes" claim independently rather than taking it, since the fix's correctness rests on it:

```
maidr/patch/lineplot.py:158   setattr(ax, DRAWN_SERIES, series)
maidr/patch/lineplot.py:181   setattr(ax, PLOT_CREATED, True)
maidr/patch/mplfinance.py:179 setattr(line, "_maidr_ma_period", …)
maidr/patch/mplfinance.py:219 setattr(line, "_maidr_date_nums", …)
maidr/patch/mplfinance.py:223 setattr(line, "_maidr_datetime_converter", …)
```

`mplfinance`'s three are on `Line2D` artists, which a clear discards, so they die with the artists. Direct `ax._maidr_*` assignment (as opposed to `setattr`) turns up nothing. Confirmed, and `_maidr_plot_created` is now a named constant beside `DRAWN_SERIES` so the pair is visibly one thing.

## The two test points

**`test_clearing_an_unregistered_figure_does_not_raise` — you're right and the test was worse than misleading.** `ax.plot()` does register a LINE layer, so the test never isolated the guard it named. It now builds a bare `fig.add_subplot()` and clears that, which puts the unregistered case where you identified it: `Axes.__init__` calls `cla()`, so the hook fires before any entry exists. Renamed to `test_clearing_a_brand_new_axes_does_not_raise`.

**The `fig.clear()` cascade** now has `test_figure_clear_empties_every_axes_on_a_multi_axes_figure`, asserting both lists end up empty across two panels however `clear_axes` and `clear()` interleave.

Plus the coverage for your main finding: `test_a_line_layer_is_registered_again_after_its_axes_is_cleared`, parametrized over `plot` and `step` — `ax.step` delegates to `Axes.plot` and goes through the same latch — and `test_clearing_an_axes_drops_the_accumulated_line_series`, since the series list would otherwise keep detached `Line2D` objects and the redrawn layer would describe lines that are no longer drawn.

Falsified, like the others:

| mutation | tests that fail |
|---|---|
| `forget_axes_state` not called | **3** (both line spellings + the series list) |
| `Axes.clear` / `cla` not patched | 5 |
| `clear_axes` drops every axes | 3 |
| `clear()` leaves `selector_ids` | 1 |

`uv run pytest` — **2342 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

One note on your closing line, that this was reviewed statically rather than by running the suite: the finding was correct anyway, and it was the kind my own tests were structurally unable to reach — every clear-and-replot case I wrote used `.bar()`, and the single `.plot()` was on a twinx axes I never cleared. A blind spot in the test matrix, not a missing assertion.

---

jooyoungseo:
Took the extensibility point in e15f661 rather than leaving it as a note, because the failure it guards is the quiet kind: the layers are dropped, the new module's latch survives, and the redraw registers nothing. That is precisely the regression this PR already shipped once and had to be caught in review — leaving the next instance to a mental note seemed like the wrong lesson to draw from that.

`test_lineplot_is_still_the_only_module_stashing_state_on_an_axes` AST-walks `maidr/patch/*.py` for `setattr(ax, ...)` / `setattr(axes, ...)` and holds the set to an explicit `allowed = {"lineplot.py"}`. AST rather than grep so a call spelled across several lines still counts, and it fails on the module that *adds* the state — where the fix belongs — instead of surfacing later as a chart that stopped being described.

Falsified by giving `barplot` a hypothetical latch:

```
AssertionError: {'barplot.py': [9]} keeps its own state on an Axes. matplotlib does
not reset attributes it does not own, so clearing the axes leaves it behind --
register its removal in `maidr.patch.lineplot.forget_axes_state` and add the module
to `allowed` here. See #499.
```

The message carries the remedy, so whoever trips it does not have to reconstruct why an unrelated test is complaining about their `setattr`.

On the docstring-length note: leaving them as they are. You read it the same way I did — they match `_AXES_WIDE_BAR_PLOTS` and `_drop_superseded_layers` in the same file — and trimming them here would make this file inconsistent with its neighbours to no benefit. If that becomes a project-wide convention it should change everywhere at once, not start with the file that just grew.

Also confirming your independent check of the `cla()`/`clear()` delegation, since you went further than I did: you are right that for a plain `Axes` the inner call resolves to the *patched* method at call time, so `ax.cla()` genuinely fires `_clear_axes` twice rather than merely being able to on some subclass. That makes the idempotency load-bearing rather than defensive, which is worth having stated on the record — `test_clearing_an_axes_that_holds_no_layers_is_a_no_op` covers it.

`uv run pytest` — **2343 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
Both gaps were real and the guard is fixed in 166ff19 rather than the docstring narrowed. You put your finger on why that was the right way round: the PR leans on this test as the safety net for the next contributor, so a guard with two documented ways around it is worse than no guard — it invites the reliance it cannot support.

I surveyed what a broad rule actually finds before rewriting, rather than patching the heuristic until your two examples passed:

```
file            line  target      attribute                    kind
lineplot.py      191  ax          DRAWN_SERIES                 setattr
lineplot.py      214  ax          PLOT_CREATED                 setattr
mplfinance.py    179  line        _maidr_ma_period             setattr
mplfinance.py    219  line        _maidr_date_nums             setattr
mplfinance.py    223  line        _maidr_datetime_converter    setattr
```

That reframed it. The target's name is the part that varies — `ax`, `instance`, `plot_ax`, anything — while **maidr's own state is always spelled `_maidr...`**. So the guard now keys on the attribute, which makes the variable name irrelevant and closes both of your cases at once instead of enumerating names. It also resolves module-level string constants, since `lineplot` spells its two through `DRAWN_SERIES` / `PLOT_CREATED` and a literal-only rule would have skipped the very module this exists for.

The allowlist records the **target** as well, because that is what decides whether cleanup is owed:

```python
allowed = {
    ("lineplot.py", "ax"):    "latch and series; forget_axes_state removes both",
    ("mplfinance.py", "line"): "on a Line2D; the clear discards the artist",
}
```

State on an axes survives the clear; state on an artist goes with it. Writing that per entry means a new one is a decision with a stated reason rather than a name added to a list. And the allowlist is asserted exhaustive, so an entry describing something that no longer exists cannot silently widen the guard — the failure mode I would otherwise have created while fixing yours.

Falsified against all four evasions, including both you named:

```
setattr, target named `a` (not ax/axes)        CAUGHT
direct assignment (your gap 1)                 CAUGHT
setattr on `instance` (your gap 2)             CAUGHT
assignment via `plot_ax` (your gap 2)          CAUGHT
```

`plot_ax` came straight from your `violinplot.py:363` reference, so that one is the real idiom rather than an invented case.

`uv run pytest` — **2343 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

On the docstring-length note: agreed with your read, and leaving them. You reached the same conclusion I did — they match `_drop_superseded_layers` and `_AXES_WIDE_BAR_PLOTS` in the same file — and if the project wants shorter ones that should change everywhere at once rather than starting with the file that just grew.

---

jooyoungseo:
Both points taken in c95c1b1. Neither changes behaviour — they were both cases of the code *claiming* more than it can support, which is the same failure this PR has now hit three times in review.

**The `fig.clear()` cascade.** You were right to flag it as version-dependent, and I checked rather than argued. On the matplotlib this was written against the cascade does fire:

```
matplotlib 3.10.9
Axes.clear fired during fig.clear(): 2 ['Axes', 'Axes']
```

But that is an observation, not a justification, and my docstring was using it as the latter — "the figure route cannot fail even with this line removed". If a matplotlib version detaches axes with `ax.remove()`, as you note some do, the cascade would not fire and the figure route *would* fail with the line removed. The docstring now leads with the durable reason (`clear()` is a public method; only calling it on its own pins what it does) and keeps the measurement as a version-scoped aside, with the `ax.remove()` caveat spelled out.

Worth noting the CI matrix cannot settle this either way: `uv sync --locked` pins one matplotlib across all four Python versions, so the matrix varies the interpreter, not matplotlib. The exposure is users on the declared `matplotlib>=3.8` floor, not CI.

**The AST guard's limits.** Fair, and the docstring now says so in as many words — it reads a name written literally or held in a module-level constant, and it sees neither a name built at runtime nor state kept *outside* the object, such as a module-level `WeakKeyDictionary` keyed by axes. That last one would need exactly the same cleanup and would pass silently.

Two rounds ago you flagged this guard as overpromising and I widened it. This round the honest move is the opposite direction — bound the claim rather than chase every evasion, because a guard whose edges nobody can see invites the reliance it cannot support. It now says what it catches and why that is the shape worth catching: it is the shape the bug actually took.

**The split idempotency rationale.** Folded into `clear_axes`'s own docstring, including the half that was only in `clear.py` — that `cla()`'s inner call resolves to the *patched* `clear()`, so a single call fires the hook twice by construction rather than merely being able to. That is your finding from the previous round; it belongs next to the property it explains.

`uv run pytest` — **2343 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
Merged as `9a4a690`.

On the registry suggestion — declining it, and the project's own rule is the reason. `CLAUDE.md` principle 1: *"No abstraction before a second caller exists."* You scoped it the same way ("not worth blocking on for a single caller"), so I think we agree on the substance; recording the decision so it does not get re-litigated.

The part I'd push back on slightly is the framing that the guard test "still requires editing `clear.py` per module". It does, but that is the property I would want: a registry makes registering cleanup *optional and easy to forget*, while a hardcoded call makes omitting it a visible edit in a file whose whole subject is clearing. The bug this PR fixed twice — once in the original patch, once in my own fix — was in both cases something not being cleaned up. Ownership living with the module that set the state is the better factoring in the abstract; being hard to forget is the better property for this specific failure mode. If a second module ever needs it, the guard test fails and whoever is there can weigh those against each other with two real callers in front of them instead of one hypothetical.

Thanks for the four rounds on this one. Two of them changed the shipped code rather than the prose — the `lineplot` latch regression especially, which would have made line and step charts silently undescribed and which my test matrix structurally could not reach, since every clear-and-replot case I had written used `.bar()`.

---

Issue #429, opened 2026-08-16
A NaN in a scatter, bar or errorbar still emits a payload the core cannot parse


jooyoungseo:
## Problem

Follow-up to #427, which fixed the line family (`MultiLinePlot` and everything extending it — line, step, smooth). The same defect is live in three other extractors.

The mechanism is unchanged: `json.dumps` writes `NaN`/`Infinity`/`-Infinity` as bare tokens, `JSON.parse` rejects all three, and the core's `catch` means `initMaidrOnElement` is never called — so the chart does not load at all.

## Measured

Sweeping plot types, after #427's fix:

```
ecdfplot            OK
line NaN gap        OK
step NaN gap        OK
boxplot             OK
violin              OK
heatmap             OK
regplot             OK
kdeplot             OK
stripplot           OK
pie                 OK

scatter with NaN    INVALID: NaN
bar with NaN        INVALID: NaN
errorbar NaN        INVALID: NaN
```

Reproduced with:

```python
plt.scatter([1, 2, float("nan"), 4], [1, 2, float("nan"), 4])
plt.bar(["a", "b", "c"], [1, float("nan"), 3])
plt.errorbar([1, 2, 3], [1, float("nan"), 3], yerr=[.1, .1, .1])
```

## Why it is filed separately rather than folded into #427

The line-family fix could drop the point without a judgement call, because in every case that reached it the non-finite coordinate was structural — seaborn's `-inf` staircase end, or matplotlib's own NaN line break. Neither names a value a reader could be told.

These three are not obviously the same:

- a **NaN bar height** is arguably "no bar for this category", which is closer to #112's missing-category question than to a line gap — and announcing the category with no value may be better than omitting it;
- a **NaN scatter point** is a point that was not drawn, so dropping looks right;
- a **NaN errorbar bound** may leave a usable estimate with only one bound, which the grammar can now express since xability/maidr#920.

So each wants its own decision, and folding them into one filter would settle three questions by accident.

## A note on scoping

It is worth considering whether the serialisation should also refuse to emit an unparseable payload — `json.dumps(..., allow_nan=False)` raises rather than writing the bare token. That converts a silently dead chart into a loud failure at render time, which is a real improvement in diagnosability but a behaviour change for anyone currently shipping such a figure. Raised here rather than decided.

## Acceptance

- Each of the three emits a payload that parses under `JSON.parse` semantics
- The choice made for each (drop the point, or keep it with the value omitted) is stated and tested, not implied
[...]

PR #501, opened 2026-08-18
fix(pointplot): keep a hued point plot's confidence intervals

### LINKED ISSUE #462: fix: a hued pointplot loses the confidence intervals it exists to show
## Description

`sns.pointplot(..., hue=…)` reads as a plain `line` layer. The estimates survive; the confidence intervals do not.

This is deliberate rather than accidental — `maidr/patch/pointplot.py` documents the choice:

> more than one estimate means a `hue` split the chart into groups while the MAIDR error bar layer carries a single series — so those intervals are dropped rather than mis-assigned.

Against today's core grammar that is the right call: `error_bar` takes one flat series, so pairing *n* groups' interval lines onto it would attach some group's bound to another group's estimate, which is worse than omitting them. But the outcome is still that a reader of a grouped point plot is given the means with nothing indicating the chart draws intervals around them — the thing a point plot exists to show, and the thing #246 added.

## Steps to reproduce

```python
sns.pointplot(data=df, x="g", y="v")             # → error_bar, intervals present
sns.pointplot(data=df, x="g", y="v", hue="grp")  # → line, intervals dropped
```

## Blocked on

xability/maidr#942 — the core grammar has no shape for a grouped error bar. Once `error_bar` can carry a series per group with a `z`, the fallback in `_register_point_layer` can be replaced by pairing each group's estimate with its own intervals.

The pairing machinery is already there and already tested: `_split`, `_pairs_up` and `_is_drawn` handle the counts, and `_pairs_up` deliberately verifies `len(intervals) == len(estimates) * groups` rather than trusting order — so the multi-estimate case is recognised today, it simply has nowhere to go.

## Related

The same core gap costs the R binding its group identity rather than its intervals: xability/r-maidr#183. Interval support: xability/maidr#789, implemented here in #246.

jooyoungseo:
## Description

`sns.pointplot(..., hue=…)` read as a plain `line` layer. The estimates survived; the intervals did not — a reader of a grouped point plot was handed the means of a chart drawn to show the uncertainty around them.

That was the right call at the time, and #462 says so: the error bar layer carried a single flat series with no field naming the group, so pairing *n* groups' interval lines onto it would have attached one group's bound to another's estimate. Omitting them was the lesser harm.

**maidr 4.4.0 removed the constraint.** `ErrorBarPoint[][]` with a `z` landed in xability/maidr#943, closing xability/maidr#942, and it is in the bundle this package ships:

```
py-maidr bundles maidr.js: 4.4.0
```

So the intervals now have somewhere to go.

## It works against the real bundle, not just on paper

Driven in Chromium against the bundled `maidr.js`, arrowing up from the first category:

```
g is a, value v is 0.82, grp is p
g is a, upper bound v is 1.15, grp is p
g is a, lower bound v is 0.77, grp is q
g is a, value v is 1.06, grp is q
g is a, upper bound v is 1.35, grp is q
```

`p`'s upper bound (1.15) and `q`'s lower (0.77) are **adjacent moves**, so a reader hears directly that the two intervals overlap — the comparison a point plot exists to support, and the group-major ordering xability/maidr#943 designed for. No page errors, and the highlight resolves.

Worth checking rather than assuming: emitting a shape the grammar documents is not the same as the shipped bundle reading it.

## How the pairing is established

Seaborn draws the interval polylines **estimate-major** — every category of the first group, then every category of the second. Measured, not assumed:

```
estimate 0: x=[0, 1, 2] y=[0.817, 2.353, 2.924]
estimate 1: x=[0, 1, 2] y=[1.062, 2.249, 3.082]
interval 0: x~[0.0] yspan=0.463..1.171     <- estimate 0, category a
interval 1: x~[1.0] yspan=1.919..2.783
interval 2: x~[2.0] yspan=2.526..3.336
interval 3: x~[0.0] yspan=0.754..1.379     <- estimate 1, category a
interval 4: x~[1.0] yspan=1.764..2.730
interval 5: x~[2.0] yspan=2.658..3.464
```

Each estimate's value falls inside the span of the interval its slice pairs it with, which is the check the tests make too — containment rather than position, since mis-pairing is the exact failure the old fallback existed to avoid.

Group names come from the legend, the only place they appear (the estimate lines carry `_child`-prefixed labels). The legend title becomes `axes.z`, so a chart grouped by treatment says "Treatment" rather than "Group". A group that cannot be named omits `z` rather than guessing.

## One regression the suite caught

A hue level missing from one category is NaN-padded by seaborn so the estimate lines stay the same length. The new path emitted that NaN straight into the JSON — which is not JSON, and stopped the chart initialising (#429). `tests/core/plot/test_non_finite_coordinates.py::test_the_payload_is_loadable` is what said so, via `parse_constant`. It now emits `null`, the same rule `MultiLinePlot._reading` applies.

### Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues

Closes #462. Depends on xability/maidr#942 / xability/maidr#943, both merged and released in v4.4.0.

## Changes Made

- `maidr/core/plot/pointplot.py` — `_extract_grouped` alongside `_extract_single`, with the per-group read factored into `_points_of` so both paths share the pairing, category labelling and dodge-rounding. `_extract_axes_data` adds `z` from the legend title, the same source `MultiLinePlot` uses. `_elements` is filled group-major so each group's rows take that group's slice.
- `maidr/patch/pointplot.py` — the multi-estimate case emits a grouped `error_bar` instead of falling through to `line`; `_group_labels` reads the legend.
- `maidr/core/plot/maidr_plot_factory.py` — dispatches on `estimates` (plural) as well as `estimate`.
- Tests — four that pinned the old fallback now pin the new behaviour, each keeping its original subject, plus a three-group case.

## Checklist

## Additional Notes

`uv run pytest` — **2344 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

**The ungrouped path is untouched** and still emits the flat form — asserted directly, since xability/maidr#943 is explicit that a single-series chart must not have to wrap itself in an array.

`test_a_horizontal_dodge_names_its_groups` moved its assertion from `y` to `x`. That is a change of key rather than of behaviour: an error bar layer carries the category as `x` in **both** orientations and lets `orientation` say which is on screen where — the convention `ErrorBarPlot` documents and the single-series path has always used. `test_a_horizontal_chart_names_its_groups_without_a_dodge` draws with `Axes.plot` for exactly this reason, and its docstring says so.

Falsified, each mutation failing only its own tests:

| mutation | tests that fail |
|---|---|
| revert to the `line` fallback | **5** |
| wrong stride slicing intervals per group | **2** |
| emit NaN instead of `null` for a padded sample | **3** |

The three-group test exists because a two-group off-by-one still lands inside the interval list and mis-pairs silently; a third group is what makes a wrong stride reach visibly wrong bounds.

---

### DIFF
--- a/maidr/core/plot/lineplot.py
+++ b/maidr/core/plot/lineplot.py
@@ -210,13 +210,9 @@ class MultiLinePlot(MaidrPlot, LineExtractorMixin):
         """
         axes_data = super()._extract_axes_data()
 
-        legend = self.ax.get_legend()
-        if legend is not None:
-            title = legend.get_title()
-            if title is not None:
-                z_label = title.get_text().strip()
-                if z_label:
-                    axes_data[MaidrKey.Z] = self._axis_config(label=z_label)
+        z_label = self._legend_title()
+        if z_label:
+            axes_data[MaidrKey.Z] = self._axis_config(label=z_label)
 
         return axes_data
 
--- a/maidr/core/plot/maidr_plot.py
+++ b/maidr/core/plot/maidr_plot.py
@@ -115,6 +115,17 @@ class MaidrPlot(ABC, FormatExtractorMixin):
         Shared rather than restated: ``MultiLinePlot`` and ``PointPlot`` both
         need it and would otherwise drift apart if the convention changed.
 
+        Read **live**, when the schema is built, while the group *names* are
+        captured once as the plotting call is patched. A caller who retitles
+        or relabels the legend in between can therefore make the two
+        disagree -- the points would carry the names the chart was drawn
+        with and the axis would carry the new title. Pre-existing, and
+        sharper for ``PointPlot`` than for ``MultiLinePlot`` because there
+        the divergence reaches per-point ``z`` values rather than one label.
+        Reconciling them means either freezing the title at registration or
+        re-reading the names at render, and both are changes to when a
+        layer decides what it says.
+
         Returns
         -------
         str
--- a/maidr/core/plot/maidr_plot.py
+++ b/maidr/core/plot/maidr_plot.py
@@ -105,6 +105,30 @@ class MaidrPlot(ABC, FormatExtractorMixin):
 
         return maidr_schema
 
+    def _legend_title(self) -> str:
+        """
+        The grouping variable's name, as the legend titles it.
+
+        A ``hue`` split names its groups in the legend entries and names the
+        *variable* in the legend title, and the title is the only place that
+        name appears -- so this is what an ``axes.z`` label is read from.
+        Shared rather than restated: ``MultiLinePlot`` and ``PointPlot`` both
+        need it and would otherwise drift apart if the convention changed.
+
+        Returns
+        -------
+        str
+            The title, or an empty string when there is no legend or no
+            title on it.
+        """
+        legend = self.ax.get_legend()
+        if legend is None:
+            return ""
+        title = legend.get_title()
+        if title is None:
+            return ""
+        return title.get_text().strip()
+
     @staticmethod
     def _axis_config(
         label: str | None = None,
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -72,7 +72,10 @@ class MaidrPlotFactory:
             # the same layer; they differ only in what the library drew.
             # `Axes.errorbar` leaves a container, while seaborn's point plot
             # leaves the lines the patch resolved and hands them over here.
-            if isinstance(kwargs.get("estimate"), Line2D):
+            # `estimates` (plural) is the `hue`-split form, one line per
+            # group; `estimate` is the single-series one. Either says the
+            # lines came from seaborn.
+            if isinstance(kwargs.get("estimate"), Line2D) or kwargs.get("estimates"):
                 return PointPlot(single_ax, **kwargs)
             return ErrorBarPlot(single_ax, **kwargs)
         elif PlotType.HEAT == plot_type:
--- a/maidr/core/plot/pointplot.py
+++ b/maidr/core/plot/pointplot.py
@@ -101,6 +101,17 @@ class PointPlot(ErrorBarPlot):
         # the cursor was in the first.
         self._elements.clear()
 
+        # `_pairs_up` establishes that the intervals divide evenly among the
+        # estimates, and the patch will not construct this type unless it
+        # does. Checked here anyway: the invariant lives in another file, so
+        # a caller constructing `PointPlot` directly -- or a future change to
+        # the patch -- would otherwise mis-slice in silence, handing one
+        # group another's bounds. That is the exact failure the `line`
+        # fallback existed to prevent, and this module raises rather than
+        # guesses everywhere else.
+        if len(self._intervals) % len(self._estimates):
+            raise ExtractionError(self.type, self.ax)
+
         per_group = len(self._intervals) // len(self._estimates)
         # All the groups are named or none is, so a layer never declares an
         # `axes.z` that some of its series do not carry. `_extract_axes_data`
--- a/maidr/core/plot/pointplot.py
+++ b/maidr/core/plot/pointplot.py
@@ -102,11 +102,15 @@ class PointPlot(ErrorBarPlot):
         self._elements.clear()
 
         per_group = len(self._intervals) // len(self._estimates)
+        # All the groups are named or none is, so a layer never declares an
+        # `axes.z` that some of its series do not carry. `_extract_axes_data`
+        # reads the same predicate.
+        named = self._named_groups()
         series: list[list[dict]] = []
 
         for index, estimate in enumerate(self._estimates):
             intervals = self._intervals[index * per_group : (index + 1) * per_group]
-            group = self._groups[index] if index < len(self._groups) else ""
+            group = self._groups[index] if named else ""
             points = self._points_of(estimate, intervals, is_vertical, labels, group)
             if not points:
                 raise ExtractionError(self.type, self.ax)
@@ -246,6 +250,12 @@ class PointPlot(ErrorBarPlot):
         treatment says "Treatment" rather than "Group". Omitted for an
         ungrouped chart, which has nothing to name.
 
+        Also omitted when the groups themselves could not be named. The two
+        have to agree: declaring a ``z`` axis while some series carry no
+        ``z`` is a shape the consumer has no reading for, and it is reachable
+        whenever the legend does not list exactly one entry per drawn group.
+        Either the whole layer is grouped-and-named or none of it is.
+
         Returns
         -------
         dict
@@ -253,18 +263,25 @@ class PointPlot(ErrorBarPlot):
         """
         axes_data = super()._extract_axes_data()
 
-        if not self._estimates:
+        if not self._estimates or not self._named_groups():
             return axes_data
 
-        legend = self.ax.get_legend()
-        if legend is not None:
-            title = legend.get_title()
-            if title is not None:
-                z_label = title.get_text().strip()
-                if z_label:
-                    axes_data[MaidrKey.Z] = self._axis_config(label=z_label)
+        z_label = self._legend_title()
+        if z_label:
+            axes_data[MaidrKey.Z] = self._axis_config(label=z_label)
 
         return axes_data
+
+    def _named_groups(self) -> bool:
+        """
+        Whether every drawn group has a name to emit as ``z``.
+
+        Returns
+        -------
+        bool
+            True when the patch supplied one name per estimate.
+        """
+        return len(self._groups) == len(self._estimates) and all(self._groups)
 
     def _is_vertical(self) -> bool:
         """
--- a/maidr/core/plot/pointplot.py
+++ b/maidr/core/plot/pointplot.py
@@ -8,7 +8,7 @@ from matplotlib.axis import Axis
 from matplotlib.lines import Line2D
 
 from maidr.core.enum import MaidrKey
-from maidr.core.plot.errorbar import ErrorBarPlot
+from maidr.core.plot.errorbar import ErrorBarPlot, _is_drawn
 from maidr.exception import ExtractionError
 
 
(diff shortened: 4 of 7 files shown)


jooyoungseo:
All four points addressed in 8a255e0. The first was a real inconsistency and the second turned out to have the same fix.

## 1 + 2: naming is now all-or-nothing

You're right that the schema could declare an `axes.z` while some series carried no `z`. `_group_labels` truncated to the group count, so a short legend named the groups it covered and left the rest unnamed, while `_extract_axes_data` emitted `axes.z` on the strength of the layer being grouped at all.

Both now read the same predicate: `_group_labels` returns nothing unless the legend lists **exactly** one entry per drawn group, and `_named_groups()` gates both the per-point `z` and the `axes.z`. Your option (a), because an unlabelled grouped chart is a shape the consumer already handles — it is what the old `line` fallback produced — while an index-based label would invent a name the chart never showed. The bounds, which are the point of the change, survive either way.

That same check answers your second point, which I'd otherwise have had to guard separately. The count mismatch **is** the cheap sanity check for legend-order-equals-draw-order, and it's now stated in the docstring rather than left implicit:

> The interval-to-estimate pairing is verified against the drawn geometry (`_pairs_up`, and the bounds bracket their estimate); the name-to-group pairing has no equivalent, because a name has no geometry to check it against. A count mismatch is the one symptom available.

That asymmetry is worth naming exactly as you framed it — the interval side is verified, the name side is assumed, and pretending otherwise would be the failure this PR is otherwise careful about.

## 3: the duplicated legend read

Extracted `_legend_title` onto `MaidrPlot`, and `MultiLinePlot` now uses it too. Worth doing now specifically because there are two callers — before this PR there was one, and extracting then would have been abstraction ahead of need.

## 4: the all-or-nothing highlight drop

Leaving as-is, and thanks for flagging rather than filing it as a defect. It matches `_extract_single`'s policy and `test_a_partial_interval_promises_no_highlight` pins the reasoning: the consumer resolves the selector to one element per point and **discards the result outright** when the count disagrees, so a partial list highlights nothing anyway. Per-group selectors would be a real improvement, but that is a change to what the layer promises rather than a fix, and it belongs in its own PR against a measured case.

## Test gaps

Both added — and the first one caught me. My initial version built the chart, then replaced the legend with a short one and asserted no `z`. **It failed**, correctly: the names are read while the call is being patched, so a later `ax.legend(...)` leaves what was captured alone. A legend short at *registration* is the case being protected against, and I could not stage it through seaborn without inventing a chart it may never draw. So the test drives the guards directly — `_group_labels` for the reader, a short `_groups` list for the emitter — which pins the actual behaviour instead of a scenario I'd contrived.

The second is a hue level that is a singleton in every category: no drawable interval of its own, and it must not cost its sibling group theirs.

Falsified: truncating instead of refusing fails 1, emitting `axes.z` for an unnamed layer fails 1. Together with the earlier three, that's five mutations each failing only its own tests.

`uv run pytest` — **2346 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

On docstring density: noted, and I'll leave it. It matches `ErrorBarPlot` and `_drop_superseded_layers` in the files being edited, and thinning only the methods this PR touched would make those files internally inconsistent.

---

jooyoungseo:
You were right to doubt that test, and it was worse than a mismatched docstring — it never reached the branch it named.

## The singleton test

Measured before rewriting. Three identical observations is not a singleton:

```
zero-variance (the old fixture): group x ['bound','bound']  group y ['bound','bound']
true n=1:                        group x ['bound','bound']  group y ['NO-bound','NO-bound']
```

Seaborn's bootstrap over a zero-variance cell still draws an interval, so **both** groups came back bounded and the "group with no drawable interval" path was never exercised.

Your point about the assertions is the more important half, and I checked it rather than taking it: under a mutation that fabricates a bound where none was drawn,

```
old fixture + any("yMin" in point ...) -> still PASSES
rewritten test                         -> FAILS
```

`any(...)` over the flattened schema is satisfied by the sibling group alone. It said nothing about the group under test, and would have gone on saying nothing.

Rewritten in 7c93d28 with a genuine single observation per category and per-group assertions: the sibling keeps every bound and brackets its estimate; the singleton group carries its estimate and category with no interval.

One thing I found while doing it — the file already has `test_a_group_without_an_interval_does_not_cost_the_others_theirs` for the **ungrouped** case, where a single *category* lacks an interval inside one series. Mine is a whole *series* lacking them, which is the only version that exercises the per-group slicing. Near-identical names for genuinely different cases, so mine is now `test_a_hue_level_with_no_drawable_interval_keeps_its_sibling_bounded` and both docstrings cross-reference.

## The legend-order assumption

Filed as **#502** rather than left as a note, because your framing of *why* it is weaker than its neighbour is the part worth preserving: the interval pairing has a real cross-check (`_pairs_up`'s count guard, plus bounds that must bracket their estimate, so a wrong stride is visibly wrong), while a name has no geometry to check it against.

I also confirmed your observation that the suite could not catch a reordering: every hue fixture uses interchangeable levels (`x`/`y`, `p`/`q`), so correct and swapped are indistinguishable by inspection. The issue records what a real test would need — a hue level whose name is recoverable from its values — and flags that CI pins one seaborn from `uv.lock`, so the exposure is users on versions CI never runs.

Agreed it is an accepted trade rather than a blocker, and it is now written down where someone bumping the seaborn floor will find it.

`uv run pytest` — **2346 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
Both code points taken in fbb744f, and the first one turned out to be sharper than either of us stated.

## The `per_group` contract

You were right that the invariant lives in another file with nothing asserting it at the consuming end. Added the check — but measuring *which* mismatches actually slip past changed what it is for:

```
5 intervals over 2 estimates: ExtractionError with or without the guard
7 intervals over 2 estimates: renders 2 groups of 3 without it, raises with it
```

A list one **short** is already caught downstream — `_points_of` raises when a group's slice does not cover its categories. A list one **long** is not: `per_group` is 3, both groups slice cleanly, and the seventh interval is dropped in silence.

So the guard is load-bearing, but for the opposite case from the one I first assumed. My initial test used the short list and **passed with the guard removed** — vacuous, caught by mutation rather than by reading it. That is twice in this PR now that a test of mine proved to be asserting nothing until a mutation said so; the pattern seems to be that I reach for the "obviously broken" input, which is exactly the one an existing check already covers.

The test now uses the long list, and the docstring records both measurements so a later reader does not mistake the choice for arbitrary.

## The `_is_drawn` collision

Real trap, and worse than it looked from one file: `core/plot/pointplot.py` imports the **scalar** `_is_drawn` from `errorbar.py`, while `patch/pointplot.py` defines a **vertex-array** one — so the two halves of this feature had one name meaning two different things. Renamed the vertex version to `_has_drawn_segment`, which says what it answers.

## The two I am not changing

**Highlighting going dark across groups** — confirming it is intended rather than an oversight, and you are right that the blast radius grew. The consumer resolves the selector to one element per point and discards the result outright when the count disagrees, so a partial list highlights nothing anyway; the choice is between "no highlight" and "no highlight, plus a promise the layer cannot keep". Per-group selectors would genuinely be better and are a change to what the layer promises rather than a fix, so they want their own PR against a measured case. You are the second reviewer to raise it, which is a reasonable signal that it deserves an issue rather than a comment — I will file one if it comes up again with a concrete chart.

**`axes.z` read at render vs `z` captured at draw** — agreed it is a real skew and agreed it is pre-existing, shared with `MultiLinePlot`. Noting rather than fixing, since changing it here would make this PR the place a `MultiLinePlot` behaviour quietly diverged.

`uv run pytest` — **2347 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings. The legend-order assumption you raised earlier is filed as #502.

---

jooyoungseo:
All three checked; one code change in 82321b0, and your shared-legend scenario turned out to be already covered — for a reason worth recording.

## The shared-legend hole does not open

You proposed an axes hosting more than one hue-encoded layer sharing a legend, where counts might coincidentally match while order does not. Measured both orderings:

```
pointplot, then a hued lineplot on the same axes
  legend entries: ['x', 'y', 'x', 'y']
  _group_labels(ax, 2) -> []

an unrelated hued layer with DIFFERENT levels drawn first
  legend at end:     ['P', 'Q', 'x', 'y']
  PointPlot._groups: []          (not ['P', 'Q'])
```

The count guard catches both, structurally rather than by luck: another hued layer contributes its own entries, so the legend grows *past* the group count. The second ordering is the one that would have been dangerous — the prior layer's `['P', 'Q']` are sitting in the legend at capture time — and it comes back refused rather than mislabelled.

It also confirms names are captured at registration, so a layer drawn afterwards cannot retroactively corrupt names already taken.

What is genuinely left needs a legend holding *exactly* the group count in a non-draw order, which requires seaborn to list one layer's own levels out of order — the original assumption, not a new route to it. Recorded on #502, which is narrower now than when I filed it.

## The unreachable guard

You were right, and it is removed. `_extract_plot_data` routes to `_extract_grouped` only when `_estimates` is non-empty, and every iteration appends or raises, so `series` cannot be empty there. Removed rather than kept with a comment defending it: a check that cannot fire is worse than none, because it sends the next reader hunting for the case that produces it. There is now a comment saying why there is no guard.

## `if group:` depending on a call-site invariant

Agreed it is not locally obvious, and agreed it is not reachable — `_named_groups()` requires `all(self._groups)`, so an empty name cannot arrive with `named=True`. Leaving it: the alternative is passing a redundant flag alongside the name, which trades one non-local invariant for two parameters that must agree.

## Docstring/type agreement

Confirmed — `list of list of dict` and `list[list[dict]]` both correct, no drift.

`uv run pytest` — **2347 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
Took the test half of your merge condition rather than the comment half, in 86d6198 — a comment would restate the assumption, and only a test can fail when it stops holding.

## The `hue_order` regression test

`test_a_group_name_lands_on_that_group_s_own_values` picks hue levels whose **names are recoverable from their values**: `low` sits near 1, `high` near 100. That is the point — every other hue fixture in the file uses interchangeable levels (`x`/`y`, `p`/`q`), so a correct naming and a swapped one read identically and nothing could tell them apart. Run with the default order and with `hue_order` reversed, since a reversal is the cheapest way seaborn could plausibly diverge and the one a caller can trigger today.

It is a real detector rather than a restatement. Reversing the name-to-group assignment:

```
FAILED test_three_groups_each_take_their_own_intervals
FAILED test_a_hue_level_with_no_drawable_interval_keeps_its_sibling_bounded
FAILED test_a_group_name_lands_on_that_group_s_own_values
```

Nothing in the suite caught that mutation before this commit.

Both orderings verified against real seaborn first, so the test pins measured behaviour rather than my expectation of it:

```
hue_order=None            legend ['low','high']  -> z='low' [1,2,3]  z='high' [100,101,102]
hue_order=['high','low']  legend ['high','low']  -> z='high' [100,101,102]  z='low' [1,2,3]
```

## The highlighting gap

Also closed — you were right that the grouped path's `_support_highlighting` fallback had assertions only on the ungrouped side. `test_a_hue_level_with_no_drawable_interval_keeps_its_sibling_bounded` now checks `_support_highlighting is False` and `elements == []`, and removing the fallback fails it.

## The empty-string hue level

Leaving as-is, and agreed it is the least-bad fallback. A hue column containing `""` would drop `z` for the whole layer rather than for that group alone — but the alternative is emitting `z: ""` for one series, which announces an unnamed group as though it were named. Dropping the naming for a chart whose groups cannot all be distinguished is the more honest of the two, and it is the same all-or-nothing rule the rest of the naming follows.

`uv run pytest` — **2348 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

---

PR #503, opened 2026-08-18
refactor(util): give CDN version resolution its own module

### LINKED ISSUE #293: Split maidr/util/dependencies.py into focused modules

jooyoungseo:
## Description

The last half of #293's suggested shape. `maidr/util/cdn.py` takes the resolver — endpoints, timeout budget, semver pin handling, the `latest` lookup and its caching, and the URL builders.

`dependencies.py` goes **1781 → 835 lines** and is finally what the issue asked for: reaching the copy bundled inside the wheel.

| #293's concern | module | done |
|---|---|---|
| Bundled asset access | `dependencies.py` | ✅ |
| **CDN version resolution** | **`cdn.py`** | ✅ **this PR** |
| Bundled-copy freshness | `bundle_freshness.py` | ✅ #497 |
| *(not in the issue)* Bundled-copy capability | `bundle_capability.py` | ✅ #494 |
| *(the issue's third note)* Shared warning policy | `warn.py` | ✅ #496 |

## The shim is weaker here, and that changed the plan

The earlier cuts leaned on `dependencies.__getattr__` for back-compat. It forwards attribute **reads** — `dependencies.get_cdn_version` still resolves, to the same object. **It cannot forward a write**, because `setattr` goes straight to the module dict:

```
read-forward works                 : True
write to dependencies reaches cdn? : False
```

So `monkeypatch.setattr(dependencies, "urlopen", ...)` would set an attribute nothing in `cdn` reads — a patch that silently does nothing rather than one that fails. Since **18 tests patch `dependencies.urlopen`** and more patch `_fetch_latest_version` / `_resolved_cdn_version`, leaving them to the shim would have produced a green suite testing nothing.

Consequences, all deliberate:

- Only the **public** names are forwarded. The resolver's state is absent from `_MOVED_TO_CDN` on purpose — listing it would make the trap look supported.
- **266 test references** were repointed at `cdn` rather than left to the shim.
- `test_the_shim_forwards_reads_but_cannot_forward_a_patch` asserts the asymmetry, so a future change that starts forwarding writes has to come and edit it.

## Measured before moving, not after

The direction is one-way: **no asset-side function references a CDN name**, checked by walking the AST rather than by reading. So `cdn` imports `dependencies` and nothing imports back.

Three names stayed behind, for the reason #293 records after the first cut:

- **`_version_key`, `_UNKNOWN_VERSION`** — used by `bundle_freshness`, so they are shared *version* primitives rather than CDN ones. Moving them would have made freshness depend on the resolver to compare two local strings.
- **`_warn_placeholder_css`** — found later, by ruff: `bundled_css_path` calls it too. It is about the placeholder *asset*, not the CDN.

## Two tests needed real updates, both predicted

**`test_no_render_path_references_the_unresolved_constants`** is the exclusion #293's third note said would need repointing "to whichever module ends up owning them". It now allows the owner **and** the shim's forwarding list — `_MOVED_TO_CDN` names the constants so the old import path resolves, and a forwarding list is not a render path.

**`maidr_css_cdn_url`'s deprecation** named `cdn_url(MAIDR_MATH_CSS_FILENAME) from maidr.util.dependencies`, which after the move sent readers to a module where neither name lives:

```
AssertionError: the deprecation for maidr_css_cdn_url names symbols that
do not resolve: ['MAIDR_MATH_CSS_FILENAME']
```

`test_deprecation_names_only_importable_symbols` caught it. The advice now names `cdn`, and the filename is re-exported there so both halves resolve where the reader is sent.

### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor (no behaviour change)

## Related Issues

Closes #293 — with the caveat below.

## Changes Made

- `maidr/util/cdn.py` — new, 1115 lines.
- `maidr/util/dependencies.py` — 1781 → 835. Gains `_MOVED_TO_CDN` and a third shim branch; keeps the version primitives and `_warn_placeholder_css`.
- `maidr/util/bundle_freshness.py` — reads `_resolver_outcome` and `_published_version` through `_cdn.`, per the from-import trap #497 recorded.
- `maidr/__init__.py`, `api.py`, `core/maidr.py`, `plotly/plotly_maidr.py`, `altair/altair_maidr.py` — repointed at the canonical module.
- 8 test files + `conftest.py` — 266 references repointed, including every `setattr(dependencies, "<moved>", ...)`.
- `tests/core/test_bundle_capability.py` — the laziness guard covers `cdn`.

## Checklist

## Additional Notes

`uv run pytest` — **2349 passed, 13 skipped**. `ruff check` unchanged at 6 pre-existing findings.

Verified rather than assumed:

- Identity preserved through the shim — `maidr.ResolverOutcome is cdn.ResolverOutcome` → `True`, which matters because `except` compares class objects.
- The write limitation reproduces exactly as documented (above).
- The laziness guard is **not vacuous**, though for this pair every module-scope import spelling fails loudly anyway (`cdn` from-imports `dependencies`, so the cycle raises `partially initialized module`). Checked directly, it flags `import maidr.util.cdn`, `from maidr.util import cdn`, and `from maidr.util.cdn import cdn_url` alike.

**On closing #293:** this completes the split it describes, but the issue named two modules and five exist. Worth a maintainer's eye on whether that shape is what was wanted before it closes — I have marked it `Closes` rather than `Refs` because the CDN cut was the last thing outstanding, not because the bookkeeping matches.

---

### DIFF
--- a/maidr/__init__.py
+++ b/maidr/__init__.py
@@ -218,21 +218,23 @@ from .util.bundle_freshness import (  # noqa: E402
     bundle_status,
     resolver_outcome,
 )
-from .util.dependencies import (  # noqa: E402
+from .util.cdn import (  # noqa: E402
     BUNDLED_TAG,
     CDN_TIMEOUT_ENV_VAR,
     CDN_VERSION_ENV_VAR,
     LATEST_TAG,
     ResolverOutcome,
+    get_cdn_version,
+    reset_cdn_version_cache,
+    set_cdn_version,
+)
+from .util.dependencies import (  # noqa: E402
     bundled_css_path,
     bundled_js_path,
     bundled_math_css_path,
-    read_bundled_math_css,
-    get_cdn_version,
     maidr_js_version,
     read_bundled_js,
-    reset_cdn_version_cache,
-    set_cdn_version,
+    read_bundled_math_css,
 )
 from .util.warn import BUNDLE_WARNING_ENV_VAR  # noqa: E402
 
--- a/maidr/altair/altair_maidr.py
+++ b/maidr/altair/altair_maidr.py
@@ -44,6 +44,8 @@ from htmltools import HTML, HTMLDocument, Tag, tags
 
 from maidr.util.dependencies import (
     MAIDR_VEGALITE_FILENAME,
+)
+from maidr.util.cdn import (
     cdn_url,
 )
 from maidr.util.environment import Environment
--- a/maidr/altair/altair_maidr.py
+++ b/maidr/altair/altair_maidr.py
@@ -56,7 +56,7 @@ from maidr.util.iframe_utils import wrap_in_iframe_plotly
 # CDN configuration for the Altair adapter. The ``vegalite.js`` adapter
 # tracks the newest published ``maidr`` release, resolved to a concrete
 # version at render time by
-# :func:`maidr.util.dependencies.get_cdn_version` so browsers never replay a
+# :func:`maidr.util.cdn.get_cdn_version` so browsers never replay a
 # stale ``@latest`` response. Pin with ``MAIDR_CDN_VERSION`` when a specific
 # version is needed.
 # ---------------------------------------------------------------------------
--- a/maidr/api.py
+++ b/maidr/api.py
@@ -248,6 +248,8 @@ def init_notebook(
         MAIDR_JS_FILENAME,
         read_bundled_js,
         read_bundled_math_css,
+    )
+    from maidr.util.cdn import (
         bundled_cdn_url,
     )
 
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -32,11 +32,13 @@ from maidr.util.bundle_freshness import warn_if_bundle_is_stale
 from maidr.util.dependencies import (
     MAIDR_JS_FILENAME,
     OFFLINE_FALLBACK_REPORT,
-    bundled_cdn_url,
     inline_bundle_tags,
     maidr_bundled_files_dependency,
     maidr_bundled_relative_dir,
     maidr_html_dependency,
+)
+from maidr.util.cdn import (
+    bundled_cdn_url,
     maidr_js_cdn_url,
 )
 from maidr.util.environment import Environment
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -43,11 +43,13 @@ from maidr.util.bundle_freshness import warn_if_bundle_is_stale
 from maidr.util.dependencies import (
     MAIDR_JS_FILENAME,
     OFFLINE_FALLBACK_REPORT,
-    bundled_cdn_url,
     inline_bundle_tags,
     maidr_bundled_files_dependency,
     maidr_bundled_relative_dir,
     maidr_html_dependency,
+)
+from maidr.util.cdn import (
+    bundled_cdn_url,
     maidr_js_cdn_url,
 )
 from maidr.util.environment import Environment
--- a/maidr/util/bundle_freshness.py
+++ b/maidr/util/bundle_freshness.py
@@ -20,10 +20,11 @@ import warnings
 from typing import NamedTuple
 
 from maidr.util import dependencies as _deps
+from maidr.util import cdn as _cdn
+from maidr.util.cdn import ResolverOutcome
 from maidr.util.dependencies import (
     _UNKNOWN_VERSION,
     _version_key,
-    ResolverOutcome,
 )
 from maidr.util.warn import (
     BUNDLE_WARNING_ENV_VAR,
@@ -138,7 +139,7 @@ def resolver_outcome() -> ResolverOutcome | None:
     >>> resolver_outcome()  # doctest: +SKIP
     ResolverOutcome(resolved='4.2.0', unreachable=(), answered_badly=())
     """
-    return _deps._resolver_outcome
+    return _cdn._resolver_outcome
 
 
 def bundle_status(*, resolve: bool = True) -> BundleStatus:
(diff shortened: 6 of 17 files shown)


jooyoungseo:
Flagging a CI oddity rather than reading it as a pass, and correcting myself along the way.

**`claude-review` reported success but posted nothing.** My first reading of the log was that it never ran, because `ANTHROPIC_API_KEY:` is empty — that was wrong. It authenticates via `CLAUDE_CODE_OAUTH_TOKEN`, and the result block shows it ran properly:

```
duration_ms: 99282   num_turns: 21   is_error: false
permission_denials_count: 15
```

So it reviewed for ~99 seconds across 21 turns and then hit **15 permission denials**, one of which was presumably the `gh pr comment` it was instructed to post with. The findings exist; they just never reached the PR. That is an infrastructure failure, not an approval, and I have re-run the workflow rather than treat a green check as a clean review.

Worth recording for #305, which already tracks this workflow's auth problems: the failure mode here is the quiet one. The check goes green, the job is marked successful, and the absence of a review looks like "nothing to say".

**Self-reviewed in the meantime**, and it found one thing (e177359):

`maidr/altair/altair_maidr.py`'s header comment told readers the version is resolved by `maidr.util.dependencies.get_cdn_version`, which after this branch lives in `maidr.util.cdn`. Exactly the same class as the `maidr_css_cdn_url` deprecation that `test_deprecation_names_only_importable_symbols` caught — prose sending someone to a module where the symbol is not. Swept `maidr/`, `tests/` and `docs/` for the shape afterwards; nothing else does it.

Three properties I checked directly rather than assuming, since the shim is the risky part of this PR:

```
public moved names: 15 | not forwarded: none
forwarded but not identical: none
maidr.__all__ entries missing: none
```

Identity matters beyond tidiness — `except MaidrBundleStaleWarning` and `isinstance` compare class objects, so a shim that returned a copy would break callers silently.

**2349 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
Merged as `a48ffc6`, **without an automated review** — disclosing that plainly rather than letting a green check imply one happened.

`claude-review` ran twice on this PR and posted nothing both times:

| run | duration | turns | cost | denials | posted |
|---|---:|---:|---:|---:|---|
| first | 99 s | 21 | $0.63 | 15 | no |
| re-run | 794 s | 12 | $5.29 | 41 | no |

Both `is_error: false`, both green. Denials went *up* on re-run, so it is deterministic rather than flaky and a third attempt would just spend more. Evidence recorded on #305, including the suggestion that a missing review should fail the check rather than pass it — a green check meaning "no review happened" silently converts an unreviewed PR into an apparently-reviewed one.

Since no external review was coming, I ran an adversarial pass myself rather than merging on green CI alone. It caught one thing, and the thing it caught was **my** error:

```
bundled URL (pinned)  : https://cdn.jsdelivr.net/npm/maidr@3.75.1/dist/maidr.js
```

I had asserted `bundled_cdn_url` returns the *bundled* version and it returned the *pinned* one. Reading the docstring showed that is the documented contract — a pin wins, deliberately, so a page cannot load two different builds of maidr.js. Pre-existing behaviour, correctly preserved; my probe's assumption was wrong. Worth stating because it is the failure mode this whole PR is about: a plausible expectation that the code contradicts for a good reason.

The corrected pass covers what the split could plausibly have broken:

```
pinned URL            : maidr@3.75.1/dist/maidr.js
bundled URL (no pin)  : maidr@4.4.0/dist/maidr.js   (bundled VERSION: 4.4.0)
invalid pins refused  : yes -- '../evil', '9.9.9; rm -rf /', '١.٢.٣'
bundle_status(offline): BundleStatus(bundled='4.4.0', published=None, ...)
```

The pin sanitisation is the one worth calling out: it is a security boundary — the guard that stops a hostile `MAIDR_CDN_VERSION` or a compromised registry response steering the `<script src>` — and it sits squarely inside the moved code. Path traversal, shell metacharacters and non-ASCII digits are all still refused, and the non-ASCII case matters because `\d` would have accepted it while the shell-side guard would not.

Plus **2349 passed, 13 skipped** across four Python versions, the browser suite green, and `ruff` unchanged at the 6 pre-existing findings.

---

PR #504, opened 2026-08-18
perf(shiny): render off the event loop, one render per figure at a time

jooyoungseo:
## Description

`maidr.render` never awaits, so on Shiny's event loop it holds it for its whole duration. Every other session on that worker waits — once per reactive flush, per input change (#454).

Measured **through the real renderer**, not the raw `savefig` the issue quotes, with the same harness before and after:

```
idle control                 longest loop gap     1.3 ms
8 renders, on the loop       longest loop gap   484.9 ms
8 renders, off the loop      longest loop gap    13.4 ms
```

## Why this works, and what it costs

Moving it to a thread works because the expensive part **releases the GIL**: `savefig` is 87–88% of a render at every chart size. Had it held the GIL throughout, `to_thread` would have relocated the work without unblocking anything, and this PR would be pointless. That was worth checking rather than assuming.

**The honest cost:** the same eight renders take **484 ms** of wall-clock on the loop and **565 ms** off it. A lone user rendering sequentially pays ~17% in thread handoff. That is the trade — one user slightly slower so concurrent users stop blocking each other — and quoting only the 484.9 → 13.4 number would oversell it.

## The lock is what makes this safe, not just faster

`savefig` writes `fig.dpi` (100 → 72 → 100) for the duration of the write. Two concurrent writes to **one** figure race on that single attribute, and the loser renders its whole chart at the other's dpi:

```
-L 640 345.6      ->  +L 460.8 345.6      (640 / (100/72) = 460.8)
```

Measured: 1 of 6 concurrent renders, no exception raised, a perfectly valid SVG at the wrong size. For a tool whose highlight overlay is positioned against that geometry, silently wrong dimensions are the bad kind of wrong — and intermittent, so it would not reproduce reliably in a bug report.

Two properties of the lock, both deliberate:

- **Per figure, not process-wide.** `savefig` on *distinct* figures is safe in parallel (verified). A single global lock would serialise unrelated sessions and give back most of the win.
- **Weak-keyed.** It adds no retention of its own, which matters while #456/#498 are open about exactly that problem. Pinned by a test.

## Two tests that were wrong before they were right

**The headline test did not run.** My first version was an `async def` with `@pytest.mark.asyncio` — this suite has no `pytest-asyncio`, so pytest collected it, **silently skipped it, and reported a pass**. It did not fail when I put the render back on the loop, because it never executed. It now goes through the file's existing `_render` helper, whose docstring already explains why the suite avoids that dependency.

**The retention test asserted the wrong thing.** It used `plt.subplots()`, which registers with `FigureManager` — and that retains every figure for the life of the process (#456). The figure stayed alive for a reason that had nothing to do with the lock map. It now uses a bare `Figure()` so only this map could hold it.

Both are the same failure this session keeps producing: a test that passes without exercising its subject.

### Type of Change

- [ ] Bug fix
- [x] Performance improvement
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor

## Related Issues

Addresses the Shiny half of #454 (its option 1). Leaves #454 open: `session.dynamic_route` — the payload size, a different cost — is untouched, and the ungrouped `maidr.render()` API is unchanged for non-Shiny callers.

## Changes Made

- `maidr/widget/shiny.py` — `_FIGURE_LOCKS` (weak-keyed), `_figure_lock`, and `_render_off_loop`; `render()` now awaits `asyncio.to_thread`.
- `tests/widget/test_shiny.py` — four tests: off-loop execution, lock stability, the unresolvable-figure fallback, and non-retention.

## Checklist

## Additional Notes

`uv run pytest` — **2353 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

Each guarantee falsified:

| mutation | tests failing |
|---|---|
| render put back on the loop | 1 |
| lock no longer stable per figure | 1 |
| lock map made strong | 1 |

An unresolvable figure gets a *fresh* lock rather than a shared sentinel-keyed one: serialising things we cannot tell apart is a step toward a deadlock, and the render is safe on its own — only the lock's scope is lost.

---

### DIFF
--- a/maidr/core/context_manager.py
+++ b/maidr/core/context_manager.py
@@ -74,12 +74,49 @@ class BoxplotContextManager(ContextManager):
 
 
 class HighlightContextManager:
+    """Carries a render's highlight wiring from extraction to the SVG writer.
+
+    Held in :class:`contextvars.ContextVar`\ s rather than as class
+    attributes, matching :class:`ContextManager` above. That is not a style
+    choice: the artist ``draw`` methods and ``XMLWriter.start`` are patched
+    *class-wide*, so every render in the process reads this state while
+    ``savefig`` walks its figure.
+
+    As plain class attributes it was safe only because every render ran
+    serialised on one thread. It stopped being safe when the Shiny renderer
+    began rendering off the event loop (#504): two **different** figures
+    drawing at once would overwrite each other's ``elements_to_highlight``,
+    and artists checked after the overwrite would match nothing, so
+    ``XMLWriter.start`` never injected their ``maidr`` attribute. Measured
+    before the change -- four concurrent renders of distinct figures went
+    from 61 selectors each to ``[7, 1, 1, 1]``. A valid SVG, with the
+    interactive layer silently gone.
+
+    ``contextvars`` is the right tool rather than a wider lock because a
+    render's wiring is genuinely per-render: ``asyncio.to_thread`` runs the
+    call in a *copy* of the context, so each render gets its own view and
+    concurrent renders of distinct figures stay parallel, which is what
+    moving off the loop was for.
+
+    One subtlety worth stating, because it is the way this fix is usually
+    got wrong: ``copy_context()`` copies the variable-to-value mapping, not
+    the values. A single shared ``dict`` left as a default would still be
+    shared through the copy, so :meth:`set_maidr_elements` installs a fresh
+    mapping for each render rather than mutating one in place.
+    """
+
     _instance = None
     _lock = threading.Lock()
 
-    elements = {}
-    elements_to_highlight = []
-    selector_ids = []
+    _elements: contextvars.ContextVar[dict] = contextvars.ContextVar(
+        "maidr_highlight_elements"
+    )
+    _elements_to_highlight: contextvars.ContextVar[list] = contextvars.ContextVar(
+        "maidr_elements_to_highlight"
+    )
+    _selector_ids: contextvars.ContextVar[list] = contextvars.ContextVar(
+        "maidr_selector_ids"
+    )
 
     def __new__(cls):
         if not cls._instance:
@@ -90,33 +127,40 @@ class HighlightContextManager:
 
     @classmethod
     def is_maidr_element(cls, id):
-        return id in cls.elements
+        return id in cls._elements.get({})
 
     @classmethod
     def get_selector_id(cls, id):
-        return cls.elements[id]
+        return cls._elements.get({})[id]
 
     @classmethod
     @contextlib.contextmanager
     def set_maidr_element(cls, element, id):
-        if element not in cls.elements_to_highlight:
+        to_highlight = cls._elements_to_highlight.get([])
+        if element not in to_highlight:
             yield
             return
 
+        elements = cls._elements.get({})
         try:
-            cls.elements[id] = cls.selector_ids[
-                cls.elements_to_highlight.index(element)
-            ]
+            elements[id] = cls._selector_ids.get([])[to_highlight.index(element)]
             yield
         finally:
-            del cls.elements[id]
+            del elements[id]
 
     @classmethod
     @contextlib.contextmanager
     def set_maidr_elements(cls, elements: list, selector_ids: list):
-        cls.elements_to_highlight = elements
-        cls.selector_ids = selector_ids
+        # A fresh mapping per render, not a shared one mutated in place:
+        # `copy_context()` copies the variable-to-value mapping, so a single
+        # dict installed once would still be shared across every concurrent
+        # render. See the class docstring.
+        token_elements = cls._elements.set({})
+        token_to_highlight = cls._elements_to_highlight.set(elements)
+        token_selectors = cls._selector_ids.set(selector_ids)
         try:
             yield
         finally:
-            cls.elements_to_highlight.clear()
+            cls._elements.reset(token_elements)
+            cls._elements_to_highlight.reset(token_to_highlight)
+            cls._selector_ids.reset(token_selectors)
--- a/maidr/widget/shiny.py
+++ b/maidr/widget/shiny.py
@@ -11,7 +11,10 @@ Requires the optional ``shiny`` extra::
 
 from __future__ import annotations
 
+import asyncio
+import threading
 import warnings
+import weakref
 from typing import Any, Literal, Optional, Union
 
 try:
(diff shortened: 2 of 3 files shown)


jooyoungseo:
The docstring finding is the best kind of catch — the code was right and the thing shipped alongside it was wrong. Fixed in e3e3e38.

## The contradiction

You're right, and it's worse than an inconsistency: **the PR body had it right and the docstring had it backwards**, which is exactly the wrong way round. The body is read once; the docstring is what someone finds six months from now.

The mismatched numbers were the cause. `602.7 / 14.9 ms` came from the raw-`savefig` experiment I ran for #454, not from this renderer — I pasted them across without re-deriving. The docstring now carries the measurement this change was actually validated on, with wall-clock beside loop-gap so the trade is visible rather than implied:

```
idle control              1.3 ms
on the loop             484.9 ms     wall 484 ms
off the loop             13.4 ms     wall 565 ms
```

and says **"It is not free"** in as many words. Also recorded the executor ceiling you noted — `min(32, cpu_count + 4)` — so a reader knows where this stops scaling instead of discovering it.

## The concurrency test

Added, and you were right that it was the gap that mattered: every other test approached the race through `_figure_lock` returning matching objects, which is not the same as showing two renders serialise.

It detects **overlap** rather than timing, so it cannot flake into a false pass — if the lock works, the second render cannot enter while the first is inside, at any speed; the sleep only widens the window a broken lock has to miss. Verified in both directions rather than assumed:

| | result |
|---|---|
| lock removed from the render path | **failed 3/3** |
| lock made per-call instead of per-figure | **failed 3/3** |
| restored | **passed 5/5** |

## The duplicate `get_axes`

Keeping it, and measuring turned "cheap" into a number:

```
get_axes median : 0.16 us
warm render     : 56.1 ms
duplicate call  : 0.00029% of one render
```

Threading it through means either making `_check_supported` return the resolved axes — turning a validator into a resolver, so it no longer does one thing — or resolving on the loop thread, which moves work back onto the loop this PR exists to keep free. Both cost more in clarity than 0.16 µs buys back. Worth measuring rather than agreeing on instinct, since "cheap" was the whole basis for the call.

Quoted annotation unquoted; the module already defers annotations.

**2354 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

One note on your closing remark about the two vacuous tests: that is now three in this session — the `async def` that was silently skipped, the retention test that measured `FigureManager`'s retention instead of the lock map's, and earlier a pointplot test that never reached the branch it named. All three passed while asserting nothing, and all three were caught by mutation rather than by reading them. Consistent enough to be a habit rather than an accident: I write the assertion I expect to hold, and it holds for the wrong reason.

---

jooyoungseo:
All five taken in 38e3ece. Point 2 was the sharp one and it was right — I measured rather than agreeing, and the measurement is worse than the review suggested.

## The `except Exception` documented a case it cannot see

```
plotly-shaped    -> returned None
None             -> returned None
int              -> returned None
str              -> returned None
empty list       -> raised StopIteration
dict             -> raised StopIteration
```

Exactly as you traced: a foreign figure matches no branch in `get_axes` and **returns** `None`, which the `getattr` default already handles. It never reaches the `except`. So the comment explained the branch by the one case that cannot get there, while the branch's real job was swallowing anything else `get_axes` might throw — immediately before an unsynchronised render.

Narrowed to `(AttributeError, StopIteration)` and logged at debug. A real bug in `get_axes` now leaves a record instead of being silently downgraded to "lock scope lost".

## Point 3 — and it is the session's recurring theme again

You're right that `test_an_unresolvable_figure_gets_a_fresh_lock_...` shows the helper works, not that anything reaches it. Two tests now drive `render()` for real, covering **both** shapes separately, because they are different paths — the one that returns `None` and the one that raises. That distinction is precisely what the old comment got wrong, so testing them as one would have re-encoded the same mistake.

## Point 1 — contention eats the same pool

Recorded in the docstring. The part worth having written down is the mechanism you identified: a render waiting on `_figure_lock` is blocked *inside* its executor thread, still holding the slot — so contention on one shared figure can make **unrelated** figures queue, which is the stall this PR removes reappearing one level down. Noted that typical Shiny usage is a figure per session, where it does not arise.

## Points 4 and 5

`_set_output_metadata` was dead — verified `render()` completes without it, and no other test in the file sets it. Dropped. Stdlib imports hoisted to the top to match the file.

## Falsification

| mutation | result |
|---|---|
| debug log removed | **1 failed** |
| catch narrowed past the real exception | **detected — as an error, not a failure** |

The second is worth flagging honestly: with the catch missing `StopIteration`, it escapes a coroutine and surfaces as `RuntimeError: coroutine raised StopIteration`, so pytest reports an error rather than a clean failure. It hung my first falsification run before I isolated it with a timeout. Detected either way — it cannot pass silently — but "fails" would have been the tidier claim and it is not the true one.

**2356 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
All three documented in e2dc408. Checking the first one turned up **a second reason the lock is needed**, which I had missed.

## `savefig` mutates two things, not one

I only knew about `fig.dpi`. Watching attributes from another thread during 30 renders:

```
fig.dpi    : 100 -> 72 -> 100
fig.canvas : FigureCanvasAgg -> FigureCanvasSVG -> FigureCanvasAgg
```

The canvas swap comes from `FigureCanvasBase._switch_canvas_and_return_print_method`, whose own docstring says it *"temporarily setting the canvas for saving the figure"*. It is the same hazard by a different attribute — two concurrent writes to one figure race on it too — and the PR's justification named only half of it.

## Which answers your backend question, rather than caveating it

You're right that GUI backends want canvas work on the main thread, and that `savefig` now runs on a worker. But **the write never reaches the GUI canvas**: `savefig` swaps in the canvas for the output format for the duration, so an SVG write goes through `FigureCanvasSVG` — pure Python, no thread affinity — whatever the figure's interactive canvas happens to be.

`maidr/backend.py` delegates to `FigureCanvasAgg` besides, and is what `import maidr` activates, so the interactive case needs a user who set `MPLBACKEND` explicitly *and* it still would not route the write through their canvas. Documented as an answer rather than an assumption.

## Non-reentrancy

Stated, with the reasoning rather than just the fact: nothing re-enters a render of one figure on one thread today, and if something ever does, **a deadlock is the better outcome than two interleaved writes** — it is the one that shows up. An `RLock` would let a nested render proceed straight into the race this exists to prevent.

## `_figure_lock(None)` gives no synchronisation

Said plainly in the docstring, including why it is fine now and how it stops being fine: plotly and altair renders touch neither `dpi` nor `canvas`, so there is nothing to race on — but a future plot type that resolves to `None` here *and* mutates shared figure state would be unprotected silently.

Thank you for tracing the two safety claims independently rather than re-reading my docstrings — `_get_plot_or_current`'s `plt.gcf()` branch being unreachable on this path is the part I would have been least able to check convincingly from inside my own change, and it is load-bearing for "distinct figures are safe in parallel".

**2356 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
**You were right, and this was a blocking defect — without your catch this PR would have shipped an accessibility regression.** Fixed in cb084f3.

## Confirmed, and worse than the description

I reproduced it before touching anything. Four concurrent renders of *distinct* figures, counting `maidr="…"` selector attributes in the output:

```
serial      [61, 61, 61, 61]
concurrent  [ 7,  1,  1,  1]
```

Not a subtle degradation — the interactive layer is almost entirely gone. Valid SVGs, on the common path (anything with patches, lines or collections), and only under concurrent traffic, so it would never have reproduced from a user's bug report.

My safety argument was wrong in exactly the way you identified: I verified `fig.dpi` and `fig.canvas`, concluded "distinct figures are safe in parallel", and never asked what *else* `savefig` touches. `HighlightContextManager` is read by class-wide patches on the artist `draw` methods and on `XMLWriter.start`, so it is in every render's path, and my per-figure lock was designed specifically not to cover distinct figures.

## The fix

`contextvars`, as you suggested, matching `ContextManager` directly above it in the same file rather than inventing anything. A render's wiring is genuinely per-render, and `asyncio.to_thread` runs the call in a *copy* of the context — so distinct figures stay parallel, which widening the lock would have given back.

One subtlety worth recording, because it is how this fix is usually got wrong and I nearly did: **`copy_context()` copies the variable-to-value mapping, not the values.** A single `dict` installed once as a default would still be shared through the copy. `set_maidr_elements` therefore installs a fresh mapping per render rather than mutating one in place. My first falsification attempt shared only that dict and the test still passed — which told me my mutation was wrong, not that the fix was unnecessary; a faithful revert to the original class attributes fails 3/3.

```
serial      [61, 61, 61, 61]
concurrent  [61, 61, 61, 61]   x6 trials
```

## The test

Added, and it is the one you described: two *different* figures rendering concurrently, asserting the selector count survives. Counts rather than diffing documents, since ids are per-render uuids and the count is what disappears.

## Smaller notes

`_FIGURE_LOCKS` is now keyed `Figure` rather than `Any` — you're right that `None` short-circuits before it. On the docstring length I'll leave it as is; it matches `_drop_superseded_layers` and `_AXES_WIDE_BAR_PLOTS` in this codebase, and trimming only the new ones would make the files inconsistent.

**2357 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

Worth saying plainly: three reviews on this PR each found something real, and this one found the thing that mattered most. I had verified two shared attributes and generalised to "distinct figures are safe", which is precisely the kind of claim I should have tested rather than reasoned to — and the test that would have caught it is the one the PR's own benchmark scenario implied.

---

jooyoungseo:
Measured your `FigureManager.figs` point rather than reasoning about it, and it lands somewhere more specific than either of us said. Filed as **#505**.

Thread ids recorded inside `create_maidr` and `get_maidr` during a real `@render_maidr` call:

```
writes (create_maidr) : loop only
reads  (get_maidr)    : on worker too
```

**Writes never leave the event loop.** Registration happens while the *user's* plotting code runs — inside `await self.fn()` — and only the render is offloaded. So there is no cross-thread check-then-act, and no corruption is reachable today.

But your instinct to flag it was right, because the safety is **incidental rather than designed**. It rests on an invariant nobody wrote down: *every write to `figs` happens on the loop thread, because plotting is done by user code before the render rather than by the render*. Three ordinary changes break it without touching `FigureManager` — a lazily-registering render path, a user whose render function offloads its own plotting with `to_thread` (a reasonable thing to do), or extending this treatment to Streamlit. In any of those, `_get_maidr` could mint two `Maidr` objects for one figure and split a chart's layers.

You also spotted that `FigureManager`'s docstring calls it thread-safe. That claim predates any real concurrent access, and it is now half-true: safe for the access pattern that exists, not the one the name implies. #505 records that, with a test-the-invariant option as the only one that would actually catch a regression.

**On your other three:**

- **`_figure_lock(None)`** — the invariant is documented at the fallback; I've left it there rather than also annotating `get_axes`, since a note in two places drifts. #505 is the better home if this needs enforcing.
- **Docstring length** — leaving as is, consistently with the other reviews that raised it. The perf numbers earn their place: they are the answer to "why a lock and not something else", which is the question a future reader will actually have. If the project wants shorter docstrings that should change everywhere at once.
- **Warning registry under threads** — agreed, and worth stating the bound: worst case is a *duplicate* warning, never a missing one, since `warn_once`'s dedup set is only ever added to. Not worth a lock for that.

**2357 passed, 13 skipped.** All 20 checks green including the browser suite. Merging.

---

PR #506, opened 2026-08-18
fix(core): make figure registration atomic

### LINKED ISSUE #505: FigureManager.figs is now read from worker threads, and its thread-safety is incidental
Raised in review of #504 and recorded rather than fixed there, because measurement narrowed it to something that is not a live defect — but is an invariant nobody wrote down and nothing enforces.

## What changed

#504 moved the Shiny render off the event loop, so `maidr.render()` now runs on a worker thread. `FigureManager.figs` is a plain, unsynchronised `dict`, and `_get_maidr` does a check-then-act on it. The concern was that concurrent sessions could now race there.

## What is actually true

Measured by recording the thread id inside `create_maidr` and `get_maidr` during a real `@render_maidr` call:

```
writes (create_maidr) : loop only
reads  (get_maidr)    : on worker too
```

**Writes stay on the event loop; only reads reach the worker.** That is because registration happens while the *user's* plotting code runs — inside `await self.fn()` — and only the render is offloaded. So there is no cross-thread check-then-act, and with the GIL making individual dict operations atomic, no corruption is reachable today.

## Why it is still worth recording

The safety is **incidental, not designed**. It depends on a property nobody stated and no test pins:

> every write to `FigureManager.figs` happens on the thread that owns the event loop, because plotting is done by user code inside the render function rather than by the render itself.

Three plausible changes break it without touching `FigureManager`:

1. A render path that registers a figure lazily — anything calling `create_maidr` from inside `maidr.render()` rather than before it.
2. A user's render function that offloads its own plotting (`await asyncio.to_thread(build_chart)`), which is a reasonable thing to do and would put `create_maidr` on a worker.
3. Extending the off-loop treatment to another integration whose registration is not already on the loop — Streamlit, or a future async API.

In any of those, `_get_maidr`'s check-then-act could mint two `Maidr` objects for one figure and split a chart's layers across them. That is the quiet kind of failure: a chart that renders, with some layers missing.

Worth noting `FigureManager`'s own docstring already describes it as thread-safe. That claim predates any real concurrent access and is now half-true — safe for the access pattern that exists, not for the one the name implies.

## Options

1. **Record the invariant** — a comment on `figs` saying writes are loop-thread-only and why, so whoever breaks it has been warned. Cheapest, and matches how #504 handled `_figure_lock(None)`.
2. **Guard the check-then-act** with the lock `FigureManager` already has for other purposes. Small, and removes the dependence on where callers happen to run.
3. **Test the invariant** — assert `create_maidr` is never reached off the loop thread during a render, which is the only option that fails when someone breaks it.

3 is the one that would actually catch a regression; 2 makes the question moot. They are not exclusive.

## Related

- #504 — where this became reachable, and which fixed the same class of bug for `HighlightContextManager` (that one was a live defect: concurrent renders of distinct figures dropped from 61 selectors to `[7, 1, 1, 1]`)
- #456 / #498 — the other open issues about this mapping's lifetime

jooyoungseo:
## Description

Closes the hole #504 opened and #505 recorded. `_get_maidr` does a check-then-act on `FigureManager.figs`:

```python
if fig not in cls.figs.keys():
    cls.figs[fig] = Maidr(fig, plot_type)
return cls.figs[fig]
```

That was safe only because every caller ran on one thread — a property nobody wrote down, and one #504 stopped guaranteeing when it moved the Shiny render off the event loop.

## What is actually true today, measured

Thread ids recorded inside `create_maidr` and `get_maidr` during a real `@render_maidr` call:

```
writes (create_maidr) : loop only
reads  (get_maidr)    : on worker too
```

Registration still stays on the loop, because plotting is the *user's* code and runs inside `await self.fn()` before the render is offloaded. **So this is not a live defect.** It is an invariant that holds by accident of where callers live rather than by anything this method does.

Three ordinary changes break it without touching `figure_manager.py`:

1. a render path that registers a figure lazily, from inside `maidr.render()`;
2. a user whose render function offloads its own plotting (`await asyncio.to_thread(build_chart)`) — a reasonable thing to write;
3. extending the off-loop treatment to Streamlit or a future async API.

The failure is the quiet kind: two `Maidr` objects for one figure, the loser dropped from `figs` while the layers registered against it are not — a chart that renders with some layers silently missing.

## The fix

`FigureManager` already carries `_lock` for its singleton guard; the check and the insert now share it. One uncontended acquire per registered layer.

I took **both** options 2 and 3 from #505 rather than either alone: the lock makes the question moot, and the test is the part that fails if someone later decides the lock is unnecessary.

### Type of Change

- [x] Bug fix (latent — not reachable through today's call paths)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor

## Related Issues

Closes #505. Follows #504, which made the surrounding code genuinely multi-threaded for the first time and fixed the same class of bug for `HighlightContextManager` — that one *was* live, dropping concurrent renders from 61 selectors to `[7, 1, 1, 1]`.

## Changes Made

- `maidr/core/figure_manager.py` — the check-then-act moves under `_lock`; the `figs` docstring records that reads now reach worker threads and what protects writes.
- `tests/core/test_figure_manager.py` — eight threads on a `Barrier` registering one figure, asserting exactly one `Maidr` results.

## Checklist

## Additional Notes

`uv run pytest` — **2358 passed, 13 skipped**. `ruff check` unchanged at the 6 pre-existing findings.

Falsified rather than assumed. A `Barrier` alone does not reliably lose the race, so the mutation widens the window with a `sleep` between the check and the insert:

| | result |
|---|---|
| unguarded check-then-act | **failed 3/3** |
| restored | **passed 3/3** |

Worth saying that the test asserts a property (one `Maidr` per figure) rather than reproducing a specific interleaving, so it stays meaningful if the internals change.

---

### DIFF
--- a/maidr/core/figure_manager.py
+++ b/maidr/core/figure_manager.py
@@ -134,7 +134,8 @@ class FigureManager:
         # before the render is offloaded -- but that is a property of where
         # callers happen to live, not of this method, and losing it would
         # mint two `Maidr` objects for one figure and split a chart's layers
-        # between them. The lock costs one uncontended acquire per layer.
+        # between them. Two uncontended acquires per registered layer, counting the
+        # paired append in `create_maidr`.
         with cls._lock:
             if fig not in cls.figs:
                 cls.figs[fig] = Maidr(fig, plot_type)
@@ -155,9 +156,14 @@ class FigureManager:
             is the change, since "No MAIDR found for figure" described maidr's
             own bookkeeping rather than anything a user could act on (#443).
         """
-        if fig not in cls.figs.keys():
-            raise UnsupportedPlotError(fig)
-        return cls.figs[fig]
+        # Locked for the same reason the writes are: this is a check-then-act,
+        # and `destroy` popping between the two lines would turn the careful
+        # `UnsupportedPlotError` below into a bare `KeyError` -- losing the
+        # message that is the whole point of raising it.
+        with cls._lock:
+            if fig not in cls.figs:
+                raise UnsupportedPlotError(fig)
+            return cls.figs[fig]
 
     @classmethod
     def destroy(cls, fig: Figure) -> None:
@@ -169,6 +175,11 @@ class FigureManager:
                 maidr = cls.figs.pop(fig)
         except KeyError:
             return
+        # Teardown runs outside the lock, deliberately -- it is real work,
+        # not a dict operation. That leaves one gap this does not close: a
+        # `create_maidr` that took its reference just before the pop can
+        # still append to an object no longer in `figs`. Its layers go
+        # nowhere, which is #456's territory rather than this lock's.
         maidr.destroy()
         del maidr
 
--- a/maidr/core/figure_manager.py
+++ b/maidr/core/figure_manager.py
@@ -28,10 +28,12 @@ class FigureManager:
         Maidr instances.
 
         Reads reach worker threads since #504, which renders off the Shiny
-        event loop. Writes go through :meth:`_get_maidr` under ``_lock``, so
-        two concurrent registrations of one figure cannot each create a
-        ``Maidr`` for it. Individual dict operations are atomic under the
-        GIL; the lock is for the check-then-act around them.
+        event loop. Both write paths -- :meth:`_get_maidr`'s insert and
+        :meth:`destroy`'s pop -- take ``_lock``, as does the paired append
+        of ``plots`` and ``selector_ids`` in :meth:`create_maidr`, which
+        must stay index-aligned. Individual dict and list operations are
+        atomic under the GIL; the lock is for the check-then-act and the
+        paired write around them.
 
     Methods
     -------
@@ -75,9 +77,27 @@ class FigureManager:
 
         # Add plot to the Maidr object associated with the plot's figure.
         maidr = cls._get_maidr(ax.get_figure(), plot_type)
+
+        # Extraction stays *outside* the lock: it is the expensive part and
+        # touches only the artists it was handed.
         plot = MaidrPlotFactory.create(axes, plot_type, **kwargs)
-        maidr.plots.append(plot)
-        maidr.selector_ids.append(Maidr._unique_id())
+
+        # The two appends do not. `plots` and `selector_ids` are separate
+        # lists held index-aligned -- `Maidr._flatten_maidr` and
+        # `_create_html_tag` both zip them, and `_drop_superseded_layers`
+        # documents what misalignment costs: every surviving layer wears its
+        # neighbour's id, so the highlight lands on the wrong mark with
+        # nothing raised.
+        #
+        # Each `append` is atomic under the GIL; the *pair* is not. Two
+        # concurrent registrations on one figure can interleave between
+        # them, which reproduces deterministically:
+        #
+        #     plots        ['plot-A', 'plot-B']
+        #     selector_ids ['id-B',   'id-A']
+        with cls._lock:
+            maidr.plots.append(plot)
+            maidr.selector_ids.append(Maidr._unique_id())
         return maidr
 
     @classmethod
@@ -141,8 +161,12 @@ class FigureManager:
 
     @classmethod
     def destroy(cls, fig: Figure) -> None:
+        # Under the same lock as the registering writes: this is the second
+        # write path into `figs`, and a `pop` racing a `_get_maidr` insert
+        # decides which of them wins by timing.
         try:
-            maidr = cls.figs.pop(fig)
+            with cls._lock:
+                maidr = cls.figs.pop(fig)
         except KeyError:
             return
         maidr.destroy()
--- a/maidr/core/figure_manager.py
+++ b/maidr/core/figure_manager.py
@@ -27,6 +27,12 @@ class FigureManager:
         A dictionary that maps matplotlib Figure objects to their corresponding
         Maidr instances.
 
+        Reads reach worker threads since #504, which renders off the Shiny
+        event loop. Writes go through :meth:`_get_maidr` under ``_lock``, so
+        two concurrent registrations of one figure cannot each create a
+        ``Maidr`` for it. Individual dict operations are atomic under the
+        GIL; the lock is for the check-then-act around them.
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Took the cleanup note in d542668. You framed it as consistent-with-practice and not a regression, which is true — but this test registers a figure purely to race on it, and the retention it was adding to is the subject of an open issue (#456). One line, so the argument for leaving it was thin.

Two things in the review I want to acknowledge specifically, because they checked claims I had not made:

**The reentrancy question.** `_lock` is non-reentrant and shared with `__new__`'s singleton guard, so a call from inside the critical section back into `FigureManager` would self-deadlock. I did not state that I had checked it; you did, and confirmed `Maidr.__init__` doesn't reference `FigureManager` at all. That is the failure this change could plausibly have introduced, and it is now on the record with the condition to watch — if `Maidr.__init__` ever grows a call back into `FigureManager`.

**The lock's scope.** You verified `MaidrPlotFactory.create(...)` and the `plots.append(...)` mutations stay *outside* the critical section, so it covers the dict check-then-insert and not plot extraction. That was deliberate but unstated; holding the lock across extraction would serialise real work rather than a dict operation.

On the global-versus-per-figure tradeoff: agreed it is a non-issue here. Registration runs once per layer, not per render, and an uncontended acquire is tens of nanoseconds — a per-figure lock would need its own weak-keyed map (as `widget/shiny.py` has) to buy nothing measurable at this call frequency. The per-figure lock in #504 exists because it guards `savefig`, which is 87–88% of a render; this one guards a dict lookup.

**2358 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
You found the same race one level up from where I put the lock, and it is worse than the one I was fixing — this one is a **live misalignment**, not a latent one. Fixed in f219740.

## Reproduced deterministically

My first probe failed to show it because all threads slept the same amount and stayed in lockstep. Forcing the interleave you described — one registration preempted between its two appends:

```
plots        ['plot-A', 'plot-B']
selector_ids ['id-B',   'id-A']
index-aligned: False
```

Exactly as you set it out. Every layer wearing its neighbour's id is the failure `_drop_superseded_layers` documents: the highlight lands on the wrong mark, nothing raised.

Both appends are now under `_lock`. `MaidrPlotFactory.create` stays outside it — the **previous** review verified that scope deliberately, and holding the lock across extraction would serialise real work rather than a list operation. Your finding and that one are compatible: the factory call out, the paired appends in.

## `destroy` — and a false claim I introduced

You're right, and it is worse than an omission: the docstring I added in the previous commit said *"Writes go through `_get_maidr` under `_lock`"*, which `destroy`'s unguarded `figs.pop` made untrue the moment I wrote it. Corrected rather than reworded, and `destroy` now takes the lock.

## The test was vacuous, again

My alignment test asserted equal lengths and unique ids. **Both hold under a scrambled order**, so it passed with the lock removed — caught only by mutating. Detecting misalignment needs the id traceable to the layer it was minted for, which a uuid is not, so each thread now mints an id naming itself and the assertion is that every layer sits opposite its own call's id. Fails 3/3 unguarded, passes 3/3 restored.

That is the **fifth** test this session that passed while asserting nothing. The pattern is now unmistakable: I assert the properties that are easy to observe rather than the one that would break, and only mutation testing tells the difference.

Imports hoisted, as flagged.

**2359 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

---

jooyoungseo:
Both taken in e8b6bd3, and chasing the second one exposed a third problem of my own making.

## `get_maidr` — you're right, and it was inconsistent

Locked. A `destroy` landing between the `in` and the `[]` turns the deliberate `UnsupportedPlotError` into a bare `KeyError`, losing the message that is the whole reason it exists. Read side and write side are symmetric now, which is what the docstring already implied.

## The racy tagging — my comment was wrong, not just the code

You identified it exactly: `create_maidr` releases the lock **before returning**, so reading `plots[-1]` afterwards can pick up another thread's layer. My comment defended it as safe "because the append happens under the same lock as the id's" — true of the pair, irrelevant to a read after the return. It could invent a misalignment the code never committed *and* mask a real one. Layers are now tagged as they are built.

## Then the fixed test turned out to be vacuous

It **passed with the lock removed**. The pacing hook I wrote was monkeypatched onto a name nothing calls, so no interleave was ever forced — the test just waited out a timeout and ran sequentially.

It now pauses inside `Maidr._unique_id`, which `create_maidr` evaluates *between* the two appends — the actual seam the race runs through, rather than a hook I invented. The wait is a deadline rather than a handshake, so it terminates either way: unguarded, B completes both appends while A is parked and the ids land crossed; guarded, B cannot enter the critical section and it simply times out.

```
unguarded : 1 failed, 1 failed, 1 failed
restored  : 1 passed, 1 passed, 1 passed
```

Deterministic, in milliseconds. Two threads with an explicit seam rather than a crowd off a barrier — a crowd is a *probabilistic* detector, and I measured it: eight threads caught the bug on 3 of 5 runs, sixteen on 5 of 5.

## And the hang that came from chasing 5/5

Raising that count is what broke the suite. A regex meant for one test bumped the **other** test's `Barrier` to 16 while it still started 8 threads, so it waited forever for arrivals that could not come — and because the `join` was unbounded, a wrong number became a hung suite rather than a failed test. The count is one constant now and every `join` is bounded, so the next mismatch fails loudly.

Worth recording as its own lesson: a broad scripted edit hit a test it was not aimed at, and the damage was invisible until the suite stopped terminating.

## `destroy`'s remaining gap

Named rather than left to look handled: teardown runs outside the lock, so a `create_maidr` holding a reference from just before the pop can still append to an object no longer in `figs`. Its layers go nowhere. That is #456's territory, not this lock's, and the comment says so.

**2359 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

That is now **six** tests this session that passed while asserting nothing, every one caught by mutation rather than by reading. The consistency is the finding: I write the assertion I expect to hold, and it holds for a reason unrelated to the code under test.

---

PR #508, opened 2026-08-18
fix(core): stop retaining every figure for the life of the process

### LINKED ISSUE #456: FigureManager.figs retains every figure for the life of the process

jooyoungseo:
Closes #456.

## The defect

`FigureManager.figs` was a class-level `dict`, so a figure stayed reachable for the life of the process once maidr registered it.

Registration happens when a chart is **plotted**, not when it is rendered, so this applied to every supported figure — including ones never rendered at all. The `plt.show()` path escaped it only because `maidr/backend.py` calls `destroy` explicitly. A Shiny or Streamlit render never goes through `plt.show`, so a long-lived server accumulated one figure, and every artist maidr extracted from it, per render.

## Measured

25 renders, each building its own figure and calling `plt.close()`:

| chart | before | after |
|---|---|---|
| 10 bars | 17.65 MB, `figs` +25 | **0.26 MB, `figs` +0** |
| 200 bars | 177.80 MB, `figs` +25 | **2.30 MB, `figs` +0** |

The `after` column matches the control #498 established (drop the entry by hand: 0.27 MB / 1.32 MB), so what remains is baseline allocation rather than retention.

## The fix, and why it is none of the three options on #456

The record now lives on the figure itself rather than in a module-level table. That makes the whole graph — figure, its `Maidr`, its layers, and the artists those layers hold — **one isolated reference cycle** once the application lets go, which the cyclic collector reclaims.

| case | figure collected |
|---|---|
| module-level dict (today) | no |
| record stored on the figure | **yes** |

This is deliberately **not** a `WeakKeyDictionary`. That was option 1, it was tried, and it freed nothing — every value reaches its own key (`Maidr._fig`, `MaidrPlot.ax`, and each subclass's own artist handles such as `BarPlot._own_bars`), so the entry keeps the figure alive and the weak key never dies. Keeping the value *on* the key sidesteps that entirely, because a value that reaches its key is exactly what a cycle *is*, and cycles are collectable as long as nothing outside points in.

It also avoids what was wrong with the other two:

- **Option 2 (bounded LRU)** evicts entries for figures the app still re-renders, stripping a live chart of its data. Here lifetime is exactly the application's own reference, so there is no bound to tune and nothing to evict early.
- **Option 3 (drop on `close_event`)** is wrong per #452 — a figure built lazily and cached (`@reactive.calc`, any memoised helper) is closed by the render that opened it but is still the live chart. Here a cached figure keeps its record for as long as the cache holds it.

### It also corrects #498's premise

#498 says "no lifetime fix in `FigureManager` can work" because `MaidrPlot` holds live artists. Measurement says otherwise: popping the `figs` entry by hand frees the figure today, cycles and all. Those references are a barrier to a *weak key*, not to the *cyclic collector*. Posting the detail on #498 separately.

## Compatibility

`figs` keeps the mapping interface it had — `figs[fig]`, `fig in figs`, `.get`, `.pop` (both signatures), `.clear`, iteration, `len` — so its callers and the 25 places the suite reads it are unchanged. `pickle` and `deepcopy` of a figure carrying a record both still work (checked).

`__len__` is implemented rather than skipped for a specific reason: without it `if figs:` is always true, which is the one way a partial mapping fails quietly instead of with a `TypeError`.

## Verification

**2361 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings.

The retention test was falsified against the previous implementation **in isolation** before being trusted — it fails on the plain dict and passes on the fix, run alone rather than only as part of the file.

Its neighbour, the mapping-contract test, is **not** a detector and no longer claims to be. An earlier draft of it asserted `len(figs) == 1`; that passed when run alone against the *old* dict and failed only in a full run, because other tests leave their figures registered. It was measuring suite state, not the contract. It now asserts only about its own figure, and its docstring says plainly that a plain dict satisfies every line of it.

Three stale comments that asserted the old behaviour as fact are corrected in the same change (`maidr/widget/shiny.py`, `maidr/backend.py`, `tests/widget/test_shiny.py`) rather than left to read as current.

---

### DIFF
--- a/maidr/backend.py
+++ b/maidr/backend.py
@@ -87,9 +87,12 @@ def show(*args: Any, **kwargs: Any) -> None:
         else:
             # clear_fig=False because the backend handles figure cleanup.
             maidr_obj.show(clear_fig=False, use_cdn=use_cdn)
-            # Clean up maidr's own tracking to prevent memory leaks.
-            # Without this, FigureManager.figs grows unboundedly because
-            # clear_fig=False skips the plt.close() -> clear.py path.
+            # Drop maidr's record now rather than waiting for the figure to
+            # be collected: `clear_fig=False` skips the `plt.close()` ->
+            # `clear.py` path, and `Gcf.destroy` below is what actually
+            # releases the figure. Since #456 the record no longer outlives
+            # the figure on its own, so this is timeliness rather than the
+            # difference between bounded and unbounded.
             MaidrFigureManager.destroy(fig)
         Gcf.destroy(manager.num)
 
--- a/maidr/core/figure_manager.py
+++ b/maidr/core/figure_manager.py
@@ -18,53 +18,49 @@ from maidr.exception.unsupported_plot_error import UnsupportedPlotError
 class _FigureRecords:
     """A mapping from ``Figure`` to ``Maidr`` that stores each entry on its key.
 
-    Behaves as the plain ``dict`` this replaced -- ``figs[fig]``, ``fig in
-    figs``, ``.get``, ``.pop``, ``.clear`` -- but the value is kept in the
-    figure's own ``__dict__`` rather than in a module-level table.
+    Reads as the plain ``dict`` this replaced -- ``figs[fig]``, ``fig in
+    figs``, ``.get``, ``.pop``, ``.clear``, iteration, ``len`` -- but the
+    value lives in the figure's own ``__dict__`` rather than in a
+    module-level table. A class-level dict kept every registered figure
+    reachable for the life of the process; storing the record on the figure
+    makes the whole graph one isolated cycle once the caller lets go, which
+    the collector reclaims (#456).
 
-    That is the whole point. A module-level dict is a strong reference from a
-    class attribute, so a figure stayed reachable for the life of the process
-    even after the application dropped it and matplotlib closed it (#456).
-    Registration happens when a chart is *plotted*, not when it is rendered,
-    so this applied to every supported figure -- and the ``plt.show()`` path
-    only escaped it because the backend calls :meth:`FigureManager.destroy`
-    explicitly. A Shiny or Streamlit render never goes through ``plt.show``.
+    Not a ``WeakKeyDictionary``: every value reaches its own key --
+    ``Maidr._fig``, ``MaidrPlot.ax``, and each subclass's artist handles --
+    so the entry would keep the figure alive and the weak key would never
+    die. #498 records that chain.
 
-    Storing the record on the figure makes the whole graph -- figure, its
-    ``Maidr``, its layers, and the artists those layers hold -- one isolated
-    reference cycle once the application lets go, which the cyclic collector
-    reclaims. Measured on a figure built, rendered and closed:
+    Invariants
+    ----------
+    A record belongs to a figure only if ``record.fig is fig``.
+        Every read goes through :meth:`_record`, which enforces it, and the
+        one write refuses a record that fails it. Without this a
+        ``copy.copy`` of a figure -- which copies ``__dict__`` entries by
+        reference -- would answer as registered and hand back the
+        *original's* chart.
 
-    ======================================  =========
-    registry                                collected
-    ======================================  =========
-    module-level dict                       no
-    record stored on the figure             yes
-    ======================================  =========
+    ``_seen`` holds every figure this mapping can enumerate.
+        It backs iteration, ``len`` and ``clear`` only, weakly, so it
+        retains nothing. :meth:`_record` adds to it, because ``deepcopy``
+        and ``pickle`` install a valid record without going through
+        :meth:`__setitem__`. **A read therefore mutates shared state**,
+        which is why every operation takes ``_guard``.
 
-    This is deliberately *not* a ``WeakKeyDictionary``, which was tried first
-    and freed nothing: every value reaches its own key -- ``Maidr._fig``,
-    ``MaidrPlot.ax``, and each subclass's own artist handles such as
-    ``BarPlot._own_bars`` -- so the entry keeps the figure alive and the weak
-    key never dies. #498 records that chain. Keeping the value *on* the key
-    sidesteps it entirely: a value that reaches its key is exactly what a
-    cycle is, and cycles are collectable as long as nothing outside points in.
+    Notes
+    -----
+    Reclamation is the cyclic collector's rather than refcounting's, so a
+    host running ``gc.disable()`` gets the growth back between collections.
+    In proportion: a bare matplotlib ``Figure`` is not refcount-reclaimable
+    either, since its artists refer back to it, so turning the collector off
+    leaks figures with or without this package.
 
-    Membership is by identity either way -- ``Figure`` inherits ``__hash__``
-    and ``__eq__`` from ``object``, so dict lookup was already identity.
-
-    ``_seen`` exists only so the mapping can be enumerated and cleared. It
-    holds weak references and no values, so it retains nothing.
-
-    Reclamation is therefore the cyclic collector's, not refcounting's --
-    so a host that runs with ``gc.disable()`` or ``gc.freeze()``, which
-    long-lived servers sometimes do for latency, gets the growth back
-    between explicit collections. Worth knowing, and worth keeping in
-    proportion: a matplotlib ``Figure`` has never been reclaimable by
-    refcount either, since its own artists refer back to it. Measured with
-    the collector off, a bare figure built and closed with no maidr
-    involvement at all is not collected. Turning the collector off leaks
-    figures whether or not this package is installed.
+    The approach leans on ``Figure`` accepting instance attributes and on
+    ``copy``/``deepcopy``/``pickle`` treating ``__dict__`` as they do today.
+    Neither is a contract matplotlib publishes. If either changes the
+    failure is quiet -- the attribute is simply absent and a figure reads as
+    unregistered -- so ``tests/core/test_figure_manager.py`` pins all three
+    behaviours rather than leaving them to be assumed.
     """
 
     #: Attribute the record is stored under on the figure.
@@ -75,60 +71,37 @@ class _FigureRecords:
 
     def __init__(self) -> None:
         self._seen: weakref.WeakSet = weakref.WeakSet()
-        # Reentrant because this class calls itself: `clear` walks `__iter__`,
-        # which asks `__contains__`, and pops as it goes.
+        # Its own lock, not `FigureManager._lock`: `figs` is reachable
+        # directly and a caller cannot be relied on to hold that one.
+        # Reentrant because `clear` walks `__iter__`, which asks
+        # `__contains__`.
         #
-        # Its own lock rather than `FigureManager._lock`, because a caller
-        # cannot be relied on to hold that one. `figs` is a public class
-        # attribute that reads like a dict, and `if fig in FigureManager.figs`
-        # is the natural thing to write -- but since the record moved onto the
-        # figure a lookup also updates `_seen`, so that apparently-read-only
-        # line races. Making each operation self-contained means the contract
-        # is enforced instead of documented.
+        # `FigureManager._lock` is still required for what this cannot do --
+        # spanning the check-then-act in `_get_maidr` and the paired append
+        # in `create_maidr`.
         #
-        # `FigureManager._lock` is still needed and still does something this
-        # cannot: it spans the *compound* operations -- the check-then-act in
-        # `_get_maidr`, and the paired append in `create_maidr` -- which no
-        # per-operation lock can make atomic.
-        #
-        # What goes wrong without this: threads adding and removing entries
-        # while others iterate raise `RuntimeError: Set changed size during
-        # iteration` out of `__iter__`. Reproduced 3 of 3 against a build
-        # with this lock removed, and 0 of 3 with it.
-        #
-        # Deliberately not covered by a test. Detection needs sustained
-        # contention -- a bounded version of that harness detected on only
-        # about half its runs, and raising the budget did not improve it,
-        # so the rate is thread-scheduling luck rather than duration. A
-        # coin-flip guard costing a busy-waiting second per suite run is
-        # worth less than this comment.
+        # Without this, concurrent add/remove during iteration raises
+        # `RuntimeError: Set changed size during iteration`. Untested here:
+        # provoking it needs sustained contention and detects about half the
+        # time. `test_one_thread_cannot_enter_the_registry_while_another_is_inside`
+        # asserts the exclusion this provides instead, deterministically.
         self._guard = threading.RLock()
 
     def _record(self, fig: Figure) -> Any:
         """This figure's own record, or ``_MISSING``.
 
-        The ownership check is not a formality. ``copy.copy`` on a ``Figure``
-        copies the ``__dict__`` entries themselves, so a shallow copy would
-        inherit the original's record and answer as registered -- handing
-        back a ``Maidr`` bound to a *different* figure, and rendering the
-        original's chart under the copy's name. The module-level dict could
-        not do that, because only the object actually inserted was ever a
-        key.
+        Returns
+        -------
+        Maidr or object
+            The record when ``record.fig is fig``, else the ``_MISSING``
+            sentinel. A missing attribute, a record naming another figure,
+            and a record whose ``Maidr`` has been destroyed (``Maidr.fig``
+            raising after ``Maidr.destroy``) are all "not registered" --
+            the last because a shallow copy can outlive the pop that
+            detached the original, and a membership test must answer a bool
+            rather than raise.
 
-        ``deepcopy`` and ``pickle`` are the opposite case and are left
-        working: both rebuild the record alongside the figure, so the copy's
-        record points at the copy and it is genuinely registered. Neither
-        goes through :meth:`__setitem__`, though -- they write ``__dict__``
-        directly -- so a legitimately owned record is added to ``_seen``
-        here. Without that a copied figure answered ``in`` but was missing
-        from ``list(figs)``, uncounted by ``len``, and survived ``clear()``.
-
-        The ``AttributeError`` guard covers a record whose ``Maidr`` has been
-        destroyed. ``FigureManager.destroy`` pops the record before calling
-        ``Maidr.destroy()``, which deletes ``_fig`` -- but it pops it off the
-        figure it was given, and a shallow copy taken beforehand still holds
-        the same object. Reading ``record.fig`` there raised out of a
-        membership test, which has to answer a bool.
+        Also brings ``_seen`` up to date, per the class's second invariant.
         """
         record = getattr(fig, self._ATTR, self._MISSING)
         if record is self._MISSING:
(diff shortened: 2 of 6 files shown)


jooyoungseo:
Both taken in `5c558f4`. The first one was a real defect, not a narrow scenario.

## Shallow copy — you were right, and it was the bad kind of wrong

Measured before touching anything:

```
shallow copy   registered=True   record.fig is it: False
deepcopy       registered=True   record.fig is it: True
unpickled      registered=True   record.fig is it: True
```

The first row is a wrong-output failure rather than a lookup miss: `figs[copy]` handed back a `Maidr` bound to a *different* figure, so the copy would have rendered the original's chart, with the original's data, announced under the copy. Exactly the shape this repo has spent the session removing.

Every read now checks the record belongs to the figure it was asked about, so a shallow copy reads as unregistered — which is what it is. Asking about a copy also no longer takes the original's record away, which a naive `pop` would have done.

The test was falsified against a build with the check removed before I trusted it.

## Pickle/deepcopy — pinned, and the reason you gave is the right one

You are correct that "checked once" is not a guard. The value is not really that it works today; it is that **a figure's pickle now carries the whole `Maidr`/`MaidrPlot` graph, which it did not before**, so something unpicklable added to either class would break pickling a plain figure with nothing to catch it. Parametrized over `deepcopy` and `pickle`, asserting both that the round trip succeeds and that the copy's record points at the copy — the latter being what separates these from the shallow case.

Worth noting the behaviour that pins: a deepcopied or unpickled figure now comes out **registered**, carrying its extracted data. That is new, and defensible — the record travelled with the figure because it is part of it — but it is a change, so it is now written down rather than incidental.

## `__len__`/`__iter__` cost

Accepted as flagged. Nothing calls either per render; the source paths use `__contains__`, `__getitem__` and `pop`, all O(1) attribute access. `_seen` exists only for enumeration. If iteration ever moves onto a hot path the `hasattr` re-check is the thing to remove.

## Docstring style

Left as prose. `_FigureRecords` is a private implementation class with no public parameters to document, and the module's NumPy-style sections are on the API surface. Happy to convert if you would rather it be uniform.

**2364 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings; `ruff format` clean on both changed files.

---

jooyoungseo:
Two more commits, neither prompted by review — both found by reading back what I had just written.

## `dc8ce16` — `del figs[fig]` disagreed with `figs.pop(fig)`

`__delitem__` deleted the attribute directly while every other read went through the new ownership check, so `del figs[copy]` succeeded on a shallow copy that `figs.pop(copy)` refused. Nothing calls `__delitem__` today, which is precisely why it would have stayed wrong until the first caller. Falsified against the previous behaviour.

## `efee9ea` — pinning the case this fix had to *not* break

#452's hazard is what ruled out two of #456's three options, and nothing in the PR asserted it still holds. Now it does, across repeated render-and-close cycles on a cached figure:

```
flush 1: registered=True  layers=1  html carries a chart=True
...
flush 5: registered=True  layers=1  html carries a chart=True
after the cache drops it, collected: True
```

Both halves, because a fix that released nothing would pass the first alone.

### And the falsification turned up something worth recording

I tried to falsify it against #456's option 3 — drop the record from a `close_event` handler — and **the test passed**. The reason is not that the test is weak:

```
canvas at connect:   FigureCanvasAgg
close_event fired:   0
```

Under `Agg`, `plt.close()` fires no `close_event` at all. So option 3 would have been **inert in a headless server** — which is exactly where the leak this issue is about happens. That is a stronger argument against it than the cached-figure hazard, and it was not on the issue.

The docstring says all of this, including that the reclaim half is the only part of that test with a falsified detector behind it (against the plain dict). I would rather it under-claim than read as a guard it is not.

**2366 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing; `ruff format` clean.

One earlier probe of mine also needs correcting for the record: an intermediate measurement of the cached case reported "after the cache drops it, collected: False". That was my own script leaving `fig` bound at module scope after the loop — the same self-pollution that made two earlier referrer dumps unreadable. With the binding released it collects. The committed test does the `del` explicitly.

---

jooyoungseo:
Three findings across the last two reviews, all real, all taken in `8c894ed`. Reproduced each before fixing it.

## The crash path — and I had explicitly reasoned it away

> `_record()` can raise an unhandled `AttributeError` for a stale shallow copy of a destroyed figure

Correct:

```
twin = copy.copy(fig); maidr.close(ax)
twin in figs  ->  RAISED AttributeError 'Maidr' object has no attribute '_fig'
```

Worth being explicit about how I missed it, because the reasoning error is the interesting part. Earlier in this PR I checked that `Maidr.destroy()` has exactly one call site and that it pops the record first, and concluded no destroyed record is ever reachable. That is true **of the figure it was popped off**. A shallow copy taken beforehand holds the same `Maidr` object, and nothing popped anything off the copy. I verified the invariant on the wrong object.

`_record` now treats a record with no live figure as unowned, so `in` answers `False` where it used to raise.

## Ghost entries — the mapping disagreed with itself

> `_seen` isn't populated for `deepcopy`/`pickle`-reconstructed figures

Also correct, and it directly falsifies a claim in this PR's own description:

```
clone in figs      : True
clone in list(figs): False
after clear(), clone still registered: True
```

`deepcopy` and `pickle` write `__dict__` directly rather than going through `__setitem__`. Your suggested fix is what landed — `_record` is the one place ownership is established, so it is where `_seen` is brought up to date. My existing round-trip test asserted `in`/`[]`/`.fig` and stopped exactly short of this; the new one asserts enumeration, `len` and `clear` too.

## `pop`/`del` atomicity

> two threads racing a `.pop()` ... could both pass the ownership check and then have the second `delattr` raise an unhandled `AttributeError` instead of the `KeyError` the class's docstring promises

Taken rather than documented, since it was three lines: both now go through one `_delete` that tolerates a lost race. The entry the caller wanted gone is gone either way, which is the answer `pop` should give.

## Also raised, and my answers

- **WeakSet iteration outside `_lock`** — left as-is and now noted. Production paths only use `__contains__`/`__getitem__`/`__setitem__`/`pop`, all under `_lock`; `__iter__`/`__len__`/`clear` are test-facing. `__iter__` already snapshots with `list()` for this reason. Same hole the plain dict had.
- **Pickle coupling deserves a downstream callout** — agreed, and it is in the PR description. Anyone pickling a maidr-instrumented figure for `multiprocessing` or joblib is now serializing the `Maidr` graph too. It works today across every subclass, and the parametrized test exists so it keeps working.
- **`Maidr.destroy()`'s ordering dependency isn't structurally enforced** — true, and the crash above is what happens when the assumption is leaned on. The `_record` guard now makes the registry safe regardless of ordering, which is a better answer than a comment asking future callers to be careful.
- **`backend.py`'s comment undersells the call** — I do not think so, and it is worth saying why rather than silently not changing it. Dropping the record for a figure the caller keeps using after `show()` is #452's hazard, not correctness. That call is timeliness; the record no longer outlives the figure without it.
- **Docstrings are a stylistic outlier** — acknowledged twice now. Happy to trim on request; leaving the *why* in place given how much of this PR has been walking back my own earlier reasoning.

**2369 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing; `ruff format` clean.

---

jooyoungseo:
One more, `5eeb095`, found by auditing the write path myself after review found three disagreements on the read paths. It is the same shape, and I would rather surface it than wait for a fourth review to.

Every read enforces ownership. The one write did not, so it could create state no read would return:

```
figs[a] = b's record
  a in figs : False
  figs[a]   : KeyError -- the write did not stick
  attribute : still set
```

A write that appears to succeed, leaves an attribute behind, and reads back as unregistered. It is now refused with a `ValueError` naming the problem.

That is a **behaviour change** and worth stating plainly: the dict this replaced would have stored a mismatched record and returned it. Refusing puts the error where the mistake is instead of at some later lookup. Nothing writes such a record today — `_get_maidr` constructs the `Maidr` from the same figure it keys on — which is precisely why it would have gone unnoticed.

Falsified against a build without the check.

**2370 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing.

That is the whole surface now: `__setitem__`, `__getitem__`, `__contains__`, `get`, `pop`, `__delitem__`, `__iter__`, `__len__`, `clear` all agree on one definition of what is registered, established in one place.

---

jooyoungseo:
Both points taken in `6151296`. The first one is real and I am recording it rather than closing it.

## The lazy `_seen` gap in `clear()`

Reproduces exactly as described:

```
clone read once before the clear -> dropped
clone never read                 -> survives
```

I had drafted a `clear()` test before this review landed — `clear` is what the suite uses for isolation and nothing asserted it — and **it would not have caught this**, because it read the clone before clearing. That is the same near-miss as my earlier round-trip test, which asserted `in`/`[]`/`.fig` and stopped just short of enumeration. Worth naming the pattern: I keep writing the setup line that happens to satisfy the invariant I am about to check.

The test now uses two clones differing only in whether they are looked up, and **asserts the gap** rather than avoiding it — including which assertion to delete if it is ever closed.

I am not closing it, and the reasoning is in `clear()`'s docstring so it does not get re-litigated: `_seen` learns of a figure when its record is written or read, and `deepcopy`/`pickle` do neither. Closing it needs a hook on the copy, which means putting registry knowledge into `Maidr.__deepcopy__`/`__setstate__`. That is a real cost for an enumeration gap that cannot outlive the clone holding it — the record is reachable only through the clone, so nothing leaks; `clear` simply does not see it. Your own read agrees ("not a real leak"), and `clear` exists for test isolation rather than as a lifetime control.

## Reads mutate shared state, and the docstring said otherwise

Correct and worth more than a nit. `FigureManager.figs`' docstring said only the write paths take `_lock`, which was true before and is not now: a lookup updates the bookkeeping behind iteration, so `in` and `get` mutate the `WeakSet`. Every production caller already holds the lock — I re-checked `_get_maidr`, `get_maidr` and `destroy` — but the docstring was actively telling the next direct caller the wrong thing. Fixed there, since that is where someone reaching for `figs` would look.

## Verification

**2371 passed, 13 skipped**, stable across four consecutive full runs (the `clear()` test wipes the registry mid-suite, so order-independence was worth checking rather than assuming). `ruff check` unchanged at 6 pre-existing; `ruff format` clean.

---

jooyoungseo:
All five points addressed in `1e9dea7`. Point 1 was right, and chasing it produced the most useful measurement in this PR.

## 1. The registry now guards itself

Your framing is the correct one — the safety net was a docstring, and `if fig in FigureManager.figs` is exactly the line someone would write. `_FigureRecords` takes its own reentrant lock now (reentrant because `clear` walks `__iter__`, which asks `__contains__`). `FigureManager._lock` still does what a per-operation lock cannot: span the check-then-act in `_get_maidr` and the paired append in `create_maidr`.

Justified by measurement rather than caution — without it:

```
RuntimeError: Set changed size during iteration
guard removed : 3/3 runs
guard active  : 0/3 runs
```

**No test ships with it, deliberately.** I wrote two and threw both away:

- The first only *read* from many threads. It passed with the lock removed — a set that never changes size cannot raise. That would have been the seventh test this session to assert nothing.
- The second churned, but finished in 0.06s and detected nothing 5 times over.

A bounded version that does detect manages roughly half its runs, and raising the budget made it **worse** (5/8 at 0.75s, 3/8 at 1.0s) — so the rate is thread-scheduling luck, not duration. A coin-flip guard costing a busy-waiting second per suite run is worth less than a comment stating what was measured, so the comment is what landed.

## 3. Not a new trade-off — checked before writing it down

You suggested documenting that reclamation now waits for a cyclic GC pass. I was about to, then measured the baseline:

```
bare matplotlib figure, no maidr, gc disabled -> collected: False
```

A matplotlib figure has never been freed by refcounting — its own artists refer back to it. So "reclaimed on a gc pass rather than instantly" describes every matplotlib figure, before and after. A doc note would have implied this PR introduced it. Not adding one.

## 2, 4, 5

- **Picklability constraint** — recorded on `Maidr` itself, since that is where someone would add the offending attribute: everything held there and on every `MaidrPlot` now participates in the figure's pickle, so a lock or an open handle added later breaks pickling a plain-looking figure somewhere the user has no reason to look.
- **Double lookup** — collapsed. `_get_maidr` and `get_maidr` asked `in` then `[]`, each doing its own ownership check; both now do one `get`.
- **`ValueError` from `__setitem__` is an API change** — agreed, and it is in the PR description and its own test. I added it in `5eeb095` for the same reason you flagged it: the alternative was a write that silently did not stick.

**2371 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing.

---

jooyoungseo:
`b84f9ad`. Point 3 was the one that mattered, and you were right in a way I should have caught myself.

## 3. A deterministic test was possible, and I had the technique already

> A deterministic variant is possible without relying on scheduler luck — e.g. patch `_record`/`_delete` to block on a `threading.Event` so two threads are forced to interleave inside the critical section.

That is exactly what #506's pairing test does, two PRs ago, for the same class of problem. I reached for the probabilistic route, measured it as flaky, and concluded no test was worth shipping — without trying the approach I had just used. "Not worth testing" was the wrong conclusion from "my first attempt was flaky."

Asserted as **mutual exclusion**, which is what the lock actually promises, rather than as the `RuntimeError` only sustained contention produces. One thread parks inside `__setitem__`'s guarded section; the other tries to enter. Guarded it cannot, and the holder pays its full deadline; unguarded it completes immediately and releases the holder early.

```
guard removed : 6/6 detected
guard restored: 6/6 passing
```

**The first version of it was vacuous**, for the same reason these keep being vacuous: I put the pacing hook on `_seen.add`, which is on the *enterer's* path too (`pop` establishes ownership before deleting), so both threads parked and neither could observe the other. It passed with the lock removed. Tagged per thread now — the same fix #506 needed, for the same mistake.

That is eight tests this session that passed while asserting nothing. Every one caught by mutation; none by reading.

## 1. `gc.disable()` — real, and worth keeping in proportion

Documented. Also measured, because the framing matters: with the collector off, a **bare matplotlib figure** built and closed with no maidr involvement is not collected either — its own artists refer back to it. Turning the collector off leaks figures whether or not this package is installed. The docstring says both halves, so nobody reads it as a cost this fix introduced.

## 2, 4, 5

- **Comment density** — third review to raise it, so I will stop defending it: fair, and the historical narrative does belong in commits. I have left it for now because this PR has walked back four of my own conclusions and the reasoning is load-bearing while that is still true. Happy to trim it in a follow-up once it has settled, if you would rather not carry it.
- **Double space** — fixed.
- **Enumeration blind spot** — agreed on all counts, including the warning about absolute `len(figs)` assertions. That is recorded in `test_the_registry_reads_as_the_mapping_it_replaced`, which explains why an earlier draft's `len(figs) == 1` was removed.

**2372 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing; `ruff format` clean on all three changed files.

---

jooyoungseo:
`9f35c26`. Taking the two you asked for my take on directly, plus the measurement.

## 5. The measurement you asked for

`_get_maidr` on an already-registered figure — the hot path, once per registered layer:

| registry | per call |
|---|---|
| module-level dict (pre-#508) | **0.37 µs** |
| record on the figure | **1.00 µs** |
| record on the figure, internal lock removed | 0.97 µs |

So 2.7× relatively, **0.63 µs absolutely**, and the `RLock` is ~0.03 µs of it — the cost is the attribute lookup and ownership check, not the lock. A three-layer figure pays under 2 µs against a render dominated by `savefig` in the tens of milliseconds. Worth having the number rather than asserting "negligible", which is the point you were making.

## 3. Comment style — fourth flag, so acting rather than defending

You are right, and I said last round I would trim it in a follow-up. Doing it now instead: **133 lines out, 86 in.** The class documents an `Invariants` section naming the two rules every method upholds, and `Notes` for couplings a reader must know. The "found in review" narrative and repro counts are gone from the source; they live in the commits and this thread, timestamped, where nobody has to maintain them.

## 2. The read side effect is now on the public surface

Moved out of a paragraph inside `__init__` and into the class's own invariants, as you suggested — `fig in figs` looks side-effect-free and is not, and that should not require finding the right paragraph.

## 4. The matplotlib coupling

Good catch, and it was the one assumption this PR never stated. Now noted: the approach leans on `Figure` accepting instance attributes and on `copy`/`deepcopy`/`pickle` treating `__dict__` as they do today, neither of which matplotlib publishes. The failure would be quiet — the attribute is simply absent and a figure reads as unregistered — so the note points at the tests that pin all three behaviours.

## 1. Scope — my take, and it is not uniform across the guards

The framing I would push back on is "guards against usage patterns nothing currently does." Most of them **repair regressions this PR would otherwise introduce**, rather than defend hypotheticals:

| guard | why |
|---|---|
| shallow-copy ownership | `copy.copy(fig)` returned the *original's* chart under the copy. The dict could not do this; storing on the figure created it. |
| stale-record read | `fig in figs` **raised** after `maidr.close()`. Created by this PR. |
| `_seen` reconciliation | a copied figure answered `in` but was uncounted and survived `clear()`. Created by this PR. |
| `del`/`pop` parity | the two disagreed *because of* the ownership check this PR added. |
| `RLock` | `figs` is reachable directly, and this PR is what made a read mutate state. |

Every one of those is repair, not speculation, and each was reproduced before being written.

**The `ValueError` on a mismatched write is the exception, and I will name it as the discretionary one.** Nothing writes such a record; `_get_maidr` constructs the `Maidr` from the figure it keys on. I added it because the alternative is a write that silently does not stick — but it is five lines defending against a caller that does not exist, and it is the one I would drop first if you want the surface smaller. Happy to remove it on request; I would not remove any of the others, because removing them ships a regression.

Every guarantee was re-falsified after the rewrite — rewriting the class is exactly when a guard gets dropped by accident. All six mutations fail their test and pass restored.

**2372 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing; `ruff format` clean on all files this PR touches.

---

PR #510, opened 2026-08-18
test(shiny): pin what the per-figure lock actually prevents

jooyoungseo:
> **Correction, before anything else.** This PR's original description claimed "removing the lock left the suite green." **That is false.** `test_two_renders_of_one_figure_do_not_overlap` already exists in this same file and already fails without the lock. I asserted the gap without running the suite against a lock-free build — I only checked that my *own* new test failed. The verification note below is what I should have run first.

Follows up on [#454](https://github.com/xability/py-maidr/issues/454#issuecomment-5321960614), where I offered a regression test for finding 3 and did not write one.

## What was actually already covered

Running the whole file against a build with `_figure_lock` removed from `_render_off_loop`:

```
FAILED test_two_renders_of_one_figure_do_not_overlap
FAILED test_concurrent_renders_of_one_figure_agree   <- this PR
2 failed, 35 passed
```

So the lock **is** guarded today. The existing test asserts the *mechanism*: with `maidr.render` monkeypatched to a sleeping stub, no two renders of one figure are ever inside at once.

## What this adds, and it is narrower than I first said

The existing test replaces the render entirely, so it cannot observe what the lock is protecting — it proves exclusion, not that the output is right. This one runs the **real** render path (real `savefig`, real SVG) and asserts the user-visible consequence: six concurrent renders of one figure produce byte-identical markup once per-render uuids and matplotlib's `dc:date` are stripped.

That matters because the failure mode is a *valid* document at the wrong scale:

```
-L 640 345.6      ->  +L 460.8 345.6
-L 640 -134.4     ->  +L 460.8 0
```

`640 / (100/72) = 460.8` — `savefig` writes `fig.dpi` for its duration and restores it. Geometry is what the highlight overlay and the tactile export are positioned against, so this is wrong in the modality a sighted reviewer checks least. A future change that keeps renders from *overlapping* but reintroduces shared mutable state some other way would pass the existing test and fail this one.

Whether a second test earns its place on that argument alone is a fair question, and a reasonable answer is no. I would rather put the corrected reasoning up and let you judge than leave the original claim standing.

## Detection

| | result |
|---|---|
| lock stubbed out, standalone harness | 10/10 mismatched |
| lock in place, standalone harness | 10/10 identical |
| lock removed, through pytest | 6/6 failed |
| lock restored, through pytest | 6/6 passed |

## Verification

**2382 passed, 13 skipped.** `ruff check` unchanged at the 6 pre-existing findings; `ruff format` clean on the changed file apart from one pre-existing hunk elsewhere in it that this PR does not touch.

### DIFF
--- a/tests/widget/test_shiny.py
+++ b/tests/widget/test_shiny.py
@@ -852,7 +852,10 @@ def test_concurrent_renders_of_one_figure_agree():
             start.wait(timeout=10)
             rendered = str(renderer._render_off_loop(ax))
             outputs.append(_VOLATILE_IN_SVG.sub("", rendered))
-        except BaseException as exc:  # reported below, not swallowed
+        except Exception as exc:  # reported below, not swallowed
+            # Narrow enough to cover everything expected --
+            # `BrokenBarrierError` is a `RuntimeError` -- while leaving
+            # `KeyboardInterrupt` and `SystemExit` free to end the thread.
             failures.append(exc)
 
     workers = [threading.Thread(target=render_once) for _ in range(racers)]
--- a/tests/widget/test_shiny.py
+++ b/tests/widget/test_shiny.py
@@ -632,7 +632,10 @@ def test_two_renders_of_one_figure_do_not_overlap(monkeypatch):
         for worker in workers:
             worker.start()
         for worker in workers:
-            worker.join()
+            # Bounded: an unbounded join turns a broken lock into a hung
+            # suite rather than a failed test (#506).
+            worker.join(timeout=60)
+            assert not worker.is_alive(), "a render deadlocked on the lock"
     finally:
         plt.close(fig)
 
@@ -767,7 +770,8 @@ def test_two_different_figures_rendering_at_once_keep_their_selectors():
         for worker in workers:
             worker.start()
         for worker in workers:
-            worker.join()
+            worker.join(timeout=60)
+            assert not worker.is_alive(), "a render deadlocked"
 
         assert [together[i] for i in range(len(figures))] == alone, (
             "a concurrent render of another figure stripped this one's "
@@ -775,3 +779,75 @@ def test_two_different_figures_rendering_at_once_keep_their_selectors():
         )
     finally:
         plt.close("all")
+
+
+#: Attributes that differ between two renders of the same chart by design --
+#: fresh uuids per layer, and the timestamp matplotlib stamps into the SVG.
+_VOLATILE_IN_SVG = re.compile(
+    r'(id="[^"]*"|url\(#[^)]*\)|<dc:date>[^<]*</dc:date>'
+    r'|xlink:href="#[^"]*"|maidr="[^"]*")'
+)
+
+
+def test_concurrent_renders_of_one_figure_agree(monkeypatch):
+    """Concurrent renders of one figure must produce the same chart.
+
+    Narrower than it first looks, and worth stating plainly:
+    ``test_two_renders_of_one_figure_do_not_overlap`` above already fails
+    without the lock, so this is not filling an unguarded gap. That one
+    monkeypatches ``maidr.render`` to a sleeping stub, so it proves the
+    lock *excludes* but cannot see what the exclusion is protecting. This
+    runs the real render and asserts the consequence.
+
+    ``savefig`` writes ``fig.dpi`` for its duration and restores it
+    afterwards, so two renders of the **same** figure at once race on one
+    mutable attribute and the loser draws the whole chart at the other
+    call's dpi (#454). Distinct figures do not race; that is why the lock
+    is per figure rather than process-wide.
+
+    The failure is the kind this project treats as worst: not garbled
+    markup and not an exception, but a complete, well-formed SVG of the
+    same chart at the wrong size. Geometry is what the highlight overlay
+    and the tactile export are positioned against, so a chart rendered at
+    72% scale is wrong in the modality a sighted reviewer checks least.
+
+    Asserted as agreement between renders rather than against a fixed size,
+    since the wrong-dpi output is internally consistent -- it is only wrong
+    relative to what every other render produced.
+
+    Measured 10 of 10 runs mismatching with the lock stubbed out and 10 of
+    10 identical with it, so this runs in CI rather than under
+    ``--run-benchmark``.
+    """
+    fig, ax = plt.subplots()
+    ax.bar([str(i) for i in range(30)], list(range(30)))
+    renderer = render_maidr(lambda: ax)
+
+    outputs: list[str] = []
+    failures: list[BaseException] = []
+
+    def render_once():
+        try:
+            rendered = str(renderer._render_off_loop(ax))
+            outputs.append(_VOLATILE_IN_SVG.sub("", rendered))
+        except BaseException as exc:  # noqa: BLE001 - reported, not swallowed
+            failures.append(exc)
+
+    workers = [threading.Thread(target=render_once) for _ in range(6)]
+    try:
+        for worker in workers:
+            worker.start()
+        for worker in workers:
+            worker.join(timeout=60)
+            assert not worker.is_alive(), "a render deadlocked on the lock"
+
+        assert not failures, f"a render raised: {failures[:2]}"
+        assert len(outputs) == len(workers)
+        assert len(set(outputs)) == 1, (
+            "concurrent renders of one figure disagreed; one of them drew "
+            "the chart at another render's dpi, which produces a valid SVG "
+            "at the wrong scale rather than an error"
+        )
+    finally:
+        FigureManager.figs.pop(fig, None)
+        plt.close(fig)
--- a/tests/widget/test_shiny.py
+++ b/tests/widget/test_shiny.py
@@ -784,7 +784,7 @@ def test_two_different_figures_rendering_at_once_keep_their_selectors():
 #: Attributes that differ between two renders of the same chart by design --
 #: fresh uuids per layer, and the timestamp matplotlib stamps into the SVG.
 _VOLATILE_IN_SVG = re.compile(
-    r'(id="[^"]*"|url\(#[^)]*\)|<dc:date>[^<]*</dc:date>'
+    r'(\bid="[^"]*"|url\(#[^)]*\)|<dc:date>[^<]*</dc:date>'
     r'|xlink:href="#[^"]*"|maidr="[^"]*")'
 )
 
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Correcting this PR's premise before anyone reviews it against a claim that is not true.

## I said removing the lock left the suite green. It does not.

`test_two_renders_of_one_figure_do_not_overlap` already exists **in this same file** and already fails without the lock:

```
FAILED test_two_renders_of_one_figure_do_not_overlap
FAILED test_concurrent_renders_of_one_figure_agree   <- this PR
2 failed, 35 passed
```

I asserted the gap after checking only that my *own* new test failed, never running the suite against a lock-free build. That is the same mistake in the opposite direction from the ones this session keeps producing: instead of a test that passes for the wrong reason, a claim that a test was missing without looking for it.

Description, commit message and the test's own docstring are all corrected. The commit was amended rather than followed by a fixup, so the false statement is not left in the history this will squash from.

## What is actually left of the argument

The existing test monkeypatches `maidr.render` to a sleeping stub. It proves the lock **excludes**; it cannot see what the exclusion protects. This one runs the real render and asserts the **consequence** — six concurrent renders of one figure produce identical markup once per-render uuids and `dc:date` are stripped.

The distinction is real but modest: a future change that stops renders overlapping while reintroducing shared mutable state some other way would pass the existing test and fail this one. Whether that earns a second test is a fair question and **a reasonable answer is no** — happy to close this if you would rather not carry it.

## One thing here is worth keeping either way

The file had two unbounded `join()` calls. An unbounded join turns a broken lock into a **hung suite** rather than a failed test — the exact failure #506 spent a debugging cycle on, and which I documented there and then left standing two functions away. Both are bounded now with an explicit liveness assertion.

**2382 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing.

---

jooyoungseo:
All three points taken in `08785ae`. The flakiness one was the substantive one and both reviews were right to lead with it.

## The timing-race concern

> a quiet/serialized scheduler could in principle let all 6 threads run without ever overlapping mid-`savefig`, silently reducing this to a no-op on that run

Exactly the failure mode this session keeps producing, arriving one step earlier than usual — a test that would pass for the wrong reason *on some machines*, which is worse than one that does it everywhere, because it looks fine here.

Both reviews suggested widening the window by growing the chart. **Measured, and it is not the variable:**

| | detection with the lock removed |
|---|---|
| 30 bars, 6 threads | 8/8 (0.58s per trial) |
| 100 bars, 6 threads | 8/8 (1.02s) |
| 200 bars, 8 threads | 8/8 (2.28s) |

Flat. Growing the chart buys runtime, not margin, so it would have been reassurance rather than a fix.

What does fix it is not depending on the scheduler at all: the six renders now start from a `threading.Barrier`, so they begin together by construction rather than by luck. A quiet runner cannot stagger them.

```
lock removed  : 8/8 detected
lock restored : 8/8 passed
```

The barrier count and thread count are one constant — a barrier expecting more arrivals than there are threads waits forever, which is the hang #506 spent a debugging cycle on.

## The two nits

Both dropped: the unused `monkeypatch` (left over from adapting the sibling test) and the `FigureManager.figs.pop` in teardown, which the autouse fixture already does and which #508 made unnecessary regardless.

## Not acted on, and why

> the leaked worker thread itself is non-daemon and keeps running in the background — Python will still block at interpreter shutdown

Correct, and left alone deliberately: it is the pre-existing pattern in this file, and changing it here would be a broader change to every concurrency test rather than something this one needs. Worth its own pass if the file ever gets one.

**2382 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing; `ruff format` clean apart from one pre-existing hunk elsewhere in the file.

The open question from my earlier correction still stands: this test is a complement to `test_two_renders_of_one_figure_do_not_overlap`, not a gap-filler, and closing it is a reasonable call if you would rather not carry a second concurrency test.

---

jooyoungseo:
All four taken in `ffc3ffb`, each verified rather than accepted on trust.

## 1. The scope note — the one that mattered

You are right that stripping `maidr="..."` discards more than per-element selector ids. Checked:

```
maidr= attributes found: 3
   {&#10;  "id": "94101a32-...",&#10;  "subplots"     <- the whole schema
   7412be45-6411-4510-8d62-e5a69a11372b
   7412be45-6411-4510-8d62-e5a69a11372b
```

So a race that scrambled the announced values without moving a coordinate would pass this test. That is acceptable for the bug it guards — `fig.dpi` lives entirely in `<path d="...">`, which the regex leaves alone — but **"produce the same chart" reads wider than what is asserted**, and on a sonification-first tool that is exactly the wrong direction to be imprecise in. The docstring now says geometry, not data.

## 2. `# noqa: BLE001` was inert

Confirmed: `extend-select = ["FA102"]` on top of ruff's defaults, so `BLE` is not enabled and the suppression implied a rule that is not in play. Replaced with a plain comment.

One more of these exists in `tests/core/test_cdn_version.py` from earlier work of mine. Left alone rather than widening this PR — flagging it so it is a decision rather than an oversight.

## 3. `\bid="` — done.

## 4. The timeout, with one adjustment

Dropped 60s → **30s**, not to 10s as suggested, because 10s would have broken an interaction: the barrier's own wait must expire *before* the join, or a thread still legitimately blocked at the barrier gets reported as a deadlocked render. So the barrier moved to 10s and the join stays above it at 30s, against a healthy run of ~1.7s.

## Re-falsified

The assertion depends on the regex I changed, so it was worth redoing rather than assuming: **8 of 8 detected** with the lock removed, **8 of 8 passing** with it.

**2382 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing.

---

jooyoungseo:
`eba8a71` — the one actionable point, taken.

## `except BaseException` → `except Exception`

Verified rather than assumed:

```
BrokenBarrierError MRO: ['BrokenBarrierError', 'RuntimeError', 'Exception', 'BaseException']
```

So `Exception` covers everything the worker can expect, while leaving `KeyboardInterrupt` and `SystemExit` free to end the thread instead of being collected into a list nothing re-raises. That was the actual hazard — a Ctrl-C during a six-thread render would have been swallowed into `failures` and reported as "a render raised".

Re-falsified after the change: **6/6 detected** with the lock removed, **6/6 passing** with it.

## Recorded, not acted on

- **Residual flake risk on a very slow runner** — real, and bounded: 10s at the barrier and 30s at the join against a ~1.7s healthy run. Tightening either trades a rare false failure for a likelier one, so the margin stays.
- **`_VOLATILE_IN_SVG` placement** — deliberate, and you called it reasonable; it is paired with exactly one test.

**2382 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing.

## Where this stands

Four review rounds, no blocking findings in any of them, and the substantive changes across them came from the reviews rather than from me: the barrier that removed the scheduler dependency, the scope note about what the comparison does not cover, the inert `noqa`, and this catch.

The question I raised in my own correction is still open and still genuinely open: this is a **complement** to `test_two_renders_of_one_figure_do_not_overlap`, not a gap-filler, and declining a second concurrency test is a defensible call. The two `join()` bounds are worth keeping either way, and I would happily split them out if you would rather take those alone.

---

jooyoungseo:
`eba4f3b` — both taken.

## The limitation belongs in the docstring, not the commit message

> the barrier only synchronizes the *start*, not the full duration … a false negative, not a flaky failure, so it wouldn't show up as CI flakiness — just reduced coverage on bad days

Right, and precisely the distinction that matters: a flaky failure gets noticed, silently reduced coverage does not. It was in the commit message and this thread but not where someone reading the test would find it.

It now says so, and names which sibling to trust on such a day — `test_two_renders_of_one_figure_do_not_overlap` detects overlap regardless of speed, so the two degrade differently under load rather than together.

This is the same principle as the geometry/data scope note one paragraph below it, and I should have applied it here without being asked: **state what the test does not cover, in the test.**

## The shadowing `import re`

Dropped. Pre-existing, and this PR's own regex already relies on the module-level import, so leaving it would have been odd.

**2382 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing, and clean on this file specifically.

## Round summary

Five review rounds, none blocking, and every substantive improvement came from them: the barrier that removed the scheduler dependency, the geometry/data scope note, the inert `noqa`, `except Exception` over `BaseException`, and now this. What I contributed unprompted was the correction to my own false premise and the two `join()` bounds.

The standing question is unchanged and still genuinely open: this **complements** `test_two_renders_of_one_figure_do_not_overlap` rather than filling a gap, and declining a second concurrency test is defensible. If you would rather take only the hardening, the two `join()` bounds and the `import re` cleanup split out cleanly.

---

jooyoungseo:
`636fde6`. The first point is a reconsideration — I declined the same observation a round ago and was wrong to.

## `daemon=True`

An earlier review noted the non-daemon threads and I waved it off as "the pre-existing pattern in this file." Your framing is better and I should have applied it myself: **this PR is already in the business of turning a hang into a failure**, and `daemon=True` is what finishes the job the bounded joins started.

A timed-out `join` fails the test but leaves the worker alive, still holding the figure's lock. Non-daemon, a real deadlock — the exact bug these tests exist to catch — wedges the interpreter at shutdown rather than failing cleanly. Daemon, the worst case degrades to a failed test. Applied to all three concurrency tests in the file, since the bounded joins already were.

"It was like that before" was a bad reason when the change is three characters and the failure it prevents is the one the test is for.

## The unsynchronised appends

Documented rather than changed: `list.append` is atomic under the GIL and both lists are read only after every worker is joined. You are right that it was implicit in a file that spells out every other concurrency decision.

## Recorded, not acted on

- **Detection numbers being representative of CI** — good check, and worth having in the thread: `ci.yml` runs plain `pytest -vvv` with no `xdist`, so each matrix job has the runner to itself, which is the condition the 8/8 and 10/10 measurements were taken under.
- **Docstring length** — fifth mention across the two PRs. Consistent with the sibling tests in this file, and I have already trimmed the equivalent prose in `figure_manager.py` where it was an outlier rather than the convention.

Re-falsified after the change: **6/6 detected** with the lock removed, **6/6 passing** with it.

**2382 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing; `ruff format` clean on this file except one pre-existing hunk this PR does not touch.

---

jooyoungseo:
`51eab35` — one change, two answers.

## 1. `list[BaseException]` → `list[Exception]`

Taken. `except Exception` guarantees only `Exception` subtypes reach the list, so the annotation was wider than anything that can land there. I noticed this myself when narrowing the catch two commits ago and left it as "correct, just loose" — the wrong call for a one-word change that a reviewer then had to spend attention on.

## 2. The barrier can cascade — accepted, and here is the trade

> if any single one of the 6 threads is slow to reach `start.wait()` … `BrokenBarrierError` propagates to *every* other waiting thread too, failing the test for a scheduling reason rather than a lock regression

Correct. Accepted deliberately, because the alternative is worse in a specific way: a false **failure** is loud and gets investigated, while the false **negative** the docstring already records — a runner slow enough that renders never overlap — is silent and just quietly stops testing anything. Against a ~1.7s healthy run, a 10s barrier budget makes the loud failure improbable; and if it ever fires, it fires visibly.

## 3. The line length — measured, since you could not run the formatter

```
workers = [threading.Thread(target=render_once, daemon=True) for _ in range(racers)]
```

**88 characters**, exactly the configured limit. That is why `ruff format` collapses it to one line rather than splitting it — an earlier commit of mine had it wrapped and the formatter joined it back. `awk 'length > 88'` over the whole file returns nothing.

**2382 passed, 13 skipped.** `ruff check` unchanged at 6 pre-existing.

---

PR #511, opened 2026-08-18
docs: stop telling Shiny users their event loop is blocked

jooyoungseo:
## The defect

`docs/index.qmd`'s "Shiny and other async servers" callout still describes the state of things **before #504**:

> **The render itself still blocks, though.** … it runs on the event loop and every other session on that worker waits. It never awaits, so nothing preempts it: for as long as one render runs, the loop does not run at all. Measured … one render holds the loop for about 55 ms

#504 moved the render to a worker thread. That paragraph has been wrong since it landed — a reader today concludes their Shiny app stalls once per reactive flush, when it does not.

Found by checking the doc against the code, which is the only way this kind of staleness surfaces; nobody files a bug against documentation that is merely out of date.

## Re-measured through the shipped renderer

50-bar chart, longest gap between successive ticks of a 1 ms `asyncio` ticker, after a warm-up (the first render pays imports and the font cache — worth roughly thirty steady-state ones, and the reason an earlier measurement on #454 had to be retracted):

| | longest loop gap |
|---|---|
| idle, no rendering | 1.4 ms |
| render on the loop, as it used to be | 609.3 ms |
| render on a worker thread, as it is now | **12.5 ms** |

Same throughput either way — 9 vs 10 renders in the window. The work did not get cheaper; it stopped being in everyone else's way.

The 609.3 ms figure independently reproduces the 602.7 ms measured on [#454](https://github.com/xability/py-maidr/issues/454#issuecomment-5321960614) with a prototype, now through the shipped code path.

## What the callout now says it does *not* mean

Two limits, so the correction does not over-claim in the other direction:

- **Same-figure renders are still serialised.** `savefig` writes `fig.dpi` for its duration, so two concurrent renders of one figure leave one chart drawn at the other's scale. Distinct figures — the usual per-session shape — run in parallel. Pinned by `test_concurrent_renders_of_one_figure_agree` (#510).
- **The chart still crosses the websocket every flush**, ~25 KB or a couple of megabytes with `use_cdn=False`. That is #454's option 2 and is still open.

## The benchmark test

`tests/core/test_render_is_synchronous.py` keeps its assertion — the underlying `maidr.render()` *is* still synchronous, and that is exactly what makes moving it off the loop necessary rather than optional. What changed is its framing: it claimed "every other session on that worker is stopped", which was a statement about `render_maidr` rather than about the function it measures. Its failure message now names both files to update if `maidr.render` itself ever stops blocking its caller.

## Verification

**2382 passed, 13 skipped**, including the benchmark under `--run-benchmark`. `ruff check` unchanged at the 6 pre-existing findings.

---

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -314,7 +314,21 @@ You give up picking up a release newer than your wheel, automatically, in an asy
 
 The same caching means this is not "async gets a different answer forever": after any resolution anywhere in the process, the loop and the main thread agree. It also means `warn_if_bundle_is_stale()` stays quiet in an app that never resolves, since it reads only an already-completed lookup.
 
-**The render itself still blocks, though.** The heading above is about the version *lookup*; do not read it as "nothing blocks the loop". `maidr.render()` is synchronous and CPU-bound — matplotlib's SVG writer, then lxml — so it runs on the event loop and every other session on that worker waits. It never awaits, so nothing preempts it: for as long as one render runs, the loop does not run at all. Measured on a trivial two-bar chart, one render holds the loop for about 55 ms, against 1.5 ms of ordinary scheduler jitter — every other session on that worker waits the whole time. A large scatter or heatmap is worse, and the cost is paid per session per input change. Re-run it with `uv run pytest tests/core/test_render_is_synchronous.py --run-benchmark`. Tracked in [#454](https://github.com/xability/py-maidr/issues/454).
+**The render no longer runs on the loop.** `maidr.render()` itself is synchronous and CPU-bound — matplotlib's SVG writer, then lxml — and it never awaits, so whoever calls it is stopped for its whole duration. `render_maidr` therefore does not call it on the loop: it hands the render to a worker thread, which works because the expensive part (`fig.savefig`, 87–88% of a render at every chart size) releases the GIL.
+
+Measured through the renderer on a 50-bar chart, longest gap between successive ticks of a 1 ms ticker:
+
+| | longest loop gap |
+|---|---|
+| idle, no rendering | 1.4 ms |
+| render on the loop, as it used to be | 609.3 ms |
+| render on a worker thread, as it is now | **12.5 ms** |
+
+Same throughput either way — the work did not get cheaper, it stopped being in everyone else's way.
+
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Answering the one thing you flagged as unverifiable, because **nothing in CI would have caught it either**.

> Worth double-checking the Quarto table renders correctly (bold `**12.5 ms**` inside a pipe-table cell) when the docs site builds, since I can't render it here

`docs.yml` runs on `workflow_dispatch` and `push` to `main` — **not on pull requests**. So a broken table would have gone green here and failed on `main` after merge. Worth knowing independently of this PR.

I checked before opening it, having noticed that my table is the **first** one inside a callout in this file — an unproven construct, not a routine one. Installed Quarto 1.6.39 and rendered `docs/index.qmd` in place, so the project's `_extensions` resolve:

```
Output created: _site/index.html
```

The table is nested inside the callout body, not hoisted out of it:

```
callout-note at 1952 | callout-body at 2220 | <table> at 7399
table is inside the callout body: True
```

Cells and the bold value:

```
td>render on a worker thread, as it is now</td> <td><strong>12.5 ms</strong></td
rendered cells: ['idle, no rendering', '1.4 ms',
                 'render on the loop, as it used to be', '609.3 ms',
                 'render on a worker thread, as it is now', '12.5 ms']
```

And the three stale strings are gone from the built HTML rather than merely from the source:

```
'about 55 ms'                    present: False
'the loop does not run at all'   present: False
'The render itself still blocks' present: False
```

(Rendered with `--no-execute`; the page's code chunks need Jupyter, which CI installs and this sandbox does not. My change touches no chunk, and the markdown path — which is what could have broken — is fully exercised. `docs/_site/` is already gitignored, so nothing is left behind.)

## The rest

- **Table row wording** — left as-is, per your read that it matches the file's existing style.
- Thank you for cross-checking the four factual claims against the source rather than the PR description. The 87–88% figure and the `_FIGURE_LOCKS` `fig.dpi` comment are exactly where I took them from, and confirming that independently is what makes a docs correction trustworthy — the failure mode here is prose drifting from code, which no test catches.

**2382 passed, 13 skipped**, including the benchmark under `--run-benchmark`. `ruff check` unchanged at 6 pre-existing.

---

PR #517, opened 2026-08-18
fix(core): stop emitting a layout gridspec's padding as empty panels

### LINKED ISSUE #513: fix: jointplot emits a 2x6 grid for three panels
## Symptom

`seaborn.jointplot` registers its three real panels correctly, but the emitted schema carries **12 grid positions, nine of them empty**:

```
grid 2 x 6  cells=12  empty=9
  [[hist],  [], [], [], [], []      ]
  [[point], [], [], [], [], [hist]  ]
```

Nothing is missing — the joint panel and both marginals are all there and all readable. But a reader arrowing between panels meets nine empty ones on the way.

(As of #512 those nine positions are at least well-formed. Before that fix they were bare `{}` and took the whole figure down on activation; this issue is only about the padding that remains.)

## Root cause

`JointGrid` lays itself out on a 6×6 gridspec to get the marginal/joint size ratios, and puts each panel on a span rather than a cell:

```
gridspec: (6, 6)
  scatter: rowspan=range(1, 6) colspan=range(0, 5)
  hist:    rowspan=range(0, 1) colspan=range(0, 5)
  hist:    rowspan=range(1, 6) colspan=range(5, 6)
```

`MaidrPlot.__init__` (`maidr/core/plot/maidr_plot.py:51`) keys each layer by the *start* of its span:

```python
self.row_index = ss.rowspan.start
self.col_index = ss.colspan.start
```

and `Maidr._flatten_maidr` (`maidr/core/maidr.py:761`) sizes the grid from the largest start seen:

```python
max_row = max(...); max_col = max(...)
```

So the three panels land at `(1,0)`, `(0,0)` and `(1,5)`, and `max_col = 5` forces six columns for two occupied ones.

## What the fix is *not*

Compressing the coordinates — renumbering the distinct row/col starts to 0..n — is the obvious move and it is **wrong**. Two cases are correct today and compression breaks both:

```
subplot2grid((2, 2)) with a spanning top row:     subplots(1, 3), middle axes undrawn:
  [[bar], []   ]                                   [[bar], [], [bar]]
  [[bar], [bar]]
```

The first depends on the spanning axes occupying one position and leaving its neighbour empty. The second's empty middle cell is a *deliberate* visual gap — the two charts really are separated by a blank panel, and a reader should meet that gap because it is on the screen. Compression would collapse both into something that no longer matches what is drawn.

The distinction is that `JointGrid`'s 6×6 is a **layout** gridspec — its extra rows and columns exist to express size ratios, not positions — whereas `subplots(1, 3)` and `subplot2grid((2, 2))` are **position** gridspecs where every index means a place a chart could be. Any fix has to tell those apart, most likely by looking at which gridspec positions are covered by *some* axes rather than by renumbering what the layers report.

## Related

- #342 (support pairplot and jointplot) — this is a gridspec artefact rather than a missing plot type, so it does not block that.
- #512 — made these positions non-fatal; this is the remaining readability cost.

## Reproduction

```python
import numpy as np, pandas as pd, seaborn as sns
import maidr

df = pd.DataFrame({"x": np.arange(30) % 7, "y": (np.arange(30) * 3) % 11})
g = sns.jointplot(data=df, x="x", y="y")
sub = FigureManager.get_maidr(g.figure)._flatten_maidr()["subplots"]
print(len(sub), "x", len(sub[0]))
for row in sub:
    print([[l["type"] for l in c["layers"]] for c in row])
```

jooyoungseo:
Closes #513.

## What breaks

`seaborn.jointplot` draws three panels and emitted **twelve** grid positions, nine of them holding nothing:

```
grid 2 x 6   cells=12   empty=9
  [[hist],  [], [], [], [], []    ]
  [[point], [], [], [], [], [hist]]
```

`JointGrid` lays its panels out on a 6×6 gridspec to get the marginal-to-joint size ratio, and they start at rows 0 and 1 and columns 0 and 5:

```
scatter: rowspan=range(1, 6) colspan=range(0, 5)
hist:    rowspan=range(0, 1) colspan=range(0, 5)
hist:    rowspan=range(1, 6) colspan=range(5, 6)
```

Sizing the grid from the largest start forces six columns for two occupied ones, so a reader arrowing between the three real panels met nine that exist only for proportion.

## The change

Key each panel by the **rank** of its span start among the starts some axes actually uses. That collapses the rows and columns no axes begins in — exactly the ones a gridspec grows for proportion rather than position. jointplot becomes the 2×2 it looks like:

```
[[hist],  []    ]
[[point], [hist]]
```

marginal above, marginal right, joint below-left, and one genuinely empty corner.

**Ranked over every axes on the figure**, not only the ones carrying a layer. That is what keeps a position the author left empty: the middle panel of a `subplots(1, 3)` drawn on twice is an axes with frame and ticks, on the screen, and a reader should meet it — its own start is what reserves it a column.

**An axes ranked but never occupied costs nothing**, because the grid is still sized from the largest rank a *panel* reaches. So a heatmap's colorbar, which carries a subplotspec of its own (row 1, column 1 of a 3×2), adds a rank rather than a row to arrow through. Ranking preserves order, so one falling between two panels separates them exactly as before.

## What is unchanged

| figure | before | after |
|---|---|---|
| `jointplot` | 2×6, 9 empty | **2×2, 1 empty** |
| `jointplot(marginal_ticks=True)` | 2×6, 9 empty | **2×2, 1 empty** |
| `subplots(1, 3)`, middle undrawn | 1×3, gap kept | 1×3, gap kept |
| `subplot2grid((2,2))` spanning top row | 2×2, 1 empty | 2×2, 1 empty |
| `subplot_mosaic` with a spanning panel | 2×2, 1 empty | 2×2, 1 empty |
| heatmap + colorbar | 1×1 | 1×1 |
| `pairplot` 3×3 | 3×3 | 3×3 |
| plain `subplots(2, 2)` | 2×2 | 2×2 |

Only the two jointplot rows move.

## Tests

Six in a new `tests/core/test_layout_gridspec_padding.py`. Each of the three design choices is covered by a test that fails without it:

| mutation | fails |
|---|---|
| ranking removed entirely | both jointplot cases |
| ranked over panels alone, not every axes | the `subplots(1, 3)` gap |
| sized from every rank rather than every panel | the colorbar case |

```
2399 passed, 1 skipped        # full suite, excluding browser
All checks passed!            # ruff
```

## Note for the record

While stress-testing this I found a **pre-existing** bug that is not touched here and reproduces identically on `main`: two `sns.heatmap` calls in a `subplots(1, 2)` collapse into a single position holding two `heat` layers, because each colorbar re-parents its heatmap into a fresh subgridspec and both then report start `(0, 0)`. Filing separately.

---

### DIFF
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -760,9 +760,15 @@ class Maidr:
         rank rather than a row to arrow through. Ranking preserves order, so
         one falling between two panels separates them exactly as before.
         """
+        # Every panel's own spec as well as the figure's axes, so a panel
+        # whose axes has been detached from the figure still gets a rank
+        # rather than falling to the front of the grid.
         spans = [
             ss
-            for ss in (ax.get_subplotspec() for ax in self._fig.axes)
+            for ss in (
+                *(ax.get_subplotspec() for ax in self._fig.axes),
+                *(plot.ax.get_subplotspec() for plot in self._plots),
+            )
             if ss is not None
         ]
         rows = {ss.rowspan.start for ss in spans}
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -9,7 +9,7 @@ import uuid
 import webbrowser
 import subprocess
 from pathlib import Path
-from typing import Any, Literal, cast
+from typing import Any, Callable, Literal, cast
 from collections import defaultdict
 
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Both notes acted on.

**The `get(index, 0)` fallback** was the better catch. You're right that nothing here detaches an axes today, but "unreachable because no caller does that" is a weaker guarantee than it needs to be — and the failure it hides is a panel silently relocating to the front of the grid, which is not something a reader could tell had happened. The ranks are now built from the panels' own specs as well as the figure's axes, so the default is unreachable by construction rather than by assumption. Costs one more pass over a list that is already short.

**The parametrize line** is wrapped.

`2399 passed, 1 skipped`, ruff clean, and no line in the new test file over 88.

---

PR #520, opened 2026-08-20
test(widget): pin that every entry point describes a figure the same way

jooyoungseo:
## Why

`maidr.render`, Shiny's `render_maidr` and Streamlit's `maidr_html` are three doors onto one figure, and a reader is **not told which one their host application used**. A grid that reads correctly through one and not another is a defect they cannot diagnose, only experience.

Nothing pins that they agree. The existing widget tests assert `"subplots" in document` — that the key survives the round trip, not that the grid has the right shape.

The doors all funnel into `Maidr._flatten_maidr` today, so agreement holds by construction rather than by anyone maintaining it. That is the reason to pin it rather than a reason not to: **#443 is the precedent in this repo.** `plt.show()` degraded gracefully for an unregistered figure while `render`/`show`/`save_html` raised `KeyError`, because the graceful path had been wired into one door and not the others. Nothing failed until a user went through the wrong one.

## What is covered

The three figures whose grid coordinates were just corrected — an authored gap (#512), a proportions gridspec (#517), and panels re-parented by their colorbars (#519). Divergence would be most likely and least visible there.

Verified through all three doors before writing the test:

| figure | `maidr.render` | `render_maidr` | `maidr_html` |
|---|---|---|---|
| `subplots(1, 3)` with a gap | `[[1, 0, 1]]` | `[[1, 0, 1]]` | `[[1, 0, 1]]` |
| `jointplot` | `[[1, 0], [1, 1]]` | `[[1, 0], [1, 1]]` | `[[1, 0], [1, 1]]` |
| two heatmaps | `[[1, 1]]` | `[[1, 1]]` | `[[1, 1]]` |

The grid is read back **out of the emitted markup**, not from the `Maidr` instance — the instance is what all three doors share, so asking it could not tell them apart.

## Two halves, because neither catches the other's failure

| mutation | agreement tests | correctness tests |
|---|---|---|
| one door post-processes the schema differently | **3 fail** | 3 pass |
| shared `_flatten_maidr` returns the wrong grid | 3 pass — they still agree, on a wrong answer | **3 fail** |

Both mutations were actually run, not reasoned about.

## Verification

```
2409 passed, 1 skipped     # full suite, excluding browser
13 passed                  # tests/browser --run-browser, on merged main
All checks passed!         # ruff, and no line over 88
```

The browser run is on `main` rather than this branch — it exercises the Shiny and Streamlit paths end to end in Chromium and confirms they are healthy after #512/#516/#517/#519. This PR adds no browser test.

## Also in this commit

Nothing else. One stale `__pycache__/test_zz_probe*.pyc` left behind by my own earlier probing was deleted from the working tree; it was git-ignored, so it does not appear in the diff.

---

### DIFF
--- a/tests/widget/test_every_door_agrees.py
+++ b/tests/widget/test_every_door_agrees.py
@@ -42,6 +42,13 @@ import seaborn as sns  # noqa: E402
 
 import maidr  # noqa: E402
 from maidr.util.environment import Environment  # noqa: E402
+
+# Guarded here as well as in `conftest.py`, matching `test_shiny.py`. The
+# directory-level skip already covers a missing shiny, but only because
+# pytest skips a whole subtree when a `conftest` raises during collection.
+# Relying on that leaves this file uncollectable on its own.
+pytest.importorskip("shiny")
+
 from maidr.widget.shiny import render_maidr  # noqa: E402
 from maidr.widget.streamlit import maidr_html  # noqa: E402
 
--- a/tests/widget/test_every_door_agrees.py
+++ b/tests/widget/test_every_door_agrees.py
@@ -1,20 +1,27 @@
-"""Every entry point describes the same figure the same way.
+"""A figure's grid survives the iframe a hosted render wraps it in.
 
-`maidr.render`, Shiny's `render_maidr` and Streamlit's `maidr_html` are
-three doors onto one figure, and a reader is not told which one their host
-application used. So a grid that reads correctly through one and not
-another is a defect they cannot diagnose, only experience.
+Under Shiny -- and anywhere else `Environment` reports a live host -- the
+chart is nested inside an iframe's `srcdoc`, which escapes the schema a
+second time on top of the escaping lxml already applied to the `<svg>`.
+That is the form a Shiny or Flask reader actually receives, and until this
+file nothing asserted the grid still reads correctly in it: every existing
+grid test builds the schema through `_flatten_maidr` directly, where no
+wrapping happens.
 
-The doors funnel into `Maidr._flatten_maidr` today, so this holds by
-construction rather than by anyone maintaining it -- which is the reason to
-pin it. #443 is the precedent: `plt.show()` degraded gracefully for an
-unregistered figure while `render`/`show`/`save_html` raised, because the
-graceful path had been wired into one door and not the others. Nothing
-failed until a user went through the wrong one.
+The three entry points are checked together, but not because they branch:
+the wrapping decision is environmental rather than per-door, so
+`maidr.render`, `render_maidr` and `maidr_html` all wrap or all do not.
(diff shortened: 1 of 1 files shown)


jooyoungseo:
You were right, and it is worse than you put it — the test was pinning nothing it claimed. Thank you for pushing on it.

## Confirmed, then found the stronger version

With no session, all three doors take the unwrapped branch exactly as you said. **Byte-identical output, 13347 characters each.**

I then tried your suggested fix — drive it through a real session so `is_shiny()` is `True`:

```
is_shiny(): True
  maidr.render   iframe=True  srcdoc=True  len=24626
  render_maidr   iframe=True  srcdoc=True  len=24626
  maidr_html     iframe=True  srcdoc=True  len=24626
```

Still identical. The wrapping decision is **environmental, not per-door**, so these three cannot diverge in production in either direction. The #443 precedent I cited is a class of bug this machinery could never have reached, and my framing was wrong on that.

## What the test was accidentally missing turned out to be the point

Under a live host the schema is nested inside `srcdoc`, escaped a second time on top of lxml's escaping of the `<svg>`. That is the form a Shiny reader actually receives, and **nothing asserted the grid still reads in it** — every other grid test goes through `_flatten_maidr` directly, where no wrapping happens.

So the file now pins that: the grid wrapped, the grid unwrapped, and the door claim stated as what it really is — not that the doors differ, but that none may grow post-processing of its own.

Your finding 2 is resolved by construction rather than by a comment: `_grid_of` now unwraps `srcdoc` first and then applies the simple attribute pattern, so there is no single-escape assumption left to document.

## Three mutations, each biting only its own tests

| mutation | fails |
|---|---|
| shared `_flatten_maidr` returns a wrong grid | all 9 |
| `maidr_html` post-processes the schema | the 3 door tests |
| iframe wrapping suppressed | the 3 wrapped tests |

The second is worth flagging because it caught me out the same way the original test did. My first attempt matched `maidr="..."`, which **does not occur** once the schema is inside `srcdoc` — so it mutated nothing, the suite passed, and I nearly recorded that as evidence the test was sound. It had to be rewritten to handle both escape depths before it bit.

```
2412 passed, 1 skipped     # full suite, excluding browser
All checks passed!         # ruff, no line over 88
```

---

jooyoungseo:
Nit taken — `pytest.importorskip("shiny")` now sits ahead of the `maidr.widget.shiny` import, matching `test_shiny.py:26`, with a comment recording why the redundancy is deliberate rather than accidental.

You put the reason better than I would have: the directory-level skip works only because pytest skips a whole subtree when a `conftest` raises during collection, so leaning on it is an implicit dependency on collection order rather than a guard. Cannot fail today given CI installs the extra, but the next file added here will copy whichever pattern it finds.

```
80 passed        # tests/widget/
9 passed         # this file
All checks passed!
```

---

PR #523, opened 2026-08-20
fix(altair): say that use_cdn=False cannot be honoured, instead of ignoring it

### LINKED ISSUE #521: use_cdn is accepted and silently discarded for Altair charts
## What happens

`maidr.render(chart, use_cdn=False)` accepts the argument on an Altair chart and then ignores it. The emitted HTML is **identical** with the flag on and off, once the generated UUIDs are normalised:

```
render(): identical once UUIDs are normalised: True
```

The only textual differences between the two outputs are two `maidr-altair-<uuid>` ids. Nothing about how the runtime is loaded changes.

So an offline user who sets `use_cdn=False` still gets a chart that needs four remote scripts:

```
https://cdn.jsdelivr.net/npm/vega@5
https://cdn.jsdelivr.net/npm/vega-lite@5
https://cdn.jsdelivr.net/npm/vega-embed@6
https://cdn.jsdelivr.net/npm/maidr@4.4.0/dist/vegalite.js
```

**No warning is raised** — measured with `warnings.catch_warnings(record=True)`: zero warnings for `use_cdn=False` on this path. The chart renders as a blank area with the runtime never defined, and nothing says why.

## Root cause

`maidr/api.py:421` returns for Altair **before** `use_cdn` is even resolved:

```python
if _is_altair_chart(plot):
    from maidr.altair import AltairMaidr

    return AltairMaidr(plot).render()      # use_cdn never passed

use_cdn = _resolve_use_cdn(use_cdn)        # only reached by other plot types
```

The same shape repeats at `api.py:480` (`show`) and `api.py:556` (`save_html`).

And `maidr/altair/altair_maidr.py` has **no handling to pass it to** — grepping that file for `use_cdn`, `bundled_js_path` or any inline/bundle logic returns nothing at all.

## How the three paths compare

Measured through `render_maidr` with `use_cdn=False`, counting remaining remote references:

| path | `use_cdn=False` honoured? | remote refs remaining |
|---|---|---|
| matplotlib | ✅ fully — bundle inlined | 0 |
| plotly | ⚠️ partly — maidr's own refs drop | 1 (`cdn.plot.ly`, plotly's own library) |
| **altair** | ❌ **not at all** | **4** |

The plotly remainder is arguably out of scope: that is plotly's own library, not maidr's. The Altair row is the problem, because `maidr@4.4.0/dist/vegalite.js` — maidr's own adapter — is among the four.

## Why this is the bad kind of failure

This is the same class as #358 and #455: a `use_cdn=False` user is by definition the one who cannot reach a CDN, and they get a blank chart with no explanation. The project already treats that as the worst shape a failure can take, and already has machinery for saying so — `_warn_if_no_runtime` on the Streamlit path, `warn_if_bundle_cannot_render` for trace types. Neither fires here.

Note the architectural context from `CLAUDE.md`: *"`maidr/altair/` delegates entirely to the upstream Vega-Lite JS adapter."* Delegation is a legitimate design, and it may well be that Altair genuinely cannot be served offline today. What is not defensible is **accepting a parameter and discarding it in silence** — the honest options are to honour it or to say it cannot be honoured.

## Suggested shape

Cheapest first:

1. **Warn.** When `use_cdn` is falsy and the plot is Altair, say that this path always loads from a CDN and name what it needs. One branch in `api.py`, no behaviour change. This alone converts a blank chart into a diagnosable one.
2. **Inline what can be inlined.** `maidr@4.4.0/dist/vegalite.js` is maidr's own artifact; whether it is shippable in the wheel like `maidr.js` is a packaging question worth answering separately from vega/vega-lite/vega-embed, which are third-party.
3. **Document it** in the offline section, alongside the `use_cdn` docs.

## Reproduce

```python
import altair as alt, pandas as pd, re, warnings
import maidr

DF = pd.DataFrame({"cat": list("abc"), "val": [1, 2, 3]})
chart = alt.Chart(DF).mark_bar().encode(x="cat", y="val")

with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    off = str(maidr.render(chart, use_cdn=False))
    print("warnings:", [str(x.message) for x in w] or "NONE")

print(sorted(set(re.findall(r"https://[^\"']+", off))))
```

## Found how

While running the browser QA that #457 asked for. That evaluation covered matplotlib only, and this is what came out of extending it to the other two paths — the Plotly and Altair keyboard QA itself could not be completed here, because Chromium in my sandbox cannot reach the CDNs those paths depend on (`ERR_CONNECTION_RESET` on every one), which is itself a consequence of the row above.

jooyoungseo:
Closes #521.

## What breaks

`maidr.render(chart, use_cdn=False)` accepts the flag on an Altair chart and discards it. The emitted HTML is **identical** with the flag on and off, once the generated UUIDs are normalised:

```
render(): identical once UUIDs are normalised: True
```

No warning is raised — measured with `warnings.catch_warnings(record=True)`, zero. So the page still needs four remote scripts:

```
https://cdn.jsdelivr.net/npm/vega@5
https://cdn.jsdelivr.net/npm/vega-lite@5
https://cdn.jsdelivr.net/npm/vega-embed@6
https://cdn.jsdelivr.net/npm/maidr@4.4.0/dist/vegalite.js
```

The reader this fails is precisely the one who set the flag: `use_cdn=False` means they cannot reach a CDN, so they get a chart that never initialises and no reason why. Same class as #358 and #455.

## Root cause

`api.py` returns for Altair **before** `use_cdn` is even resolved:

```python
if _is_altair_chart(plot):
    from maidr.altair import AltairMaidr
    return AltairMaidr(plot).render()      # use_cdn never passed

use_cdn = _resolve_use_cdn(use_cdn)        # only reached by other plot types
```

Repeated at `show` and `save_html`. And `maidr/altair/altair_maidr.py` has nothing to pass it to — that file contains no reference to `use_cdn`, `bundled_js_path`, or any inline handling at all. The path renders through the upstream Vega-Lite adapter, published only on a CDN, so there is genuinely nothing to inline.

| path | `use_cdn=False` honoured? | remote refs left |
|---|---|---|
| matplotlib | fully — bundle inlined | 0 |
| plotly | partly | 1 (`cdn.plot.ly`, plotly's own library) |
| **altair** | **not at all** | **4** |

## What this changes

Nothing about behaviour — it says so instead of staying quiet. Whether the flag could one day be honoured is a packaging question about maidr's own `vegalite.js`, separate from the third-party vega bundle, and is left open on #521 as options 2 and 3.

Two deliberate narrowings:

- **Only an explicit `False` warns.** `"auto"` has no fallback on this path either, but a reader on `"auto"` has not said they are offline, and warning on the common case would be noise.
- **The check resolves through `_resolve_use_cdn`.** Reading the caller's argument alone would let `set_use_cdn(False)` through silently — the same discard, reached by the other door.

## Tests

Seven, in a new file. Three mutations, each biting only its own tests:

| mutation | fails |
|---|---|
| warning removed entirely | the 3 that expect it |
| raw argument read instead of the resolved mode | only the process-default test |
| warn on every mode | only the 3 "stays quiet" cases |

The matplotlib control is there because it honours the flag by inlining, so it must stay silent — that is what keeps the warning from being wired to the wrong condition.

```
2419 passed, 1 skipped     # full suite, excluding browser
All checks passed!         # ruff
```

## Found how

Extending the #457 browser QA to the Plotly and Altair paths. That extension could not be completed — Chromium in my sandbox cannot reach the CDNs those paths need — and this is why there was no offline mode to fall back to. Recorded on [#457](https://github.com/xability/py-maidr/issues/457#issuecomment-5360409488).

---

### DIFF
--- a/maidr/api.py
+++ b/maidr/api.py
@@ -189,9 +189,11 @@ def _warn_altair_ignores_use_cdn(
         caller because ``None`` defers to the process default, which may
         itself be ``False``.
     stacklevel : int
-        Passed through to :func:`warnings.warn`. Keyword-only for the same
-        reason :func:`maidr.util.fallback.warn_unsupported` does it: the
-        frame count differs per entry point.
+        Passed through to :func:`warnings.warn`. Keyword-only because it is
+        a frame count rather than a value about the chart, and the three
+        entry points that pass it sit at the same depth today -- so a
+        positional argument would read as one more thing about `use_cdn`
+        and silently become wrong the moment one of them grew a wrapper.
     """
     if _resolve_use_cdn(use_cdn) is not False:
         return
--- a/maidr/api.py
+++ b/maidr/api.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import os
+import warnings
 from typing import Any, Literal
 
 from htmltools import Tag
@@ -157,6 +158,54 @@ def _resolve_use_cdn(
     if value is None:
         return get_use_cdn()
     return value
+
+
+#: Named in the warning so a reader can see what would have to travel with
+#: the page, rather than being told only that something will not.
+_ALTAIR_REMOTE_RUNTIME = (
+    "vega, vega-lite, vega-embed and maidr's own vegalite.js"
+)
+
+
+def _warn_altair_ignores_use_cdn(
+    use_cdn: bool | Literal["auto"] | None, *, stacklevel: int
+) -> None:
+    """Say that ``use_cdn=False`` cannot be honoured for an Altair chart.
+
+    The Altair path delegates to the upstream Vega-Lite adapter, which is
+    loaded from a CDN and has no bundled counterpart -- so the flag was
+    accepted and discarded, and the reader who most needs to know is the
+    one it fails for: ``use_cdn=False`` means they cannot reach a CDN, and
+    without this they get a chart that never initialises and no reason why
+    (#521).
+
+    Only an explicit ``False`` warns. ``"auto"`` is the CDN with a fallback
+    that this path does not have either, but a reader on ``"auto"`` has not
+    said they are offline, and warning them on every render would be noise
+    on the common case.
+
+    Parameters
+    ----------
+    use_cdn : bool, {"auto"}, or None
+        The caller's value, unresolved. Resolved here rather than by the
+        caller because ``None`` defers to the process default, which may
+        itself be ``False``.
+    stacklevel : int
+        Passed through to :func:`warnings.warn`. Keyword-only for the same
+        reason :func:`maidr.util.fallback.warn_unsupported` does it: the
+        frame count differs per entry point.
+    """
+    if _resolve_use_cdn(use_cdn) is not False:
+        return
+
+    warnings.warn(
+        "maidr: use_cdn=False cannot be honoured for an Altair chart. That "
+        "path renders through the upstream Vega-Lite adapter, which is only "
+        f"published on a CDN, so the page still loads {_ALTAIR_REMOTE_RUNTIME} "
+        "remotely and will not initialise without network access. Render the "
+        "same data through matplotlib or seaborn for an offline chart.",
+        stacklevel=stacklevel,
+    )
 
 
 def init_notebook(
@@ -421,6 +470,8 @@ def render(
     if _is_altair_chart(plot):
         from maidr.altair import AltairMaidr
 
+        _warn_altair_ignores_use_cdn(use_cdn, stacklevel=3)
+
         return AltairMaidr(plot).render()
 
     use_cdn = _resolve_use_cdn(use_cdn)
@@ -479,6 +530,8 @@ def show(
     """
     if _is_altair_chart(plot):
         from maidr.altair import AltairMaidr
+
+        _warn_altair_ignores_use_cdn(use_cdn, stacklevel=3)
 
         return AltairMaidr(plot).show(renderer)
 
@@ -555,6 +608,8 @@ def save_html(
     """
     if _is_altair_chart(plot):
         from maidr.altair import AltairMaidr
+
+        _warn_altair_ignores_use_cdn(use_cdn, stacklevel=3)
 
         return AltairMaidr(plot).save_html(
             file,
--- a/maidr/api.py
+++ b/maidr/api.py
@@ -162,9 +162,7 @@ def _resolve_use_cdn(
 
 #: Named in the warning so a reader can see what would have to travel with
 #: the page, rather than being told only that something will not.
-_ALTAIR_REMOTE_RUNTIME = (
-    "vega, vega-lite, vega-embed and maidr's own vegalite.js"
-)
+_ALTAIR_REMOTE_RUNTIME = "vega, vega-lite, vega-embed and maidr's own vegalite.js"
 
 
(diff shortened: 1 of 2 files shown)


jooyoungseo:
All three notes taken, including the one you asked to be addressed before merge.

**The `render()` docstring gap was real** — I checked rather than assumed. The caveat sits at `api.py:523` (`show`) and `:601` (`save_html`) and was absent from `render`, so a reader of `render`'s docs alone would meet the new warning with no explanation for it. It now carries the same note, extended to say what the warning does:

> Altair charts always use the CDN — this argument is not plumbed through that adapter, so `False` warns rather than taking effect (#521).

**The parens** around `_ALTAIR_REMOTE_RUNTIME` are gone; you were right that they wrapped a single literal with nothing to concatenate.

**`test_save_html_says_it_too`** now asserts exactly one warning like its `render` sibling, rather than a non-empty list. The looser form would have passed if the helper were ever called twice on one entry point, which is the kind of thing this test should catch.

Thanks also for independently confirming the `stacklevel=3` frame count and the `_resolve_use_cdn` "other door" — both were deliberate and it is useful to have them checked rather than taken on trust.

```
2419 passed, 1 skipped     # full suite, excluding browser
All checks passed!         # ruff
```

---

jooyoungseo:
The coverage gap was the important one, and you were right that nothing would have caught it.

**`test_show_says_it_too` added**, and mutation-checked the way you framed the risk — removing the warning call from `show` *alone*:

```
FAILED test_show_says_it_too
1 failed, 7 passed
```

Only that test, which is what makes it worth having. The renderer is stubbed rather than driven, since the warning fires before the hand-off and letting `show` proceed would make the test depend on a display.

**The `stacklevel` rationale was wrong and I've corrected it.** I had copied `warn_unsupported`'s justification — "the frame count differs per entry point" — and you're right that it does not hold here, since all three call sites sit at the same depth. The honest reason to keep it keyword-only is that it is a frame count rather than a fact about the chart, so a positional argument would read as one more thing about `use_cdn` and quietly become wrong if an entry point ever grew a wrapper. That is now what the docstring says.

**The 90-column line** is wrapped.

```
2420 passed, 1 skipped     # full suite, excluding browser
All checks passed!         # ruff, no line over 88
```

---

jooyoungseo:
Keeping the file in `tests/core/`, and worth saying why rather than just leaving it.

You named the counter-argument yourself — it exercises `maidr.api` entry points rather than `AltairMaidr` internals — and the directory is already organised that way. Every other `use_cdn` test lives there:

```
tests/core/test_cdn_event_loop.py
tests/core/test_cdn_version.py
tests/core/test_offline_bundle.py
tests/core/test_offline_fallback_report.py
```

while `tests/altair/test_altair_maidr.py` is about the adapter itself. So the grouping is by concern rather than by plot library, and someone asking "what does `use_cdn` actually do" finds these together. Moving this one would split that.

On the second nit — `_ALTAIR_REMOTE_RUNTIME` drifting if the adapter's peer dependencies change — that is a real hazard and nothing here pins it. It is a warning string, so a drift is cosmetic rather than a wrong behaviour, but it would mean naming a script that is no longer fetched. Not worth a test that asserts against a hardcoded list on both sides; flagging it so it is on the record.

The third I have left as you suggested.

```
20/20 checks green, including the browser suite
2420 passed, 1 skipped
```

Merging.

---

PR #525, opened 2026-08-20
fix(core): read a bivariate histogram as the heatmap it draws

### LINKED ISSUE #522: A bivariate `sns.histplot` is silent, though the same QuadMesh read through `sns.displot` is a heatmap
`sns.histplot(df, x=..., y=...)` draws a 2-D histogram as a `QuadMesh`, exactly as `ax.hist2d` and `ax.pcolormesh` do. Those two are read as `heat`. This one raises `UnsupportedPlotError` and the figure is unreadable.

## Measured

seaborn 0.13.2, matplotlib 3.9.4, through `FigureManager.get_maidr`:

| call | layers |
|---|---|
| `sns.histplot(df, x='a', y='b')` | **silent** (`UnsupportedPlotError`) |
| `sns.displot(df, x='a', y='b')` | `['heat']` |
| `sns.jointplot(df, x='a', y='b', kind='hist')` | `['hist', 'hist']` — **the margins only** |
| `ax.pcolormesh(...)` | `['heat']` |
| `ax.hist2d(a, b)` | `['heat']` |
| `sns.heatmap(...)` | `['heat']` |
| `sns.histplot(df, x='a')` | `['hist']` |
| `sns.displot(df, x='a')` | `['hist']` |

The two rows that matter are the first two. **The same chart, drawn from the same data, is readable through `displot` and silent through `histplot`** — and the only difference between them is that `histplot` is patched and `displot` is not.

The `jointplot(kind='hist')` row is the same bug wearing a worse symptom: the reader gets both marginal histograms and *not* the joint panel, which is the thing a joint plot is drawn for. Nothing raises, so it reads as a complete chart.

## Why

Three correct decisions compose into a wrong one.

1. `common()` sets the internal context before running the draw, so that a patched seaborn function calling a patched matplotlib one does not register twice (`maidr/patch/common.py:231-236`).
2. Inside the draw, seaborn calls `ax.pcolormesh` — confirmed by spying on `Axes.pcolormesh`, which records exactly one call. The `heat()` patch sees `ContextManager.is_internal_context()` and draws quietly, registering nothing (`maidr/patch/heatmap.py:93-94`).
3. Back in `common()`, `_drew_bars` correctly answers *no* — a bivariate histplot draws no `BarContainer` — so the histogram patch declines. That decline is deliberate and right; its docstring records why (#388): registering `hist` for a QuadMesh "promises a reading nothing can produce, and extraction then took the whole figure down with it."

Each step is defensible. Together they mean the one registration the chart could have had was suppressed by a guard that assumed the outer patch would make its own — and then the outer patch declined.

`displot` escapes because it is not patched at all, so step 1 never happens and the inner `pcolormesh` registers on its own account. That is the accident that makes the two spellings disagree.

## Proposed

Have the seaborn histogram patch register `HEAT` when the draw produced a `QuadMesh` and no bars, rather than declining outright. It is local, it asks the artists rather than the arguments — the discipline `_drew_bars` already established — and it leaves the recursion guard alone, which is doing its job everywhere else.

The alternative, making the guard carry *what* the outer patch intends to register so an inner registration survives an outer decline, is more machinery and would want its own issue. Worth naming because the same composition can bite any patched function that declines after suppressing a supported inner artist; `histplot` is the one that currently does.

## Not in scope

`sns.jointplot(kind='kde')` returns `['smooth', 'smooth']` — again the margins only. That one is not this bug: its joint panel is a `QuadContourSet`, a plot type py-maidr does not read at all. Filed separately.

jooyoungseo:
Closes #522.

`sns.histplot(df, x=..., y=...)` draws a `QuadMesh` of joint counts, exactly as `ax.hist2d` and `ax.pcolormesh` do. Those read as `heat`. This raised `UnsupportedPlotError` and the figure was unreadable — while the *same chart* through `sns.displot(df, x=..., y=...)` read as `heat` all along.

## Why the two spellings disagreed

Three correct decisions composed into a wrong one.

1. `common()` sets the internal context before the draw, so a patched seaborn call that reaches a patched matplotlib one does not register twice.
2. Inside the draw seaborn calls `ax.pcolormesh` — confirmed by spying on `Axes.pcolormesh`, which records exactly one call. The `heat()` patch sees the context and draws quietly, registering nothing.
3. `_drew_bars` then answers *no*, and the histogram patch declines. That decline is right, and #388's tests pin it: `hist` promises one bin per bar with a count, which a mesh has neither of.

The guard in step 2 assumes the outer patch will make a registration of its own. In step 3 the outer patch declined, and the one the chart could have had went with it.

`displot` escaped only by accident: it drives `_DistributionPlotter.plot_bivariate_histogram`, which nothing patches, so the inner `pcolormesh` runs outside the context and registers on its own account.

## The fix

Having declined `hist`, the seaborn histogram patch now registers `heat` when the call added a mesh. `_drew_mesh` asks what *this call* drew, the way `_drew_bars` does, so a heatmap already on the axes is not claimed a second time.

## A second defect it had to land on

The first version of this passed my own probe and then failed an existing test — `test_a_scatter_under_a_2d_histogram_still_renders` — with `ExtractionError … from <class 'matplotlib.collections.PathCollection'>`.

`extract_scalar_mappable` took the first `ScalarMappable` on the axes, and **a scatter's `PathCollection` is a `ScalarMappable` too** — that is how a colour-mapped scatter works. So a heatmap drawn beside a scatter was read from an artist with no grid.

This is not new and not mine: on unmodified `main`,

```python
sns.scatterplot(...); ax.pcolormesh(...)
```

already fails exactly this way. It now prefers a `QuadMesh`, `PolyQuadMesh` or `AxesImage` and falls back to the first mappable of any kind, so nothing that resolved before resolves differently. It also returns `None` on an axes holding none rather than raising `StopIteration` through a bare `next()` — which is the `Optional` it has always been annotated with, what `HeatPlot` is written for (`if data is None`), and the same shape of failure #388 removed from `extract_container`.

## Four existing expectations move

Every one of them is the point of the fix, and none is a weakened assertion:

| test | was | now |
|---|---|---|
| `test_a_scatter_under_a_2d_histogram_still_renders` | `[SCATTER]` | `[SCATTER, HEAT]` |
| `test_a_jointplot_of_histograms_renders` | `[HIST, HIST]` | `[HEAT, HIST, HIST]` |
| `test_bars_already_on_the_axes_do_not_make_it_a_histogram` | `[BAR]` | `[BAR, HEAT]` |
| `test_a_bivariate_histogram_still_declines` | `[]` | `[HEAT]` |

The jointplot row is the one worth pausing on. It returned both marginal histograms and nothing for the joint panel — so nothing raised, and a reader was handed a complete-sounding chart with the thing a joint plot is drawn for silently missing. Each test keeps its original point: none of them now asserts a `hist` for a mesh.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — 2,412 passed, 18 skipped. (`ruff check` reports 51 pre-existing findings across the example notebooks on a clean tree; the four files this touches are clean.)

| mutation | result |
|---|---|
| the mesh registration removed (the reported bug) | 6 failed |
| ownership dropped — register whenever *any* mesh is present | 1 failed |
| the pre-draw snapshot emptied, so everything looks new | 1 failed |
| extractor back to the first mappable of any kind | 2 failed |
| extractor indexes an empty list instead of returning `None` | 1 failed |

The ownership case took two attempts to make honest. My first test drew `sns.heatmap` then a bivariate `histplot`, which passed the mutation as well as the real code — a bivariate call always adds a mesh of its own, so nothing distinguishes "this call's" from "any". The reachable case is a call that declines bars *and* adds no mesh: `sns.histplot(x=[])`. That one reddens both mutations.

## Not in scope, found while measuring

Two things this touches but does not fix, both pre-existing and both reproducible without any of the above:

- **A numeric-axis heatmap's row and column labels are the axis ticks, not the bins.** `ax.hist2d(a, b, bins=(3, 2))` emits a 2×3 grid with **nine** x labels and seven y labels; `pcolormesh` and `imshow` do the same. Only `sns.heatmap` lines up, because its axis is categorical. A reader navigating column 2 of 3 is told a label from a list of 9.
- **Two meshes on one axes are both read from the first.** Two `pcolormesh` calls emit two `heat` layers with identical data. `extract_scalar_mappable`'s own comment has always admitted the assumption — *"We assume only one ScalarMappable is present to avoid plot clutter"* — and resolving a layer's artist per layer rather than per axes is a larger change than this one.

I'll file both rather than widen this PR.

---

### DIFF
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -244,9 +244,17 @@ def sns_hist(wrapped, instance, args, kwargs) -> Axes:
         # reason than that it is unpatched, so the inner call runs outside the
         # context -- which is what made the same figure readable through one
         # spelling and silent through the other (#522).
-        axes = drawn if isinstance(drawn, Axes) else FigureManager.get_axes(drawn)
+        axes = FigureManager.get_axes(drawn)
         if _drew_mesh(axes, meshes):
-            FigureManager.create_maidr(axes, PlotType.HEAT)
+            # Named from `stat` rather than hardcoded, because it is what the
+            # cells actually hold: the default is a count, but
+            # `stat="density"` or `"probability"` makes them something else,
+            # and a reader hearing "count" for a density has been told the
+            # wrong thing about every cell. `HexbinPlot` labels its own `z`
+            # the same way and for the same reason.
+            FigureManager.create_maidr(
+                axes, PlotType.HEAT, z_label=str(kwargs.get("stat", "count"))
+            )
         return drawn
 
     # Register the histogram as HIST as before
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -7,7 +7,7 @@ import wrapt
 
 import numpy as np
 from matplotlib.axes import Axes
-from matplotlib.collections import QuadMesh
+from matplotlib.collections import PolyQuadMesh, QuadMesh
 from matplotlib.container import BarContainer
 from matplotlib.patches import Polygon
 from matplotlib.lines import Line2D
@@ -156,7 +156,16 @@ def _drew_bars(plot: Any, before: list) -> bool:
 
 def _meshes_of(ax: Axes | None) -> list:
     """
-    The ``QuadMesh`` artists already on an axes.
+    The mesh artists already on an axes.
+
+    Both mesh classes, not only the one seaborn currently draws through.
+    Measured on seaborn 0.13.2, a bivariate ``histplot`` reaches
+    ``Axes.pcolormesh`` and so produces a ``QuadMesh`` -- but naming only that
+    would tie this to a seaborn internal, and the failure if it ever moved to
+    ``Axes.pcolor`` would be the silent one this whole change exists to
+    remove: the mesh would go unrecognised, the call would decline as before,
+    and the chart would be quiet again. ``PolyQuadMesh`` costs nothing to
+    accept and is a heatmap by the same argument.
 
     Held as a list of the artists themselves rather than of their ids, for the
     reason ``_containers_of`` gives: an id compared after the object it named
@@ -176,7 +185,7 @@ def _meshes_of(ax: Axes | None) -> list:
     return [
         artist
         for artist in (getattr(ax, "collections", ()) or ())
-        if isinstance(artist, QuadMesh)
+        if isinstance(artist, (QuadMesh, PolyQuadMesh))
     ]
 
 
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -7,6 +7,7 @@ import wrapt
 
 import numpy as np
 from matplotlib.axes import Axes
+from matplotlib.collections import QuadMesh
 from matplotlib.container import BarContainer
 from matplotlib.patches import Polygon
 from matplotlib.lines import Line2D
@@ -153,6 +154,57 @@ def _drew_bars(plot: Any, before: list) -> bool:
     )
 
 
+def _meshes_of(ax: Axes | None) -> list:
+    """
+    The ``QuadMesh`` artists already on an axes.
+
+    Held as a list of the artists themselves rather than of their ids, for the
+    reason ``_containers_of`` gives: an id compared after the object it named
+    was freed can be matched by an unrelated artist allocated at the same
+    address.
+
+    Parameters
+    ----------
+    ax : Axes or None
+        The axes to look at, or None when it does not exist yet.
+
+    Returns
+    -------
+    list
+        The meshes on it, in draw order.
+    """
+    return [
+        artist
+        for artist in (getattr(ax, "collections", ()) or ())
+        if isinstance(artist, QuadMesh)
+    ]
+
+
+def _drew_mesh(ax: Axes | None, before: list) -> bool:
+    """
+    Whether *this call* drew a mesh of joint counts.
+
+    Asked the same way ``_drew_bars`` asks its question, and for the same
+    reason: an axes that already held a heatmap -- ``sns.heatmap(ax=ax)``
+    first, then a bivariate ``histplot(ax=ax)`` -- would otherwise have
+    someone else's mesh claimed as this call's.
+
+    Parameters
+    ----------
+    ax : Axes or None
+        The axes the call drew on.
+    before : list
+        The meshes that were on it beforehand.
+
+    Returns
+    -------
+    bool
+        True when a mesh appeared that was not there before.
+    """
+    seen = {id(mesh) for mesh in before}
+    return any(id(mesh) not in seen for mesh in _meshes_of(ax))
+
+
 def sns_hist(wrapped, instance, args, kwargs) -> Axes:
     """
     Patch seaborn.histplot to register HIST and (if kde=True) SMOOTH layers for MAIDR.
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Both points taken, both in 7e882c6.

**`_meshes_of` naming only `QuadMesh`.** You're right that it coupled the reading to a seaborn internal, and right about the failure mode: it would regress to the *original* silent decline rather than fail loudly, which is the one shape of failure this change exists to remove. `_meshes_of` now names `PolyQuadMesh` as well — measured, `ax.pcolor` produces one — which also makes it agree with `extract_scalar_mappable`, where both were already named. The docstring says why rather than leaving it to be re-derived.

**Two stacked bivariate histograms.** Added, and worth being precise about what it does and does not do:

```
sns.histplot(x='v', y='w'); sns.histplot(x='w', y='v')   →   ['heat', 'heat']
```

The layer count is right, and that is all the test asserts. Both layers read their values from the **first** mesh — which is #527, reproduces on two plain `ax.pcolormesh` calls with none of this involved, and is stated in the test body so the assertion is not read as endorsing the data.

It is also, honestly, not a mutation-killer: with the ownership check widened to "any mesh on the axes", this case still gives two layers, because a bivariate call always adds a mesh of its own. That is exactly why the ownership test uses `sns.histplot(x=[])` instead — a call that declines bars *and* adds nothing is the only shape that separates "this call's mesh" from "a mesh". So the new test rounds out coverage of the real-world case, and the empty one is what holds the guard up.

Full suite after both: 2,413 passed, 18 skipped. `ruff check` clean on every file this touches (the 6 findings under `maidr/`+`tests/` are pre-existing `__init__.py` re-export warnings, unchanged).

On the third point — agreed, and thank you for confirming it by reading `_extract_scalar_mappable_data`/`extract_level` rather than taking the description's word for it. Both are now filed: #526 for the tick-versus-cell labels, #527 for the shared-artist resolution.

---

jooyoungseo:
All three actioned — the first two in 5d67d62, the third filed.

**1. `z_label` lost for the auto-registered heatmap.** Real gap, and the right fix turned out to be slightly better than a constant. What the cells hold is in `stat`, and only the default is a count:

```
sns.histplot(x, y)                      →  axes.z = {"label": "count"}
sns.histplot(x, y, stat="density")      →  axes.z = {"label": "density"}
sns.histplot(x, y, stat="probability")  →  axes.z = {"label": "probability"}
```

Hardcoding `"count"` would have put the wrong word on every number in the grid for the other two. Read from `stat`, the way `HexbinPlot` labels its own `z`, and pinned by a test.

**2. Unguarded `FigureManager.get_axes(drawn)`.** You're right about the asymmetry, and checking it turned up that the branch beside it could never run either way: `get_axes` returns an `Axes` unchanged (`elif isinstance(artist, Axes): return artist`), so `drawn if isinstance(drawn, Axes) else get_axes(drawn)` and `get_axes(drawn)` are the same call. Dropped the conditional rather than wrapping it — a `try/except` around a branch no input can reach is a guard no test can hold up, and removing it fixes the asymmetry too.

**3. `extract_collection`'s bare `next()`.** Confirmed and filed as #529. It reproduces directly:

```python
fig, ax = plt.subplots()
ax.plot([1, 2, 3], [3, 2, 1])          # a line: no PathCollection
CollectionExtractorMixin.extract_collection(ax, PathCollection)
→ StopIteration
```

and `ScatterPlot._extract_plot_data` is written for `None` all the way through (`_extract_point_data` opens with `if plot is None`), so the handling can never run — the same sentence #388 wrote about `extract_container`. Third instance of the pattern, all three in one file. Left out of this PR because its caller is a different chart type and it deserves its own test.

On the `id()` note in your first paragraph — agreed, and thank you for actually checking the lifetime rather than pattern-matching the warning. The snapshot list holds the artists for the whole scope, so the ids in `_drew_mesh` are compared while their objects are alive. I've left `_meshes_of`'s docstring saying why the list holds artists rather than ids, since that is where the reasoning belongs.

Full suite: 2,414 passed, 18 skipped. `ruff` clean on every file touched.

---

Issue #530, opened 2026-08-20
A figure mutated while another session renders it produces an SVG the payload does not describe


jooyoungseo:
Rendering off the event loop (#454 / #504) left a window the per-figure lock does not cover: the lock serialises **renders**, and a mutation is not a render. While one session's `savefig` runs on a worker thread, the loop is free — which is the whole point — so another session's render function can draw into the same figure at the same time.

The result is not an error. It is a chart whose **SVG shows one thing and whose maidr payload announces another**.

## Measured, through the shipped `render_maidr.render()`

Two renderers in one session context over a module-level figure — the pattern `_close_new_figures` documents as supported (*"an app that builds one figure at module scope and returns it on each flush keeps the figure it owns"*). One returns the figure; the other labels it first, as an app would for its own user:

```python
FIG, AX = plt.subplots()
AX.bar([str(i) for i in range(200)], list(range(200)))

@render_maidr(use_cdn=True)
def reader():
    return AX

@render_maidr(use_cdn=True)
def writer():
    AX.set_title("SESSION-A-LABEL")
    return AX

await asyncio.gather(reader.render(), writer.render())
```

Reading `reader`'s own output:

```
B's SVG carries A's label = True      B's payload has it = False
```

3/3 runs. Nothing raised.

## The control says it is the off-loop move that opened the window

Same script with `asyncio.to_thread` replaced by an inline call — the shape `render()` had before #504:

| | B's SVG carries A's label |
|---|---|
| off the loop (shipped) | **True** — 3/3 |
| on the loop (pre-#504) | False |

On the loop the two renders could not overlap, so there was no window. This is a consequence of the fix, not something it missed, and it is the honest cost line to put beside #454's 533 ms → 24 ms.

## Why it matters more here than in a plotting library

The payload is extracted from the artists; the SVG is written by `savefig`. Anything drawn between those two points lands in one and not the other. A sighted user sees the title; a screen-reader user is never told it exists. Every check that reads the payload passes.

The same window covers data, not just decoration. Adding a bar mid-render put the extra `<path>` in the SVG while the payload still described 200 bars, and the count of maidr-tagged elements moved between runs (415 → 413 → 412 across otherwise identical runs) — so which element a given payload index highlights is not stable either. That part is timing-dependent and I am not reporting a fixed number for it; the title case above is the deterministic one.

## Reachability

[...]

PR #531, opened 2026-08-20
fix(streamlit): render one figure at a time, as the Shiny door already does

jooyoungseo:
The per-figure render lock added for Shiny in #504 guards one door. Streamlit renders through the other one, unsynchronised — and Streamlit is if anything the more exposed of the two: it runs **every session's script in its own ScriptRunner thread**, and `@st.cache_resource` is its documented way to share one object across them.

## The failure

`savefig` writes `fig.dpi` for its duration and restores it afterwards, so two renders of the same figure at once race on that one attribute and the loser draws its whole chart at the other call's dpi. Six concurrent `maidr_html()` calls on one figure, five trials, diffed against a serial render with per-render uuids and timestamps normalised away:

```
trial 0: 1/6 differ from a serial render [1]
trial 1: 0/6 ...
total mismatching concurrent renders: 1/30
```

The one that differs is not garbled markup and not an exception. Its background path traces a different rectangle from every other render's:

| | the same element's `d` attribute |
|---|---|
| the other five renders | `M 0 345.6  L 460.8 345.6  L 460.8 0  L 0 0  z` |
| the loser of the race | `M 0 345.6  L 640 345.6  L 640 -134.4  L 0 -134.4  z` |

`640 / (100/72) = 460.8` — a complete, well-formed SVG of the same chart at 72% scale. Geometry is what the highlight overlay and the tactile export are positioned against, so this is wrong in the modality a sighted reviewer checks least — and it only happens under concurrent traffic, so it would not reproduce from a bug report.

## The fix

The lock moves to `maidr/util/figure_lock.py` and both doors take it. **One registry, not one per integration:** two registries would let a Shiny render and a Streamlit render of the same figure overlap, which is precisely the case the lock exists to prevent.

## A second gap, found while moving the resolution

Deciding *which* figure to lock moved too, and two shapes were silently missing:

| value handed to the renderer | old resolution | now |
|---|---|---|
| `ax` | the figure | the figure |
| `fig` | **`None`** | the figure |
| `None` (current figure) | **`None`** | the figure |

A `Figure` resolves to a *list* of axes, and `getattr(list, "figure", None)` is `None` — which `figure_lock` answers with a fresh, unshared lock. So **a Shiny render function returning `fig` rather than `ax` raced exactly as if there were no lock at all**, and `maidr_html()` with no argument would have done the same. Resolution now goes through `maidr.render`'s own `_get_plot_or_current`, because a resolver that disagrees with the renderer takes a lock on a figure the render never touches — synchronised-looking and not synchronised.

Review raised the follow-on: resolving `plt.gcf()` separately for the lock and for the render leaves the same window one level up, since `Gcf` is process-global. `maidr_html` now resolves once and renders *that* figure (`a1c5e9c`).

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,430 passed, 14 skipped**. (`ruff check` reports 6 pre-existing re-export findings in `maidr/core/**/__init__.py` on a clean tree; every file this touches is clean.)

Every guard falsified by mutation rather than assumed:

| mutation | result |
|---|---|
| the Streamlit lock removed (the reported bug) | new test fails **5 of 5** runs |
| lock restored | passes **10 of 10** runs |
| resolver back to the old logic | Shiny concurrency test fails its new `figure` arm **3 of 3**, `axes` arm passes |
| the resolve-once step dropped | new current-figure test fails **3 of 3** |

The third row is the point of parametrising that test: the two arms were genuinely different before this change, and only one of them was covered.

## Tests

- `tests/util/test_figure_lock.py` — the lock's own unit tests move here with it (per-figure, stable, weak-keyed), plus `test_every_way_of_naming_one_chart_takes_the_same_lock` pinning the table above.
- `tests/widget/test_streamlit.py::test_concurrent_renders_of_one_figure_agree` — mirrors the Shiny sibling, barrier-synchronised so it does not depend on the scheduler being busy.
- `tests/widget/test_streamlit.py::test_the_current_figure_is_resolved_once_and_rendered` — moves the current figure *inside* `figure_lock`, which is the interleaving a second thread would produce.
- `tests/widget/test_shiny.py` — the three moved unit tests are gone from here; the tests that say the *render path* reaches the shared lock and resolver stay, with the log assertion repointed at the shared module's logger.

## Scope

This is the concurrent-**render** case, through the two doors this package ships. Two things it deliberately does not do, both filed with their own measurements:

- the window where an app mutates a figure *while* a render is in flight, which no lock keyed on renders can close — #530
- the same race reached through `maidr.render` directly, which is where Flask and any other threaded caller meet it — #532

Refs #454.

### DIFF
--- a/maidr/api.py
+++ b/maidr/api.py
@@ -429,6 +429,13 @@ def _get_plot_or_current(plot: Any | None) -> Any:
     """
     Get the plot object or current matplotlib figure if plot is None.
 
+    Underscored but not private to this module:
+    :func:`maidr.util.figure_lock.resolve_figure` and
+    :func:`maidr.widget.streamlit.maidr_html` both call it, so that the
+    figure a render *locks* is the figure it goes on to write. A change
+    here that made this answer differently would put those locks on a
+    figure the render never touches -- synchronised in appearance only.
+
     Parameters
     ----------
     plot : Any or None
--- a/maidr/util/figure_lock.py
+++ b/maidr/util/figure_lock.py
@@ -153,6 +153,8 @@ def resolve_figure(value: Any) -> Any:
         # matching ``render``'s own loop over the list.
         axes = axes[-1] if axes else None
 
+    # ``axes`` is an ``Axes`` or ``None`` by here: ``get_axes`` returns one
+    # of those or a list, and the list is unwrapped above.
     return getattr(axes, "figure", None)
 
 
--- a/maidr/util/figure_lock.py
+++ b/maidr/util/figure_lock.py
@@ -143,10 +143,11 @@ def resolve_figure(value: Any) -> Any:
         return None
 
     if isinstance(axes, list):
-        # A ``Figure`` resolves to its list of axes, and so does a list of
-        # artists. They share one figure in every ordinary case; where they
-        # do not, ``render`` writes the last one's, so that is the one to
-        # hold.
+        # Only a ``Figure`` lands here: ``get_axes`` returns its ``.axes``
+        # property, while a list of artists takes a different branch and
+        # comes back as a single ``Axes``. A figure's axes all share that
+        # figure, so which one is picked does not matter -- the last,
+        # matching ``render``'s own loop over the list.
         axes = axes[-1] if axes else None
 
     return getattr(axes, "figure", None)
--- a/maidr/util/figure_lock.py
+++ b/maidr/util/figure_lock.py
@@ -1,0 +1,155 @@
+"""One lock per matplotlib ``Figure``, shared by every threaded entry point.
+
+Not process-wide: ``savefig`` on distinct figures is safe in parallel, and
+a single lock would serialise unrelated sessions and throw away most of
+what threading buys.
+
+Why a lock is needed at all: ``savefig`` mutates the figure it is writing,
+for the duration of the write. Two things, both measured by watching the
+attribute from another thread while renders ran:
+
+* ``fig.dpi`` goes 100 -> 72 -> 100, so two concurrent writes race on that
+  one attribute and the loser renders its whole chart at the other's dpi.
+  A 640x480 chart came out 460.8x345.6 -- exactly the 100/72 ratio -- as a
+  valid SVG, raising nothing, on 1 of 6 concurrent renders (#454).
+* ``fig.canvas`` is swapped to a canvas that supports the output format and
+  swapped back (``FigureCanvasAgg`` -> ``FigureCanvasSVG`` ->
+  ``FigureCanvasAgg``), by
+  ``FigureCanvasBase._switch_canvas_and_return_print_method``.
+
+That second one also answers a question this raises: a render may run on a
+worker thread, and GUI backends generally want canvas work on the main
+thread. It does not reach the GUI canvas -- the write goes through the
+format's own canvas, which for SVG is pure Python and has no thread
+affinity. maidr's own backend delegates to ``FigureCanvasAgg`` besides,
+and is what ``import maidr`` activates.
+
+The registry lives here rather than in one integration because more than
+one door renders on threads. Shiny renders off the event loop through
+``asyncio.to_thread``; Streamlit runs every session's script in its own
+ScriptRunner thread. Two doors with two registries would let a Shiny
+render and a Streamlit render of the *same* figure overlap, which is the
+case the lock exists to prevent -- so there is one registry, keyed by
+figure, for the process.
+"""
+
+from __future__ import annotations
+
+import logging
+import threading
+import weakref
+from typing import Any
+
+_logger = logging.getLogger(__name__)
+
+#: The locks themselves, keyed weakly so a closed figure's lock goes with
+#: it. The lock does not reference the figure, so this adds no retention
+#: (#498).
+#:
+#: ``threading.Lock``, not ``RLock``: nothing re-enters a render of the
+#: same figure on the same thread today, and if something ever does, a
+#: deadlock is a better outcome than two interleaved writes to one figure,
+#: because it is the one that shows up.
+_FIGURE_LOCKS: weakref.WeakKeyDictionary[Any, threading.Lock] = (
+    weakref.WeakKeyDictionary()
+)
+
+#: Guards creation of the per-figure locks above, which is itself a
+#: check-then-act on a shared mapping.
+_FIGURE_LOCKS_GUARD = threading.Lock()
+
+
+def figure_lock(figure: Any) -> threading.Lock:
+    """Return the lock for ``figure``, creating it on first use.
+
+    Parameters
+    ----------
+    figure : Any
+        The ``matplotlib`` figure about to be rendered, or ``None`` when it
+        could not be resolved.
+
+    Returns
+    -------
+    threading.Lock
+        A lock unique to that figure. An unresolvable figure gets a fresh
+        lock rather than a shared one -- serialising things we cannot tell
+        apart would be a guess in the direction of a deadlock, and the
+        render is safe on its own.
+
+        Note what that means: an unresolvable value gets **no**
+        synchronisation at all. Correct today because the values that land
+        here -- plotly, altair -- are rendered without touching a
+        ``matplotlib`` figure's ``dpi`` or ``canvas``, so there is no
+        shared state to race on. A future plot type that resolves to
+        ``None`` here *and* mutates shared figure state would be
+        unprotected silently, which is the reason to say so rather than
+        leave it to be inferred.
+    """
+    if figure is None:
+        return threading.Lock()
+    with _FIGURE_LOCKS_GUARD:
+        lock = _FIGURE_LOCKS.get(figure)
+        if lock is None:
+            lock = threading.Lock()
+            _FIGURE_LOCKS[figure] = lock
+        return lock
+
+
+def resolve_figure(value: Any) -> Any:
+    """Return the ``matplotlib`` figure ``value`` belongs to, or ``None``.
+
+    Resolves the way :func:`maidr.render` does, deliberately: ``None``
+    means the current figure there, and a value naming several axes is
+    rendered as the last one's figure. A resolver that disagreed with the
+    renderer would take a lock on a figure the render never touches, which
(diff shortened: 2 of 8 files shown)


jooyoungseo:
Both notes were worth acting on, and one of them was a real (if narrow) gap I introduced. Pushed as `a1c5e9c`.

## 1. `resolve_figure(None)` and `plt.gcf()` — fixed rather than bucketed

You asked whether this belongs in #530's bucket. It does not, and thank you for not letting me file it there: **#530 is about a figure being mutated during a render; this is about locking the wrong figure entirely**, which is worse, because the render is then unsynchronised while looking synchronised.

The side effect itself is benign — `plt.gcf()` creating a figure where none exists just moves a creation `render` would have done a few lines later, and both calls land on the same object. The divergence you describe is the real part: `Gcf` is process-global, so another thread's `plt.figure()` between the two resolutions leaves the lock on one figure and the write on another.

`maidr_html` now resolves once and renders *that* figure:

```python
if plot is None:
    current = resolve_figure(None)
    if current is not None:
        plot = current
```

An empty current figure resolves to no axes and is left as `None`, so `render` rejects it on its own account exactly as before.

Pinned by `test_the_current_figure_is_resolved_once_and_rendered`, which does not simulate a thread — it moves the current figure *inside* `figure_lock`, which is the interleaving a second thread would produce, and then asserts both that the locked figure was the original and that the rendered chart is the original's data. Falsified by mutation: removing the resolve-once step fails it 3 of 3 runs.

## 2. The list-branch comment — corrected

You are right that it was imprecise, and the imprecision was mine rather than a shorthand: I wrote that a list of artists also resolves to a list, when `get_axes` sends it through `next(...)` and returns a single `Axes`. Only a `Figure` reaches that branch. The comment now says so, and says why `[-1]` is arbitrary-but-fine (a figure's axes all share that figure) rather than implying it is choosing between candidates.

## 3. `figure_lock(None)` handing back unshared locks

Agreed, and deliberately left as is. Two values that resolve to nothing are not known to be the same thing, so sharing a lock between them would be a guess in the direction of a deadlock — and today the values that land there (plotly, altair) touch no matplotlib figure state. The docstring says both halves of that, including that a future matplotlib-backed path resolving to `None` would be silently unprotected. If one appears, the fix is to make it resolve, not to widen the lock.

## Verification after the change

`pytest 0 / ruff 0 / uv lock --check 0` — **2,430 passed, 14 skipped**. The earlier mutation results still hold: Streamlit lock removed fails 5 of 5, restored passes 10 of 10, resolver reverted fails the Shiny `figure` arm 3 of 3.

---

jooyoungseo:
The residual TOCTOU is real and is now closed rather than documented — you traced it correctly. Pushed as `771b039`.

## The axes-less case

You are right about the mechanism: `resolve_figure(None)` answers *which figure to lock*, and it returns `None` for two different situations — no current figure, and a current figure that has no axes yet. Substituting only when the answer was non-`None` left the second `plt.gcf()` in place for the second situation. Narrowed, not closed, and the commit message claiming "resolve once" would have been wrong for it.

Rather than comment the gap, `maidr_html` now substitutes through the same helper `render` itself calls:

```python
plot = _get_plot_or_current(plot)
```

That is the identity for anything not `None`, so the current figure is resolved exactly once whatever it holds. An axes-less figure now reaches `render` as a `Figure` rather than as `None`, and `render` rejects it the same way it always did — `resolve_figure` still answers `None` for it, so it takes an unshared lock, which is correct: there is nothing to serialise on a figure with no chart in it.

Pinned by `test_an_axes_less_current_figure_is_still_resolved_only_once`, which **counts resolutions rather than racing them** — a race test here would be measuring the scheduler. Restoring the version you flagged fails it 3 of 3 runs.

## The `get_axes` branch-order coupling

Agreed and added. The comment previously read as though the distinction came from matplotlib's class hierarchy; the actual dependency is that `get_axes` tries `Artist` before anything `Figure`-specific, so a new branch ahead of it would change which shape arrives. That is now stated where the branch is, so a future reordering has something to trip over.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,437 passed, 14 skipped**, rebased onto `fe5a46b` (#525) and re-run there rather than only against the older base.

Mutation results now cover four guards:

| mutation | result |
|---|---|
| Streamlit lock removed | fails **5 of 5** |
| lock restored | passes **10 of 10** |
| resolver back to the old logic | Shiny `figure` arm fails **3 of 3**, `axes` arm passes |
| resolve-once dropped entirely | fails **3 of 3** |
| resolve-once narrowed to the non-`None` case (this round) | fails **3 of 3** |

---

jooyoungseo:
Note 3 was the one with teeth, and it was right. Pushed as `0479bbb`.

## The cross-door guarantee is now asserted rather than inferred

You put it exactly: the PR's stated motivation — one registry, so a Shiny render and a Streamlit render of one figure cannot overlap — rested on both doors importing the same module-level dict, which is a fact about Python's import semantics and not about this code staying correct.

`test_the_two_doors_exclude_each_other_on_one_figure` now drives both doors at one figure from a barrier and asserts the renders do not overlap. It asserts *exclusion* rather than output equality, since the two doors wrap the chart differently and their markup is not comparable; the consequence of failing to exclude is already covered per door by the two `test_concurrent_renders_of_one_figure_agree` tests.

**The mutation is the part worth reporting.** I gave `streamlit.py` a registry of its own — the exact regression you describe:

| | result |
|---|---|
| `test_shiny.py` + `test_streamlit.py` (72 tests) | **all pass** |
| the new cross-door test | **fails 3 of 3** |

So the gap you identified was real and total: every per-door test would have stayed green through it.

## The other three

**1. `_get_plot_or_current` as a cross-module private.** Fair, and it is now three callers rather than one. Not doing it here: it is a separate logical change to `api.py`'s surface, and there is precedent in-tree for a deferred cross-module private import (`core/maidr.py` reaches for `maidr.api.init_notebook` the same way). Worth promoting to a named helper in `maidr/util/` when someone touches that surface — happy to do it as its own PR if you want it, rather than bundling it into a concurrency fix.

**2. `get_axes` called twice.** Agreed and accepted, for the reason you give. Worth being explicit about why it is not worth restructuring: the alternative is threading the resolved axes from the lock site into `render`, which couples the doors to `render`'s internals to save a few `isinstance` checks. The redundancy is the cheaper half of that trade.

**3. `_FIGURE_LOCKS_GUARD` as a process-wide guard.** Right to flag the scope change. The critical section is a dict `get` plus a `set`, and it is held only on *creation* — a steady-state app with a figure per session touches it once per figure, not once per render. If it ever became hot the fix is a striped or lock-free registry rather than removing the guard, since the check-then-act is what it is there for.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,438 passed, 14 skipped**, on `fe5a46b` (#525).

---

jooyoungseo:
Both actionable notes taken, `bdb678d`.

**The `_get_plot_or_current` comment.** You framed this better than I did in my last reply: the issue is not that a private symbol is imported, it is that a future edit at its *definition* has no way to know two lock sites depend on its answer. Its docstring now says so, and says what breaks — locks landing on a figure the render never touches, which is the exact failure mode this series is about. That is where a maintainer will actually be standing when they change it, so it beats a note in `figure_lock.py`.

**`monkeypatch.setattr` for the render stub.** Switched. The hand-rolled save/restore worked, but a `finally` that has to be right is worse than a fixture that cannot be wrong — and inconsistency inside one PR is its own small cost. Re-ran the registry-split mutation afterwards to be sure the swap did not weaken the test: still fails **3 of 3**.

**`figure_lock(None)`.** Noted, no change. Nothing on the matplotlib path resolves to `None` today — the two shapes that used to (`Figure`, `None`) are what this PR fixed, which is the second time that particular hole has been found by walking into it. If a future plot type resolves to `None` *and* touches shared figure state, the fix is to make it resolve rather than to widen the lock.

## Final state

`pytest 0 / ruff 0 / uv lock --check 0` — **2,438 passed, 14 skipped**, on `fe5a46b`.

Five mutations, each falsifying a specific guard:

| mutation | result |
|---|---|
| Streamlit lock removed | fails 5 of 5 |
| lock restored | passes 10 of 10 |
| resolver back to the old logic | Shiny `figure` arm fails 3 of 3, `axes` arm passes |
| resolve-once dropped, then narrowed to the non-`None` case | fails 3 of 3 each |
| registry split per door | 72 per-door tests still pass, cross-door test fails 3 of 3 |

Merging once CI is green on this commit.

---

jooyoungseo:
Both points taken, `725cfc2`. Merging on green — the last three rounds have converged on nits and each has been applied.

## The short-circuit you asked me to sanity-check

Your read is correct: `render()` returns on `value is None` before `_render_off_loop`, so a `None` never reaches the resolution, and nothing calls `plt.gcf()` for a blank output. You also named the right worry — that short-circuit is now load-bearing for a **second** reason, and only the first was pinned. `test_none_renders_nothing` asserted the payload was `None`, which stays true whether or not a stray figure gets opened on the way.

It now asserts the figure count too. Falsified by mutation: removing the short-circuit fails it **3 of 3** runs, where previously it would have sailed through.

That is the better answer than a comment, because the thing at risk is not a reader's understanding but a refactor.

## The `getattr` precondition

Added. The file front-loads its rationale into the module docstring and the branch comments, and that last line was the one place a reader had to reconstruct the state themselves — `axes` is an `Axes` or `None` by then, the list having been unwrapped above.

## On not running the suite

Worth saying plainly since you flagged the constraint: the numbers in this thread are from local runs, and CI has run the same suite on every commit — currently green across all 20 checks on `bdb678d`, including the browser accessibility job. The mutation results are the part CI cannot give you, which is why they are in the description rather than just "tests pass".

## Final

`pytest 0 / ruff 0 / uv lock --check 0` — **2,438 passed, 14 skipped**, on `fe5a46b`.

Six mutations across the series, each falsifying one guard: Streamlit lock removed (5/5), resolver reverted (Shiny `figure` arm 3/3), resolve-once dropped (3/3), resolve-once narrowed (3/3), registry split per door (3/3, while all 72 per-door tests still passed), and now the `None` short-circuit removed (3/3).

---

PR #533, opened 2026-08-20
feat(core): read Axes.broken_barh as the gantt chart it draws

### LINKED ISSUE #524: Coverage sweep: `broken_barh`, `contour` and `stairs` are unread, and each already has a trace waiting for it
Ran the matplotlib and seaborn catalogues through `FigureManager.get_maidr` to find what is still silent. seaborn 0.13.2, matplotlib 3.9.4, 32 calls.

Most of #345's checkboxes are stale — `stripplot`, `swarmplot`, `residplot`, `ecdfplot`, `lmplot`, `pairplot`, `jointplot`, `catplot`, `relplot` and `displot` all read now, through the axes-level patches they route into. Nine calls do not, and they fall into three groups.

## Readable, and the trace already exists

Each of these is exactly recoverable from the artist — no pixel inversion, no colour to invert, no inference.

### `ax.broken_barh` → `GANTT`

This is matplotlib's gantt chart, and MAIDR has had a gantt trace since xability/maidr#801.

```python
ax.broken_barh([(1, 3), (6, 2)], (10, 4))
ax.broken_barh([(2, 5)], (20, 4))
```

draws one `PolyCollection` per call, one path per interval, and the vertices are the numbers:

```
PolyCollection #1   x: 1.0 -> 4.0   y: 10.0 -> 14.0
                    x: 6.0 -> 8.0   y: 10.0 -> 14.0
PolyCollection #2   x: 2.0 -> 7.0   y: 20.0 -> 24.0
```

`GanttPoint` claims `{lane, start, end}`, which is exactly what is drawn: the `yrange` is the lane and each `xrange` is an interval in it. A collection per lane also nests the way `GanttData` wants without any grouping heuristic.

### `ax.contour` / `ax.contourf` / bivariate `sns.kdeplot` → `CONTOUR`

MAIDR has had a contour trace since xability/maidr#802, and matplotlib hands the level values over directly:

```python
cs = ax.contour(z)
cs.levels    # [0.0, 2.0, 4.0, 6.0, 8.0, 10.0, 12.0, 14.0]
cs.allsegs   # the polylines, per level
```

Worth stating plainly, because it is what blocks the same chart elsewhere: **the level is a number here, not a colour.** In Observable Plot a contour keeps its magnitude only in a continuous fill, which is why xability/maidr#1084 left `contour` and `density` unread there. `QuadContourSet` has no such problem — `.levels` is the data.

`sns.kdeplot(df, x=..., y=...)` is the same artist with computed levels (`[0.0059, 0.0234, 0.0406, …]`), so it comes along with it. That also finishes the other half of `sns.jointplot(kind='kde')`, which currently returns `['smooth', 'smooth']` — the two marginals, and nothing for the joint panel the chart is drawn for.

### `ax.stairs` → `HISTOGRAM` (or `STEP`)

`ax.stairs` is how modern matplotlib draws a **pre-binned** histogram, and `StepPatch` keeps both halves exactly:

```python
sp = ax.stairs([1, 3, 2], [0, 1, 2, 3])
sp.get_data()   # StairData(values=[1, 3, 2], edges=[0, 1, 2, 3], baseline=0)
```

Counts and bin edges, unrounded, straight off the artist. Contrast xability/maidr#1055, where a Google Charts histogram had to be declined because its bin edges existed only as inverted pixels — here there is nothing to invert. `ax.step` already reads as `step`, so this is also a parity gap between two spellings of a staircase.

## Correctly declined, and worth recording as such

- **`ax.quiver` and `ax.barbs`.** A vector field's payload is the *direction*, and there is no field in the grammar for a bearing. Reading only the magnitudes would announce half a chart without saying which half. This is the same call, on the same reasoning, that xability/maidr#1098 made for `Plot.vector` with a `rotate` channel.
- **`sns.rugplot` and `ax.eventplot`.** Both draw one tick per observation along a single axis — a `LineCollection` of 200 two-point segments for the rug, an `EventCollection` with exact `get_positions()` for the event plot. The numbers are recoverable; what is missing is the second field. A one-dimensional distribution has no category to announce against, which is the reason xability/maidr#1069 gives for declining a `Plot.tickX` with no cross-channel. A rug drawn *beside* another plot is a different question and a harder one, since it would be a composite.

## Not this issue

`sns.histplot(x=, y=)` is silent too, but for an unrelated reason — the recursion guard suppresses a registration the outer patch then declines. Filed as #522, because it is a defect in something that already works rather than a plot type nobody has read yet.

## Suggested order

`broken_barh` first: it is a whole chart type that matplotlib users reach for by name, the data needs no interpretation, and the trace is already in the runtime. Then `stairs`, which is small and closes a parity gap. Then the contour family, which is the largest of the three because it introduces a plot type py-maidr has never emitted.

jooyoungseo:
Closes #524 (the first of the three it names).

`Axes.broken_barh` is matplotlib's gantt chart. It was read as nothing, though the runtime has had a gantt trace since xability/maidr#801 and the pinned bundle already knows it — `bundle_trace_types()` lists `gantt` among its 60.

## Why it fits without inference

The call is already shaped like the trace. **One call is one lane**: the `yrange` places it, each `(start, width)` in `xranges` is an interval in it. So `{lane, start, end}` is what the caller wrote, not something recovered from a drawing.

The numbers come off the artist exactly:

```python
ax.broken_barh([(110, 30), (150, 10)], (10, 9))
ax.broken_barh([(10, 50), (100, 20), (130, 10)], (20, 9))
ax.set_yticks([15, 25], labels=["Bill", "Jim"])
```

```json
{"points": [[{"x": "Bill", "start": 110.0, "end": 140.0},
             {"x": "Bill", "start": 150.0, "end": 160.0}],
            [{"x": "Jim", "start": 10.0, "end": 60.0},
             {"x": "Jim", "start": 100.0, "end": 120.0},
             {"x": "Jim", "start": 130.0, "end": 140.0}]],
 "lanes": ["Bill", "Jim"]}
```

Nothing inverted through a scale, nothing rounded — the `PolyCollection`'s vertices *are* the corners.

## Two decisions, both measured

**A chart is several calls and is still one chart.** My first working version registered a layer per call, and a two-lane schedule came out as two one-lane charts. That defeats the point: `GanttData.points` is nested by lane so one layer holds the whole schedule, and up/down are how a reader moves *between* lanes. A later call on the same axes now becomes another lane of the layer already there.

**A lane's name is not its position, and the tick is not at its centre.** The chart matplotlib's own documentation shows names its lanes *after* drawing them, which works only because extraction runs when the schema is first asked for rather than when the call was patched — verified. But an exact match finds nothing:

| bar | spans | centre | tick |
|---|---|---|---|
| Bill | 10–19 | 14.5 | **15** |
| Jim | 20–29 | 24.5 | **25** |

So the tick that names a lane is the one *inside* it. And only an explicit tick counts, which took two measurements to get right:

- left alone, matplotlib's `AutoLocator` puts **several** ticks inside every bar — an unlabelled version of that chart offers "8", "10", "12", "14", "16" for the lane spanning 10–19;
- but "several" is not a safe test on its own. Widen the axis and exactly one automatic tick lands inside — measured, a bar spanning 10–14 under `ylim=(0, 50)` has the single tick "10" in it, and taken as a name the lane would be called "10", which is the axis reading itself out.

`set_yticks` installs a `FixedLocator`; matplotlib's own choice is an `AutoLocator`. That is the difference between a name and a coordinate. A lane with no single explicit tick of its own is named by its position, which is always true and never a guess.

## Verification

`pytest 0 / ruff 0` — 2,431 passed, 18 skipped (up from 2,414; the 6 `ruff` findings across `maidr/` + `tests/` are the pre-existing `__init__.py` re-export warnings, and every file this adds or touches is clean).

| mutation | result |
|---|---|
| lanes not merged — a layer per call | 6 failed |
| a lane always named by its position | 3 failed |
| the interval's end reads as its start | 1 failed |
| the `FixedLocator` check dropped | 1 failed |
| "the first tick inside" instead of "exactly one" | 1 failed |

The last two were **dead on the first pass** and are the reason two more fixtures exist. With only the documentation chart and its unlabelled twin, "exactly one tick inside" already caught everything the locator check was there for, and "the first" was indistinguishable from "exactly one". They are now pinned by the `ylim=(0, 50)` chart above and by a lane holding two labelled ticks, which is named by neither.

## Not in scope

The other two items #524 names — `ax.stairs` and the `contour` family — are separate readings and want their own PRs. `ax.quiver`, `ax.barbs`, `sns.rugplot` and `ax.eventplot` stay declined, on the reasoning already established in xability/maidr#1098 and #1069.

---

### DIFF
--- a/maidr/core/enum/maidr_key.py
+++ b/maidr/core/enum/maidr_key.py
@@ -58,6 +58,13 @@ class MaidrKey(str, Enum):
     # Step plot keys. The per-point ordinal level name reuses LABEL above.
     STEP_DIRECTION = "stepDirection"
 
+    # Gantt keys. An interval carries the two ends of its span; its lane is
+    # `X`, and the lane names live in `LANES` so that a lane holding nothing
+    # still has somewhere to carry one.
+    START = "start"
+    END = "end"
+    LANES = "lanes"
+
     # Histogram plot keys.
     X_MIN = "xMin"
     X_MAX = "xMax"
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -15,6 +15,11 @@ class PlotType(str, Enum):
     COUNT = "count"
     DODGED = "dodged_bar"
     ERRORBAR = "error_bar"
+    #: A schedule of intervals in lanes. `Axes.broken_barh` is matplotlib's
+    #: gantt chart: one call per lane, drawing that lane's intervals as a
+    #: `PolyCollection`, so the shape the trace wants -- a lane and the two
+    #: ends of each interval in it -- is what the call was given.
+    GANTT = "gantt"
     HEAT = "heat"
     #: A hexagonal bin lattice: the standard answer to an overplotted scatter.
     #: Read as a grid of cells each carrying a count, which is a heatmap --
--- a/maidr/core/plot/gantt.py
+++ b/maidr/core/plot/gantt.py
@@ -1,0 +1,203 @@
+from __future__ import annotations
+
+import uuid
+
+import numpy as np
+from matplotlib.axes import Axes
+from matplotlib.collections import Collection
+from matplotlib.ticker import FixedLocator
+
+from maidr.core.enum import MaidrKey, PlotType
+from maidr.core.plot import MaidrPlot
+from maidr.exception import ExtractionError
+
+
+class GanttPlot(MaidrPlot):
+    """
+    A schedule of intervals in lanes, as ``Axes.broken_barh`` draws it.
+
+    ``broken_barh(xranges, yrange)`` is matplotlib's gantt chart, and it is
+    shaped like one already: **one call is one lane**. The `yrange` says where
+    that lane sits and each `(start, width)` in `xranges` is an interval in it,
+    so the trace's `{lane, start, end}` is what the caller wrote rather than
+    something inferred from a drawing.
+
+    The numbers come off the artist exactly. A `PolyCollection` holds one path
+    per interval and each path's vertices are the corners, so a bar written
+    `(1, 3)` reads back as 1 to 4 with nothing to invert and nothing to round.
+
+    Naming the lane
+    ---------------
+    The `yrange` is a position, not a name, and the chart the matplotlib
+    documentation shows names its lanes afterwards::
+
+        ax.broken_barh([(110, 30), (150, 10)], (10, 9))
+        ax.broken_barh([(10, 50), (100, 20)], (20, 9))
+        ax.set_yticks([15, 25], labels=["Bill", "Jim"])
+
+    Two things make that readable. Extraction runs when the schema is first
+    asked for rather than when the call was patched, so ticks set *after* the
+    bars are in place by then. And the tick that names a lane is the one
+    **inside** it: measured, the example above puts its ticks at 15 and 25
+    while the bars span 10-19 and 20-29, so their centres are 14.5 and 24.5
+    and an exact match finds nothing.
+
+    Only a ``FixedLocator`` counts, which is what tells a name from an axis.
+    ``set_yticks`` installs one; left alone, matplotlib uses an ``AutoLocator``
+    and puts several ticks inside every bar -- measured, an unlabelled chart
+    offers "8", "10", "12", "14" and "16" for a lane spanning 10 to 19, none of
+    which is that lane's name. A lane with no single tick of its own is named
+    by its position, which is always true and never a guess.
+    """
+
+    def __init__(self, ax: Axes, **kwargs) -> None:
+        self._collections: list[Collection] = list(kwargs.pop("collections", ()) or ())
+        super().__init__(ax, PlotType.GANTT)
+
+    def add_lane(self, collection: Collection) -> None:
+        """
+        Take another ``broken_barh`` call on the same axes as another lane.
+
+        A gantt chart is drawn one call per lane, so a two-lane schedule is
+        two calls. Registering each as its own layer would give a reader two
+        one-lane charts to switch between rather than one chart to move up and
+        down inside, which is the whole of what the trace is for -- `points`
+        is nested by lane precisely so one layer holds them all.
+
+        The schema is dropped rather than amended, because it is built lazily
+        on first access and a lane arriving after it was built would otherwise
+        never appear.
+
+        Parameters
+        ----------
+        collection : Collection
+            The `PolyCollection` the new call drew.
+        """
+        self._collections.append(collection)
+        self._schema = {}
+
+    def _extract_plot_data(self) -> dict:
+        """
+        Read each collection as one lane of intervals.
+
+        Returns
+        -------
+        dict
+            ``{"points": [[{"x", "start", "end"}, ...], ...], "lanes": [...]}``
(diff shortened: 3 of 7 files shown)


jooyoungseo:
The lock finding is right and it is a real defect, not a style point. Fixed in a811d6c. The selector question was the right one to ask and the answer is "it works" — but I checked rather than assumed, and it is worth writing down why.

## `_lane_of` reading `figs` unlocked

Correct on both counts, including the part that is easy to wave away. The manager's docstring is explicit that a *read* is not safe either:

> Since #456 a lookup also updates the bookkeeping behind iteration, so `in` and `get` mutate shared state rather than only observing it.

Routed through `FigureManager.get_maidr()` instead, which takes `_lock` itself, catching the `UnsupportedPlotError` it raises for an unregistered figure as the ordinary first call.

And you were right that the read was only half of it. The check-then-act is now taken under a lock belonging to this module. `FigureManager._lock` cannot be borrowed: it is a plain `Lock`, not an `RLock`, and `create_maidr` takes it — so holding it across the decision would deadlock rather than race, which is worth stating since it is the obvious first thing to reach for.

**On testing it:** I wrote the eight-thread version first and it does not falsify — it passes just as readily with the `with` removed, three runs out of three. The window is a few bytecodes wide and the GIL rarely lands inside it, so racing it is a smoke test wearing a guard's clothes. What *can* be asserted exactly is the invariant the lock exists for, so the test now records `_lanes.locked()` at the moment the decision is taken and asserts it both times. That reddens when the `with` goes.

## Selectors per lane rather than per point

Checked against the JS trace and against the rendered SVG, because you are right that assuming would be how a highlight-only bug ships.

`GanttTrace.mapToSvgElements` (`src/model/gantt.ts`) **flattens first and re-nests by lane length**:

```ts
const flat = typeof selectors === 'string'
  ? Svg.selectAllElements(selectors)
  : (selectors as string[]).flatMap(one => Svg.selectAllElements(one));

if (flat.length !== this.lanes.flat().length) {
  return null;
}

let taken = 0;
return this.lanes.map(lane => flat.slice(taken, taken += lane.length));
```

So what it needs is not one selector per interval but *one element per interval after flattening*, in lane order. A per-lane selector that resolves to that lane's intervals gives exactly that — and measured on the rendered SVG for the documentation chart, it does:

```
maidr-cffcfa44…  ->  2 direct <path> children     (Bill: 2 intervals)
maidr-7f811e44…  ->  3 direct <path> children     (Jim: 3 intervals)
```

Five elements, five intervals, sliced back into 2 and 3. Note the trace withdraws highlighting entirely on a count mismatch rather than highlighting the wrong task, so a slip here would fail safe — but it does not slip.

## The other two

- **`add_lane` invalidating a schema already read.** Right, and the ordering assumption is now in the docstring. It is real but narrow: the schema is built lazily and read at HTML-export time, by which point every `broken_barh` call has run. A caller who rendered mid-plot and then drew another lane would get a rebuilt schema on next access — which is the correct content, and the `id` is re-stamped by `set_id` on the same path that assigns it.
- **Docstring density.** Fair, and deliberate here: `HexbinPlot` next door carries a class docstring of much the same length and for the same reason — the non-obvious decisions are the expensive part to re-derive. I would rather it match its neighbour than the average.

`pytest 0 / ruff 0` — 2,432 passed, 18 skipped.

---

Issue #427, opened 2026-08-16
sns.ecdfplot emits -Infinity, which is not valid JSON, so the chart never initialises


jooyoungseo:
## Problem

`sns.ecdfplot()` registers as a `STEP` layer, and its first point carries `x = -inf` — seaborn starts the ECDF's staircase at negative infinity so the first step has somewhere to come from.

`json.dumps` writes that as the bare token `-Infinity`, which is **not** valid JSON. It reaches the rendered SVG:

```
maidr="… &quot;data&quot;: [ [ { &quot;x&quot;: -Infinity, …"
```

and `src/index.tsx:113` in the core parses that attribute:

```ts
const maidr = JSON.parse(json) as Maidr;
```

Verified in node:

```
JSON.parse rejects -Infinity: No number after minus sign in JSON at position 7
```

The `catch` around it logs to the console and returns, so `initMaidrOnElement` is never called.

## Impact

This is not a degraded reading — it is a chart that does not load. Audio, text, braille and highlight are all absent, and the only trace is a `console.error` a screen reader user has no reason to be looking at.

## Measured

```
ECDF first/last x: -inf 2.002392583645255
non-finite x count: 1
bare -Infinity in rendered output: yes
```

## Scope

The immediate cause is specific to the ECDF's leading point, but the general shape is that a non-finite value anywhere in extracted data produces a payload the core cannot parse. `NaN` and `Infinity` serialise the same way and `JSON.parse` rejects all three. Worth fixing where the schema is built rather than in the step extractor alone, so any future extractor that meets a masked or infinite value degrades to a missing point instead of a dead chart.

## Note

`maidr/core/maidr.py:388` embeds the same schema as `var maidr = {…}` in a `<script>` tag, where `-Infinity` **is** a legal JavaScript literal. So whether this reproduces depends on which embedding path is taken — the SVG-attribute path (`embed_data`) is the one that breaks.

PR #536, opened 2026-08-20
feat(core): read Axes.stairs as the pre-binned histogram it draws

### LINKED ISSUE #535: `ax.stairs` is silent, though it is matplotlib's own pre-binned histogram and `ax.step` and `ax.hist` both read
The second of the three readings #524 recorded; the first landed in #533. Measured on matplotlib 3.9.4.

## The gap

Three ways to draw a staircase, one of them silent:

| call | layers registered |
|---|---|
| `ax.step([0, 1, 2, 3], [1, 3, 2, 2], where="post")` | `['step']` |
| `ax.hist([0.5, 1.5, 1.5, 2.5], bins=[0, 1, 2, 3])` | `['hist']` |
| `ax.stairs([1, 3, 2], [0, 1, 2, 3])` | **`UnsupportedPlotError`** |

`ax.stairs` is the spelling matplotlib's own documentation reaches for once the binning has already happened:

```python
counts, edges = np.histogram(values, bins=8)
ax.stairs(counts, edges)
```

That chart is a static picture today.

## Nothing has to be inferred

`StepPatch` keeps both halves exactly:

```python
ax.stairs([1, 3, 2], [0, 1, 2, 3]).get_data()
# StairData(values=array([1, 3, 2]), edges=array([0, 1, 2, 3]), baseline=array(0))
```

Counts and bin edges, unrounded, straight off the artist — which is the same pair `HistPlot` recovers from its bars, so the two spellings can emit the *same* payload rather than two that resemble each other. Contrast xability/maidr#1055, where a Google Charts histogram had to be declined because its bin edges existed only as inverted pixels.

`baseline` does not complicate it. It takes `None`, a scalar or a per-bin array, and `get_data()` returns `values` unchanged in every case — it is a drawing choice, not a reading. Nor does it reach anything announced: the core's `Histogram` trace takes its bin range from `xMin`/`xMax` on a vertical chart and from `yMin`/`yMax` on a horizontal one, and never reads the pair on the other axis.

## Why `HistPlot` cannot just be pointed at it

`ax.stairs` leaves no `BarContainer` — it adds a single `StepPatch` to `ax.patches`, and `HistPlot._extract_plot_data` looks a container up. Two `stairs` calls leave two patches side by side, so the artist has to be handed over by the patch rather than searched for on the axes.

## The limit, stated up front

A staircase is **one** `<path>` for the whole chart where `ax.hist` is one per bar. Measured on the same three-bin chart:

| call | `<path>` elements under `axes_1` for the data |
|---|---|
| `ax.hist` | 3 — one per bar |
| `ax.stairs` | 1 — the whole outline, `fill=True` and `baseline=None` alike |

So the chart can announce every bin and highlight none of them. A selector matching the one path would be worse than no selector: it would outline the entire staircase identically at every bin, telling a low-vision reader nothing about where they are. This is the blind spot xability/maidr#814 names, and the same trade already made in xability/r-maidr#193 for `geom_raster` — a blind reader gains every number, nobody loses a highlight they had, and `ax.hist` draws the same histogram with a patch per bin for anyone who needs both.

## Found alongside, not this issue

**`sns.histplot(element="step")` and `element="poly"` are silent too**, and not through `ax.stairs`:

```
element=bars   patches=[Rectangle × 5]  containers=[BarContainer]  → ['hist']
element=step   patches=[]  collections=[PolyCollection]            → nothing
element=poly   patches=[]  collections=[PolyCollection]            → nothing
```

Seaborn draws those through a fill rather than through bars, so `sns_hist` reaches `_drew_bars` → False and declines — the third branch of the same decline #522 fixed for the bivariate mesh. It needs its own reading (a `PolyCollection` whose vertices are a staircase) and its own issue.

## Remaining from #524

`ax.contour` / `ax.contourf` / bivariate `sns.kdeplot` → `CONTOUR`. The largest of the three, since it introduces a plot type py-maidr has never emitted.

jooyoungseo:
Closes #535 — the second of the three readings #524 recorded, after #533.

Three ways to draw a staircase, one of them silent:

| call | before |
|---|---|
| `ax.step([0, 1, 2, 3], [1, 3, 2, 2], where="post")` | `['step']` |
| `ax.hist([0.5, 1.5, 1.5, 2.5], bins=[0, 1, 2, 3])` | `['hist']` |
| `ax.stairs([1, 3, 2], [0, 1, 2, 3])` | **`UnsupportedPlotError`** |

`ax.stairs` is the spelling matplotlib's documentation reaches for once `np.histogram` has already done the binning, and that chart was a static picture.

## Nothing inferred

`StepPatch.get_data()` returns `StairData(values=[1, 3, 2], edges=[0, 1, 2, 3], baseline=0)` — counts and bin edges, unrounded, straight off the artist. Nothing inverted from a pixel, nothing reconstructed from bar spacing, so author-set uneven thresholds survive exactly.

`baseline` does not complicate it. It takes `None`, a scalar or a per-bin array, and `get_data()` returns `values` unchanged in all four cases I measured — it is a drawing choice. Nor does it reach anything announced: the core's `Histogram` trace takes its bin range from `xMin`/`xMax` on a vertical chart and `yMin`/`yMax` on a horizontal one, and **never reads the pair on the other axis** (`src/model/histogram.ts:22,38,56`). That is what settles the question rather than a convention.

## The parity is structural, not coincidental

`ax.stairs` leaves no `BarContainer` — it adds a single `StepPatch` — so `HistPlot`'s container lookup cannot find it and `StairsPlot` reads the artist the patch hands over. But the point *builder* is now shared (`HistPlot._bin_point`), so the two spellings emit the same payload for the same chart:

```
stairs [{"y": 1.0, "x": 0.5, "xMin": 0.0, "xMax": 1.0, "yMin": 0, "yMax": 1.0}, …]
hist   [{"y": 1.0, "x": 0.5, "xMin": 0.0, "xMax": 1.0, "yMin": 0, "yMax": 1.0}, …]
                                                                       equal ✓
```

That extraction is what a second caller earns, and the mutation table below shows it guards both.

The patch's own artist is handed over rather than looked up, for the reason `maidr/patch/gantt.py` gives: a second `stairs` call leaves a second `StepPatch` beside the first, and "the patch on this axes" would describe the opening call twice.

## The limit, stated rather than buried

A staircase is **one** `<path>` for the whole chart. Measured on the same three-bin chart:

| call | `<path>` elements for the data |
|---|---|
| `ax.hist` | 3 — one per bar |
| `ax.stairs` | 1 — the whole outline; `fill=True` and `baseline=None` alike |

So the layer emits **no selectors**: the chart announces every bin and highlights none. A selector matching that one path would be worse than none — it would outline the entire staircase identically at every bin, telling a low-vision reader nothing about where they are. This is the blind spot xability/maidr#814 names, and the same trade made in xability/r-maidr#193: a blind reader gains every number, nobody loses a highlight they had, and `ax.hist` draws the same histogram with a patch per bin for anyone who needs both.

It is asserted rather than described — `test_a_staircase_announces_its_bins_but_has_nothing_to_highlight` — so it reddens the day per-bin elements appear. The layer is checked to *exist* first, because a chart that registered nothing would also have no selectors and would pass for the wrong reason (the defect I had to fix in xability/r-maidr#193's equivalent test).

## Two non-finite cases, measured rather than assumed

matplotlib accepts an infinite bin edge and rejects a NaN one (`ValueError: Nan values in "edges" are disallowed`), and accepts a NaN count. They are not the same case:

- a **count** of NaN is how a staircase leaves a bin blank — the bin still has a position, so it is kept and its count emitted as `null`, the gap the core's bar family already reads;
- an **edge** that is infinite leaves the bin nowhere to be, so it is dropped. Not a contrivance: `sns.ecdfplot` opens its staircase at `-inf` for exactly this reason, which is what #427 was about.

Both keep the payload loadable. `json.dumps` writes bare `NaN`/`Infinity`, which `JSON.parse` rejects — and the chart then never initialises at all.

## Verification

`pytest 0 / ruff 0` — 2,445 passed, 18 skipped, up from 2,432. Every file this adds or touches is `ruff`-clean; the 6 repo-wide findings are the pre-existing `__init__.py` re-export warnings.

| mutation | result |
|---|---|
| the patch is never imported | 12 of 13 failed |
| the factory ignores `step_patch` and always builds a `HistPlot` | 11 failed |
| read the axes' first patch instead of the call's own | 1 failed |
| a NaN count is emitted raw instead of as `null` | 2 failed |
| a NaN count drops the bin instead of nulling it | 1 failed |
| the non-finite **edge** guard dropped | 1 failed |
| the empty-staircase guard dropped | 1 failed |
| `_support_highlighting` left on | 1 failed |
| orientation always vertical | 2 failed |
| the shared `_bin_point` mutated | 4 stairs tests **and** `test_bar_orientation.py::test_vertical_histogram_bins_run_along_x` |

The last row is the one worth reading: mutating the shared builder reddens the existing `hist` suite as well, which is what makes the parity structural rather than a coincidence two tests happen to agree on.

**One mutation passed and led to a real fix.** Emitting the NaN raw left `test_a_blank_bin_leaves_the_payload_parseable` green, because Python's `json.loads` *accepts* the bare `NaN` token that `JSON.parse` rejects — so the round trip proved nothing about the payload the browser has to read. The test now parses with a `parse_constant` hook that refuses all three tokens, which is what makes the two agree, and it reddens.

## Found alongside, not in this PR

**`sns.histplot(element="step")` and `element="poly"` are silent too**, and not through `ax.stairs`:

```
element=bars   containers=[BarContainer]     → ['hist']
element=step   collections=[PolyCollection]  → nothing
element=poly   collections=[PolyCollection]  → nothing
```

Seaborn draws those through a fill rather than through bars, so `sns_hist` reaches `_drew_bars` → False and declines — the third branch of the same decline #522 fixed for the bivariate mesh. It needs its own reading and its own issue; recorded in #535.

`ax.contour` / `ax.contourf` / bivariate `sns.kdeplot` → `CONTOUR` is what remains of #524, and is the largest of the three since it introduces a plot type py-maidr has never emitted.

---

### DIFF
--- a/maidr/core/plot/histogram.py
+++ b/maidr/core/plot/histogram.py
@@ -53,6 +53,61 @@ class HistPlot(MaidrPlot, ContainerExtractorMixin, DictMergerMixin):
 
         return data
 
+    @staticmethod
+    def _bin_point(
+        orientation: str, bin_start: float, bin_size: float, count: float | None
+    ) -> dict:
+        """
+        Build one bin's point, whichever way the histogram was drawn.
+
+        Shared with :class:`~maidr.core.plot.stairs.StairsPlot`, so the two
+        spellings of a histogram -- a row of bars from ``Axes.hist``, a
+        staircase from ``Axes.stairs`` -- emit the same payload for the same
+        chart rather than two that merely resemble one another.
+
+        Parameters
+        ----------
+        orientation : str
+            ``"horz"`` when the bins run up the y axis, ``"vert"`` otherwise.
+        bin_start : float
+            The bin's lower edge, along the axis the bins run on.
+        bin_size : float
+            The width of the bin, on that same axis.
+        count : float or None
+            What the bin holds, or ``None`` when it holds nothing measurable.
+
+        Returns
+        -------
+        dict
+            The point, with the bin edges on the axis the bins run on.
+
+        Notes
+        -----
+        The bounds across the bins -- ``yMin``/``yMax`` on a vertical
+        histogram, ``xMin``/``xMax`` on a horizontal one -- are emitted for
+        the shape's sake and are not read: the core's ``Histogram`` trace
+        takes its bin range from the pair on the *binned* axis and never
+        looks at the other.
+        """
+        if orientation == "horz":
+            return {
+                MaidrKey.X.value: count,
+                MaidrKey.Y.value: bin_start + bin_size / 2,
+                MaidrKey.Y_MIN.value: bin_start,
+                MaidrKey.Y_MAX.value: bin_start + bin_size,
+                MaidrKey.X_MIN.value: 0,
+                MaidrKey.X_MAX.value: count,
+            }
+
+        return {
+            MaidrKey.Y.value: count,
+            MaidrKey.X.value: bin_start + bin_size / 2,
+            MaidrKey.X_MIN.value: bin_start,
+            MaidrKey.X_MAX.value: bin_start + bin_size,
+            MaidrKey.Y_MIN.value: 0,
+            MaidrKey.Y_MAX.value: count,
+        }
+
     def _extract_bar_container_data(
         self, plot: BarContainer | None
     ) -> list[dict] | None:
@@ -66,35 +121,23 @@ class HistPlot(MaidrPlot, ContainerExtractorMixin, DictMergerMixin):
             # height and the count off its width. The vertical case is the
             # mirror of that.
             if self._orientation == "horz":
-                count = float(patch.get_width())
-                bin_start = float(patch.get_y())
-                bin_size = float(patch.get_height())
-
                 data.append(
-                    {
-                        MaidrKey.X.value: count,
-                        MaidrKey.Y.value: bin_start + bin_size / 2,
-                        MaidrKey.Y_MIN.value: bin_start,
-                        MaidrKey.Y_MAX.value: bin_start + bin_size,
-                        MaidrKey.X_MIN.value: 0,
-                        MaidrKey.X_MAX.value: count,
-                    }
+                    self._bin_point(
+                        "horz",
+                        float(patch.get_y()),
+                        float(patch.get_height()),
+                        float(patch.get_width()),
+                    )
                 )
                 continue
 
-            count = float(patch.get_height())
-            bin_start = float(patch.get_x())
-            bin_size = float(patch.get_width())
-
             data.append(
-                {
-                    MaidrKey.Y.value: count,
-                    MaidrKey.X.value: bin_start + bin_size / 2,
-                    MaidrKey.X_MIN.value: bin_start,
-                    MaidrKey.X_MAX.value: bin_start + bin_size,
-                    MaidrKey.Y_MIN.value: 0,
-                    MaidrKey.Y_MAX.value: count,
-                }
+                self._bin_point(
+                    "vert",
+                    float(patch.get_x()),
+                    float(patch.get_width()),
+                    float(patch.get_height()),
+                )
             )
 
         # Tag the elements for highlighting
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from matplotlib.axes import Axes
 from matplotlib.lines import Line2D
+from matplotlib.patches import StepPatch
 from maidr.core.enum import PlotType
 from maidr.core.plot.areaplot import AreaPlot
 from maidr.core.plot.barplot import BarPlot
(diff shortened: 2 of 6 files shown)


jooyoungseo:
The baseline gap is real and is now closed in 175854e. It was a claim with no guard behind it — exactly the kind this PR's own mutation table is meant to catch, and it slipped through because I argued it in prose instead of asserting it.

`test_the_baseline_is_a_drawing_choice_and_not_a_reading` runs the same three bins under `baseline=None`, `baseline=2` and `baseline=[0, 1, 0]` and requires all three payloads to equal the default's. Falsified by subtracting the baseline from every value inside `_extract_plot_data`, which reddens that test and only that test:

```
M11: the baseline is folded into the reading
FAILED tests/core/test_stairs.py::test_the_baseline_is_a_drawing_choice_and_not_a_reading
1 failed, 13 passed
```

On the other point: the `histogram.ts` lines are current — I read `src/model/histogram.ts` at `xability/maidr` main today, and `binRangeMin`/`binRangeMax` and the `text` getter take `xMin`/`xMax` on a vertical chart and `yMin`/`yMax` on a horizontal one, never the pair on the other axis. It is a cross-repository argument either way, so it is worth re-checking whenever the pinned bundle moves.

The two style notes are both deliberate and read correctly, so I have left them: the import lists are already ordered by when things were added rather than alphabetically, and `patch/stairs.py` cannot use `common()` because the empty-staircase early return has to happen between the draw and the registration.

---

PR #538, opened 2026-08-20
fix(core): render one figure at a time for every caller, not just two doors

### LINKED ISSUE #530: A figure mutated while another session renders it produces an SVG the payload does not describe

### LINKED ISSUE #532: The dpi race is in `maidr.render`, so every threaded caller has it — not only the two doors that now hold a lock
#531 gives the Streamlit door the per-figure lock the Shiny door has had since #504, and puts both on one registry. That closes the two integrations this package ships. It does not close the defect, because the defect is a level down.

## Measured through `maidr.render`, with no widget involved

Six threads, barrier-synchronised, rendering one figure; outputs compared with per-render uuids and the SVG timestamp normalised away:

```
trial 0: 2 distinct outputs from 6 renders
trial 1: 1 distinct outputs from 6 renders
trial 2: 1 distinct outputs from 6 renders
trial 3: 1 distinct outputs from 6 renders
trial 4: 1 distinct outputs from 6 renders
trials with disagreeing renders: 1/5
```

Same root cause as #454: `savefig` writes `fig.dpi` for its duration and restores it, so the loser of the race draws the whole chart at the other call's dpi — a complete, well-formed SVG at 72% scale, raising nothing.

## Who is exposed

Anyone rendering one figure from more than one thread:

- **Flask.** `Environment.is_flask()` exists, so this is a supported embedding, and Werkzeug serves threaded by default.
- Any app calling `maidr.render` / `save_html` from a thread pool.
- A notebook doing its own threading.

None of them go through `maidr/widget/`, so none of them get the lock #531 installs.

## The shape a fix would take

Put the lock where the mutation is — around the `savefig` in `maidr/core/maidr.py` — and both doors drop theirs. That is smaller than what exists now, not larger, and covers every caller by construction.

The one thing to get right is that it must be taken at **exactly one level**. `show()` and `save_html()` both call into the render path, so a lock held at the API level and again underneath would deadlock on a non-reentrant `threading.Lock` — and `RLock` is the wrong answer here, since the point of the plain lock is that a re-entering render is a bug that should show up rather than interleave (see the note in `maidr/util/figure_lock.py`).

The registry to reuse is already in `maidr/util/figure_lock.py` after #531; nothing new is needed for it. What needs deciding is whether the core render path is willing to take a lock at all, which is a design call rather than a mechanical move — it changes the concurrency contract of the public API for every caller, not only for the two integrations.

Not doing it in #531, because that PR fixes a reported defect on the Streamlit door and this widens the blast radius from `maidr/widget/` to `maidr/core/`.

Follow-up from #454 and #531.

jooyoungseo:
Closes #532.

#504 and #531 put a per-figure lock on the Shiny and Streamlit doors. **The race is a level below them**, so those two are the only callers it protects.

## Measured, with no widget involved

Six threads, barrier-synchronised, through `maidr.render` on one figure; outputs compared with per-render uuids and the SVG timestamp normalised away:

```
trial 0: 2 distinct outputs from 6 renders
trial 1: 1 distinct outputs from 6 renders
...
trials with disagreeing renders: 1/5
```

Same root cause as #454 — `savefig` writes `fig.dpi` for its duration and restores it, so the loser of the race draws its whole chart at the other call's dpi. A complete, well-formed SVG at 72% scale, raising nothing.

Who was exposed: a threaded **Flask** app (`Environment.is_flask` makes it a supported embedding, and Werkzeug serves threaded by default), anything calling `render` or `save_html` from a thread pool, a notebook doing its own threading. None of them go through `maidr/widget/`.

## The fix is smaller than what it replaces

`Maidr._create_html_tag` takes the lock. Every render of a matplotlib figure funnels through it — `render`, `save_html` and `_create_html_doc` all call it and none re-enters it — so **one lock site covers every caller**, and a plain `Lock` stays safe. The doors drop theirs, because a second lock above this one would deadlock against it rather than nest, exactly as the module docstring warned.

The body moves to `_build_html_tag` so the lock wraps it without re-indenting sixty lines.

## It also deletes the bug class the last two reviews kept circling

The doors had to work out *which* figure a value named before they could lock it. When that answer was wrong — as it was for a `Figure`, and for the current figure — the lock was taken on nothing while the render went ahead (#531). Two reviewers then flagged the resolver's reliance on `api._get_plot_or_current` and on `get_axes`'s branch order.

Locking by `self._fig`, which the `Maidr` instance already holds, **cannot disagree with the render**. So `resolve_figure` goes, its coupling goes, and the `maidr_html` resolve-once dance goes with them.

| | before | after |
|---|---|---|
| lock sites | 2 (per door) | 1 (core) |
| callers covered | Shiny, Streamlit | every caller |
| figure resolved for locking | yes, twice, by two modules | never — the figure is already known |
| lines | — | **−192** |

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,442 passed, 14 skipped**. (`ruff check` reports the 6 pre-existing re-export findings in `maidr/core/**/__init__.py`; every file this touches is clean.)

| mutation | result |
|---|---|
| the core lock removed | **5 of 6** concurrency guards fail, 3 runs of 3 |
| …the sixth being `test_two_figures_still_render_in_parallel`, which correctly does not care | passes |
| the lock made process-wide instead of per-figure | **that sixth guard fails**, 3 of 3 |

The second row is the point of the sixth test: a process-wide lock passes every "do they agree?" test while throwing away the parallelism that rendering on a thread exists for.

## Tests

- `tests/core/test_render_serialises.py` — new. Concurrent renders through `maidr.render` agree (parametrised over naming the chart by `ax` and by `fig`), and two *distinct* figures still overlap.
- `tests/widget/test_shiny.py`, `tests/widget/test_every_door_agrees.py` — the two overlap tests now stub `Maidr._build_html_tag` rather than `maidr.render`. Stubbing `render` would replace the lock along with the work and leave them asserting nothing; this is the shape of false pass worth naming.
- `tests/util/test_figure_lock.py` — the resolver's tests go with the resolver. What they protected (every way of naming one chart reaches one lock) is now structural, and is asserted at the render level instead.
- The per-door `test_concurrent_renders_of_one_figure_agree` tests are untouched and still pass, now through the core lock — which is the point: the doors' *behaviour* is unchanged.

## Scope

Still does not close #530 — a figure mutated *while* a render is in flight is not a concurrent render, and no lock keyed on renders can cover it.

---

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -326,7 +326,7 @@ Measured through the renderer on a 50-bar chart, longest gap between successive 
 
 Same throughput either way — the work did not get cheaper, it stopped being in everyone else's way.
 
-Two things this does not do. Renders of the **same figure** are serialised, because `savefig` writes `fig.dpi` for its duration and two concurrent writes leave one chart drawn at the other's scale; distinct figures, which is the usual per-session shape, run in parallel. That lock is shared with the Streamlit path rather than private to this one, so a figure reached from both is still rendered one at a time. And the chart still travels over the websocket on every flush — roughly 25 KB, or a couple of megabytes with `use_cdn=False`; that part is tracked in [#534](https://github.com/xability/py-maidr/issues/534).
+Two things this does not do. Renders of the **same figure** are serialised, because `savefig` writes `fig.dpi` for its duration and two concurrent writes leave one chart drawn at the other's scale; distinct figures, which is the usual per-session shape, run in parallel. That lock is held by the render path itself rather than by this integration, so a figure is rendered one at a time however it is reached — through Shiny, through Streamlit, or from a thread of your own. And the chart still travels over the websocket on every flush — roughly 25 KB, or a couple of megabytes with `use_cdn=False`; that part is tracked in [#534](https://github.com/xability/py-maidr/issues/534).
 
 `uv run pytest tests/core/test_render_is_synchronous.py --run-benchmark` re-measures the underlying `maidr.render()`, which is still synchronous and still blocks its caller — that is what makes moving it off the loop necessary rather than optional.
 :::
--- a/maidr/api.py
+++ b/maidr/api.py
@@ -429,13 +429,6 @@ def _get_plot_or_current(plot: Any | None) -> Any:
     """
     Get the plot object or current matplotlib figure if plot is None.
 
-    Underscored but not private to this module:
-    :func:`maidr.util.figure_lock.resolve_figure` and
-    :func:`maidr.widget.streamlit.maidr_html` both call it, so that the
-    figure a render *locks* is the figure it goes on to write. A change
-    here that made this answer differently would put those locks on a
-    figure the render never touches -- synchronised in appearance only.
-
     Parameters
     ----------
     plot : Any or None
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -432,6 +432,15 @@ class Maidr:
         six threads on one figure: 1 of 5 trials came back with two
         distinct outputs.
 
+        The lock covers the whole render rather than the ``savefig`` call
+        alone. That is wider than the mutation strictly needs, and
+        deliberate: the schema is extracted from the same artists
+        ``savefig`` is writing, so a narrower lock would let one render's
+        payload be built while another render mutates the figure under it.
+        It is also what the two doors held before this, so nothing about
+        contention changes for them. In a single-threaded script the lock
+        is never contended and costs an uncontended acquire per render.
+
         Locking by ``self._fig`` also removes a resolution step that could
         disagree with the render. The integrations had to work out *which*
         figure a value named before they could lock it, and when that
@@ -459,11 +468,13 @@ class Maidr:
         data_in_svg: bool = True,
         use_cdn: bool | Literal["auto"] = "auto",
     ) -> Tag:
-        """Render the chart. Call :meth:`_create_html_tag`, which holds the lock.
+        """Render the chart. Callers want :meth:`_create_html_tag`, which locks.
 
-        Separate only so that the lock wraps the whole of this rather than
+        Split out only so that the lock wraps the whole of this rather than
         this being indented under a ``with``: the body is long, and the
-        rename keeps the two concerns readable apart.
+        split keeps the two concerns readable apart. Nothing but
+        :meth:`_create_html_tag` should call this -- a caller reaching past
+        it renders without the lock.
         """
         # Before the artists are collected, not after: reading `plot.elements`
         # renders the layer, and a superseded `BarPlot` on a stacked axes
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -24,6 +24,7 @@ from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.plot import MaidrPlot
 from maidr.core.plot.barplot import BarPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
+from maidr.util.figure_lock import figure_lock
 from maidr.util.bundle_capability import (
     schema_trace_types,
     warn_if_bundle_cannot_render,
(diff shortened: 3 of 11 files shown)


jooyoungseo:
Both nits taken. Thank you especially for checking the pickling constraint — that the lock is acquired through `figure_lock(self._fig)` rather than stored on `self` is load-bearing (a `Lock` attribute would break `pickle.dumps(fig)`, and `Maidr` instances ride on the figure since #456), and it was not something I had written down anywhere.

**The critical section's width.** You are right that it is wider than the `savefig` call the docstring justifies it with, and the docstring now says so and why: the schema is extracted from the same artists `savefig` is writing, so a narrower lock would let one render's payload be built while another render mutates the figure under it. Two facts that were in my head and not in the file.

One correction to the framing, since it matters for anyone reading this later: a single-threaded script is not affected at all. There is no second holder, so the lock is never contended — it costs one uncontended acquire per render, which is nanoseconds. The blocking case needs a *concurrent* render of the same figure, which is the case that was silently producing a wrong-scale chart before.

**The `_build_html_tag` docstring.** Reworded — it read as an instruction to the method rather than guidance to a maintainer. It now says plainly that nothing but `_create_html_tag` should call it, and what happens if something does: a render without the lock.

## Verification after the change

`pytest 0 / ruff 0` — 83 passed across `tests/core/test_render_serialises.py` and `tests/widget`, and the full suite was **2,442 passed, 14 skipped** on the commit you reviewed. Docstrings only in this commit, so the mutation results stand: lock removed fails 5 of 6 guards 3/3, lock made process-wide fails the sixth 3/3.

Merging once CI is green.

---

jooyoungseo:
The stale docstring is a straight miss, and the kind worth catching: I reworded every file the diff touched and left the one that stopped calling `figure_lock` still pointing at it, with the import gone. A reader following that `:func:` reference would have found nothing.

Fixed. The paragraph's actual point survives the move and is the part that matters, so it stays and now says where the lock is: `_render_off_loop` no longer *takes* a lock, but it still **waits** on one inside its executor thread, holding the slot — so enough sessions rendering one shared figure can still make unrelated figures queue for a thread.

The patch-style inconsistency is fixed too — `test_two_figures_still_render_in_parallel` uses `monkeypatch.setattr` like its siblings instead of restoring by hand. Re-ran the process-wide-lock mutation afterwards to be sure the swap did not weaken it: still fails **3 of 3**.

## Verification

`pytest 0 / ruff 0` — **2,442 passed, 14 skipped**.

Thank you for tracing the call paths independently, including `api.save_html`'s stacked/list branch reaching `_create_html_doc` per figure. That branch is the one I would have been least sure of from memory, and it is exactly where a missed path would have left a silent hole.

---

PR #540, opened 2026-08-20
feat(core): read Axes.contour as the scalar field it draws

### LINKED ISSUE #539: `ax.contour` and bivariate `sns.kdeplot` are silent, and the level is a number rather than a colour
The last of the three readings #524 recorded, after #533 (`broken_barh`) and #536 (`stairs`). Measured on matplotlib 3.9.4, seaborn 0.13.2.

## Silent today

```
ax.contour(X, Y, Z, levels=[0.2, 0.5, 0.8])   → UnsupportedPlotError
sns.kdeplot(data=df, x="a", y="b")            → UnsupportedPlotError
sns.jointplot(data=df, x="a", y="b", kind="kde")  → ['smooth', 'smooth']
```

That last one is the sharpest: the jointplot reads its two **marginals** and nothing at all for the joint panel the chart is drawn for.

## Nothing has to be inverted from a colour

MAIDR has carried the contour trace since xability/maidr#802, and matplotlib hands both halves over directly:

```python
cs = ax.contour(X, Y, Z, levels=[0.2, 0.5, 0.8])
cs.levels      # [0.2, 0.5, 0.8]
cs.get_paths() # one Path per level: 65, 41 and 25 vertices
```

This is what separates the same chart here from the same chart elsewhere. xability/maidr#1084 left Observable Plot's `contour` unread because there the magnitude survives only in a continuous fill colour, so several distinct levels render as the same pixel. `QuadContourSet` has no such problem: **`.levels` is the data.**

`sns.kdeplot(x=, y=)` is the same artist with computed levels (`[0.0059, 0.0234, 0.0406, …]`), so it comes along with it — and finishes the jointplot.

## Three things the drawing forces, all measurable

**A level is not one curve.** A field with two peaks crosses a level twice, and matplotlib draws both islands in a *single compound path with two `MOVETO`s*:

```
levels: [0.5]   paths: 1
  path 0: 86 vertices, 2 MOVETO(s)
```

Read as one series they would be joined by a straight line running between the peaks — a curve announced across ground the field never took, which is the defect xability/maidr#1079 describes for a gappy line. Each drawn polyline has to become its own series; `ContourPoint` carries the level on every point, so several series sharing a level costs nothing.

**A level nothing reaches still gets a path.** Measured, `levels=[0.2, 0.5, 2.0]` on a field whose maximum is 1 gives three paths with 65, 41 and **0** vertices. A series with no points is a row a reader can land on and be told nothing about.

**A filled contour is a different chart.** `ax.contourf` with three levels draws **two** paths — the bands *between* the levels — and each outline runs along two different level curves stitched together. Announcing one as "the 0.2 contour" would be right for half of its points. The same applies to `sns.kdeplot(..., fill=True)`.

## Highlighting

Setting a gid on the contour set produces one `<g>` holding a `<path>` per level, so a curve is addressable by the level it belongs to. Two islands of one level then name the same element — a reader on either sees that level outlined, which is what the drawing can say when the islands have no elements of their own.

## The recursion guard is in the way for seaborn

`sns.kdeplot` reaches `ax.contour` internally, but the `kdeplot` patch sets the internal context around its own call, so a patched `Axes.contour` declines and the outer patch — which only knows how to register curves — declines too. The chart goes out silent. That is the same shape as #522, and the fix is the same: the seaborn patch makes the registration itself, for the sets *that call* added.

jooyoungseo:
Closes #539 — the last of the three readings #524 recorded, after #533 (`broken_barh`) and #536 (`stairs`).

| call | before | now |
|---|---|---|
| `ax.contour(X, Y, Z, levels=[0.2, 0.5, 0.8])` | `UnsupportedPlotError` | `['contour']` |
| `sns.kdeplot(data=df, x="a", y="b")` | `UnsupportedPlotError` | `['contour']` |
| `sns.jointplot(data=df, x="a", y="b", kind="kde")` | `['smooth', 'smooth']` | `['contour', 'smooth', 'smooth']` |

That third row is the sharpest: the jointplot read its two **marginals** and nothing for the joint panel the chart is drawn for.

## The level is a number, not a colour

```python
cs = ax.contour(X, Y, Z, levels=[0.2, 0.5, 0.8])
cs.levels       # [0.2, 0.5, 0.8]
cs.get_paths()  # one Path per level: 65, 41 and 25 vertices
```

Both halves come off the artist. This is exactly what separates the same chart here from the same chart elsewhere: xability/maidr#1084 left Observable Plot's `contour` unread because there the magnitude survives only in a continuous fill, so distinct levels render as the same pixel. `QuadContourSet` has no such problem.

The curves are checked against the field rather than against themselves — a gaussian bump's 0.5 contour is the circle of radius `√(-ln 0.5)`, and every emitted point sits on it to within 0.05.

## Three things the drawing forces, all measured

**A level is not one curve.** A field with two peaks crosses a level twice, and matplotlib draws both islands in a *single compound path*:

```
levels: [0.5]   paths: 1
  path 0: 86 vertices, 2 MOVETO(s)
```

Read as one series they would be joined by a straight line across the saddle between the peaks — a curve announced over ground the field never took, which is the defect xability/maidr#1079 describes. Each drawn polyline becomes its own series; `ContourPoint` carries the level on every point, so several series sharing one costs nothing.

**A level nothing reaches still gets a path.** `levels=[0.2, 0.5, 2.0]` on a field whose maximum is 1 gives three paths with 65, 41 and **0** vertices. A series with no points is a row a reader can land on and be told nothing about, so it contributes none — and a set that reaches no level at all registers nothing rather than an empty layer (#421's shape).

**A filled contour is a different chart.** `contourf` with three levels draws **two** paths — the bands *between* the levels — and each outline runs along two different level curves stitched together. Announcing one as "the 0.2 contour" would be right for half its points. Declined, and so is `sns.kdeplot(..., fill=True)`.

## Why seaborn needed its own registration

`sns.kdeplot` reaches `ax.contour` internally, but the `kdeplot` patch sets the internal context around its own call — so the patched `Axes.contour` declines, and the outer patch, which only knows how to register curves, declined too. The chart went out silent. Same shape as #522, same fix: the seaborn patch registers the sets **that call added**, so an `ax.contour(...)` already on the axes is not claimed twice.

## Highlighting

Tagging the set gives one `<g>` holding a `<path>` per level, so a curve is addressed by the level it belongs to. Two islands of one level name the same element: a reader on either sees that level outlined, which is the honest answer when the islands have no elements of their own. Asserted against the **rendered SVG**, not just the payload — a selector matching nothing is xability/maidr#814's blind spot, where everything reads and nothing lights up.

## A pin turned red, as designed

`tests/core/plot/test_hist2d.py::test_bivariate_kdeplot_is_not_registered_yet` was written to fail the day a contour trace arrived — *"a contour plot is not a heatmap with different colours: the level is the navigable object, not the cell"*. It has, so it is replaced by the two cases it stood in for, and that file's module docstring is corrected: it still said the bivariate kdeplot "does NOT read".

## Verification

`pytest 0 / ruff 0` — 2,473 passed, 18 skipped, up from 2,445. Every file added or touched is `ruff`-clean; the 6 repo-wide findings are the pre-existing `__init__.py` re-export warnings.

| mutation | result |
|---|---|
| the contour patch is never imported | 9 of 16 failed |
| a level is one series (no `MOVETO` split) | 1 failed |
| the empty-path guard in the patch dropped | 1 failed |
| a filled kdeplot is claimed too | 1 failed |
| the before-snapshot in the kdeplot patch dropped | 1 failed |
| selectors numbered by series rather than by level | 1 failed |
| the contour set is never tagged | 2 failed |
| the level is not carried on the points | 6 failed |

**One mutation passed, and correcting the code comment was the right response rather than adding a test.** Taking the vertex sitting in the `CLOSEPOLY` slot instead of repeating the curve's first point left every test green. The reason is measurable: matplotlib writes the subpath's *start point* into that slot, so on this artist the two are identical —

```
CLOSEPOLY slot 42 holds [-1.92308, -0.71732]   subpath start: [-1.92308, -0.71732]   equal: True
CLOSEPOLY slot 85 holds [ 1.15385, -0.75847]   subpath start: [ 1.15385, -0.75847]   equal: True
```

— and no fixture can tell them apart. The code keeps `current[0]` because matplotlib documents that vertex as **ignored**, so relying on what it happens to hold is relying on something outside the contract. The comment claimed it was a placeholder; it now says what was measured, and says plainly that this is the one choice here nothing can falsify.

## Not in scope

`sns.histplot(element="step")` and `element="poly"` are still silent — seaborn draws those through a fill rather than through bars, so `sns_hist` reaches `_drew_bars` → False and declines. A third branch of the same decline, recorded in #535, and a different reading (a `PolyCollection` whose vertices are a staircase).

---

### DIFF
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -12,6 +12,12 @@ class PlotType(str, Enum):
     #: distinct type because :attr:`BOX` is fixed at one rung, and depth is the
     #: whole point of the chart.
     BOXEN = "boxen"
+    #: A scalar field drawn as curves of constant value. The level is a
+    #: number here rather than a colour -- `QuadContourSet.levels` is the
+    #: data, and `get_paths()` gives one path per level -- which is what
+    #: separates `Axes.contour` from the same chart in a renderer that keeps
+    #: its magnitude only in a fill.
+    CONTOUR = "contour"
     COUNT = "count"
     DODGED = "dodged_bar"
     ERRORBAR = "error_bar"
--- a/maidr/core/plot/contour.py
+++ b/maidr/core/plot/contour.py
@@ -1,0 +1,209 @@
+"""ContourPlot — a scalar field read as the iso-value curves it is drawn as."""
+
+from __future__ import annotations
+
+import math
+import uuid
+
+import numpy as np
+from matplotlib.axes import Axes
+from matplotlib.contour import ContourSet
+from matplotlib.path import Path
+
+from maidr.core.enum import MaidrKey, PlotType
+from maidr.core.plot import MaidrPlot
+
+
+class ContourPlot(MaidrPlot):
+    """
+    A contour plot: one curve per level of a scalar field.
+
+    ``Axes.contour`` is the one chart in this family whose value is a **number
+    rather than a colour**. ``QuadContourSet.levels`` holds the levels the
+    caller asked for, and ``get_paths()`` returns one path per level, so both
+    halves invert exactly and nothing has to be recovered from a fill. That is
+    what separates it from the same chart elsewhere: xability/maidr#1084 left
+    Observable Plot's ``contour`` unread because there the magnitude survives
+    only in a continuous fill colour.
+
+    Two things about the drawing shape the reading.
+
+    **A level is not one curve.** A field with two peaks crosses a level twice,
+    and matplotlib draws both islands in a *single* compound path with two
+    ``MOVETO`` commands. Read as one series they would be joined by a straight
+    line running between the peaks -- a curve announced across ground the field
+    never took, which is the defect xability/maidr#1079 describes for a gappy
+    line. So each drawn polyline becomes its own series, and several series may
+    share a level. ``ContourPoint`` carries the level on every point precisely
+    so that costs nothing.
+
+    **A filled contour is not this chart.** ``contourf`` draws the *bands*
+    between levels, so its outlines run along two different level curves
+    stitched together and there is one fewer of them than there are levels --
+    measured, three levels give two paths. Announcing such an outline as "the
+    0.2 contour" would be right for half of its points. The patch declines it;
+    see ``maidr/patch/contour.py``.
+
+    Parameters
+    ----------
+    ax : Axes
+        The axes the field was drawn on.
+    contour_set : ContourSet
+        The artist this call produced, handed over rather than searched for:
+        two ``contour`` calls on one axes leave two sets, and "the contour set
+        on this Axes" would read the first one twice.
+    **kwargs : dict
+        Ignored; accepted so the factory can forward what it was given.
+    """
+
+    def __init__(self, ax: Axes, contour_set: ContourSet, **kwargs) -> None:
+        self._contour_set = contour_set
+        #: The level index behind each emitted series, in emission order. This
+        #: is what keeps the selectors aligned when one level draws several
+        #: islands and so contributes more than one series.
+        self._series_levels: list[int] = []
+        super().__init__(ax, PlotType.CONTOUR)
+
+    def _extract_plot_data(self) -> list[list[dict]]:
+        """
+        Read one series per drawn polyline off the contour set.
+
+        Returns
+        -------
+        list of list of dict
+            One series per polyline, each point carrying its level. Levels
+            nothing reached contribute no series -- matplotlib still emits an
+            empty path for them, and a series with no points is a row a reader
+            can land on and be told nothing about.
+        """
+        levels = np.asarray(self._contour_set.levels, dtype=float)
+        paths = self._contour_set.get_paths()
+
+        self._series_levels = []
+        data: list[list[dict]] = []
+        for index, path in enumerate(paths):
+            if index >= len(levels):
+                break
+            level = float(levels[index])
+            if not math.isfinite(level):
+                continue
+            for polyline in _polylines(path):
+                data.append(
+                    [
+                        {
+                            MaidrKey.X.value: float(vertex[0]),
+                            MaidrKey.Y.value: float(vertex[1]),
+                            MaidrKey.LEVEL.value: level,
+                        }
+                        for vertex in polyline
+                    ]
+                )
+                self._series_levels.append(index)
+
+        return data
+
+    def _get_selector(self) -> list[str]:
+        """
+        Return one selector per emitted series.
+
+        matplotlib draws the whole set as one group holding a ``<path>`` per
+        level, so a series is addressed by the level it belongs to. Two series
+        of one level therefore name the same element: a reader on either island
+        of the 0.5 contour sees the 0.5 contour outlined, which is the honest
+        answer when the drawing gives the islands no elements of their own.
+
+        Returns
+        -------
+        list of str
+            One selector per series, in emission order.
+        """
+        gid = self._contour_set.get_gid()
+        if gid is None:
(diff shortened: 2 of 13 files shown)


jooyoungseo:
Both correctness points were worth chasing, and the second one caught me making the bug while trying to fix it. f371d7e.

## The unreached-level indexing — and the mistake it exposed

The case you named, a level nothing reaches *between* two that are reached, cannot occur. matplotlib rejects the input outright:

```
ax.contour(X, Y, Z, levels=[0.2, 2.0, 0.5])
# ValueError: Contour levels must be increasing
```

and with an increasing list, any level strictly inside the field's range is crossed somewhere, so an unreached one is always past an end. The reachable form is **first**, and you were right that nothing tested it.

Measured with `levels=[-1.0, 0.2, 0.5]` on a field whose maximum is 1:

```
path vertex counts: [0, 65, 41]
group children:     3 <path> elements
   the first: <path clip-path="url(#…)" style="fill: none; stroke: #440154; …"/>
```

The unreached level **does** reach the document — as a `<path>` carrying a style and no `d` attribute at all — so the elements and the paths stay in step even though the series do not, and numbering over `get_paths()` was right.

I did not get there first. My initial probe counted with a regex requiring `d="`, so it skipped the very element it was looking for, reported two children, and I "fixed" the numbering to count emitted series instead. The new test then failed with `assert 3 == 2` and the fix turned out to be the bug. Reverted, and `test_a_level_nothing_reaches_still_holds_its_place_in_the_document` now pins both halves — the selectors are `(2)` and `(3)`, and the group's first child is the `d`-less one. Renumbering over the series reddens it.

Thank you for that one. It would have shipped as a chart that reads correctly and outlines the level below.

## The highlight — measured, and it is the other mechanism

There are two selector shapes in this codebase, and they use different machinery:

| layer | selector | needs `_elements` |
|---|---|---|
| `bar` (base `_get_selector`) | `g[maidr='true'] > path` | **yes** — the attribute is injected via `HighlightContextManager` |
| `hexbin` (#341) | `g[id='{gid}'] > g:nth-of-type(n) > use` | no |
| `gantt` (#533) | `g[id='{gid}'] > path` | no |
| `contour` (this PR) | `g[id='{gid}'] > path:nth-of-type(n)` | no |

The gid form addresses the group matplotlib itself writes, from a gid the layer stamps. Counted in the rendered HTML:

```
bar      [('svg', schema), ('g', 'f0cc21cb…'), ('g', 'f0cc21cb…')]
gantt    [('svg', schema)]
hexbin   [('svg', schema)]
contour  [('svg', schema)]
```

and on the frontend side `Svg.selectAllElements` (`src/util/svg.ts:123`) is a plain `document.querySelectorAll` — it has no dependence on a `maidr=` attribute. So the highlight resolves for contour by the same route it resolves for the two gid-based readings already in main, and adding `ContourSet` to `highlight.py`'s wrapped classes would tag an element nothing selects by.

If that route ever turns out not to work, it is one defect across hexbin, gantt and contour rather than something introduced here — worth its own issue, and I would rather find it with a browser than assert it either way from the schema.

## The `prospective_axes` duplication

Fair, and it was my third copy. Hoisted to `maidr/patch/common.py` with the `isinstance(ax, Axes)` guard — `regplot`'s version, the stricter of the two — and all three patches now call it. One test monkeypatched `regplot._prospective_axes` by name and was updated to the new one.

Full suite: 2,474 passed, 18 skipped, `ruff` clean on every file touched.

---

PR #541, opened 2026-08-20
feat(core): say when a figure was drawn into while it was being rendered

jooyoungseo:
Takes option 4 from #530. The issue stays open — this does not stop the wrong chart, it stops it being **silent**, which is the part that matters most here.

## The window

The schema is read from the artists; the SVG is written from them afterwards. Anything that draws into the figure between those two points lands in one and not the other — a chart that shows something it never announces, or announces something it does not show.

The per-figure lock (#504, #531, #532) stops another *render* from doing that. It cannot stop the application itself, on a figure it still holds. Measured through the shipped Shiny renderer, with one session's render in flight and another session's render function drawing into a shared module-level figure:

```
B's SVG carries A's label = True      B's payload has it = False
```

3 of 3 runs, nothing raised. A sighted user sees the change; a screen-reader user is never told it exists; **every check that reads the payload passes**. That is the shape of bug this project exists to prevent, arriving through the door meant to prevent it.

## Reported, not repaired — deliberately

Re-reading the schema after the SVG would race the same way, and only the caller knows whether the figure was supposed to be still. So the warning names both remedies rather than guessing one:

> maidr: this figure was drawn into while it was being rendered, so the chart's SVG and the data maidr announces for it may not match. Finish plotting before rendering, or render a figure no other thread is using.

## The detector

A census of artist counts and axis labels per axes, taken either side of the render. It has to be cheap enough to take twice per render and only has to answer *did this change* — not what.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,466 passed, 14 skipped**.

Falsified in both directions:

| mutation | result |
|---|---|
| the check removed | 2 of the new tests fail, **3 runs of 3** |
| the census reduced to counts without labels | 1 fails, **3 of 3** |
| **false positives** | **0** occurrences across the whole suite |

The false-positive direction is what decides whether this can ship at all — a detector that reported the render's own work would fire on every chart, and a warning that always fires is one nobody reads. Beyond the zero-across-the-suite count there is an explicit silent-render test over six chart shapes, plus a colorbar: the one construction that adds an axes of its own, which is exactly what the census counts.

## A note on how it is tested

The mid-render case is driven from **inside `_get_svg`**, not by racing two threads. A race would make it a test of the scheduler — passing where the timing happened to land inside the window, and silently testing nothing where it did not. Mutating from inside the render puts the change exactly where a concurrent application would have to put it to go unnoticed, every time.

I also had to correct my own probe while measuring this: the first version set the same title on every trial, so after the first there was nothing to detect, and it reported 1 of 5 where the truth is 5 of 5. Distinct mutations per trial is the fix, and the test comment says so.

---

### DIFF
--- a/maidr/__init__.py
+++ b/maidr/__init__.py
@@ -211,6 +211,7 @@ from .patch import (  # noqa: E402, F401
     violinplot,
 )
 from .util.bundle_capability import MaidrBundleTraceWarning  # noqa: E402
+from .util.render_census import MaidrRenderRaceWarning  # noqa: E402
 from .util.bundle_freshness import (  # noqa: E402
     STALE_MINOR_GAP,
     BundleStatus,
@@ -267,6 +268,7 @@ __all__ = [
     "LATEST_TAG",
     "MaidrBundleStaleWarning",
     "MaidrBundleTraceWarning",
+    "MaidrRenderRaceWarning",
     "STALE_MINOR_GAP",
     "bundle_status",
     "resolver_outcome",
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -532,6 +532,13 @@ class Maidr:
                 "for it may not match. Finish plotting before rendering, "
                 "or render a figure no other thread is using.",
                 UserWarning,
+                # No stacklevel points at the caller here, and it is worth
+                # saying so rather than implying otherwise: the depth from
+                # a user's call varies by entry point -- `render`,
+                # `save_html` and `show` reach this through different
+                # numbers of frames. 2 names maidr's own render machinery,
+                # which is at least a frame the reader recognises. What
+                # went wrong is in the message rather than in the location.
                 stacklevel=2,
             )
 
@@ -941,7 +948,7 @@ class Maidr:
             "subplots": subplot_grid,
         }
 
-    def _artist_census(self) -> tuple:
+    def _artist_census(self) -> tuple[int, tuple[tuple[Any, ...], ...]]:
         """A cheap description of what is currently drawn on the figure.
 
         Compared either side of a render to notice a figure being drawn
@@ -949,11 +956,24 @@ class Maidr:
         artists themselves: it has to be cheap enough to take twice per
         render, and it only has to answer "did this change", not what.
 
+        **What it does not see.** Everything here is O(1) per axes, which
+        buys the cheapness and costs the guarantee: a mutation that changes
+        an *existing* artist in place -- ``patch.set_height``,
+        ``line.set_ydata``, ``collection.set_offsets`` -- moves no count
+        and no label, and passes unnoticed. So this is a detector for
+        artists appearing, disappearing, or being relabelled, not for the
+        data inside them, and silence from it is not a promise that a
+        figure was still. Catching a data change would mean hashing the
+        data on every render, which is the cost this is shaped to avoid.
+        Both mutations measured on #530 -- a title set and a bar added --
+        are the kind it sees.
+
         Measured stable across a render for bar, heatmap, hist, line,
         scatter, boxplot, a mesh with a colorbar -- which re-parents its
         axes -- and a seaborn countplot, so it does not report the render's
-        own work as interference. Tagging for highlight sets attributes on
-        artists that already exist; nothing here adds one.
+        own work as interference. ``tests/core/test_mid_render_mutation.py``
+        holds that to those shapes; tagging for highlight sets attributes
+        on artists that already exist, and adds none.
         """
         return (
             len(self._fig.axes),
@@ -963,7 +983,14 @@ class Maidr:
                     len(ax.patches),
                     len(ax.collections),
                     len(ax.images),
-                    ax.get_title(),
+                    len(ax.texts),
+                    ax.get_legend() is not None,
+                    # Three titles, not one: `get_title()` reads the centre
+                    # one only, so a left- or right-placed title moved
+                    # during a render would otherwise be invisible here.
+                    ax.get_title(loc="center"),
+                    ax.get_title(loc="left"),
+                    ax.get_title(loc="right"),
                     ax.get_xlabel(),
                     ax.get_ylabel(),
                 )
--- a/maidr/core/maidr.py
+++ b/maidr/core/maidr.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import json
-import warnings
 
 import io
 import os
@@ -26,6 +25,7 @@ from maidr.core.plot import MaidrPlot
 from maidr.core.plot.barplot import BarPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
 from maidr.util.figure_lock import figure_lock
+from maidr.util.render_census import artist_census, warn_if_figure_changed
 from maidr.util.bundle_capability import (
     schema_trace_types,
     warn_if_bundle_cannot_render,
@@ -496,7 +496,7 @@ class Maidr:
                 selector_ids.append(self.selector_ids[i])
 
         # Build schema once so id stays consistent across SVG and global var
-        drawn_before = self._artist_census()
+        drawn_before = artist_census(self._fig)
         schema = self._flatten_maidr()
 
         # Ask the bundle whether it can draw what this render is about to
@@ -517,30 +517,8 @@ class Maidr:
 
         # The schema was read from the artists; the SVG was written from
         # them afterwards. Anything that drew into the figure in between
-        # is in one and not the other -- a chart that shows something it
-        # never announces, or announces something it does not show. The
-        # lock stops another *render* from doing that; it cannot stop the
-        # application itself, on a figure it still holds (#530).
-        #
-        # Reported rather than repaired: re-reading the schema would race
-        # the same way, and the caller is the only one who knows whether
-        # the figure was supposed to be still.
-        if self._artist_census() != drawn_before:
-            warnings.warn(
-                "maidr: this figure was drawn into while it was being "
-                "rendered, so the chart's SVG and the data maidr announces "
-                "for it may not match. Finish plotting before rendering, "
-                "or render a figure no other thread is using.",
-                UserWarning,
-                # No stacklevel points at the caller here, and it is worth
-                # saying so rather than implying otherwise: the depth from
-                # a user's call varies by entry point -- `render`,
-                # `save_html` and `show` reach this through different
-                # numbers of frames. 2 names maidr's own render machinery,
-                # which is at least a frame the reader recognises. What
-                # went wrong is in the message rather than in the location.
-                stacklevel=2,
-            )
+        # is in one and not the other -- see :mod:`maidr.util.render_census`.
+        warn_if_figure_changed(drawn_before, self._fig)
 
         # Generate external payload if data is not embedded in SVG
         maidr = None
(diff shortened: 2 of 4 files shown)


jooyoungseo:
This review found a genuine hole and two of my own testing errors. All of it acted on.

## The coverage gap — half widened, half admitted

You were right that `get_title()` reads the centre title only. Same for legends, annotations and `ax.texts`. All four are O(1) per axes, so they cost nothing the shape of this check was avoiding, and each is now a dimension with **its own test that fails when that dimension is deleted**.

The rest of your list — `set_height`, `set_ydata`, `set_offsets` — is a real limit and stays one. Catching it means hashing the data on every render, which is the cost this detector is shaped to avoid. So it is now stated where the reader is, in the docstring:

> …a mutation that changes an *existing* artist in place moves no count and no label, and passes unnoticed. So this is a detector for artists appearing, disappearing, or being relabelled, not for the data inside them, and **silence from it is not a promise that a figure was still**.

And the admission has a test of its own — `test_a_value_changed_in_place_is_not_detected` — so it cannot quietly become false. If someone strengthens the census later, that test fails and tells them the docs need updating rather than leaving prose describing an older detector.

On your specific question: **yes, the measured Shiny race is caught.** It was `ax.set_title(...)` on a shared figure, and the added-bar probe was `ax.bar(...)` — a centre title and an artist count. Both are dimensions this sees. Worth confirming rather than assuming, since a motivating example the detector missed would have been the worst possible outcome.

## Two errors of mine, found by following your notes

**The legend case was not testing the legend.** It plotted a labelled line *and* called `ax.legend()`, so `len(ax.lines)` moved and the case passed with the legend dimension deleted — decoration by exactly the standard your review sets. The labelled artist is now drawn before the render, so only the legend dimension can catch it; dropping that dimension now fails **3 of 3**.

**The docstring claimed a seaborn countplot was measured and nothing exercised one.** I had measured it in a throwaway probe and written the claim from that. It is now a case in the silent-render test, along with a legend and a left-placed title.

## The nits

- **`stacklevel=2`** now carries a rationale, and the honest one: no single value points at the caller, because `render`, `save_html` and `show` reach this through different numbers of frames. 2 at least names maidr's own render machinery. What went wrong is in the message, not the location.
- **The discarded `FigureManager.get_maidr(fig)`** was leftover. Removed.
- **Patch style** consolidated on `monkeypatch.setattr`.
- **Return annotation** is now `tuple[int, tuple[tuple[Any, ...], ...]]`.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,474 passed, 14 skipped**, still **zero** occurrences of the warning across the suite.

| mutation | result |
|---|---|
| the check removed | 2 tests fail, 3 of 3 |
| census without labels | 1 fails, 3 of 3 |
| `len(ax.texts)` dropped | 1 fails |
| `ax.get_title(loc="left")` dropped | 1 fails |
| `ax.get_legend()` dropped | 1 fails, 3 of 3 *(passed before the test was fixed)* |

---

jooyoungseo:
Both findings were right, and the second one is the better catch of the two — it would have shipped a feature that works once per process and looks like it works always.

## 1. Figure-level text

Correct and now fixed. `_flatten_maidr` reads `suptitle`/`supxlabel`/`supylabel`, the census did not, and a suptitle changed mid-render was the reported bug one level up, unreported. All three are in the census with a case each, and dropping `get_suptitle()` fails **3 of 3**.

## 2. The once-per-location rule

Verified before fixing, because the fix depends on the mechanism:

```
three collisions in one process, filters untouched → 2 warnings shown
```

Two rather than your predicted one, and the reason is worth recording: maidr's own plot patches call `warnings.simplefilter("ignore")` inside a `catch_warnings` block, which mutates filter state and invalidates the registry as a side effect. That is an accident, not a guarantee, and relying on it would be exactly the sort of thing this PR is trying to stop.

Fixed two ways, because neither covers the other's case:

- **Its own category with an always-filter registered at import** (`MaidrRenderRaceWarning`, re-exported so `-W error` users can silence it alone). This is what covers *the same figure* racing repeatedly. Measured 3 of 3 with it, 2 of 3 without.
- **The message names the figure.** The default rule keys on message text, so this keeps *distinct* figures from collapsing into one report — which matters because anything that rebuilds the filter list drops the registration, and **pytest does that between tests**. Without this, the guarantee would evaporate in exactly the environment where most people would try to observe it.

The census and warning moved to `maidr/util/render_census.py` with the category, following `bundle_freshness` and `bundle_capability`.

## A test of mine that could not fail

Worth reporting rather than quietly fixing. My first version of the repeat-reporting test counted warnings across three renders and **passed with the figure naming removed** — because building each figure runs those same plot patches, invalidating the registry, so the count measured the accident rather than the fix.

It is now two tests that can each fail:

- distinct figures produce distinct message texts — falsified by dropping the naming, **3 of 3**
- repeat collisions on one figure are all reported, counted in a **subprocess** — falsified by dropping the registration, **3 of 3**

The subprocess is not decoration: pytest's filter handling means in-process the test would be measuring pytest, and the docstring says so.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,480 passed, 14 skipped**, still **zero** occurrences of the warning across the suite.

---

jooyoungseo:
Point 1 is correct and it was the right thing to block on: my "every collision is reported" claim was proven only for renders that do not overlap a plot call, which is not the scenario this feature exists for. You also spotted the tell in my own writeup — the 2-of-3 result I attributed to "an accident" *was* this race.

## The lock

`warn_if_figure_changed` now takes the same `_FILTER_LOCK` that `patch/common.py` takes around its `catch_warnings` block. Falsified rather than assumed — eight collisions against a thread plotting continuously:

| | result |
|---|---|
| with the lock | all 8 reported |
| without it | warnings lost, **3 runs of 3** |

That is now a test (`test_collisions_are_reported_while_another_thread_is_plotting`), in a subprocess for the same reason as its sibling: in-process it would measure pytest's filter handling rather than the lock.

On deadlock, since it is the obvious next question: a plot call holds `_FILTER_LOCK` and never renders, so it cannot be waiting on `figure_lock` while a render waits on `_FILTER_LOCK`. There is no cycle, and that reasoning is written where the lock is taken rather than left to be re-derived.

## The other two

**The stale symbol** — `Maidr._artist_census` stopped existing when the census moved module, and it was the failure message of the test whose entire purpose is to stop the docs and the code drifting apart. Fixed; a wrong pointer there is worse than in most places.

**Import-time global mutation** — now stated in the module docstring rather than left to be discovered at the call site: importing this module registers a filter permanently, every other filter change in the package is scoped to a `catch_warnings` block, and this one cannot be because it has to be in force at a moment none of this code controls.

**The `id()` reuse nit** — real, and left as is with the reasoning recorded: the always-filter covers the repeat-on-one-figure case, and a counter would name the *report* rather than the chart, which is not what a reader trying to find the wrong chart needs. Line length fixed with `ruff format`.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,481 passed, 14 skipped**, zero occurrences of the warning across the suite.

The interleaving test costs ~35s, which is why it is 8 collisions rather than the 20 I started with — 8 still fails 3 of 3 without the lock, so the sensitivity survives the trim.

---

jooyoungseo:
All three taken. Thank you for verifying the lock ordering against the actual code rather than against my claim about it — `figure_lock` → `_FILTER_LOCK` in one direction only, with plot calls never taking the first, is exactly the reasoning, and it is better for having been checked independently.

**The residual gap is real and the docstring now says it.** You are right that "only misses in-place mutations" was narrower than the truth: the census read is not atomic against the application either, so the "before" it captures can be a figure mid-mutation rather than a clean one. Both are the same trade — a detector that locked the application out would be a *repair*, and repairing this is what the module deliberately does not do — but a reader should not have to derive that.

**The unused `bool` return is gone.** It was speculative API with no caller, which `CLAUDE.md` argues against directly. The warning is the whole output.

**The 600s timeout was misleading and that is my fault.** It had not been observed to be borderline — a healthy run is ~35s and I simply picked a generous number, which then read as evidence of a problem. Now 180s, with the healthy cost in the docstring so a future flake is recognisable as one rather than as "it was always slow."

On the pyplot-from-two-threads point: it is deliberate, because a patched plot call on another thread *is* the situation under test, and the docstring now says so. If it ever goes flaky, that note is what should stop someone from concluding the lock is at fault.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,481 passed, 14 skipped**, zero occurrences of the warning across the suite.

Merging on green.

---

jooyoungseo:
**CI went red on my last commit, and the earlier review was right where I was not.**

When it was flagged that the 600s timeout "suggests it's already been observed to be slow/borderline", I answered that the number was merely generous and cut it to 180. Python 3.10 and 3.12 then timed out on the two-core runners. The test really was that slow there — my trim is what exposed it, and the original note was the correct read of the evidence.

## The cause was the test, not the timeout

A thread plotting in a tight loop to manufacture overlap **burns a whole core**, which on a two-core runner is taken from the renders it is supposed to overlap. Throttling it made the run affordable — 8s on two cores — and gutted the sensitivity:

```
without the lock, throttled traffic:  1 failure in 5 runs
```

A guard that detects its own bug one time in five is worse than no guard, because it reads as coverage.

## Held open instead of raced for

The render's own `_get_svg` now starts a thread that takes `_FILTER_LOCK` and enters the same ignore-everything `catch_warnings` block a patched plot call enters, holding it across the warning. That is the situation under test, produced directly instead of asked of the scheduler.

| | before | after |
|---|---|---|
| detection with the lock removed | 1 of 5 | **5 of 5** |
| cost on two cores | >180s (timeout) | **~5s** |

Deterministic turned out to be both stronger and cheaper, which is usually the way with tests like this.

## Verification

Run under `taskset -c 0,1` — what the runners have, rather than my four cores:

`pytest 0 / ruff 0 / uv lock --check 0` — **2,481 passed, 14 skipped**, zero occurrences of the warning across the suite.

I should have measured on a constrained CPU the first time, given the test's whole subject is contention.

---

jooyoungseo:
Thank you for tracing the deferred import rather than accepting my comment about it — you reconstructed the exact cycle (`maidr/__init__` → `.patch` → `patch/common` → `figure_manager` → `core.maidr` → `render_census`, arriving while `patch/common` is still running its own imports and before `_FILTER_LOCK` is assigned). That reasoning is now written where the import is, since "deferred because of a cycle" is the kind of comment a future reader deletes.

**Note 3 taken.** `warnings.warn` calls `showwarning` synchronously and that can be a user hook — `logging.captureWarnings` installs one — so the filter lock spans it. Accepted rather than fixed, because releasing before the warning is delivered is precisely the window the lock exists to close; the trade is now stated next to the lock instead of left to be worked out.

**Notes 1 and 2 belong in the release notes, and that is where I am putting them.** `CHANGELOG.md` is owned by semantic-release here, so the merge commit body is the channel — it will carry both:

- the detector sees artists appearing, disappearing and being relabelled, **not** data changed in place, so its *silence* is not a guarantee that a render was clean;
- importing maidr now registers a permanent `always` filter for `MaidrRenderRaceWarning`, which downstream projects asserting on `warnings.filters`, or running `-W error` without an exemption, will want to know about.

Your first note is the one I would most want a reader to see, and you are right that the PR's "stops it being silent" framing overstates it: that holds for the structural subset only. The merge message says so plainly rather than leaving the qualification in a docstring.

**On CI budget (note 4)** — the interleaving test is now ~5s rather than ~35s, because it holds the plot window open deliberately instead of racing for it. That change came out of the timeout failure earlier in this thread; the two subprocess tests together now cost about 7s.

## Final state

All 20 checks green on `725cfc2`'s successor, across Python 3.9–3.12 and the browser suite. Locally, under `taskset -c 0,1` to match the runners: **2,481 passed, 14 skipped**, zero occurrences of the warning across the suite.

Merging.

---

PR #547, opened 2026-08-20
feat(core): read ax.tricontour as the field it draws

### LINKED ISSUE #546: ax.tricontour reads as nothing, though the contour reader already handles it
## What happens

```python
ax.tricontour(x, y, z, levels=[1.0, 2.0, 3.0])
```

registers no layer at all. `ax.contour` on the same field is fully navigable since #540.

## Why

`Axes.tricontour` is a separate method and only `Axes.contour` is patched. Nothing else differs — measured:

```python
cs = ax.tricontour(x, y, z, levels=[1.0, 2.0, 3.0])
type(cs).__name__            # 'TriContourSet'
isinstance(cs, ContourSet)   # True
cs.filled                    # False
list(cs.levels)              # [1.0, 2.0, 3.0]
len(cs.get_paths())          # 3
```

and the existing reader already handles it unchanged, with no edit to `ContourPlot`:

```python
tag(cs)
ContourPlot(ax, contour_set=cs)._extract_plot_data()
# 3 series, of 29 / 41 / 43 points
```

So the layer is a patch away, not a reading away.

## The chart

`tricontour` contours a field sampled at **scattered** points rather than on a grid — it triangulates them first. That is the same chart `ax.contour` draws and the same thing a reader navigates: the level is the object, and `TriContourSet.levels` carries it as a number, which is what made the contour trace the right home in the first place (xability/maidr#802).

## Scope

- Patch `Axes.tricontour` the way `Axes.contour` is patched, sharing the same body rather than copying it.
- `Axes.tricontourf` stays unpatched for the same reason `contourf` does: it draws the bands *between* levels, so one outline runs along two different level curves and announcing it as a level's own curve would be right for half its points. Worth an explicit test, since the two names sit next to each other and the pair is easy to widen by accident.
- The empty-set guard in the existing patch applies unchanged — a triangulation where every level falls outside the data draws nothing, and registering it would put a phantom layer in the schema (#421).

Measured on matplotlib 3.9.4.

jooyoungseo:
Closes #546.

## The defect

`ax.tricontour` registered nothing, while `ax.contour` on the same field has been fully navigable since #540. The reason was the patch, not the reading: `tricontour` is its own method and only `contour` was wrapped.

## Nothing else differs

```python
cs = ax.tricontour(x, y, z, levels=[1.0, 2.0, 3.0])
type(cs).__name__            # 'TriContourSet'
isinstance(cs, ContourSet)   # True
cs.filled                    # False
list(cs.levels)              # [1.0, 2.0, 3.0]
len(cs.get_paths())          # 3
```

and `ContourPlot` read it correctly **with no edit at all** — three series of 29, 41 and 43 points — before this wrapped anything:

```python
tag(cs)
ContourPlot(ax, contour_set=cs)._extract_plot_data()
```

So the patch function is shared rather than copied: the two calls differ only in where the field was sampled, `tricontour` triangulating scattered points instead of taking a grid. That is the same chart and the same navigable object — the level, carried as a number, which is what made the contour trace the right home (xability/maidr#802).

## `tricontourf` stays refused

For the reason `contourf` does: it draws the bands *between* levels, so one outline runs along two different level curves and announcing it as a level's own curve would be right for half its points.

Pinned in its own test rather than left to `contourf`'s, because the two names sit beside each other in the same module and are exactly the pair a later widening would take together by accident — measured, patching it turns that test red on its own.

## The curves are checked against the field, not against a recording

The fixture is a bowl (`z = x² + y²` at 120 scattered points), so the curve at level L is the circle of radius √L — checkable in closed form without trusting the triangulation to have produced anything in particular:

```python
radii = [np.hypot(p["x"], p["y"]) for p in curve]
assert np.allclose(radii, np.sqrt(2.0), atol=0.15)
```

The tolerance is the triangulation's own: a contour on a triangular mesh interpolates linearly along each edge, so a curve sits slightly inside a convex level set.

## Falsification

| mutation | result |
|---|---|
| don't patch `tricontour` | 4 tests red |
| also patch `tricontourf` | 1 red |
| drop the empty-set guard | 2 red |

Every guard earns its place. The empty-set guard is the existing one, re-exercised on this path: a triangulation where every level falls outside the data draws nothing, and registering it would leave a layer with no curves for a reader to walk into (#421).

Selectors are asserted against the rendered SVG, not as strings — a selector matching nothing reads correctly through audio, braille and text while lighting nothing up (xability/maidr#814), and that is asked separately of `tricontour` because a different artist writes the group.

## Checks

`uv run pytest`: 2502 passed, 18 skipped. `ruff check`: clean.

---

### DIFF
--- a/maidr/core/enum/maidr_key.py
+++ b/maidr/core/enum/maidr_key.py
@@ -20,6 +20,11 @@ class MaidrKey(str, Enum):
     # Plot legend keys.
     CAPTION = "caption"
     LABEL = "label"
+    #: What this layer *is*, when a figure holds several of a kind: the hue
+    #: group a split scatter reads, say. Distinct from `TITLE`, which names
+    #: the chart and is carried by every layer of a figure alike
+    #: (xability/maidr#828).
+    NAME = "name"
     SUBTITLE = "subtitle"
     TITLE = "title"
 
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from matplotlib.axes import Axes
 from matplotlib.lines import Line2D
+from matplotlib.collections import PolyCollection
 from matplotlib.patches import StepPatch
 from maidr.core.enum import PlotType
 from maidr.core.plot.areaplot import AreaPlot
@@ -22,6 +23,7 @@ from maidr.core.plot.pointplot import PointPlot
 from maidr.core.plot.scatterplot import ScatterPlot
 from maidr.core.plot.regplot import SmoothPlot
 from maidr.core.plot.stairs import StairsPlot
+from maidr.core.plot.stepped_histogram import SteppedHistPlot
 from maidr.core.plot.stepplot import StepPlot
 from maidr.core.plot.violin_kde_plot import ViolinKdePlot
 from maidr.core.plot.violin_box_plot import ViolinBoxPlot
@@ -97,6 +99,11 @@ class MaidrPlotFactory:
             # find and no per-bin artist to look for.
             if isinstance(kwargs.get("step_patch"), StepPatch):
                 return StairsPlot(single_ax, **kwargs)
+            # `sns.histplot(element="step"/"poly")` draws the same
+            # distribution as one closed outline, so there is no container to
+            # find and the call hands its `PolyCollection` over instead.
+            if isinstance(kwargs.get("collection"), PolyCollection):
+                return SteppedHistPlot(single_ax, **kwargs)
             return HistPlot(single_ax)
         elif PlotType.LINE == plot_type:
             if PlotDetectionUtils.is_mplfinance_line_plot(single_ax, **kwargs):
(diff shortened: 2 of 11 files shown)


jooyoungseo:
Good question on the tolerance — it deserved measuring rather than defending. Tightened in fbf9857.

Per level, on this fixture:

```
level 1.0   want r=1.0000   max deviation 0.1603   gap to next level 0.4142
level 2.0   want r=1.4142   max deviation 0.0733   gap to next level 0.3178
level 3.0   want r=1.7321   max deviation 0.0383   gap to next level 0.2679
```

Two things fall out. The test asks about level 2.0, whose worst point is 0.073 inside the true circle — so `atol=0.15` was about twice as loose as it needed to be. And level 1.0's deviation is 0.160, which would have *failed* at 0.15; the mesh error is largest on the smallest ring, where the same edge length subtends more of the curve. Worth knowing before anyone extends the test to the other levels.

Now `atol=0.1`: clears the mesh with room at level 2.0 and is still more than three times tighter than the 0.318 it would have to swallow to accept a neighbouring level. Checked in the direction that matters rather than assumed — asserting the level-2 curve against level 1's radius fails at this tolerance, so it does discriminate rather than pass for almost anything.

The numbers and the reasoning are now in the test's docstring, so the next person to touch it does not have to re-derive them.

---

PR #549, opened 2026-08-20
docs: tell Shiny readers what the render lock does not cover

jooyoungseo:
`docs/index.qmd`'s "Shiny and other async servers" callout now ends on the render lock — that renders of one figure are serialised, however the figure is reached. Read straight through, that sounds like shared figures are handled.

They are not. The lock serialises **renders against each other**. It does nothing about an app's own render function drawing into a figure another session is in the middle of rendering, which produces a chart whose SVG and announced data disagree: visible to a sighted reader, invisible to a screen-reader one.

Since #541 that condition warns at the moment it happens, and **the audience for that warning is exactly the reader of this callout** — someone running concurrent sessions against a figure they cache. The callout now shows the message they will see and how to silence it if the sharing is deliberate.

It also states the limit the warning cannot state about itself: it notices artists appearing, disappearing or being relabelled, not a value changed in place on an existing one, so **hearing nothing is not a guarantee that a render was clean**. That qualification is in the merge commit for #541 and in the module docstring; a reader of the docs should not have to find either.

## Scope

One paragraph added to an existing callout, prose only. No code.

Refs #530.

---

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -328,6 +328,12 @@ Same throughput either way — the work did not get cheaper, it stopped being in
 
 Two things this does not do. Renders of the **same figure** are serialised, because `savefig` writes `fig.dpi` for its duration and two concurrent writes leave one chart drawn at the other's scale; distinct figures, which is the usual per-session shape, run in parallel. That lock is held by the render path itself rather than by this integration, so a figure is rendered one at a time however it is reached — through Shiny, through Streamlit, or from a thread of your own. And the chart still travels over the websocket on every flush — roughly 25 KB, or a couple of megabytes with `use_cdn=False`; that part is tracked in [#534](https://github.com/xability/py-maidr/issues/534).
 
+**What the lock cannot do is stop your app.** It serialises renders against each other; it does not stop a render function drawing into a figure another session is in the middle of rendering. That produces a chart whose SVG and announced data disagree — visible to a sighted reader, invisible to a screen-reader one — so maidr says so, naming the figure:
+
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Thank you for checking the quote against the f-string character by character — that is the check this PR most needed, and it is the one I could most easily have got subtly wrong.

**The drift observation is worth more than a note, so I took it.** You are right that there is no doc/code sync check in the repo, and that quoting the message makes it part of the public-facing contract. It is also the same class as two things already fixed today — a docs link to a closed issue (#537) and a test naming a symbol that had moved (#541) — which is three instances of "the prose kept describing the old thing" in one day.

`tests/core/test_docs_quote_the_real_warning.py` now raises the real warning, extracts the quoted line from `docs/index.qmd`, and compares them with the figure's name removed from both sides — that being the one part which legitimately differs between a quote and any occurrence of it.

Falsified from both directions, which matters because drift can start on either side:

| mutation | result |
|---|---|
| reword the message, leave the docs | fails **3 of 3** |
| reword the docs, leave the message | fails **3 of 3** |

**On the blockquote being a new pattern in this file:** noted, and I left it. A fenced code block would read as something to copy and run, which this is not — it is output to recognise. If the docs site later settles on a convention for quoting log output, this is one line to change and the test above will keep it honest.

## Verification

`pytest 0 / ruff 0 / uv lock --check 0` — **2,526 passed, 14 skipped**. (`ruff check` reports the 6 pre-existing re-export findings in `maidr/core/**/__init__.py`; the files here are clean.)

---

Issue #480, opened 2026-08-17
Every horizontal plotly bar chart is silent and announced inside out


jooyoungseo:
## Summary

`go.Bar(y=cats, x=vals, orientation="h")` emits its points in the **horizontal**
arrangement — the measure in `x` — but never emits an `orientation` key. The
core defaults a missing orientation to vertical and then reads `point.y` as the
magnitude, which on a horizontal layer is the category name.

The result is a chart that loads, highlights and navigates normally, sounds
**nothing**, and announces the measure and the category the wrong way round.

## Measured

The emitted layer, keys in full:

```
keys: ['MaidrKey.AXES', 'MaidrKey.DATA', 'MaidrKey.ID',
       'MaidrKey.SELECTOR', 'MaidrKey.TITLE', 'MaidrKey.TYPE']

data: [{x: 30, y: 'apple'}, {x: 70, y: 'banana'}, {x: 50, y: 'cherry'}]
```

No `orientation`. `horz` appears nowhere in the rendered HTML.

Feeding exactly that layer to a real `BarTrace`:

```
{"empty": false,
 "raw": null,
 "text": {"main":  {"label": "X", "value": 30},
          "cross": {"label": "Y", "value": "apple"},
          "mainAxis": "x", "crossAxis": "y"}}
```

- `freq.raw` is **`null`** — there is no magnitude to pitch, so every bar is
  silent.
- The announcement is inverted: `30` is given as the point's identity and
  `"apple"` as its value.

## Scope

Three types, all in the bar family:

| type | `orientation` emitted | first point (horizontal) |
|---|---|---|
| `bar` | **no** | `{"x": 30, "y": "apple"}` |
| `stacked_bar` | **no** | `{"x": 30, "z": "u", "y": "apple"}` |
| `dodged_bar` | **no** | `{"x": 30, "z": "u", "y": "apple"}` |
| `hist` | yes | — |
| `box` | yes | — |

The data itself is right in every case — `paired_axes()` is deliberately
symmetric and hands back the horizontal arrangement correctly. Only the key
that says how to read it is missing.

## This is the mirror of a bug already fixed twice

- xability/r-maidr#184: the key without the layout — `"horz"` declared over
  vertical points. Same symptom, `freq.raw = null`.
- Here: the layout without the key.

Both produce a silent chart, from opposite halves of the same pairing. The
contract is now written down at `MaidrLayer.orientation` (xability/maidr#947):
*the bar family swaps `x` and `y`; nothing else does.*

[...]

PR #550, opened 2026-08-20
feat(core): read ax.eventplot as the raster of event times it draws

### LINKED ISSUE #548: ax.eventplot reads as nothing, though every row's event times are on the artist
## What happens

```python
ax.eventplot([times_a, times_b, times_c])
```

registers no layer. The chart is a raster plot — one row per series, a tick at each event time — and it is the standard way to draw spike trains, event timelines and arrival processes.

## The data is fully on the artist

`eventplot` returns one `EventCollection` per row, and each carries everything a reading needs:

```python
coll.get_positions()    # [2.252 3.002 6.251 7.757 8.972]   the events
coll.get_lineoffset()   # 0                                  this row's place
coll.get_linelength()   # 1                                  tick height
coll.get_orientation()  # 'horizontal'                        which axis is time
```

for three rows of 5, 3 and 6 events, and `orientation` flips to `'vertical'` under `ax.eventplot(..., orientation="vertical")`.

## What it should read as

**A scatter per row**, not a spike. The Observable adapter's `spike` reading (#1098 upstream) is for a mark that *stands a magnitude at a place* — the spike's length is data. An eventplot's ticks are all the same length: `get_linelength()` is one number for the whole row, so the height is decoration and only the position is data.

So each row is a series of positions at a fixed offset on the other axis, which is exactly a scatter — and `ScatterPoint.yLabel` (xability/maidr#927) has somewhere to put the row's name when the axis carries one. One layer per row also matches how the project already handles a chart drawn as several artists: one collection, one layer, named so two of a kind can be told apart (xability/maidr#828).

Reading all rows as a single layer would be wrong for the reason #426 gives about scatters: the rows are separate series, and merging them announces one cloud where the chart shows three.

## What needs care

- **Orientation.** `orientation="vertical"` puts the events on y and the row offsets on x. The layer has to swap both the data and say `orientation` rather than only one of them, which is the shape of the bug r-maidr#189 and py-maidr#480 both had.
- **Selectors.** Not yet measured — the chart currently produces no SVG at all through `maidr.render`, so the element structure has to be established before anything is promised. If a per-event element exists, each row's layer names its own; if not, the layer should announce and highlight nothing and say so, as the `geom_raster` reading does in r-maidr#193.
- **An empty row.** `ax.eventplot([[], times])` draws a collection with no segments. It should not become a layer with no points to walk into (#421).

Measured on matplotlib 3.9.4.

jooyoungseo:
Closes #548.

## The defect

`ax.eventplot([...])` registered nothing. It is the standard way to draw a spike train, an arrival timeline or a log of occurrences — a tick at every event time, one row per series — and a blind reader got a picture.

## Everything a reading needs is already on the artist

`eventplot` returns one `EventCollection` per row:

```python
coll.get_positions()    # [2.252 3.002 6.251 7.757 8.972]
coll.get_lineoffset()   # 0
coll.get_linelength()   # 1
coll.get_orientation()  # 'horizontal'
```

## A scatter, not a spike

A spike stands a **magnitude** at a place and its length is the data — that is what the Observable adapter's `spike` reading (xability/maidr#1098) is for. An event plot's ticks are all one length: `get_linelength()` is a single number for the whole row, so the height is decoration and only the position is data. What the reader navigates is a series of positions.

**One layer per row**, not one for the chart, for the reason #426 gives about scatters drawn as several collections: the rows are separate series, and merging them announces one cloud where the chart shows several.

## Orientation is in the coordinates, not in a key

```python
ax.eventplot(rows)                          # (1.0, 0.0) (4.0, 0.0) (7.0, 0.0)
ax.eventplot(rows, orientation="vertical")  # (0.0, 1.0) (0.0, 4.0) (0.0, 7.0)
```

No `orientation` key is emitted, and that is deliberate: a scatter is symmetric and its trace does not read the key, so the swap has to be in the data or it does not happen. Declaring one without the other is the defect #480 and xability/r-maidr#189 both had — declaring it *as well* would be the same mistake from the other side.

## Three things the drawing decides

**Highlighting works.** Measured, a row's collection is written as one `<g>` carrying one `<path>` per event, in position order:

```
<g id="maidr-…">
 <path d="M 73.83 93.89 L 73.83 174.53"/>
 <path d="M 110.05 93.89 L 110.05 174.53"/>
</g>
```

so an event has an element of its own and a row addresses only its own. The test resolves every selector against the rendered SVG rather than trusting the string — a selector that lands in another row's group reads correctly through audio, braille and text while outlining the wrong ticks, which is the blind spot xability/maidr#814 names. It also asserts the two layers resolve into **two** groups, since a shared one would mean both were addressing the same elements while announcing different events.

**An empty row is not a layer.** `ax.eventplot([[], times])` draws a collection with no positions; registering it would leave a layer to walk into and find nothing (#421). The rows that did draw keep their own names, so skipping one does not rename the rest — which is why the name is looked up by the row's own index rather than by its place among the registered layers.

**A row is not named after its own number.** An unlabelled axis still has ticks, and measured, their text is the offset — `'0.0'`, `'1.0'`, with matplotlib's Unicode minus `−`. Taking it would "name" every row the coordinate the payload already carries, so a reader switching layers would hear a number they were about to be told anyway. Named ticks are used:

```python
ax.set_yticks([0, 1], ["neuron A", "neuron B"])   # names: 'neuron A', 'neuron B'
ax.eventplot(rows)                                 # names: None, None
```

The comparison is numeric rather than string-wise, because the same offset is written `0`, `0.0` or `−0.5` depending on the formatter.

## The stacked axis is named

"Row" where the caller named nothing, asked of the axes rather than of the payload the base builds — that fills `"X"` and `"Y"` as placeholders, so a check on what came back would leave a reader told the series are stacked along "Y". A caller who sets `ax.set_ylabel("neuron")` keeps their label.

## Falsification

| mutation | result |
|---|---|
| register only the first row | 7 red |
| ignore `get_orientation()` | 1 red |
| register empty rows too | 1 red |
| name a row after its own number | 1 red |
| leave the stacked axis as the placeholder | 1 red |
| skip the gid assignment (selectors go empty) | 1 red |

Every guard is load-bearing. The gid is assigned at registration rather than read at render for the reason `HexbinPlot` and `contour.tag` give — matplotlib stamps one at *draw* time and the schema is built first, so reading it later finds `None` and the layer ships announcing correctly and highlighting nothing.

## Checks

`uv run pytest`: 2536 passed, 18 skipped. `ruff check`: clean.

---


jooyoungseo:
All three points applied in bd1cf36 — and measuring the second one turned up something worse underneath it.

**The `lineoffsets` bug was real, exactly as described.** Measured:

```python
ax.set_yticks([0, 2], ["neuron A", "neuron B"])
out = ax.eventplot(rows, lineoffsets=2)
[c.get_lineoffset() for c in out]   # [0, 2]
# names before: ['neuron A', None]   ← "neuron B" at y=2, looked up at 1.0
# names after:  ['neuron A', 'neuron B']
```

Every coordinate stayed right while the layer went unnamed, which is exactly the kind of gap no assertion about the data would show. Fixed by keying on `row.get_lineoffset()` rather than recomputing from `lineoffsets` — the artist's own answer, so it needs no assumption about how a scalar is broadcast. `test_named_rows_survive_a_non_default_spacing` locks it in, and reverting to the index lookup turns it red.

**The `reads()` point was right, but the real problem was bigger than a phantom layer — it was a crash.** `get_positions()` does not merely mis-handle a non-finite row, it **raises**:

```python
ax.eventplot([[2.0, float("nan"), 5.0]])
coll.get_segments()   # [(2, 2), (2, 2), (0,)]
coll.get_positions()  # IndexError: too many indices for array
```

matplotlib gives a non-finite event a degenerate segment of shape `(0,)` and `get_positions` indexes it as `segment[0, pos]`. So it is not the all-non-finite corner — **one** NaN in an otherwise ordinary row is enough, and this would have raised inside the caller's own `ax.eventplot` call on a chart that draws fine today. Thank you for pulling the thread; I would have shipped it.

The reader now works off `get_segments()` throughout, which answers for every row, and `reads()` asks whether anything drawable came back.

Two things fell out of that measurement and are now written into the code rather than assumed:

- matplotlib emits **one `<path>` per segment**, degenerate ones included, so a row with a missing value has one element more than it has events. The selectors are keyed on segment indices for that reason, and `test_a_missing_value_leaves_an_element_the_reader_is_never_sent_to` asserts the surplus against the rendered document.
- `EventCollection.set_positions` sorts with `np.sort`, which puts every non-finite value **last**. So a degenerate segment can never precede a drawable one, and the segment index and a count of the announced points cannot currently disagree — the mutation that swaps them stays green. The segment index is kept anyway, because it is what the document is numbered by and the version that stays right if that sort ever changes, and the comment says plainly that no test separates them because no chart can.

The sort also means a row is announced left to right rather than in the order the caller wrote it, which the test now pins.

The import ordering nit is fixed too.

Suite: 2540 passed, 18 skipped.

---

jooyoungseo:
Three of the four applied in 7a67f70; the fourth declined with a reason.

**The docstring point was right, and it was backwards in the way you said.** I checked rather than took it: `_legend_title()` is called from `_extract_axes_data()`, which `render()` runs on every render, so the legend title genuinely is read live and a later retitle does show up. `_row_labels` is the opposite — frozen at registration. So the comment cited the one precedent that contradicts it. It now cites `scatterplot.hue_groups()`, which is captured at patch time as this is, and says outright that the legend title is the opposite case, so the next reader who checks the claim finds it holds instead of finding it inverted.

**`_marks` is declared in `__init__`** and read directly rather than through a `getattr` default. Same behaviour, one less thing inferred from call order.

**The multi-`Axes` gap was real, and it is now covered.** `test_a_raster_beside_another_chart_stays_in_its_own_panel` draws a two-row raster in the left panel and an ordinary scatter in the right, and asserts the grid holds `[2, 1]` layers with the right coordinates in each cell. Falsified: replacing `FigureManager.get_axes(rows[0])` with `plt.gca()` in the patch turns exactly that one test red and nothing else — so nothing else in the suite was watching panel placement, which is what you predicted.

**`events()` being computed twice I left alone.** Threading it through `create_maidr`'s kwargs would put a derived list in the same dict the caller's own kwargs travel in, and the recompute is a walk over one row's segments. Not worth the coupling at event-plot sizes.

Suite: 2541 passed, 18 skipped. `ruff check --diff` clean.

---

PR #551, opened 2026-08-21
fix(core): name a heatmap's cells from the grid, not from its axis ticks

### LINKED ISSUE #526: A numeric-axis heatmap is labelled with its axis ticks, not its cells
`HeatPlot._extract_plot_data` fills `x` and `y` from `extract_level(self.ax, ...)`, which reads the axes' **tick labels**. On a categorical axis those are the cells. On a numeric one they are not, and there are usually more of them than there are cells.

## Measured

matplotlib 3.9.4, seaborn 0.13.2. Every case draws a **2 × 3** grid:

| call | grid | x labels | y labels | |
|---|---|---|---|---|
| `ax.hist2d(a, b, bins=(3, 2))` | 2 × 3 | **9** | **7** | mismatch |
| `ax.pcolormesh(z)` where `z` is 2 × 3 | 2 × 3 | **7** | **9** | mismatch |
| `ax.imshow(z)` | 2 × 3 | **7** | **9** | mismatch |
| `sns.heatmap(z)` | 2 × 3 | 3 | 2 | ok |

```
ax.hist2d      x: ["−2.0","−1.5","−1.0","−0.5","0.0","0.5","1.0","1.5","2.0"]
               y: ["−3","−2","−1","0","1","2","3"]
sns.heatmap    x: ["0","1","2"]
               y: ["0","1"]
```

`sns.heatmap` lines up because its axis is categorical and its ticks *are* the cells. Every numeric-axis heatmap disagrees.

## Why it matters

The grid is navigated by cell. A reader moving to the second of three columns is read a label drawn from a list of nine — a number off matplotlib's tick locator, chosen to look tidy on an axis, with no relationship to which cell the cursor is in. The count is right and the values are right; the coordinates are somebody else's.

It is quiet, too. Nothing raises, the layer renders, and the counts are correct, so the reading sounds authoritative.

## What the labels should be

Every one of these artists knows its own cell boundaries:

- `pcolormesh` / `hist2d` — the `QuadMesh`'s coordinates are the bin edges; `hist2d` also returns `xedges` and `yedges` directly.
- `imshow` — the extent, defaulting to `-0.5 … n-0.5`, so cell *i* is centred on *i*.
- `sns.histplot(x=, y=)` — the same `QuadMesh`, once #522 lands and it is read at all.

So a cell can be named by its edges (`xMin`/`xMax`, as `hist` already does) or by its centre. Which of the two is a decision about what a reader wants to hear at a heatmap cell, and is worth making deliberately rather than by whichever is easier to reach — hence filing rather than guessing.

## Related

Found while measuring #522, which routes `sns.histplot(x=, y=)` into this same reading. That PR makes a previously silent chart readable; it does not make this labelling any better or worse, and the two are independent — this reproduces on `ax.hist2d` alone, with nothing from #522 involved.

jooyoungseo:
Closes #526.

## What was wrong

`HeatPlot._extract_plot_data` filled `x` and `y` from `extract_level`, which reads the axes' **tick labels**. Reproduced on matplotlib 3.9.4 / seaborn 0.13.2, every case drawing a **2 × 3** grid:

| call | grid | x labels | y labels | |
|---|---|---|---|---|
| `ax.hist2d(a, b, bins=(3, 2))` | 2 × 3 | **9** | **7** | ✗ |
| `ax.pcolormesh(z)` | 2 × 3 | **7** | **9** | ✗ |
| `ax.imshow(z)` | 2 × 3 | **7** | **9** | ✗ |
| `sns.heatmap(z)` | 2 × 3 | 3 | 2 | ✓ |

```
hist2d      x: ['−2.0','−1.5','−1.0','−0.5','0.0','0.5','1.0','1.5','2.0']
            y: ['−3','−2','−1','0','1','2','3']
```

`sns.heatmap` lines up because its axis is categorical and its ticks *are* the cells. Every numeric-axis heatmap disagreed, and nothing raised — the counts and values were right, so a reader moving to the second of three columns was read a number off matplotlib's tick locator with no way to tell it was somebody else's coordinate.

## After

```
hist2d      x: ['-1.6648', '-0.19792', '1.26896']   y: ['-2.0627', '1.35646']
pcolormesh  x: ['0.5', '1.5', '2.5']                y: ['0.5', '1.5']
imshow      x: ['0', '1', '2']                      y: ['0', '1']
sns.heatmap x: ['0', '1', '2']                      y: ['0', '1']      (unchanged)
```

## Centre, not edges — and why

The issue left this open deliberately, so here is the reasoning rather than a preference.

`HeatmapData.x` and `.y` in the core grammar are `string[]` — **one label per column**, with no per-cell `min`/`max` the way `HistPoint` has. Both a centre and an edge range fit in a string, so the question is which a reader wants at a cell.

`HexbinPoint` already answers it for the same family, and its own documentation gives the reason: *"The centre is carried per bin rather than derived from a lattice origin and a cell size"* — a bin is named by where it is. The label is announced on **every move of the cursor**, an edge range roughly doubles that announcement's length, and consecutive centres already give the width by their spacing. So: centre, consistent with the neighbouring decision.

Formatted `.6g`, and deliberately **not** `self._fmt`: that is the caller's format for the cell *values* — what `sns.heatmap` annotates a cell with — and a coordinate is not one of those.

## Which answer wins

The ticks are checked against the cells rather than assumed to be them, so a chart that names its own grid keeps its names. The check is on **where the ticks sit**, not on how many there are — a locator that happened to draw one tick per column would otherwise have its numbers taken as the names, and there is a test for exactly that.

`extract_level_positions` is new, next to `extract_level` and filtered by the same rule, so the labels and their positions stay index-aligned however that rule changes.

## An image's row order

`get_extent()` reports its two ends bottom-then-top **whichever way up the image is drawn**:

```
origin="upper"   extent = (-0.5, 2.5,  1.5, -0.5)
origin="lower"   extent = (-0.5, 2.5, -0.5,  1.5)
```

Row 0 is at −0.5 in both, so which end it sits against comes from `origin`, not from the order. Reading them in the given order names the rows backwards for one of the two, and the values are emitted in array order either way — so the names have to be. There is a test per origin, and each pins one direction of the constant.

## Falsification

| mutation | result |
|---|---|
| go back to the tick labels unconditionally | 8 of 10 red |
| accept the ticks whenever the *count* matches, without checking where they are | the coincidence test red |
| `first, last = bottom, top` (ignore `origin`) | the default-`imshow` test red |
| `first, last = top, bottom` (the opposite constant) | the `origin="lower"` test red |

Suite: 2552 passed, 18 skipped. `ruff check --diff` clean.

## Noted, out of scope

For `ax.imshow(z, origin="lower")` array row 0 is drawn at the **bottom**, and it is still emitted first — while `HeatmapData` documents rows as top-first. That is the xability/maidr#971 family of defect, it predates this change, and this change neither causes nor hides it: each row is now named with its own coordinate, which is true either way; only which direction <kbd>↑</kbd> walks is affected. Worth filing separately rather than folding in here.

---

### DIFF
--- a/maidr/core/plot/heatmap.py
+++ b/maidr/core/plot/heatmap.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
+import numpy as np
 import numpy.ma as ma
 from matplotlib.axes import Axes
 from matplotlib.cm import ScalarMappable
 from matplotlib.collections import PolyQuadMesh, QuadMesh
+from matplotlib.image import AxesImage
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Three of the four applied, and the second one turned out to be worth more than a comment.

**1. The truncated comparison is guarded.** `len(at) == len(centres)` is now required alongside the labels' length. I could not build a chart that separates the two — `extract_level_positions` has no `i < len(labels)` guard, so it is the longer of the pair and `zip` reaches every centre — but that is an argument about today's filter rather than about the code, and the guard costs nothing and states the invariant. No test pins it, because I could not construct one honestly; saying so rather than writing a test that passes for a different reason.

**2. The rectilinear assumption was real, and declining beats commenting.** Measured a sheared `pcolormesh(X, Y, Z)`, 2 × 3:

```
x row 0: [0.  1.  2.  3. ]
x row 1: [0.3 1.3 2.3 3.3]
x row 2: [0.6 1.6 2.6 3.6]
```

Naming every column after row 0 gives a reader `0.5, 1.5, 2.5` — true of one row of the grid and of no other, delivered with the same confidence as a correct name. That is the shape of #526 itself, reintroduced in a corner.

So the branch now checks that the rows agree about x and the columns about y, and returns `None` when they do not — which falls back to the ticks, the same path an unrecognised artist takes. That fallback is *still wrong* for this chart in exactly the way #526 describes, and the test says so: a grid whose cells cannot be located is left as it was rather than given names invented for it. Reading a curvilinear mesh properly is a different question from this one.

**3. `pcolormesh` exact centres are pinned** — `["0.5", "1.5", "2.5"]` / `["0.5", "1.5"]` — so the mesh branch is as precise as the other two.

**Also, on the `.6g` collision you flagged as theoretical: it is reachable.** Centres around 1e9 spaced by 1 all format to `1e+09`, and two cells with one name is worse than a long name — a reader moving between them is told they have not moved. The precision now escalates (6 → 12 → 17) until the names are distinct, with a test at 1e9.

Falsification of the two new guards:

| mutation | result |
|---|---|
| drop the axis-alignment check | the curvilinear test red |
| never raise the precision past `.6g` | the 1e9 test red |

On the docstring length: I left them. They are carrying the reasoning that makes the ticks-vs-cells and centre-vs-range decisions auditable, which is the same thing you found useful in the review, and the file's neighbouring comments (`_extract_scalar_mappable_data`'s note on the two mesh classes) are written the same way.

The lint failure was mine and unrelated: an `import maidr` the test did not need. CI pins ruff 0.3.4 and the sandbox has 0.15.8, which did not flag it — running `uvx ruff@0.3.4 check --diff .` locally now, and it is clean.

Suite: 2555 passed, 18 skipped.

---

jooyoungseo:
Round two, both points taken — and one of them turned into a gain rather than a decline.

**The two tick filters.** You were right that `extract_level_positions` held its own copy of `extract_level`'s dataLim filtering, and that the docstring's claim the two "stay index-aligned however that rule changes" was true only for as long as somebody edited both copies together. That is a claim about a future editor, not about the code. Both now route through one `_ticks_in_view(ax, key)` returning `(position, label)` pairs, so alignment holds by construction. `extract_level`'s shared-xtick fallback stays where it was; the one deliberate divergence — a *sibling* axes' labels have no positions on this axes to pair against — is now documented in the positions docstring rather than mirrored, and `_cell_names` already guards it by checking both lengths.

**Gouraud shading.** I went to measure this rather than reason about it, and the measurement inverted the fix. On matplotlib 3.9.4, a 2 × 3 `pcolormesh(z, shading="gouraud")` returns coordinates of shape `(2, 3, 2)` — against the flat-shaded `(3, 4, 2)`. That is one coordinate per *cell*, not one per boundary: with gouraud the values sit **at** the coordinates instead of filling the quads between them, so each coordinate already is a cell's position and there is no midpoint to take. So rather than declining to name a gouraud mesh, `_cell_centres` now returns the coordinates as-is when `len(edges) == count`, and such a mesh is named `'0','1','2'` / `'0','1'` exactly like `imshow` and `sns.heatmap`. Previously it fell through to nine tick labels for three columns — the very failure this PR exists to fix, reachable by one keyword argument.

Two tests cover the new ground: `test_a_gouraud_shaded_mesh_is_named_at_its_coordinates_not_between_them`, and `test_a_pcolor_grid_is_named_like_the_mesh_it_shares_a_contract_with` for `pcolor`'s `PolyQuadMesh`, which reaches the same branch through a different class.

Verified: `ruff check` clean, `uvx ruff@0.3.4 check --diff .` (CI's pin) exit 0, full suite **2557 passed, 18 skipped**.

---

PR #565, opened 2026-08-21
fix(heatmap): tell a picture from a grid of values instead of dying on both

### LINKED ISSUE #564: `ax.imshow()` of a photo kills the whole figure, and so does `ax.spy()` — a heat layer assumes every grid is scalar
Found by sweeping the matplotlib `Axes` plotting methods that have no patch, the way #214 came out of sweeping base R.

## Measured

Two shapes of array reach `HeatPlot` and raise at **render**, after the plotting line has already returned:

```python
fig, ax = plt.subplots()
ax.imshow(np.zeros((2, 2, 3)) + 0.5)      # an RGB image
maidr.render(fig)._repr_html_()
#> ValueError: could not convert string to float: '[0.5 0.5 0.5]'

fig, ax = plt.subplots()
ax.spy(np.eye(3))                          # a sparsity pattern
maidr.render(fig)._repr_html_()
#> ValueError: could not convert string to float: 'True'
```

`plt.show()` raises the same way. And it is not confined to the offending axes — **every other chart in the figure goes with it**:

```python
fig, (a, b) = plt.subplots(1, 2)
a.bar(["x", "y"], [1, 2])
b.imshow(photo)
maidr.render(fig)
#> ValueError: could not convert string to float: '[0.5 0.5 0.5]'
```

A perfectly readable bar chart, lost because a picture sits beside it.

| call | array | today |
| --- | --- | --- |
| `imshow(m)` — 2-D numeric | `(2, 2) int64` | reads as `heat` |
| `imshow(rgb)` | `(2, 2, 3) float64` | **ValueError at render** |
| `imshow(rgba)` | `(2, 2, 4) float64` | **ValueError at render** |
| `imshow(rgb_uint8)` | `(2, 2, 3) uint8` | **ValueError at render** |
| `imshow(bool_mask)` | `(2, 2) bool` | **ValueError at render** |
| `spy(m)` | `(3, 3) bool` | **ValueError at render** |
| `pcolormesh(bool_mask)` | `(2, 2) bool` | **ValueError at render** |
| `spy(m, marker=".")` | — | no scalar-mappable artist; nothing registers, no crash |

## Why it happens

One line, `maidr/core/plot/heatmap.py:295`:

```python
return [list(map(lambda x: float(format(x, self._fmt)), row)) for row in array]
```

`self._fmt` defaults to `""` (it exists for `seaborn.heatmap`'s cell-annotation format), and `format(x, "")` is `str(x)`:

- for a numpy bool, `format(np.True_, "")` is `'True'` — note that `format(np.True_, ".2f")` is `'1.00'`, so this only bites at the default format;
- for a row of an RGB image, `x` is a length-3 array and `format` gives `'[0.5 0.5 0.5]'`.

`float()` then raises, from inside `render`, on a figure the user has already drawn.

## They need different answers

**A boolean grid is readable and should be read.** `True`/`False` is 1/0, and that is exactly what a sparsity pattern is — `spy()` exists to show you where the non-zeros are. Coercing a boolean array to numbers before formatting gives the chart the reading it should always have had.

**An RGB or RGBA image is not a heatmap at all.** The last axis is colour, not a value: there is no scalar per cell to announce, and no colourbar meaning to give the `z` axis. The right outcome is for the layer **not to register**, so the figure renders as a plain SVG and anything else on it keeps working — the same outcome `quiver`, `barbs` and `streamplot` already have, which produce no layer and no crash. Inventing a reading for it (luminance? the mean channel?) would announce a number the chart does not show.

## Also turned up by the same sweep, not proposed here

Recorded so the sweep is written down somewhere. None of these raises; each is silence or an asymmetry, which is a different and much smaller problem than losing the figure:

| call | today | note |
| --- | --- | --- |
| `quiver`, `barbs`, `streamplot` | no layer | vector fields — no trace type in the core |
| `fill_betweenx(y, x1, x2)` | **no layer** | `fill_between` reads as `area` (#339); the transposed spelling does not |
| `fill(x, y)` | no layer | a closed polygon |
| `hlines` / `vlines` | no layer | while `axhline` / `axvline` on a charted axes leave the chart readable |
| `tricontourf`, `tripcolor` | no layer | the line form `tricontour` reads as `contour`; the filled forms do not — the same line/filled split r-maidr #211 records for `geom_contour_filled` |
| `acorr`, `xcorr` | no layer | drawn with `vlines`, so they inherit that gap |
| `matshow`, `psd`, `csd`, `specgram`, `magnitude_spectrum`, `phase_spectrum`, `angle_spectrum`, `triplot`, `clabel`, `loglog`, `semilogx`, `semilogy` | already read | via the artists they draw through |

`fill_betweenx` is the one of those worth its own issue.

jooyoungseo:
Closes #564.

## What was wrong

`ax.imshow` accepts three shapes: `(M, N)` of numbers is a heatmap, and `(M, N, 3)` / `(M, N, 4)` are **pictures**, their last axis colour rather than value. A boolean `(M, N)` mask is a fourth case, and the one `ax.spy()` draws.

All three of the last ones raised at **render**, after the plotting line had returned:

```python
ax.imshow(np.zeros((2, 2, 3)) + 0.5)
#> ValueError: could not convert string to float: '[0.5 0.5 0.5]'

ax.spy(np.eye(3))
#> ValueError: could not convert string to float: 'True'
```

and it was never confined to the offending axes — the render walks every layer, so a bar chart drawn beside a photograph died with it:

```python
fig, (a, b) = plt.subplots(1, 2)
a.bar(["x", "y"], [1, 2])
b.imshow(photo)
maidr.render(fig)      #> ValueError
```

One line of `HeatPlot` assumed every grid it was handed was scalar:

```python
[list(map(lambda x: float(format(x, self._fmt)), row)) for row in array]
```

`self._fmt` defaults to `""` — it exists for `seaborn.heatmap`'s cell annotations — and `format` with an empty spec is `str`. For a numpy bool that is `'True'`; for a row of an RGB image, `x` is a length-3 array and it is `'[0.5 0.5 0.5]'`.

| call | array | before | after |
| --- | --- | --- | --- |
| `imshow(m)` | `(2, 2) int64` | `heat` | unchanged |
| `imshow(bool)` | `(2, 2) bool` | **ValueError** | `heat`, 1 and 0 |
| `spy(m)` | `(3, 3) bool` | **ValueError** | `heat`, the sparsity pattern |
| `pcolormesh(bool)` | `(2, 2) bool` | **ValueError** | `heat`, 1 and 0 |
| `imshow(rgb)` / `imshow(rgba)` / `imshow(rgb_uint8)` | `(2, 2, 3\|4)` | **ValueError** | no layer; figure renders |
| `bar` beside `imshow(rgb)` | — | **ValueError** | the bar chart reads |

## The two need different answers

**A mask is read.** `True` and `False` are 1 and 0, and showing where a matrix is non-zero is the whole purpose of `spy()`. The array is converted rather than the formatting special-cased, because the defect was only ever at the *default* format — `format(np.True_, ".2f")` was already `"1.00"`, so a fix keyed on the empty spelling would leave a seaborn caller's `fmt` producing something different from everyone else's. Both spellings now give the same numbers, which is asserted.

**A colour image is not registered.** There is no number per cell to announce and nothing for the colourbar the `z` axis describes to mean. Declining leaves the image drawn and unregistered, which is what `ax.quiver` and `ax.streamplot` already do: the figure renders, emits the existing "plot type(s) not yet supported … falling back to static image" warning, and anything else on it keeps working. Inventing a reading — a luminance, a channel mean — would announce a number the chart does not show.

The test is on the **third axis, not the dtype**: `imshow` takes RGB as floats in 0..1 or as `uint8` in 0..255, and a dtype-shaped guard would let the integer spelling through. A two-dimensional greyscale image stays a heatmap, which is the line the decline must not cross.

## Tests

`tests/core/plot/test_image_grid_kinds.py`, 11 cases, each driven through a real `maidr.render()`.

| mutation | result |
| --- | --- |
| the boolean coercion removed | 3 failures — `spy`, boolean `imshow`, boolean `pcolormesh` |
| `_is_colour_image` always False | 5 failures, including the bar chart beside the picture |
| `ndim >= 4` instead of `>= 3` | the same 5 |
| `ndim >= 2`, declining every grid | 7 failures — every reading case, including the greyscale image |

Full suite: `2636 passed, 18 skipped`. `ruff check --diff` clean on the changed files.

## Found by a sweep, recorded in the issue rather than fixed here

The issue carries the rest of the sweep over unpatched `Axes` methods. None of the others raises; each is silence or an asymmetry. The one worth its own issue is **`fill_betweenx`**, which registers no layer while `fill_between` reads as `area` (#339) — the same chart written the other way round. `tricontourf` and `tripcolor` are the same line-versus-filled split, with `tricontour` already reading as `contour`.

---

### DIFF
--- a/maidr/core/plot/heatmap.py
+++ b/maidr/core/plot/heatmap.py
@@ -292,4 +292,15 @@ class HeatPlot(
         else:
             self._support_highlighting = False
 
+        # A mask is a grid of numbers as much as a grid of counts is, and
+        # `ax.spy()` draws nothing else: its whole purpose is to show where a
+        # matrix is non-zero. It only failed at the default format --
+        # `self._fmt` is `""` unless seaborn's caller set one, and
+        # `format(np.True_, "")` is `"True"`, which `float` refuses. With any
+        # numeric format it already worked: `format(np.True_, ".2f")` is
+        # `"1.00"`. Converting first gives the same 1 and 0 under every
+        # format instead of only some of them (#564).
+        if array.dtype == bool:
+            array = array.astype(float)
+
         return [list(map(lambda x: float(format(x, self._fmt)), row)) for row in array]
--- a/maidr/patch/heatmap.py
+++ b/maidr/patch/heatmap.py
@@ -162,7 +162,9 @@ def heat(wrapped, _, args, kwargs) -> Axes | AxesImage | Collection:
     Axes or AxesImage or Collection
         Whatever the wrapped function returned: an ``Axes`` from seaborn, an
         ``AxesImage`` from ``imshow``, or the mesh the two ``pcolor`` variants
-        render.
+        render. The draw always happens; the return is the same either way.
+        No layer is registered when the artist turns out to hold a colour
+        image rather than a grid of values -- see :func:`_is_colour_image`.
     """
     # `seaborn.heatmap` draws through `Axes.pcolormesh`, and both are patched
     # here. Without this guard the inner call registers a second HEAT layer for
--- a/maidr/patch/heatmap.py
+++ b/maidr/patch/heatmap.py
@@ -61,6 +61,40 @@ def _grid_of(plot, ax: Axes | None):
     return drawn[-1] if drawn else None
 
 
+def _is_colour_image(grid) -> bool:
+    """
+    Whether the artist holds a picture rather than a grid of values.
+
+    ``ax.imshow`` accepts three shapes: an ``(M, N)`` array of scalars, which
+    is a heatmap, and ``(M, N, 3)`` / ``(M, N, 4)`` arrays whose last axis is
+    **colour**. The last two are photographs and rendered images, and there is
+    no number per cell to announce -- no value, and nothing for the colourbar
+    the ``z`` axis describes to mean.
+
+    Registered as a heatmap they did not merely read badly, they killed the
+    figure: ``HeatPlot`` formats each cell with ``float(format(x, fmt))``, and
+    for a row of an RGB image ``x`` is a length-3 array, so ``render`` raised
+    ``ValueError: could not convert string to float: '[0.5 0.5 0.5]'`` and
+    took every other chart on the figure with it (#564).
+
+    Declining leaves the image drawn and unregistered, which is what
+    ``ax.quiver`` and ``ax.streamplot`` already do -- the figure renders, and
+    a bar chart beside the picture keeps working.
+
+    Parameters
+    ----------
+    grid : Any
+        The artist :func:`_grid_of` found, or None.
+
+    Returns
+    -------
+    bool
+        True when the artist's array carries colour rather than values.
+    """
+    array = getattr(grid, "get_array", lambda: None)()
+    return array is not None and getattr(array, "ndim", 0) >= 3
+
+
 def _declares_fmt(wrapped: Callable) -> bool:
     """
     Whether a patched function takes ``fmt`` as a parameter of its own.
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Both review suggestions applied in 105bdbb.

- `heat()`'s **Returns** section now says the function can come back without registering a layer, and that the draw and the return value are unaffected either way — the inline comment at the branch is no longer the only place that says so.
- `_types()` catches `UnsupportedPlotError` rather than `Exception`. You are right that it is the better test: it is the exception `render`/`show`/`save_html` already catch for this case, and narrowing it means a *different* failure inside `get_maidr` fails the test instead of being read as "no layers registered", which was exactly the answer those cases expect.

`11 passed` on the file, `ruff check --diff` clean on both changed files.

---

PR #567, opened 2026-08-21
fix(area): name the axis a sideways band was actually drawn against

### LINKED ISSUE #566: `ax.fill_betweenx` announces its heights under the horizontal axis' title and its magnitudes under the vertical one
Split out of the sweep in #564, where I first mis-described this function (corrected [there](https://github.com/xability/py-maidr/issues/564#issuecomment-5374579218)).

## Measured

Two calls that draw transposes of each other produce **identical** payloads:

```python
fig, ax = plt.subplots()
ax.set_xlabel("horizontal")
ax.set_ylabel("vertical")

ax.fill_between(X, V)     # X = [1,2,3,4], V = [2,4,3,5]
#   axes:   {"x": {"label": "horizontal"}, "y": {"label": "vertical"}}
#   points: x=1 y=2, x=2 y=4, x=3 y=3, x=4 y=5

ax.fill_betweenx(X, V)
#   axes:   {"x": {"label": "horizontal"}, "y": {"label": "vertical"}}
#   points: x=1 y=2, x=2 y=4, x=3 y=3, x=4 y=5      <- byte for byte the same
```

`fill_betweenx(y, x1)` fills between the vertical positions `y` and the horizontal curve `x1`. So in that second call `[1, 2, 3, 4]` are **heights** on the vertical axis and `[2, 4, 3, 5]` are the **magnitudes** reaching out along the horizontal one. Every number is announced under the other axis' title: a reader is told "horizontal 1, vertical 2" where the chart shows the point at vertical 1, horizontal 2.

## Why it happens

`maidr/patch/fillbetween.py` reads the arguments correctly — `fill_betweenx` is wired as `_fill(..., position="y", value="x1", other="x2")`, which is right — and then hands them over positionally:

```python
FigureManager.create_maidr(ax, PlotType.AREA, x=positions, series=[values], ...)
```

`AreaPlot` writes `positions` into `MaidrKey.X` and `values` into `MaidrKey.Y`, and takes both axis titles straight off the axes. The two spellings therefore differ nowhere after the patch. The docstring says as much, and it is half right:

> The same chart with the axes exchanged: the shared axis is `y` and the magnitude runs along `x`. Emitted as an area either way, since **what a band measures does not change with which way it is drawn.**

What a band *measures* does not change. Which axis each number is *read against* does, and that is what the payload gets wrong.

## Proposed: swap the axis titles, not the data

The obvious-looking fix — put the magnitudes in `x` and the positions in `y`, so each number sits under its own axis — is wrong, and measurably so. The core sonifies an area trace's `y` and navigates along its `x`. Swapping the data would pitch `[1, 2, 3, 4]`: a rising ramp on every horizontal area chart ever drawn, regardless of the data, because those are positions.

The other obvious-looking fix — emit `orientation: "horz"` — would be a promise the core does not keep. `src/util/orientation.ts` marks `TraceType.AREA` as **not oriented**, deliberately:

> An area trace is a line with a fill, and it is navigated as one: along the series, then between series. Which way the band is drawn changes nothing about that, so — as with LINE, STEP and SMOOTH — there is no orientation to announce.

and the record's own header warns that reading it as though it answered the data-layout question "is how two bindings came to emit a horizontal bar with a category name where the magnitude belongs".

So: **swap the two `AxisConfig` entries** for `fill_betweenx`, leaving the data where the trace expects it. The core's Vega-Lite adapter already does exactly this, for exactly this reason, on the waterfall — a type whose axes it cannot swap either:

```ts
// A waterfall drawn along x holds those the other way round, so the two
// titles swap here — otherwise the step name is announced under the
// amount axis' title and the amount under the step's.
if (resolveRangedFields(encoding).isHorizontal) {
  [axes.x, axes.y] = [axes.y, axes.x];
}
```

After that swap: the announcement pairs each number with the axis it came from, and the pitch still tracks the magnitude. The format config has to travel with the title rather than with the key, so the swap must happen after `extract_format` merges — a currency formatter set on the x axis describes the horizontal numbers, which are the ones now in `y`.

## What that still leaves

Navigation. Left/right moves along the trace's `x`, which for a horizontal band is the *vertical* axis, so the right arrow walks the reader up the chart. That is what `orientation` exists to fix, and it cannot be said for an area trace today. Whether `AREA` should join the oriented types is a core question with a rationale already written down against it, so it belongs in xability/maidr rather than here; this issue is about the half that is wrong regardless of how that is answered.

Not in scope either: the **three**-argument form of both spellings, `fill_between(x, lo, hi)` and `fill_betweenx(y, lo, hi)`, which register nothing. That is #339's deliberate decision — the band's content is the gap rather than either edge — and it is symmetric across the two.

jooyoungseo:
Closes #566.

## What was wrong

Two calls that draw transposes of each other produced **identical** payloads:

```python
ax.set_xlabel("horizontal"); ax.set_ylabel("vertical")

ax.fill_between(X, V)     # X = [1,2,3,4], V = [2,4,3,5]
ax.fill_betweenx(X, V)
# both: axes x="horizontal" y="vertical", points (1,2) (2,4) (3,3) (4,5)
```

`fill_betweenx(y, x1)` fills between the vertical positions `y` and the horizontal curve `x1`. So in the second call `[1, 2, 3, 4]` are **heights** and `[2, 4, 3, 5]` are the **magnitudes** reaching out along x. A reader was told "horizontal 1, vertical 2" where the chart draws that point at vertical 1, horizontal 2 — every number under the other axis' title.

The patch already read the arguments correctly (`_fill(..., position="y", value="x1", other="x2")`) and then handed them over as `x=positions, series=[values]`, so the two spellings differed nowhere after it. The docstring said as much, and it was half right: what a band *measures* does not change with which way it is drawn — which axis each number is *read against* does.

## Why the titles move and not the data

Both of the obvious-looking fixes are wrong, and each is pinned by a test.

**Moving the data** would put the positions in `y`, which the core sonifies. Every sideways band would then play `[1, 2, 3, 4]` — a rising ramp whatever the data says, because those are positions, not magnitudes.

**Emitting `orientation: "horz"`** would be a promise the core does not keep. `src/util/orientation.ts` marks `TraceType.AREA` as **not** oriented, deliberately:

> An area trace is a line with a fill, and it is navigated as one: along the series, then between series. Which way the band is drawn changes nothing about that, so — as with LINE, STEP and SMOOTH — there is no orientation to announce.

and that record's own header warns that reading it as though it answered the data-layout question "is how two bindings came to emit a horizontal bar with a category name where the magnitude belongs".

So the two `AxisConfig` entries are exchanged instead, **after** the format config is merged into each — a currency formatter set on the x axis describes the horizontal numbers, which are the ones that move. The core's Vega-Lite adapter does exactly this, for exactly this reason, on a horizontal waterfall — a type whose axes it cannot swap either.

| call | axes before | axes after | points |
| --- | --- | --- | --- |
| `fill_between(x, y)` | x=horizontal, y=vertical | unchanged | unchanged |
| `fill_betweenx(y, x)` | x=horizontal, y=vertical | **x=vertical, y=horizontal** | unchanged |
| `stackplot(x, y)` | x=horizontal, y=vertical | unchanged | unchanged |

## What it does not fix

Navigation. Left and right still walk the trace's `x`, which for a sideways band is the vertical axis, so the right arrow moves the reader up the chart. That is what `orientation` would carry, and whether `AREA` should join the oriented types is a core question with a rationale already written down against it — recorded in #566 rather than decided here.

## Tests

`tests/core/plot/test_sideways_area.py`, 10 cases.

Two of them exist to pin the halves that must **not** move: `test_the_magnitudes_stay_where_the_trace_pitches_them` and `test_no_orientation_is_claimed`. Another compares the two spellings directly rather than each against a literal, so they cannot quietly converge back into the identical payloads this fixes.

The last one pins something the sweep got wrong on the way here: the **three**-argument form of both spellings registers nothing, symmetrically, which is #339's deliberate decision — the band's content is the gap rather than either edge. I first mis-read that as an asymmetry between `fill_between` and `fill_betweenx` and said so in #564; [corrected there](https://github.com/xability/py-maidr/issues/564#issuecomment-5374579218), and now pinned so nobody re-reads it that way.

| mutation | result |
| --- | --- |
| `fill_betweenx` not marked transposed | 3 failures |
| only the label strings swap, not the whole `AxisConfig` | the number-format test fails |
| the data is swapped instead of the titles | the magnitudes test and the two-spellings test fail |
| `transposed` defaults to True | `stackplot` is transposed and fails |

Full suite: `2646 passed, 18 skipped`. `ruff check --diff` clean on the changed files.

---

### DIFF
--- a/maidr/core/plot/areaplot.py
+++ b/maidr/core/plot/areaplot.py
@@ -57,6 +57,61 @@ class AreaPlot(MaidrPlot):
         self._series = list(kwargs.pop("series", []))
         self._labels = list(kwargs.pop("labels", []))
         self._collections = list(kwargs.pop("collections", []))
+        # `fill_betweenx` draws the same band with the axes exchanged: the
+        # positions run down the page and the magnitudes out along x. The
+        # numbers still go into the fields the trace reads them from -- see
+        # `render` for why the titles move instead of the data.
+        self._transposed = bool(kwargs.pop("transposed", False))
+
+    def render(self) -> dict:
+        """
+        Build the layer, exchanging the two axis titles for a sideways band.
+
+        `fill_betweenx(y, x1)` fills between the vertical positions `y` and
+        the horizontal curve `x1`, so its positions belong to the y axis and
+        its magnitudes to the x axis -- the mirror of every other chart this
+        class reads. Emitted unchanged, the two spellings produced byte-
+        identical payloads for charts that are transposes of each other, and
+        every number was announced under the other axis' title (#566).
+
+        The **titles** move rather than the data, which is deliberate twice
+        over:
+
+        - the core sonifies an area trace's `y` and steps along its `x`, so
+          putting the positions in `y` would pitch `[1, 2, 3, 4]` -- a rising
+          ramp on every sideways band ever drawn, whatever the data says;
+        - `orientation` is the field that says a chart is drawn sideways, and
+          `src/util/orientation.ts` marks `AREA` as not oriented on purpose,
+          so emitting one would be a promise the core does not keep.
+
+        The same exchange, for the same reason, is what the core's Vega-Lite
+        adapter does to a horizontal waterfall.
+
+        What it does not fix is navigation: left and right still walk the
+        trace's `x`, which for a sideways band is the vertical axis. That is
+        what `orientation` would carry, and it cannot be said for an area
+        trace today.
+
+        Returns
+        -------
+        dict
+            The layer, with `axes.x` and `axes.y` exchanged when the band was
+            drawn sideways.
+        """
+        maidr_schema = super().render()
+
+        if not self._transposed:
+            return maidr_schema
+
+        # After `super().render()` rather than in `_extract_axes_data`,
+        # because the format config is merged into each `AxisConfig` there: a
+        # currency formatter set on the x axis describes the horizontal
+        # numbers, which are the ones this moves.
+        axes = maidr_schema.get(MaidrKey.AXES)
+        if isinstance(axes, dict) and MaidrKey.X in axes and MaidrKey.Y in axes:
+            axes[MaidrKey.X], axes[MaidrKey.Y] = axes[MaidrKey.Y], axes[MaidrKey.X]
+
+        return maidr_schema
 
     def _extract_plot_data(self) -> list[list[dict]]:
         if self._x is None or not self._series:
--- a/maidr/patch/fillbetween.py
+++ b/maidr/patch/fillbetween.py
@@ -243,9 +243,9 @@ def fill_betweenx(wrapped, instance, args, kwargs) -> Collection:
     The same chart with the axes exchanged: the shared axis is ``y`` and the
     magnitude runs along ``x``. Emitted as an area either way, since what a
     band measures does not change with which way it is drawn -- but which
-    axis each number is *read against* does, and saying so is what
-    ``transposed`` carries. Without it the two spellings produced identical
-    payloads for charts that are transposes of each other (#566).
+    axis each number is *read against* does, and ``transposed`` is what says
+    so. See :meth:`maidr.core.plot.areaplot.AreaPlot.render` for what it
+    moves and why that rather than the data (#566).
     """
     return _fill(wrapped, instance, args, kwargs, "y", "x1", "x2", transposed=True)
 
--- a/maidr/patch/fillbetween.py
+++ b/maidr/patch/fillbetween.py
@@ -142,7 +142,16 @@ def _magnitudes(wrapped, args: tuple, kwargs: dict, position: str, value: str):
     return positions, values
 
 
-def _fill(wrapped, instance, args, kwargs, position: str, value: str, other: str):
+def _fill(
+    wrapped,
+    instance,
+    args,
+    kwargs,
+    position: str,
+    value: str,
+    other: str,
+    transposed: bool = False,
+):
     """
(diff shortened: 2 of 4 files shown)


jooyoungseo:
All three nits applied in 629a5b3, and one of them turned out to be worth more than a nit.

**`test_fill_betweenx_is_the_same_chart_turned_over`.** You checked it for a *regression* and correctly found none — it never asserted on labels. But that is the test whose name says what the two spellings share, and it was silent on the one thing they no longer do. It now sets the two axis titles and asserts the exchange, with a sentence saying the data assertions above it are unchanged *because* the titles move and the numbers do not. Nothing was rewritten; a reader landing there is no longer told half the story by omission.

**The redundant assertion.** Confirmed rather than taken on faith — `class MaidrKey(str, Enum)`, so the two checks are the same check. One assertion now, with a comment saying why one is enough.

**The local import.** Hoisted to module scope.

**On the duplicated rationale** — fair, and trimmed where it was a third copy: `fill_betweenx`'s docstring now points at `AreaPlot.render` for what moves and why. The two long ones that remain are deliberate and are the pattern this codebase uses: the method that does it, and the test module that pins it. Collapsing those two into one would leave whichever survives as the only place a reader is told, and they are read by different people at different times.

`24 passed` across both files, `ruff check --diff` clean.

Also confirming your `MaidrPlotFactory` note: `STACKED_AREA` reaches the same class and could carry `transposed`, but nothing passes it — `stackplot` has no sideways spelling in matplotlib, so there is no caller to give it one.

---

PR #569, opened 2026-08-21
fix(hue): name a pairplot's diagonals from the legend it builds afterwards

### LINKED ISSUE #561: A `pairplot` hue leaves its diagonal panels anonymous, because the legend does not exist yet when they register
Found while finishing #558. The two fixes there — #559 for the histogram split, #560 for the KDE naming — both work by matching an artist's colour against the legend swatch that names it. `sns.pairplot(df, hue=...)` is the one chart where that cannot fire, and the reason is **timing**, not the match.

## Measured

Every diagonal panel of a hue-grouped `pairplot` registers two `smooth` layers with no name:

```
[0][0][0] smooth  (no name)   [0][0][1] smooth  (no name)
[1][1][0] smooth  (no name)   [1][1][1] smooth  (no name)
```

while the off-diagonal scatters on the same figure come out named:

```
[0][1][0] point  name='y'     [0][1][1] point  name='x'
```

Instrumenting the naming helper at the moment each diagonal registers:

```
_curve_names: 2 fills, ax legend=False, fig legends=0
_curve_names: 2 fills, ax legend=False, fig legends=0
after:        fig legends=1, axes legends=[False, False, False, False, False, False]
```

There is **no legend anywhere** when the panels register — not on the axes, not on the figure. `PairGrid.add_legend()` builds the figure-level one after every panel has been drawn. By the time it exists, the layers are long since registered.

(The diagonals are drawn as filled `PolyCollection` bands rather than lines, which is why the count above says "fills". #560 names those too where a legend is present — a direct `kdeplot(hue=..., fill=True)` comes out `x, y` — so the artist type is not what is missing here.)

## Why the scatters are named and the diagonals are not

`hue_groups()` reads the axes legend at registration, and the off-diagonal panels do have one at that moment; `add_legend()` later moves it to the figure. The diagonals never get one of their own.

## What it would take

Both options are changes to **when** a layer decides what it says, which is why this is filed rather than folded into #560:

1. **Defer the match to render time.** The figure legend exists by then. It would also fix the case `hue_groups` already documents — a manual `ax.scatter(c=[...])` followed by `ax.legend(handles=[...])`, which is read as one layer today because the legend is built a line too late. The cost is that a caller who retitles or relabels the legend between drawing and rendering changes what the chart says, which is the same divergence `MaidrPlot._legend_title` records for `axes.z`.
2. **Read the figure's legend as well as the axes'.** Cheaper, and does nothing here on its own: at registration `fig.legends` is empty too.

So (1) is the only one that would work, and it is a real design decision about registration versus render.

## Not urgent

Nothing is lost — both distributions on every diagonal panel are announced, with their values. What is missing is which group each belongs to, so a reader stepping between two identical-sounding layers has to infer it. The off-diagonal panels of the same figure are named, which makes the gap visible rather than uniform.

## Related

- #558, #559, #560 — the hue split and naming this came out of
- #544 — `hue_groups()` and the registration-time legend read
- xability/maidr#828 — `MaidrLayer.name`, the field being filled

jooyoungseo:
Closes #561.

## What was wrong

#559 and #560 name a hue's layers by matching each artist's colour against the legend swatch that names it, and both did the match **at registration**. `sns.pairplot(hue=...)` is the one chart where that cannot fire, and the reason is timing rather than the match — `PairGrid.add_legend()` builds one figure-level legend after every panel has been drawn, so when the diagonals register there is no legend anywhere. Measured on the same figure:

```
smooth  (no name)   smooth  (no name)     <- the diagonals
smooth  (no name)   smooth  (no name)
point   name='x'    point   name='y'      <- the scatters beside them
```

The scatters are named because the off-diagonal panels *do* have an axes legend at that moment; `add_legend()` later moves it to the figure. The diagonals never get one of their own. Nothing is lost — both distributions are announced with their values — but a reader stepping between two identical-sounding layers has to infer which is which, on a figure whose other panels say.

## Three things, and none of them works alone

**1. `GROUP_NAME` takes a callable as well as a string**, and the two classes that honour it resolve one at render. #561 called this a real design decision about registration versus render, and the precedent settles it: `MaidrPlot._legend_title` already resolves `axes.z` at render and already accepts the same divergence — a caller who relabels the legend between drawing and rendering changes what the chart says.

**2. The legend is looked for on the figure** when the axes has none of its own, which is where a `PairGrid` puts it — and only when the figure carries **exactly one**, since two cannot say which names this axes' colours.

**3. `_group_names` delegates to `kdeplot._names_for`** rather than matching inline. Not tidying: a pairplot draws its bars translucent and its swatches opaque — measured, alpha 0.5 against 1.0 with identical hues — so the exact-RGBA comparison it made named nothing, and the guarded hue-only second pass the KDE side already had is what answers. Without it the `diag_kind="hist"` spelling stayed anonymous while the KDE one was fixed.

| chart | before | after |
| --- | --- | --- |
| `pairplot(hue=…)` diagonals | `smooth` ×4, unnamed | `y, x, y, x` |
| `pairplot(hue=…, diag_kind="hist")` diagonals | `hist` ×4, unnamed | `y, x, y, x` |
| the scatters beside them | `x, y, x, y` | unchanged |
| `kdeplot(hue=…)`, `histplot(hue=…)`, `histplot(hue=…, kde=True)` | named | unchanged |
| an ungrouped chart | unnamed | unnamed |

## Not fixed, and deliberately

The manual `ax.scatter(c=[...])` followed by a late `ax.legend(handles=[...])` that #561 also mentions. A hue-grouped scatter is split into one layer **per group** at registration, so deferring its name would mean deferring how many layers a figure has — a different change from deferring what one of them is called.

## Tests

`tests/core/plot/test_pairplot_group_names.py`, 12 cases.

Two are there to make the others bite. **"one figure legend names the axes below it"** builds a real figure legend from the drawn colours and asserts it names them — which is what makes **"two figure legends name nothing"** a test of the guard rather than of an empty legend. Its labels are supplied in seaborn's own reversed order, so a match by position would give each curve the other group's name and both would agree for the wrong reason.

**"a translucent bar is named by an opaque swatch"** states the colour half on its own: the bars' RGBA is disjoint from the swatches', and their three-channel hues are equal.

| mutation | result |
| --- | --- |
| the figure legend is not read | 5 failures |
| any number of figure legends accepted | "two figure legends name nothing" fails |
| the KDE names resolve at registration again | 3 failures |
| the histogram matches exact RGBA only | the `diag_kind="hist"` case fails |

Full suite: `2658 passed, 18 skipped`. `ruff check --diff` clean on the changed files.

---

### DIFF
--- a/maidr/core/plot/histogram.py
+++ b/maidr/core/plot/histogram.py
@@ -39,8 +39,13 @@ class HistPlot(MaidrPlot, ContainerExtractorMixin, DictMergerMixin):
         # could name it. `None` is the ungrouped case and every chart that
         # drew no legend. Opted into here rather than read by `MaidrPlot`;
         # see `GROUP_NAME` for why that is per class.
+        # A string names the group now; a callable names it at render, which
+        # is what a `pairplot` needs -- its legend does not exist until every
+        # panel has been drawn (#561).
         group_name = kwargs.pop(GROUP_NAME, None)
-        self._group_name = group_name if isinstance(group_name, str) else None
+        self._group_name = (
+            group_name if isinstance(group_name, str) or callable(group_name) else None
+        )
         super().__init__(ax, PlotType.HIST)
         self._orientation = "vert"
 
@@ -79,8 +84,9 @@ class HistPlot(MaidrPlot, ContainerExtractorMixin, DictMergerMixin):
         # populated by `_extract_plot_data`, which that call runs.
         base_schema = super().render()
         hist_orientation = {MaidrKey.ORIENTATION: self._orientation}
-        if self._group_name:
-            hist_orientation[MaidrKey.NAME] = self._group_name
+        name = self._group_name() if callable(self._group_name) else self._group_name
+        if name:
+            hist_orientation[MaidrKey.NAME] = name
         return DictMergerMixin.merge_dict(base_schema, hist_orientation)
 
     def _extract_plot_data(self) -> list[dict]:
--- a/maidr/core/plot/maidr_plot.py
+++ b/maidr/core/plot/maidr_plot.py
@@ -30,6 +30,14 @@ import uuid
 #: ``ScatterPlot`` and ``EventPlot`` name their layers by other means -- a hue
 #: split and row labels -- and are unaffected, being passed neither this key
 #: nor anything that collides with it.
+#:
+#: The value may be a **string** or a **callable returning one**. A string is
+#: the name the patch already resolved; a callable is resolved at render, for
+#: a chart whose legend does not exist yet when its layers register. A
+#: ``pairplot(hue=...)`` is that chart: ``PairGrid.add_legend()`` runs after
+#: every panel is drawn, so at registration there is no legend anywhere and
+#: every diagonal came out anonymous while the scatters beside it were named
+#: (#561). Each class that honours this key resolves the callable itself.
 GROUP_NAME = "_maidr_group_name"
 
 
--- a/maidr/core/plot/regplot.py
+++ b/maidr/core/plot/regplot.py
@@ -43,8 +43,13 @@ class SmoothPlot(MaidrPlot):
         # A `kdeplot(hue=...)` draws one curve per level and both were
         # announced identically before (#558). Opted into here rather than
         # read by `MaidrPlot`; see `GROUP_NAME` for why that is per class.
+        # A string names the group now; a callable names it at render, which
+        # is what a `pairplot` needs -- its legend does not exist until every
+        # panel has been drawn (#561).
         group_name = kwargs.get(GROUP_NAME, None)
-        self._group_name = group_name if isinstance(group_name, str) else None
+        self._group_name = (
+            group_name if isinstance(group_name, str) or callable(group_name) else None
+        )
         super().__init__(ax, PlotType.SMOOTH)
         self._smooth_gid = None
         self._regression_line = kwargs.get("regression_line", None)
@@ -60,8 +65,9 @@ class SmoothPlot(MaidrPlot):
         a reader hearing the identical announcement twice.
         """
         schema = super().render()
-        if self._group_name:
-            schema[MaidrKey.NAME] = self._group_name
+        name = self._group_name() if callable(self._group_name) else self._group_name
+        if name:
+            schema[MaidrKey.NAME] = name
         return schema
 
     def _get_selector(self):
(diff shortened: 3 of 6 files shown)


jooyoungseo:
The `legend_of` caveat is real — I measured it rather than reasoning about it, and it does misname:

```python
fig, (first, second) = plt.subplots(1, 2)
sns.kdeplot(data=left,  x="v", hue="g", ax=first,  legend=False)   # groups p, q
sns.kdeplot(data=right, x="v", hue="h", ax=second, legend=False)   # groups s, t
fig.legend(handles=[…first's colours…], labels=["p", "q"])

# -> [('smooth', 'p'), ('smooth', 'q'), ('smooth', 'p'), ('smooth', 'q')]
```

The second panel's groups are `s` and `t` and they are announced as `p` and `q`. Nothing in the artists can say otherwise: both panels draw the same default colour cycle, so a legend built for the first matches the second exactly.

That is a confident wrong name, which is the one thing every decline in this module exists to avoid — so it deserved more than a shrug. I looked for a discriminator and there isn't one: a `PairGrid`'s panels and two independently-hued panels are indistinguishable from the drawn colours, the artist counts, and the swatch counts alike.

**Kept, with the trade stated where it is made** (eaf2b54). The mitigation was already in the code and is now named as one: **the axes' own legend always wins**, so the figure's is consulted only when a panel has none. That means the ambiguity needs a figure built by hand with *every* panel's legend suppressed — not something any seaborn call does on its own, since a panel that keeps its legend is named by it. Against that sits every `pairplot`, whose whole grid does share one hue mapping and which reads as nothing today.

Both halves are now pinned rather than left to be discovered, which is the regression test you suggested:

- `test_an_axes_own_legend_wins_over_the_figures` — the mitigation.
- `test_one_figure_legend_names_every_panel_below_it` — the cost, asserted as `["p", "q", "p", "q"]` with a comment saying the second panel's groups are `s` and `t`.

If you would rather the module keep its "decline rather than name wrongly" rule absolutely, the lever is one line — drop the figure-legend fallback in `legend_of` — and the price is that `pairplot` goes back to anonymous diagonals. I have taken the other side of that, and the code now says so out loud so the choice can be reversed on purpose rather than by accident.

On the two you flagged for awareness only: the deferred-resolution timing is deliberate and documented against the `_legend_title` precedent, and `deferred_names`' per-call memo is scoped exactly as you traced — one fresh closure per group of artists.

`14 passed` on the file, `ruff check --diff` clean.

---

PR #570, opened 2026-08-21
feat(gantt): read hlines and vlines as the schedule of intervals they draw

### LINKED ISSUE #568: `ax.hlines` / `ax.vlines` draw a chart of intervals and register nothing, so a figure made of them is only a picture
From the same sweep as #564, measured further.

## Measured

`Axes.hlines` and `Axes.vlines` return a `LineCollection` whose `get_segments()` gives both ends of every segment exactly — nothing inverted, nothing rounded:

```python
ax.hlines([1, 2, 3], [0, 2, 4], [5, 7, 6])
# LineCollection, segments:
#   [[0, 1], [5, 1]]     lane 1, from 0 to 5
#   [[2, 2], [7, 2]]     lane 2, from 2 to 7
#   [[4, 3], [6, 3]]     lane 3, from 4 to 6
# layers: none — UnsupportedPlotError
```

That is a gantt: one interval per lane, with the lane on the other axis. A figure drawn entirely of `hlines` therefore falls back to a static image, and the numbers it states are in the call.

## The three shapes, and how they separate

The rule is already settled twice in the core, for Observable's `Plot.rule` (xability/maidr#1100) and Vega-Lite's `rule` (xability/maidr#1122):

> if every line shares an end, that end is the frame or the baseline rather than anything measured, and the mark is handed back

Measured, matplotlib's spellings separate exactly the same way:

| call | segments | reading |
| --- | --- | --- |
| `hlines([1,2,3], [0,2,4], [5,7,6])` | ends all differ | **a gantt of three lanes** |
| `vlines([1,2,3], [0,2,4], [5,7,6])` | ends all differ | the same, lanes on x |
| `vlines([1,2,3], 0, [5,7,6])` | every segment starts at `0` | a lollipop's stems — the markers beside them carry the value |
| `hlines([1,2,3], 0, 5)` | every segment is `0`→`5` | reference lines drawn across the frame |

The last two must keep registering nothing, for the reason #1100 gives: read as spans they announce "0 to 8" where the chart means "8".

## Related, already read

- `ax.broken_barh` reads as a gantt (#524), and is matplotlib's *documented* gantt — one call per lane.
- `ax.eventplot` reads as a scatter per row (#548), measured beside these: `eventplot([[1,2],[3,4]])` gives `['point', 'point']`.

`GanttPlot` as it stands is shaped for `broken_barh` — it reads corners off a `PolyCollection` and takes one call as one lane. `hlines` draws every lane in one call and hands back a `LineCollection` of segments, so this needs its own extraction rather than a flag on that class. That is the size of the work, and the reason this is filed rather than folded into #564.

## The rest of the sweep, unchanged

`quiver`, `barbs` and `streamplot` register nothing and have no trace type in the core to register as. `ax.fill(x, y)` draws a closed polygon, which is not a series. `tricontourf` and `tripcolor` are the filled halves of a split whose line half (`tricontour`) already reads as a `contour` — the same line-versus-filled question r-maidr #211 records for `geom_contour_filled`. `acorr` and `xcorr` draw through `vlines`, so they inherit whatever this issue settles.

jooyoungseo:
Closes #568.

## What was wrong

`Axes.hlines` and `Axes.vlines` draw one segment per row of the data and hand back a single `LineCollection` whose `get_segments()` gives both ends exactly — a collection stores its ends rather than serialising a path, so there is nothing to invert and nothing to round. Measured on matplotlib 3.9.4:

```python
ax.hlines([1, 2, 3], [0, 2, 4], [5, 7, 6])
# [[0, 1], [5, 1]]   [[2, 2], [7, 2]]   [[4, 3], [6, 3]]
```

Three lanes at 1, 2 and 3 running 0–5, 2–7 and 4–6. That is the chart `broken_barh` draws and the layer it emits (#524), and it registered **nothing at all** — so a figure made of them fell back to a static image whose numbers were sitting in the call.

## Its own class, sharing one rule

`broken_barh` draws **one lane per call** and hands over the corners of a `PolyCollection`; these draw **every lane in one call**. `GanttPlot` is shaped for the first and `SpanPlot` for the second, and the factory picks by the artist the patch passes rather than by type, since both emit the same layer.

The lane-naming rule is shared rather than copied: a lane is called by the single **explicit** tick inside it and by its position otherwise (the rule #533 settled for `broken_barh`), generalised to read whichever axis the lanes were laid out on — y for `hlines`, x for `vlines`.

| call | reads as |
| --- | --- |
| `hlines([1,2,3], [0,2,4], [5,7,6])` | `gantt`, lanes `1, 2, 3` |
| `vlines([1,2,3], [0,2,4], [5,7,6])` | the same, lanes along x |
| the above with `set_yticks(labels=[…])` | lanes `design, build, ship` |
| `vlines([1,2,3], 0, [5,7,6])` — lollipop stems | nothing |
| `hlines([1,2,3], 0, 5)` — reference lines | nothing |
| `hlines([1], [0], [5])` — one segment | nothing |
| `broken_barh(…)` | unchanged |

## What is not a schedule

The rule xability/maidr#1100 settled for Observable's `rule` mark and #1122 for Vega-Lite's:

> if every line shares an end, that end is the frame or the baseline rather than anything measured, and the mark is handed back

- a lollipop's stems all start at zero, and read as spans they announce "0 to 8" where the chart means "8" — which the markers at their tips already carry;
- reference lines all span the same interval, which no row of the data states;
- a single segment cannot be told from either, since one end trivially agrees with itself. That is the cost the same rule pays on both other adapters, paid here for the same reason.

**The decision is made in the patch, before anything registers.** A layer that refuses at extraction takes the whole figure with it — the defect #564 was about — so a declined call registers nothing at all, and a bar chart drawn beside a lollipop still reads. That is asserted.

## Highlighting

One selector over the whole collection, not one per lane: there is one artist, and `GanttTrace` slices a flat element list into lanes by their lengths.

It is **withheld** when a lane's segments are not consecutive in draw order. `hlines([1, 2, 1], …)` draws lane 1 either side of lane 2; the element count still matches, so the core would not withdraw the selectors itself, and its slicing would hand lane 2 somebody else's interval. Leaving the layer unhighlighted is what a schedule already does when its lanes do not follow row order.

## Tests

`tests/core/plot/test_spanplot.py`, 12 cases.

The pair worth naming is **"a lane drawn twice out of order is left unhighlighted"** and **"a lane drawn twice in a row keeps its highlighting"** — together they make the first a check on the *ordering* rather than on repeats, which a single test could not distinguish.

| mutation | result |
| --- | --- |
| every shape accepted as a schedule | 4 failures — the stems, the reference lines, the single span, and the bar chart beside them |
| the contiguity guard removed | the out-of-order lane keeps selectors and fails |
| the decision moved out of the patch | the same 4 — the layer registers and then refuses at render |
| `vlines` told its lanes run the other way | the vertical case fails |
| lanes never named by their ticks | the named-lane case fails |

Full suite: `2672 passed, 18 skipped`. `ruff check --diff` clean on the changed files.

## Left where it was

The rest of the sweep in #568 is unchanged: `quiver`, `barbs` and `streamplot` have no trace type in the core to register as, `ax.fill` draws a closed polygon, and `tricontourf`/`tripcolor` are the filled half of a split whose line half already reads. `acorr` and `xcorr` draw through `vlines`, so they now inherit whatever their segments say — measured, both draw from a common baseline and are declined.

---

### DIFF
--- a/maidr/core/plot/gantt.py
+++ b/maidr/core/plot/gantt.py
@@ -158,29 +158,34 @@ class GanttPlot(MaidrPlot):
             float(finite[:, 1].max()),
         )
 
-    def _lane_name(self, low: float, high: float) -> str | float:
+    def _lane_name(self, low: float, high: float, axis=None) -> str | float:
         """
         What to call the lane a bar spanning ``low`` to ``high`` sits in.
 
         Parameters
         ----------
         low, high : float
-            The bar's y extent.
+            The bar's extent along the lane axis.
+        axis : Axis, optional
+            The axis the lanes are laid out on. The y axis by default, which
+            is where ``broken_barh`` puts them; ``SpanPlot`` passes the x axis
+            for a set of vertical spans, whose lanes run the other way.
 
         Returns
         -------
         str or float
             The tick label naming it, or its centre when no single tick does.
         """
+        axis = self.ax.get_yaxis() if axis is None else axis
         centre = (low + high) / 2
-        if not isinstance(self.ax.get_yaxis().get_major_locator(), FixedLocator):
+        if not isinstance(axis.get_major_locator(), FixedLocator):
             # An axis matplotlib chose the ticks for. Several land inside a
             # bar and none of them is its name.
             return centre
 
         inside = [
             text.get_text()
-            for position, text in zip(self.ax.get_yticks(), self.ax.get_yticklabels())
+            for position, text in zip(axis.get_ticklocs(), axis.get_ticklabels())
             if low <= position <= high and text.get_text()
         ]
         # Exactly one, or the label is a guess between candidates rather than
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -13,7 +13,6 @@ from maidr.core.plot.contour import ContourPlot
 from maidr.core.plot.errorbar import ErrorBarPlot
 from maidr.core.plot.eventplot import DRAWN_EVENTS, EventPlot
 from maidr.core.plot.gantt import GanttPlot
-from maidr.core.plot.spanplot import DRAWN_SPANS, SpanPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
 from maidr.core.plot.heatmap import HeatPlot
 from maidr.core.plot.hexbinplot import HexbinPlot
@@ -24,6 +23,7 @@ from maidr.core.plot.pieplot import PiePlot
 from maidr.core.plot.pointplot import PointPlot
 from maidr.core.plot.scatterplot import ScatterPlot
 from maidr.core.plot.regplot import SmoothPlot
+from maidr.core.plot.spanplot import DRAWN_SPANS, SpanPlot
 from maidr.core.plot.stairs import StairsPlot
 from maidr.core.plot.stepped_histogram import SteppedHistPlot
 from maidr.core.plot.step_histogram import STEP_COUNTS, STEP_EDGES, StepHistPlot
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -13,6 +13,7 @@ from maidr.core.plot.contour import ContourPlot
 from maidr.core.plot.errorbar import ErrorBarPlot
 from maidr.core.plot.eventplot import DRAWN_EVENTS, EventPlot
 from maidr.core.plot.gantt import GanttPlot
+from maidr.core.plot.spanplot import DRAWN_SPANS, SpanPlot
 from maidr.core.plot.grouped_barplot import GroupedBarPlot
 from maidr.core.plot.heatmap import HeatPlot
 from maidr.core.plot.hexbinplot import HexbinPlot
@@ -76,6 +77,14 @@ class MaidrPlotFactory:
         elif PlotType.BOXEN == plot_type:
             return BoxenPlot(single_ax, **kwargs)
         elif PlotType.GANTT == plot_type:
+            # Two calls draw this chart and they are shaped oppositely.
+            # `broken_barh` draws one lane per call and hands back a
+            # `PolyCollection` of corners; `hlines`/`vlines` draw every lane
+            # in one call and hand back a `LineCollection` of segments. The
+            # layer is the same either way, so the type cannot say which, and
+            # the artist the patch passes does (#568).
+            if DRAWN_SPANS in kwargs:
+                return SpanPlot(single_ax, **kwargs)
             return GanttPlot(single_ax, **kwargs)
         elif PlotType.CONTOUR == plot_type:
             return ContourPlot(single_ax, **kwargs)
--- a/maidr/core/plot/spanplot.py
+++ b/maidr/core/plot/spanplot.py
@@ -16,11 +16,54 @@ DRAWN_SPANS = "_maidr_spans"
 #: Key saying which way the spans run: ``True`` for ``hlines``.
 SPANS_ALONG_X = "_maidr_spans_along_x"
 
-#: How close two coordinates must be to count as the same lane, relative to
-#: the span of the lane axis. Segment ends come back exactly as the caller
-#: gave them -- a `LineCollection` stores them, it does not serialise a path
-#: -- so this exists for a caller who computed a lane rather than typing it.
-_TOLERANCE = 1e-9
+#: How far apart a segment's two lane coordinates may be and still count as
+#: level. Absolute, and deliberately left that way: the scale this wants is
+#: the *magnitude* of the coordinates, where float spacing lives, and the
+#: only scale available here is the spread of the lanes, which is a
+#: different quantity -- three lanes a second apart at epoch scale spread by
+#: 2 while their float spacing is nearer 2e-7. So a relative version built
+#: from what this function knows would not answer the question it looks like
+#: it answers. Unresolved rather than solved; it costs nothing today because
+#: `hlines` and `vlines` give both ends of a segment the *same* float, so
+#: the comparison is exactly zero and any tolerance passes. It would matter
+#: only for a hand-built `LineCollection` at large magnitude.
+_LEVEL_TOLERANCE = 1e-9
+
+#: How close two of a chart's ends must be to count as the same end, as a
+#: fraction of the extent the chart covers along the axis the intervals run
+#: on. Relative because the question is about the chart's own scale, and
+#: because the shared coordinate is often zero, which has no magnitude to be
+#: relative to.
+#:
+#: Exact equality was not enough. `hlines` and `vlines` hand back the
+#: caller's own floats untouched, so a literal baseline is bit-identical
+#: across segments -- but a *computed* one need not be. Measured:
+#: ``tops * 0.1 - tops / 10`` is elementwise zero in arithmetic and
+#: ``[5.55e-17, 0.0, 0.0, 0.0]`` in floats, so four lollipop stems stopped
+#: sharing a baseline and were read as a schedule announcing "0 to 8" for a
+#: chart that means "8".
+_SHARED_END_TOLERANCE = 1e-9
+
+
+def _same_within(values: list, extent: float) -> bool:
+    """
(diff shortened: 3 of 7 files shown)


jooyoungseo:
Pushed `ced4d6c`. The first finding was real, and worse than the review put it — thanks for catching it.

## The `broken_barh` + `hlines` collision

Measured before touching anything:

```python
ax.hlines([1, 2, 3], [0, 2, 4], [5, 7, 6])
ax.broken_barh([(0, 3)], (10, 9))
```

```
SpanPlot  lanes=[1.0, 2.0, 3.0]
```

Four bands drawn, three announced. The `broken_barh` lane at 14.5 is gone, and nothing raised — `_lane_of` matched the `SpanPlot` on `isinstance`, `add_lane` appended the `PolyCollection` to `self._collections`, and `SpanPlot._extract_plot_data` reads `self._spans` and never looks there. Exactly the silent-drop shape the audit comment in `histogram.py` records for two `ax.hist()` calls.

Fixed by matching on `type(plot) is GanttPlot`, because what `_lane_of` wants is not "a gantt-shaped layer" but "a layer `add_lane` can actually extend", and those stopped being the same set when `SpanPlot` arrived. Both orders now read everything:

```
SpanPlot  lanes=[1.0, 2.0, 3.0]   sel=g[id='maidr-9417…'] > path
GanttPlot lanes=[14.5]            sel=g[id='maidr-a04d…'] > path
```

I did **not** take the "raise in `SpanPlot.add_lane`" option. A refusal there would have to propagate out of the caller's own `ax.broken_barh(...)` to be heard, and a plotting call that throws is worse than two layers — the reverse drawing order already produced two layers and rendered fine.

Pinned parametrized over both orders, deliberately: only one direction was ever broken, so a single-order test would pass on a "fix" that broke the other. Under the reverted mutation the `[True]` case fails and `[False]` still passes, which is the whole point. `test_two_broken_barh_calls_still_merge_into_one_chart` guards the behaviour `_lane_of` exists for, and fails if the lookup is narrowed any further.

## Shared-start schedules

Right that this is a false negative, and it now says so on the function rather than only in the class docstring, plus `test_a_schedule_whose_tasks_share_a_start_is_declined`.

Keeping the behaviour. `hlines([1, 2, 3], [0, 0, 0], [5, 7, 6])` and a lollipop's stems are geometrically identical — one shared start, differing ends — so no rule reads one and declines the other. One of the two has to lose, and announcing "0 to 8" for a chart that means "8" is the worse loss; the declined schedule still gets a picture. Same trade xability/maidr#1100 and #1122 made.

## Two `hlines` calls not merging

Deliberate, now commented and pinned. The asymmetry follows from the calls being shaped differently: `broken_barh` draws *one* lane, so lanes must accumulate across calls for a schedule to exist at all, whereas one `hlines` draws the whole schedule — merging a second in would join two complete charts into one neither call made. It also keeps an `hlines` and a `vlines` on one axes separable, which matters since they lay lanes out on opposite axes.

Import ordering moved to its alphabetical slot.

Full suite `2679 passed, 18 skipped`; `ruff check --diff` clean. Four mutations falsified: restoring `isinstance` (2 fail), `_lane_of` returning `None` (2 fail), relaxing `states_an_interval` to `or` (5 fail), and inverting the two-`hlines` expectation (1 fail).

---

jooyoungseo:
Pushed `c690425` for the two coverage gaps. Both were claims **I wrote in prose** with nothing holding the code to them, which is the useful kind of catch.

**`hlines` + `vlines` on one axes.** The `patch/spanplot.py` docstring asserts they lay lanes out on opposite axes; measured with different tick labels on each:

```python
ax.set_yticks([1, 2, 3], labels=["alpha", "beta", "gamma"])
ax.set_xticks([10, 20], labels=["left", "right"])
ax.hlines([1, 2, 3], [0, 2, 4], [5, 7, 6])
ax.vlines([10, 20], [0, 1], [4, 9])
```

```
SpanPlot along_x=True   lanes=["alpha", "beta", "gamma"]
SpanPlot along_x=False  lanes=["left", "right"]
```

Holds. The test deliberately labels the two axes differently, because the load-bearing part isn't that two layers appear — it's that neither borrows the other's names. Naming both off the y axis fails it.

**Whole-call decline on one bad segment.** `ax.hlines([1, 2, 3], [0, 2, np.nan], [5, 7, 6])` registers nothing and the figure still renders as a picture. Now pinned; changing `return None` to `continue` in `read_spans` fails it. Agreed this is a real asymmetry with `GanttPlot._corners`, and it stays deliberate — a reader handed two of three tasks has no way to know a third existed, whereas a picture at least doesn't lie about completeness.

On the lane-name exact-equality note: correct, and worth having written down. `SpanPlot` passes `low == high == lane`, so `GanttPlot._lane_name`'s `low <= position <= high` collapses to exact equality, where `broken_barh` gets real slack from the bar's extent. It holds for the ordinary case because the same literals coerce to `float64` identically down both paths, and it would be brittle if ticks and lanes came from separate computations. Not changing it here — widening the match needs a tolerance whose scale I'd want to pick against a real case rather than invent, and the `_TOLERANCE` discussion in this same review is the same open question. If it bites, it bites as "lane announced as `2.0000000001` instead of its label", which is a degraded name rather than wrong data.

Full suite `2681 passed, 18 skipped`; `ruff check --diff` clean.

---

jooyoungseo:
Pushed `86b3075`. Point 2 was a real defect — measured, fixed, pinned. Point 3 measured clean. And chasing point 2 turned up a mistake of my own, described below.

## The float-noise dodge is real

```python
tops = np.array([3.0, 8.0, 5.0, 9.0])
baseline = tops * 0.1 - tops / 10.0   # elementwise zero in arithmetic
# [5.55111512e-17, 0.0, 0.0, 0.0]     # in floats
ax.vlines([1, 2, 3, 4], baseline, tops)
```

Before: `layers: ['gantt']`. Four lollipop stems that stopped sharing a baseline by 5.55e-17 and got read as a schedule — announcing "0 to 8" where the chart means "8", which is the precise outcome `states_an_interval` exists to prevent. After: `layers: []`.

Fixed by comparing ends against the extent the chart covers on that axis rather than as exact float sets. Relative because the question is about the chart's own scale, and because the shared coordinate is usually zero — which has no magnitude to be relative to, so a magnitude-relative test would be the wrong tool here. Extent zero collapses to exact equality, which is the right answer when there's no scale.

## A change of mine that didn't survive its own mutation

I initially made the *levelness* tolerance relative too, on the reasoning both reviews raised about epoch-scale coordinates. Restoring the absolute form as a mutation **passed all 25 tests** — so I looked at why instead of letting it go, and the rationale I'd written was wrong: the only scale available in `read_spans` is the *spread* of the lanes, not their magnitude. Three lanes a second apart at epoch scale spread by 2, while their float spacing is nearer 2e-7. A relative tolerance built from the spread doesn't answer the question it appears to answer.

So that half is reverted. The constant is now `_LEVEL_TOLERANCE`, still absolute, with the reasoning and the open question recorded on it rather than a fix that only looks like one. It costs nothing today — `hlines`/`vlines` give both ends of a segment the *same* float, so the comparison is exactly zero and any tolerance passes. It would matter only for a hand-built `LineCollection` at large magnitude, and picking the right form there wants a real case.

The epoch-scale test stays, with its comment corrected: it exercises the shared-end tolerance not being so wide that a real schedule's differing ends look shared (extent 700, tolerance 7e-7, starts differing by 200), not levelness.

## `ax.stem`

Measured: `layers: ['line']` — it registers a line layer off its marker tips and this reading never claims it. Unchanged from before the PR, now pinned so the span patch can't start claiming it silently.

## On scope (point 1)

Intentional, and worth stating plainly: this promotes any `hlines`/`vlines` call with varying starts *and* varying ends, not only literal schedules. Manually-drawn intervals — confidence bands, custom error bars — are variable-width per-row intervals, and reading them as spans announces what they show. The shapes where that would be *wrong* are exactly the ones sharing an end, which is what the decline covers. Previously all of these fell back to a static image.

CLAUDE.md's plot-type list is a pre-existing gap and I've left it alone rather than widen this PR; happy to do it separately.

Full suite `2685 passed, 18 skipped`; `ruff check --diff` clean. Mutations: exact float equality restored fails `test_a_computed_baseline_is_still_a_lollipop`; `_SHARED_END_TOLERANCE = 1.0` fails 14.

---

PR #571, opened 2026-08-21
feat(rug): read a seaborn rug plot as the observations it marks

jooyoungseo:
Closes #250.

## What was wrong

`sns.rugplot` draws one short tick per observation against the frame and hands back a plain `LineCollection` that no patch claimed. Measured before the change:

```python
sns.rugplot(df, x="value")     # NO LAYERS
```

So a figure whose only layer was a rug fell back to a picture, and — the case that matters more — a rug drawn over a density curve left the raw observations unread. That is the one thing the curve does not state: a KDE is smoothed, the rug is where the data actually fell.

While measuring the neighbouring issues I also checked #252 and #251: `stripplot` and `swarmplot` already read as three `point` layers each, so this PR is only about the rug.

## What it does now

```
RugPlot  name=value   axes={"x": {"label": "value"}, "y": {"label": "Rug"}}
                      data=[{"x": 1.0, "y": 0}, {"x": 2.5, "y": 0}, ...]
```

Read as a **scatter**, for the reason `EventPlot` gives about an event plot's ticks: `height` is one number for the whole call, so a tick's length is decoration and only its position is data. A rug's ticks are held constant on the axis carrying the observations and stretched across the other, measured on seaborn 0.13.2:

```
sns.rugplot(x=[10.251, ...])
[[10.251, 0.0], [10.251, 0.025]]
```

so the *constant* coordinate is the measurement.

## Three judgement calls, stated

**The strip is renamed, not left carrying the chart's label.** An event plot's row sits at `lineoffsets`, a real place on a real axis worth announcing. A rug's ticks sit against the frame at a fraction of the axes height, which measures nothing. So that axis becomes "Rug" even when the caller labelled it — a rug over a `kdeplot` has a genuine "Density" label there, and every point this layer emits sits at 0, so keeping it would announce each observation as "Density 0", a number the chart does not show. This differs from `EventPlot`'s "Row", which only fills a blank.

**One layer per collection, not per call.** `rugplot(x=..., y=...)` marks two margins and returns one collection each; merging them would announce a single series whose coordinates come off two different axes. Each is named from the column it marks.

**A rug beside another layer is registered, not declined as a duplicate.** This is the opposite call to the one xability/maidr#1124 made for Vega-Lite text overlays, and deliberately: those labels sit *on* the marks they duplicate, whereas a rug occupies its own strip and, next to a density curve or histogram, is the only thing in the figure stating where the observations are. It carries `name` (xability/maidr#828) so a reader can tell it from its neighbour.

## A bug found in this PR's own code

`MaidrPlot.render()` builds the **axes** payload before the data, so resolving the orientation inside `_extract_plot_data` left `rugplot(y=...)` labelling the y axis "Rug" while the observations it carries were announced under "X". Now resolved in the constructor; `test_a_rug_on_y_names_the_other_axis` pins it.

## Fallback tests

`tests/core/test_unsupported_fallback.py` used a rug as its example of an unsupported chart, which it no longer is. It now uses `ax.barbs` — no trace type in the core to be read as, since a barb states a speed *and* a direction at a place — so it will not quietly stop testing the fallback. `test_seaborn_color_probe.py`'s rug-over-scatter case now expects two layers; the count is still what pins the phantom-layer regression, since a third layer there would be the probe back.

## Verification

Full suite `2697 passed, 18 skipped`; `ruff check --diff` clean. Highlighting checked against the rendered SVG — 4 ticks, 4 `<path>` children under the group, selector resolves.

Five mutations, each applied alone:

| mutation | result |
|---|---|
| resolve the orientation in extraction again | `test_a_rug_on_y_names_the_other_axis` fails |
| keep the chart's own label on the strip axis | 2 fail |
| emit the tick base instead of a constant | 2 fail |
| sweep the axes instead of taking this call's collections | `test_a_second_rug_on_the_same_axes_reads_only_its_own_ticks` fails |
| drop the layer name | 3 fail |

The fourth of those first **survived**: a scatter beside a rug is safe by accident, since a `PathCollection` is skipped anyway. Two rug calls on one axes is the case that actually bites, and that test was added after the mutation exposed the gap.

---

### DIFF
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -23,6 +23,7 @@ from maidr.core.plot.pieplot import PiePlot
 from maidr.core.plot.pointplot import PointPlot
 from maidr.core.plot.scatterplot import ScatterPlot
 from maidr.core.plot.regplot import SmoothPlot
+from maidr.core.plot.rugplot import DRAWN_RUG, RugPlot
 from maidr.core.plot.spanplot import DRAWN_SPANS, SpanPlot
 from maidr.core.plot.stairs import StairsPlot
 from maidr.core.plot.stepped_histogram import SteppedHistPlot
@@ -139,6 +140,11 @@ class MaidrPlotFactory:
             # through a class of its own under the same type (#548).
             if kwargs.get(DRAWN_EVENTS) is not None:
                 return EventPlot(single_ax, **kwargs)
+            # A rug's ticks are a scatter of positions for the same reason,
+            # and arrive as a plain `LineCollection` -- not an
+            # `EventCollection`, so `EventPlot` cannot read one (#250).
+            if kwargs.get(DRAWN_RUG) is not None:
+                return RugPlot(single_ax, **kwargs)
             return ScatterPlot(single_ax, **kwargs)
         elif PlotType.DODGED == plot_type or PlotType.STACKED == plot_type:
             return GroupedBarPlot(single_ax, plot_type, **kwargs)
--- a/maidr/core/plot/rugplot.py
+++ b/maidr/core/plot/rugplot.py
@@ -115,14 +115,17 @@ class RugPlot(MaidrPlot):
         )
         self._label = kwargs.get(RUG_LABEL, None)
 
-        # Which axis the positions lie along, resolved here rather than
-        # during extraction. `MaidrPlot.render()` builds the *axes* payload
-        # before the data, so a value settled in `_extract_plot_data` is
-        # still the default when `_extract_axes_data` reads it -- measured, a
-        # `rugplot(y=...)` then named the y axis "Rug" and left the
-        # observations it carries labelled "X".
+        # Read once, here, rather than again per render. Two reasons, and
+        # the first is correctness: `MaidrPlot.render()` builds the *axes*
+        # payload before the data, so an orientation settled in
+        # `_extract_plot_data` is still the default when
+        # `_extract_axes_data` reads it -- measured, a `rugplot(y=...)` then
+        # named the y axis "Rug" and left the observations it carries
+        # labelled "X". The second is that `render()` can run more than once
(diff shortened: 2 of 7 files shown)


jooyoungseo:
Pushed `ef69b4c` — all four points acted on, the two behavioural ones measured first.

**4 — hue-grouped rugs.** Measured; the worry doesn't materialise. seaborn draws one `LineCollection` either way and only recolours it, so a hue rug is one layer holding every observation:

```python
sns.rugplot(df, x="value", hue="sex")
# RugPlot  [1.0, 2.5, 3.0, 7.25]   ← all four, one layer
```

Also checked the layered case this reading is actually for — `kdeplot(hue=…)` then `rugplot(hue=…)` gives `['smooth', 'smooth', 'point']` with the rug carrying all four. Both pinned rather than left as ad-hoc measurements, since the answer is seaborn's to change and not mine.

**2 — `_name_for` taking a stale label.** You're right, and it reproduces:

```python
sns.scatterplot(df, x="value", y="other")
sns.rugplot(x=np.array([7.0, 8.0]))
# rug name: "value"   pts: [7.0, 8.0]
```

Keeping the behaviour and documenting it. The artist carries no record of the column it came from, so the alternative is naming nothing at all — and a name off the shared axis is at least the axis these ticks stand on. The layer's *data* is unaffected; only what it's announced as. Documented on `_name_for` and pinned by `test_a_name_can_come_off_an_axis_another_plot_labelled`, so it's known rather than discovered.

**1 — redundant `read_rug`.** Fixed; positions and orientation are both read once in the constructor. Your note that `render()` can run more than once is what made it worth doing rather than cosmetic — it was revalidating every segment each time for a collection that cannot change underneath it.

**3 — selector shape.** Now `[]`, matching `EventPlot`.

Full suite `2700 passed, 18 skipped`; `ruff check --diff` clean. Two more mutations: reverting the orientation to lazy fails 2 (including the two-margin case, which the caching change had to not break), and truncating the segment list fails 3 including both new hue tests.

---

PR #573, opened 2026-08-22
fix(triplot): decline a triangulation mesh instead of reading it as a line

### LINKED ISSUE #572: `ax.triplot` announces a triangulation mesh as a line chart of its edge list
Found while working through what #568 left open. Of everything in that remainder, this is the only one that is actually **wrong** rather than merely unread — it does not fall back to a picture, it confidently announces something the chart does not show.

## Measured

`main` at `c07b98c`, eight scattered points:

```python
rng = np.random.default_rng(0)
tx, ty = rng.uniform(0, 1, 8), rng.uniform(0, 1, 8)
ax.triplot(tx, ty)
```

```
type:   line
series: 1
points in first series: 32          ← the chart draws 8
first 6: [{"x": 0.0410, "y": 0.8159},
          {"x": 0.6370, "y": 0.5436},
          {"x": 0.0410, "y": 0.8159},   ← the first point again
          {"x": 0.2698, "y": 0.9351},
          {"x": 0.0165, "y": 0.0027},
          {"x": 0.6370, "y": 0.5436}]
```

Thirty-two "observations" for eight drawn points, x running 0.04 → 0.64 → 0.04 → 0.27 → 0.02 → 0.64, and the same point appearing twice in the first six.

## Cause

`triplot` draws the mesh by handing the flattened edge list to `ax.plot`:

```python
tri_lines = ax.plot(tri_lines_x.ravel(), tri_lines_y.ravel(), **kw_lines)
```

so the line patch sees an ordinary `Axes.plot` and registers a LINE layer. The "series" is an edge traversal — each triangle contributing three vertices and a separator — not a sequence of observations.

## Why this is worse than the rest of the remainder

A line trace tells a reader there is a trend through ordered observations, and offers to play it as one. Here there is no order: the x values are the mesh's traversal, they revisit points, and the count bears no relation to the data. A reader is given a shape to follow that the chart does not contain.

Everything else left in #568 declines correctly, and I checked each rather than assuming:

| call | today | verdict |
| --- | --- | --- |
| `contourf` | no layers | **correct** — see below |
| `tricontourf` | no layers | correct, same reason |
| `tripcolor` | no layers | correct — pseudocolour on an irregular triangulation, no trace carries it |
| `ax.fill` | no layers | correct — a closed polygon states no series |
| `quiver` / `barbs` / `streamplot` | no layers | correct — a vector at a place has no trace type |
| **`triplot`** | **line, 32 points** | **wrong** |

On `contourf` specifically, since it looked like the obvious gap: it is not one. Measured against `contour` on the same field with `levels=[1, 2, 4]`:

- `contour` — 3 paths, each sitting at **one** level (z ≈ 1.0 / 2.0 / 4.0)
- `contourf` — 2 paths, each spanning **two** (path 0 runs z = 1.0 → 2.0)

A filled contour draws the bands *between* levels, so announcing one of those outlines as a level's own curve would be right for about half its vertices. That is exactly the rule xability/r-maidr's `Ggplot2ContourLayerProcessor` already states for `geom_contour_filled()`, reached there by reading `level` as a factor of intervals rather than a number. Both bindings agreeing is worth keeping.

## Proposed

Decline `triplot`: draw it inside the internal context so the inner `ax.plot` calls register nothing, the way every other "drawn but not a series" case is handled. The figure then falls back to a picture, or reads whatever else is on the axes — which is what it should have done all along.

The mesh is genuinely not a chart maidr has a trace for. `triplot` is usually drawn *under* a `tricontour`, and that half already reads.

## Related

- #568 — the sweep this was left over from
- #539 — the contour reading `contourf` is deliberately not part of
- xability/r-maidr#198 — the ggplot2 side's stated refusal of filled contours

jooyoungseo:
Closes #572. This finishes what #568 left over — and the outcome is not what I expected when I listed that remainder.

## The one that was actually wrong

Everything left in #568 was on my list as "unread". Measuring found that six of the seven decline **correctly**, and the seventh doesn't decline at all — it reads, and says something false.

```python
rng = np.random.default_rng(0)
tx, ty = rng.uniform(0, 1, 8), rng.uniform(0, 1, 8)
ax.triplot(tx, ty)
```

```
type:   line
points: 32                     ← the chart draws 8
first 6 x: 0.0410, 0.6370, 0.0410, 0.2698, 0.0165, 0.6370
                   ↑ the first point again, third in the series
```

`triplot` draws the mesh by handing the flattened edge list to `Axes.plot`:

```python
tri_lines = ax.plot(tri_lines_x.ravel(), tri_lines_y.ravel(), **kw_lines)
```

so the line patch sees an ordinary plot call. The "series" is an edge traversal — three vertices per triangle and a separator — not a sequence of observations. A line trace tells a reader there is a trend through ordered values and offers to play it; here there is no order, points repeat, and the count bears no relation to the data. **Worse than being unread, because nothing about it looks wrong.**

Declined by drawing inside the internal context so the inner `plot` calls register nothing. It has to happen in the patch: a layer that refuses while the schema is built takes the whole figure with it (#564).

## What I got wrong about the rest

I had `contourf` down as an obvious gap — the filled twin of a call that reads. It isn't. Measured on one field with `levels=[1, 2, 4]`:

| | paths | each path's z range |
| --- | --- | --- |
| `contour` | 3 | one level each (1.0 / 2.0 / 4.0) |
| `contourf` | 2 | **two** — path 0 runs 1.0 → 2.0 |

A filled contour draws the bands *between* levels, so announcing one of those outlines as a level's own curve would be right for about half its vertices. That is exactly the rule xability/r-maidr's `Ggplot2ContourLayerProcessor` already states for `geom_contour_filled()`, reached there by reading `level` as a factor of intervals. Both bindings agreeing is worth keeping, so this PR **asserts the measurement** rather than describing it — the test compares path z-spans directly.

Final state of the remainder, each checked rather than assumed:

| call | today | verdict |
| --- | --- | --- |
| `contourf`, `tricontourf`, `tripcolor` | no layers | correct — bands between levels |
| `ax.fill` | no layers | correct — a closed polygon states no series |
| `quiver`, `barbs`, `streamplot` | no layers | correct — no trace holds a vector at a place |
| **`triplot`** | **line, 32 points** | **fixed** |

All six were declining correctly but nothing pinned it, so a future reading for any of them would have been an accident rather than a decision. They're pinned now.

## Neighbours

A `triplot` is usually drawn *under* a `tricontour`. That half still reads, in either drawing order. An ordinary `plot` after a `triplot` still reads too — which pins that the internal context doesn't leak, a failure that would be silent and total.

## Verification

Full suite `2714 passed, 18 skipped`; `ruff check --diff` clean.

Two mutations: unregistering the patch fails 3 (the mesh reads again), and leaking the context by entering it without a `with` fails the same 3 — the second because the *contour* then stops registering too, which is the leak the third test guards.

---

### DIFF
--- a/maidr/patch/__init__.py
+++ b/maidr/patch/__init__.py
@@ -26,6 +26,7 @@ from . import (  # noqa: E402, F401
     pieplot,
     pointplot,
     scatterplot,
+    triplot,
     spanplot,
     stairs,
     regplot,
--- a/maidr/patch/triplot.py
+++ b/maidr/patch/triplot.py
@@ -2,13 +2,14 @@ from __future__ import annotations
 
 import wrapt
 from matplotlib.axes import Axes
+from matplotlib.lines import Line2D
 
 from maidr.core.context_manager import ContextManager
 from maidr.patch.common import _draw_quietly
 
 
 @wrapt.patch_function_wrapper(Axes, "triplot")
-def triplot(wrapped, instance, args, kwargs):
+def triplot(wrapped, instance, args, kwargs) -> list[Line2D]:
     """
     Draw a patched ``Axes.triplot`` and register nothing for it.
 
@@ -52,8 +53,8 @@ def triplot(wrapped, instance, args, kwargs):
 
     Returns
     -------
-    list
-        Whatever ``triplot`` returned.
+    list of Line2D
+        Whatever ``triplot`` returned -- the mesh's lines and its markers.
     """
     if ContextManager.is_internal_context():
         return _draw_quietly(wrapped, args, kwargs)
--- a/maidr/patch/triplot.py
+++ b/maidr/patch/triplot.py
@@ -1,0 +1,62 @@
+from __future__ import annotations
+
+import wrapt
+from matplotlib.axes import Axes
+
+from maidr.core.context_manager import ContextManager
+from maidr.patch.common import _draw_quietly
+
+
+@wrapt.patch_function_wrapper(Axes, "triplot")
+def triplot(wrapped, instance, args, kwargs):
+    """
+    Draw a patched ``Axes.triplot`` and register nothing for it.
+
+    A triangulation mesh is not a series, and until this it was announced as
+    one. ``triplot`` draws the mesh by handing the flattened edge list to
+    ``Axes.plot``::
+
+        tri_lines = ax.plot(tri_lines_x.ravel(), tri_lines_y.ravel(), ...)
+
+    so the line patch saw an ordinary plot call and registered a LINE layer.
+    Measured on eight scattered points: a line of **thirty-two** points, x
+    running 0.04 -> 0.64 -> 0.04 -> 0.27, the first point appearing again as
+    the third (#572).
+
+    That is the mesh's edge traversal -- three vertices per triangle and a
+    separator -- not a sequence of observations. A line trace tells a reader
+    there is a trend through ordered values and offers to play it as one;
+    here there is no order, points repeat, and the count bears no relation to
+    the data. Worse than being unread, because nothing about it looks wrong.
+
+    Declined rather than given a reading. The mesh states which points were
+    joined, and no trace in the core carries that -- the same conclusion
+    `quiver`, `barbs` and `streamplot` reach for a vector at a place. A
+    `triplot` is usually drawn *under* a `tricontour`, and that half reads;
+    on its own the figure falls back to a picture, which is what a chart maidr
+    cannot read is supposed to do.
+
+    Suppressed by drawing inside the internal context, so the inner ``plot``
+    calls register nothing -- the decline has to happen here rather than at
+    extraction, because a layer that refuses while the schema is built takes
+    the whole figure with it (#564).
+
+    Parameters
+    ----------
+    wrapped : Callable
+        ``Axes.triplot``.
+    instance : Any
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Pushed `1528198` — all three points.

**Return annotation.** Fixed: `-> list[Line2D]`, matching what `Axes.triplot` returns and what every sibling patch carries.

**The `regplot.patched_plot` interaction.** Your trace is right, and I measured it rather than take it on the trace, because "two patches on one method" is exactly the thing that breaks silently:

```
triplot(label='fit'):        []
triplot(label='smooth'):     []
triplot(label='regression'): []
triplot(label=None):         []
```

Nothing leaks — the `common()` call that would register the SMOOTH layer checks the context itself. Pinned as a parametrized test, since the reasoning depends on a guard living one call further in than the patch that would trip it.

**The matplotlib-geometry test.** Fair, and worth having in the file rather than only in the PR. That test now says outright that it reads matplotlib's own output with no maidr patch in the way, so a future matplotlib could fail it without anything of ours having regressed — and that it wants reading rather than just re-running. Keeping it: the decision not to patch `contourf` rests on that measurement, and an assumption nothing checks is how the two bindings drift apart.

Full suite `2717 passed, 18 skipped`; `ruff check --diff` clean.

---

Issue #502, opened 2026-08-18
A hue group's name is matched to its series by legend order, which nothing verifies


jooyoungseo:
Raised in review of #501 and recorded rather than fixed there, since it is an accepted trade rather than a defect today.

## The assumption

`_group_labels` (`maidr/patch/pointplot.py`) reads the legend entries and hands them to `PointPlot` positionally, against the estimate lines in `ax.lines` draw order. That is only correct if seaborn lists its legend handles in the order it drew the series.

It does today. It is not part of seaborn's public API.

## Why this one is weaker than its neighbour

The same module makes a second positional assumption — that interval polylines arrive estimate-major — and that one **has a correctness cross-check**:

- `_pairs_up` verifies `len(intervals) == len(estimates) * categories` before anything is paired.
- The tests assert each estimate's value falls **inside** the span of the interval its slice pairs it with, so a wrong stride produces a visibly wrong bound.

The naming has no equivalent, because a name has no geometry to check it against. `_group_labels` requires the legend to list exactly one entry per drawn group — added in #501 for a different reason — and a count guard is all that is available. **A future seaborn that keeps the count and reorders the legend would attach every name to the wrong series, silently.**

The announcement would be confidently wrong in the way this project treats as the worst outcome: `grp is p` on `q`'s numbers, with the estimates and bounds all correct and nothing to indicate the swap.

## What the tests cannot currently see

Every hue fixture in `tests/core/plot/test_pointplot.py` uses interchangeable category-like levels (`x`/`y`, `p`/`q`). Correct and swapped orderings are indistinguishable by inspection, so the suite would not catch a reordering either.

Making it visible needs a fixture where the group *name* is recoverable from the group's *values* — e.g. a hue level whose observations are an order of magnitude apart, so `"low"` paired with values around 100 is detectable. That is a real test, not a change to the code.

## Options

1. **Leave it, with the assumption named.** Where it is now — the docstring states it and says the count check is the only symptom available. Cheapest, and consistent with how this codebase already reverse-engineers seaborn internals elsewhere.
2. **Add a value-based regression test** as above. Catches a reordering on the version CI runs, which is the version that matters most.
3. **Verify at runtime**, if seaborn exposes a mapping from handle to artist that can be read without depending on more internals than the ordering itself. Worth a look before assuming it does not; `legend.legend_handles` may or may not be stable enough to be an improvement rather than a second assumption.

[...]

PR #576, opened 2026-08-22
fix(line): stop announcing matplotlib's own label as a series name

### LINKED ISSUE #575: A line labelled `_nolegend_` is announced by that name
`_nolegend_` is matplotlib's documented way to keep an artist out of the legend: any artist whose label is that string is skipped when the legend is built. It is a sentinel, not a name. py-maidr announces it as one.

```python
fig, ax = plt.subplots()
ax.plot([1, 2, 3], [4, 9, 2], label="_nolegend_")
```

emits `z: '_nolegend_'` on every point of the series, and a reader is told the series is called `_nolegend_`.

The same reading applies to anything matplotlib labels that way itself. `Axes.stem` is the case that surfaced this — it labels both its marker line and its baseline `_nolegend_`, so a stem plot announces two series by that name (filed separately).

## Why it is only this sentinel

`LinePlot._extract_line_data` already guards the *other* one:

```python
elif not label.startswith("_child"):
    line_type = label
```

`_childN` is what matplotlib names an unlabelled artist, and suppressing it is right for the same reason suppressing `_nolegend_` is: neither is a name the caller chose. Measured across four cases:

| call | `get_label()` | announced `z` |
| --- | --- | --- |
| `ax.plot(x, y)` | `_child0` | `None` — correct |
| two `ax.plot` calls | `_child0`, `_child1` | `None`, `None` — correct |
| `ax.plot(x, y, label="mine")` | `mine` | `mine` — correct |
| `ax.plot(x, y, label="_nolegend_")` | `_nolegend_` | `_nolegend_` — **wrong** |

matplotlib's own rule is broader than either check: `Legend` skips any label starting with an underscore. That is the rule worth matching, rather than adding a second literal beside `_child` — a leading underscore is how matplotlib spells "internal" throughout, and a caller who wants a series named `_x` announced has no way to ask matplotlib to put it in a legend either.

## Expected

A series whose label matplotlib would keep out of its own legend carries no name, exactly as an unlabelled one already does.

jooyoungseo:
Closes #575.

`_nolegend_` is matplotlib's documented way to keep an artist out of the legend — `Legend` skips any artist labelled that way. It is a sentinel, not a name, and py-maidr announced it as one:

```python
ax.plot([1, 2, 3], [4, 9, 2], label="_nolegend_")
```

emitted `z: "_nolegend_"` on every point, telling a reader the series is called `_nolegend_`. matplotlib uses the sentinel itself, so callers reached this without ever typing the string: `Axes.stem` labels both its marker line and its baseline that way.

## Why only this one leaked

`LinePlot._extract_line_data` already guarded the *other* sentinel, `_childN` — what an unlabelled artist gets — and suppressing that is right for the same reason. Measured across four cases before the change:

| call | `get_label()` | announced `z` |
| --- | --- | --- |
| `ax.plot(x, y)` | `_child0` | `None` — correct |
| two `ax.plot` calls | `_child0`, `_child1` | `None`, `None` — correct |
| `ax.plot(x, y, label="mine")` | `mine` | `mine` — correct |
| `ax.plot(x, y, label="_nolegend_")` | `_nolegend_` | `_nolegend_` — **wrong** |

## The change

`maidr/util/artist_label.py` adds `series_name(artist)`, which matches matplotlib's own rule — a leading underscore — rather than adding a second literal beside `_child`. The list of internal spellings is not closed, and a caller who wants a series named `_x` announced has no way to put it in a matplotlib legend either.

`MplfinanceLinePlot` carried the same `_child`-only test and now shares the helper, which also stops it recomputing the label once per point.

The detection heuristics in `patch/regplot.py` and `patch/pointplot.py` are deliberately left alone: those look for the `_child` prefix to work out *which artist is which*, which is a different question from what to call a series.

## Verification

`tests/core/plot/test_series_naming.py`, 9 tests. Each mutation was applied alone against a restored baseline; all five were killed:

| mutation | result |
| --- | --- |
| revert to the `_child`-only rule | 4 failed |
| drop the non-string guard | 1 failed |
| drop the missing-`get_label` guard | 1 failed |
| `lineplot` ignores the helper | 6 failed |
| `lineplot` never names a series | 2 failed |

Full suite: 2744 passed, 18 skipped. `ruff check` clean on every touched file.

---

### DIFF
--- a/maidr/core/plot/lineplot.py
+++ b/maidr/core/plot/lineplot.py
@@ -102,6 +102,24 @@ def _reading(y: object) -> object:
         return y
 
 
+
+def _drew_something(line: Line2D) -> bool:
+    """
+    Whether a line has any coordinates at all.
+
+    Parameters
+    ----------
+    line : Line2D
+        One of the axes' lines.
+
+    Returns
+    -------
+    bool
+        True when the line carries at least one point.
+    """
+    xydata = line.get_xydata()
+    return xydata is not None and bool(getattr(xydata, "size", 0))
+
 class MultiLinePlot(MaidrPlot, LineExtractorMixin):
     """
     A class for extracting and processing data from line plots.
@@ -262,6 +280,12 @@ class MultiLinePlot(MaidrPlot, LineExtractorMixin):
             and line identifiers for one line, or None if the plot data is invalid.
         """
         all_lines = self._series()
+
+        # Only the lines that drew something take part, in both the pairing
+        # below and the loop. A seaborn `hue` split leaves probe lines with
+        # no data among the real ones, and counting those against the legend
+        # would pair its names with series that are never announced.
+        all_lines = [line for line in all_lines if _drew_something(line)]
         if not all_lines:
             return None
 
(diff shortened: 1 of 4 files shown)


jooyoungseo:
The legend-path bug is real — measured before touching anything, and it reproduces exactly as described:

```python
ax.plot([1, 2, 3], [4, 9, 2], label="_nolegend_")
ax.plot([1, 2, 3], [1, 2, 3], label="revenue")
ax.legend()
# legend texts: ['revenue']   line labels: ['_nolegend_', 'revenue']
# announced:    ['revenue', 'revenue']
```

Fixed here rather than deferred. The issue is "a line labelled `_nolegend_` is announced by that name"; with a legend present it still was, and by a *worse* name — its neighbour's. Closing #575 while the headline symptom survived `ax.legend()` would have been closing it in name only.

**The rule.** Equal lengths keep the positional pairing, because two cases rely on it and both are right to: `ax.legend(["A", "B"])` renames every series positionally and means to, and a seaborn `hue` split names its groups in the legend while the lines carry `_child` sentinels (#502). Only a *shorter* legend is matched against the lines matplotlib would have put in it.

Two things the measurement changed about the fix:

**The `named` branch is not redundant.** A mutation deleting it passed all 14 tests, since the per-line fallback already gives the right answer for everything I had written. Investigating rather than shipping it found the case that separates them:

```python
ax.legend(["Renamed"])   # beside a hidden line
# announced: [None, 'Renamed']    — not [None, 'revenue']
```

One entry, and it belongs to the visible series. That is now `test_a_legend_shorter_than_the_series_still_renames_the_one_it_names`, and it kills the mutation.

**The pairing has to run over the lines that are actually announced.** My first version counted `named` over the raw series list and regressed seaborn's hue naming from `["a", "b"]` to `[None, None]` — seaborn leaves probe lines with no data among the real ones (four `Line2D` for a two-group hue), so the legend's names were paired with series that are never emitted. The drawn-line filter now runs once, before both the pairing and the loop.

Tests: 15 in `test_series_naming.py`, including the two draw orders, a hidden line either side of a named one, and the seaborn hue case. All four mutations killed after the addition above. Full suite green.

On the two smaller points:

- **`ax.legend(handles=[q, p])`** — I checked this while I was in here. It announces `['q', 'p']` for lines drawn `p, q`: the names are *swapped*. `legend_handles` are proxy artists rather than the drawn lines (`h is line` is `False`), so there is no identity to map through, and this is a different defect from the one here — wrong order rather than wrong count. Pre-existing on `main`, unchanged by this PR, and deliberately not fixed in it. Filing separately.
- **mplfinance coverage** — fair; that path's behaviour change is exercised by no test. It is a straight reuse of the already-tested helper, so I have left it rather than build an mplfinance fixture inside a naming PR, and noted it.

---

PR #585, opened 2026-08-22
feat(hist): read the seaborn histogram outlines drawn without fill

### LINKED ISSUE #583: An unfilled outlined histogram is silent
`sns.histplot(..., fill=False)` is how you draw a histogram as an outline rather than a filled shape — the standard way to overlay two distributions without one hiding the other. With `element="step"` or `element="poly"` it registers **no layer at all**, and the figure falls back to a picture.

Measured across the element/fill matrix, ten cases:

| call | layers | artists drawn |
| --- | --- | --- |
| `element="step"` | `['hist']` | 1 `PolyCollection` |
| `element="step"`, `hue` | `['hist', 'hist']` | 2 `PolyCollection` |
| `element="step", fill=False` | **none** | 1 `Line2D` |
| `element="step", fill=False`, `hue` | **none** | 2 `Line2D` |
| `element="poly"` | `['hist']` | 1 `PolyCollection` |
| `element="poly"`, `hue` | `['hist', 'hist']` | 2 `PolyCollection` |
| `element="poly", fill=False` | **none** | 1 `Line2D` |
| `element="poly", fill=False`, `hue` | **none** | 2 `Line2D` |
| `element="bars", fill=False` | `['hist']` | 5 patches |
| `element="bars", fill=False`, `hue` | `['hist', 'hist']` | 10 patches |

The pattern is exactly the fill: `element="bars"` keeps its patches whether or not it is filled, so it reads either way. `step` and `poly` swap their `PolyCollection` for a `Line2D` when unfilled, and #557's `SteppedHistPlot` looks for the collection.

The counts and bin edges are unchanged by `fill` — it is a purely visual choice — so the chart says exactly what its filled twin says, and is read for one spelling and not the other. That is the same shape as #556, where four `histtype`s of `ax.hist` were read for two of them.

## Expected

`fill=False` reads as the same `hist` layer its filled twin does, from the `Line2D` outline rather than from the collection. One layer per group under `hue`, as the filled spelling already gives.

jooyoungseo:
Closes #583.

## The defect

`sns.histplot(..., fill=False)` draws its distribution as a `Line2D` outline
rather than as patches, and nothing in `maidr/patch/histogram.py` looked at
lines. The chart registered **no layer at all** — not a wrong reading, silence.

Measured across the element × fill × hue matrix, before this change:

| call | layers |
|---|---|
| `element="bars"`, `fill=True` | `["hist"]` |
| `element="bars"`, `fill=False` | `[]` |
| `element="step"`, `fill=True` | `["hist"]` |
| `element="step"`, `fill=False` | `[]` |
| `element="poly"`, `fill=True` | `["hist"]` |
| `element="poly"`, `fill=False` | `[]` |
| `element="step"`, `fill=False`, `hue=` (2 levels) | `[]` |
| `element="poly"`, `fill=False`, `hue=` (2 levels) | `[]` |
| `element="step"`, `fill=False`, `kde=True` | `[]` |
| `element="poly"`, `fill=False`, `kde=True` | `[]` |

Four of the ten were silent for want of a reading; `element="bars", fill=False`
still draws unfilled patches and was already read.

## The reading

New `maidr/core/plot/outlined_histogram.py`:

- `_read_line(line)` decides which of the two coordinate columns is the binned
  axis by which one ascends, so a `y=`-oriented histogram reads without a
  separate code path.
- `_bins_from(positions, values, stepped)` turns the vertices back into bins. A
  **stepped** outline traces bin *edges*, so the last value is the closing zero
  and is dropped. A **poly** outline traces bin *centres*, so the edges are
  recovered from the spacing — and when that spacing is uneven the centres do
  not determine the edges, so it returns `None` and the layer declines rather
  than inventing them.
- `OutlinedHistPlot` subclasses `HistPlot` and reuses its bin payload, so the
  emitted schema is byte-identical to the filled equivalent. Highlighting is off
  (`_support_highlighting = False`): a single `<path>` cannot be selected per
  bin.

`maidr/patch/histogram.py` gains `_drew_outline_lines(ax, before, kde)`, which
takes the lines seaborn added during the call and keeps the outlines. Each is
registered as its own `PlotType.HIST` layer, named from the legend by colour via
the existing `_names_for`, so a `hue=` outline names its groups the way the
`element="bars"` spelling does.

## What this deliberately does not read

`element="poly", fill=False, kde=True` stays silent, and a test pins that.

A poly outline and a KDE curve are both `drawstyle="default"` and differ only in
vertex count — which both `gridsize=` and the bin count move, so no threshold
separates them. A positional split does not work either: an XML read of the axes
shows seaborn interleaves them per hue group, `[outline, kde, outline, kde]`,
not outlines-then-curves. Announcing a density as a distribution is worse than
announcing nothing, and the chart was already silent, so this is not a
regression. `element="step", fill=False, kde=True` **is** read — a stepped
outline is unambiguous by drawstyle, which is exactly the discriminator the
filter uses.

## Review round: a name computed and dropped

`OutlinedHistPlot.__init__` called `HistPlot.__init__(ax)` without its kwargs,
so the `GROUP_NAME` the patch matched from the legend never reached the layer —
a `hue=` chart came out with two anonymous `hist` layers. Fixed in `cafe1f8`,
and the hue test now asserts the names rather than the layer count.

Measuring that also showed the **filled** `element="step"`/`"poly"` path names
nothing either, and for a wider reason: its caller never computes a name at all.
That is older than this branch, so it is filed as #587 rather than folded in.

| call | layer names |
|---|---|
| `element="bars"`, `fill=True` | `["b", "a"]` |
| `element="step"`, `fill=True` | `[None, None]` (#587) |
| `element="step"`, `fill=False` | `["b", "a"]` |

## Verification

- Unfilled agrees with filled — same bin points, same `orientation` — for all
  four element × orientation combinations.
- `steps-post` and `steps-pre` produce the same array layout, so the edge-drop
  is not drawstyle-specific.
- Uneven bins, `element="step"`: edges `0/1/5/10` read back exactly as counts
  `0/6/2`. Uneven bins, `element="poly"`: declines.
- Six mutations applied one at a time against a restored baseline, all killed:
  dropping the last stepped value, the ascending-column choice, the even-spacing
  guard, the drawstyle filter, the horizontal swap, and the kwargs forward.
- `tests/core/plot/test_unfilled_histogram.py` — 20 tests.
- Full suite: **2795 passed, 18 skipped**. `ruff check` clean.

### DIFF
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -19,6 +19,7 @@ from maidr.core.plot.hexbinplot import HexbinPlot
 from maidr.core.plot.histogram import HistPlot
 from maidr.core.plot.lineplot import MultiLinePlot
 from maidr.core.plot.lollipop import LollipopPlot
+from maidr.core.plot.outlined_histogram import OUTLINE_LINE, OutlinedHistPlot
 from maidr.core.plot.maidr_plot import MaidrPlot
 from maidr.core.plot.pieplot import PiePlot
 from maidr.core.plot.pointplot import PointPlot
@@ -117,6 +118,11 @@ class MaidrPlotFactory:
             # find and the call hands its `PolyCollection` over instead.
             if isinstance(kwargs.get("collection"), PolyCollection):
                 return SteppedHistPlot(single_ax, **kwargs)
+            # The same two elements drawn `fill=False`, which swaps the
+            # collection for a bare `Line2D` and left the chart silent (#583).
+            outline = kwargs.get(OUTLINE_LINE)
+            if outline is not None:
+                return OutlinedHistPlot(single_ax, outline, **kwargs)
             # `ax.hist(histtype="step"/"stepfilled")` draws a `Polygon` per
             # dataset and leaves no container either, so the call hands over
             # the counts and edges it already returned (#555).
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -19,8 +19,8 @@ from maidr.core.plot.hexbinplot import HexbinPlot
 from maidr.core.plot.histogram import HistPlot
 from maidr.core.plot.lineplot import MultiLinePlot
 from maidr.core.plot.lollipop import LollipopPlot
-from maidr.core.plot.outlined_histogram import OUTLINE_LINE, OutlinedHistPlot
 from maidr.core.plot.maidr_plot import MaidrPlot
+from maidr.core.plot.outlined_histogram import OUTLINE_LINE, OutlinedHistPlot
 from maidr.core.plot.pieplot import PiePlot
 from maidr.core.plot.pointplot import PointPlot
 from maidr.core.plot.scatterplot import ScatterPlot
--- a/maidr/core/plot/outlined_histogram.py
+++ b/maidr/core/plot/outlined_histogram.py
@@ -75,7 +75,14 @@ class OutlinedHistPlot(HistPlot):
 
     def __init__(self, ax: Axes, line: Line2D, **kwargs) -> None:
         self._line = line
-        super().__init__(ax)
+        # Forwarded, not dropped: the patch names each outline from the
+        # legend swatch its colour matches and hands the name over under
+        # `GROUP_NAME`, which `HistPlot.__init__` is the thing that reads.
+        # Calling it with the axes alone computed every name and threw them
+        # away, leaving a `hue=` chart with two "hist" layers and no way to
+        # tell which group either announced -- the same silence this class
+        # exists to end, one level down.
+        super().__init__(ax, **kwargs)
         # One `<path>` for the whole outline, as `SteppedHistPlot` and
         # `Axes.stairs` have: there is no per-bin element for a selector to
         # name, and one naming the line would light the whole distribution up
--- a/maidr/core/plot/outlined_histogram.py
+++ b/maidr/core/plot/outlined_histogram.py
@@ -14,6 +14,23 @@ from maidr.core.plot.histogram import HistPlot
 #: layer -- the defect #527 fixed for containers.
 OUTLINE_LINE = "_maidr_outline_line"
 
+#: The keyword the histogram patch says which axis the distribution runs
+#: along under, as ``True`` for a ``histplot(y=...)`` and ``False`` for the
+#: ordinary one.
+#:
+#: Handed over because the drawing cannot always answer it and the caller
+#: always can. A stepped outline gives itself away -- its counts column
+#: closes on a repeated value, so it is never strictly ascending -- but a
+#: ``poly`` outline carries one bin centre and one count per vertex with
+#: nothing repeated, and a histogram whose counts happen to climb evenly has
+#: two ascending, evenly spaced columns and no way to tell them apart. Two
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Confirmed by measurement and fixed in `cafe1f8`.

`histplot(hue="g", element="step", fill=False)` came out `[None, None]` while the patch had computed `["b", "a"]` for it — `super().__init__(ax, **kwargs)` now forwards them, and `test_each_hue_group_carries_the_name_its_colour_is_given` asserts the values rather than the layer count, so removing the forward turns it red again (verified: both parametrisations fail).

Measuring it also settled the `SteppedHistPlot` question you raised. The filled `element="step"`/`"poly"` path is not merely latent — it is silent today:

| call | layer names |
|---|---|
| `element="bars"`, `fill=True` | `["b", "a"]` |
| `element="step"`, `fill=True` | `[None, None]` |
| `element="step"`, `fill=False` | `["b", "a"]` (after this change) |

Its caller never computes a name, so fixing it needs the colour match added as well as the kwargs forwarded. That is wider than #583 and older than this branch, so it is filed as #587 rather than folded in — and the PR description's "the same way a filled one does" is corrected to name the `bars` spelling, which is the one that actually does.

Import ordering tidied too. The blank lines before `_drew_outline_lines` measured as two, matching the rest of the module, so I left them.

---

jooyoungseo:
It reproduces. Confirmed and fixed in `7fa848a`.

```
sns.histplot(df, y="v", bins=2, element="poly", fill=False)   # counts 2 and 5
  xydata      [[2.0, 0.3125], [5.0, 0.7375]]
  orientation 'vert'
  data        xMin 0.5, xMax 3.5, y 0.3125
```

Exactly as predicted: the bin edges built out of the counts, the bin centre 0.3125 announced as the count. And it is not a corner case — a single gap is evenly spaced whatever it holds, so **every** two-bin `poly` is ambiguous, not only one with monotonic counts. The four-bin shape needs the counts to climb *evenly* (1/2/3/4), which is also ordinary.

Your diagnosis of why `step` escapes is right too, and it is what the fix leans on: the counts column closes on a repeated value, so it is never strictly ascending and one column alone reads as bins.

The fix does not reuse geometry, because there is none left to reuse — that is the difference from `SteppedHistPlot`, whose `PolyCollection` carries the baseline that gives the shared coordinate away. A bare `Line2D` does not have it. So the patch hands the orientation over (`OUTLINE_HORIZONTAL`, from `histplot`'s keyword-only `y=`) and the reading consults it **only to break the tie**: where one column alone reads as bins the drawing still decides, and where both read and nothing was handed over it declines rather than picks — the rule an uneven `poly` already follows.

Tests added: the two-bin and four-bin reproducers in both elements, each pinned against its filled twin, plus a direct `_read_line` test for the decline. Five mutations applied one at a time against a restored baseline, all killed — including "take the first reading", which is the pre-fix behaviour.

Two smaller things from the same measurement:

- `kwargs.get("x") is None` in the orientation test was dead. A bivariate `histplot(x=, y=)` draws no `Line2D` at all — one `QuadMesh`, read as a heatmap — so it never reaches the outline branch. A mutation removing the condition survived; it is gone.
- Blank line before `_drew_mesh` fixed, and the two `outlined_histogram` imports left as separate lines because one is a plain name and one is aliased.

Full suite: 2800 passed, 18 skipped.

---

Issue #544, opened 2026-08-20
A hue-grouped scatter is announced as one undifferentiated cloud


jooyoungseo:
## What happens

```python
sns.scatterplot(data=df, x="a", y="b", hue="g")
```

reads as a single `point` layer holding every observation, with no trace of `g` anywhere in the payload:

```python
layer.keys()   # axes, data, id, selectors, title, type
layer["axes"]  # {'x': {'label': 'a', …}, 'y': {'label': 'b', …}}   — no z
layer["data"][:2]
# [{'x': 0.034…, 'y': 1.259…}, {'x': 1.359…, 'y': -0.028…}]
```

A sighted reader sees three colours and a legend titled `g`. A blind reader gets 60 points and no way to tell which group any of them is in — the grouping *is* the chart, and it is the one thing that does not survive.

No data is lost: all 60 points are present, in draw order. What is lost is which series each belongs to.

## Where it bites

Same reading for `sns.scatterplot(hue=)`, `sns.relplot(kind="scatter", hue=)`, `sns.jointplot(hue=)` and `sns.pairplot(hue=)`. Measured on a 2-group frame, both grid charts:

```
jointplot(hue):  joint panel -> 1 point layer, 60 points, no group
pairplot(hue):   each off-diagonal cell -> 1 point layer, no group
```

The marginals of both are already correct — they emit **one `smooth` layer per hue level** — so within a single chart the KDE curves are navigable by group and the scatter they describe is not.

## Why it happens

`seaborn` draws a hue-grouped scatter as **one** `PathCollection` with per-point colours, not one collection per group:

```python
coll = ax.collections[0]
len(coll.get_offsets())          # 12
len(coll.get_facecolors())       # 12
len({tuple(c) for c in coll.get_facecolors()})   # 3
```

`ScatterPlot` reads the collection's offsets and stops there. Its constructor comment already records the single-collection fact, but in service of *which* collection to read rather than of what the colours mean.

## The grouping is fully recoverable

The legend carries the map, complete with the variable's name:

```python
leg = ax.get_legend()
leg.get_title().get_text()                  # 'g'
[t.get_text() for t in leg.get_texts()]     # ['x', 'y', 'z']
[h.get_facecolor() for h in leg.legend_handles]
# [0.122 0.467 0.706 1], [1.0 0.498 0.055 1], [0.173 0.627 0.173 1]
```

and those are exactly the three distinct colours in `get_facecolors()`, so every point's group is a lookup away.

[...]

PR #588, opened 2026-08-22
feat(scatter): read the hue a strip or swarm plot was grouped by

### LINKED ISSUE #586: A hue-grouped stripplot or swarmplot is read as though it had no hue
## What happens

`sns.stripplot(..., hue=...)` and `sns.swarmplot(..., hue=...)` emit exactly the
same schema as the same call without `hue=`. The grouping is not merely
unnamed — it is absent. Nothing in the payload distinguishes the two subgroups,
so a reader has no way to tell which points are which, on a chart where sighted
readers are told by colour and by a legend.

```python
df = pd.DataFrame({
    "cat": ["a"]*6 + ["b"]*6 + ["c"]*6,
    "val": [1,2,3,4,5,6, 10,11,12,13,14,15, 20,21,22,23,24,25],
    "hue": (["x"]*3 + ["y"]*3) * 3,
})
sns.stripplot(data=df, x="cat", y="val", hue="hue")
```

| call | layers | names | `z` |
|---|---|---|---|
| `stripplot(x, y)` | `point`, `point`, `point` | — | — |
| `stripplot(x, y, hue=)` | `point`, `point`, `point` | — | — |
| `swarmplot(x, y, hue=)` | `point`, `point`, `point` | — | — |
| `catplot(kind="strip", hue=)` | `point`, `point`, `point` | — | — |

Byte for byte identical: the three layers are the three *categories*, each
holding its own six points, and the legend `['x', 'y']` that seaborn drew never
reaches the schema.

Compare `sns.scatterplot(..., hue=...)` on the same frame, which #544 already
splits into one named layer per hue level with the hue variable on `z`. The same
data drawn two ways reads two different ways, and only one of them is right.

## Why

Both charts register through `maidr/patch/scatterplot.py`, at the inner
`Axes.scatter` calls seaborn makes — one per category. `hue_groups()` needs two
things from the collection at that moment, and **neither exists yet**. Measured
by spying on `Axes.scatter`:

```
strip hue  at each scatter return: legend present = False, facecolor rows = 1, unique = 1
           after the call:         legend = ['x', 'y'], unique colours per collection = 2
```

Seaborn's `_CategoricalPlotter.plot_strips` ends with

```python
points = ax.scatter(sub_data["x"], sub_data["y"], color=color, **plot_kws)
if "hue" in self.variables:
    points.set_facecolors(self._hue_map(sub_data["hue"]))
...
self._configure_legend(ax, _scatter_legend_artist, common_kws=plot_kws)
```

so the per-point hue colours are assigned *after* `ax.scatter` returns, and the
legend is built after every category has been drawn. `hue_groups()` sees one
uniform colour and no legend, declines on both counts, and the layer is
registered ungrouped. This is not a bug in `hue_groups` — it is being asked
before the answer exists.

`sns.scatterplot` avoids it only because it is wrapped at the seaborn level:
the whole call runs inside the internal context and the collection is read
afterwards. `stripplot`/`swarmplot` have no such wrapper.

## Where the fix belongs

`_CategoricalPlotter.plot_strips` / `plot_swarms`, the way `plot_bars` and
`plot_boxens` are already wrapped — `catplot` drives the plotter directly and
imports neither public function, so a patch on `seaborn.stripplot` reaches
three of the four rows above and a patch on the method reaches all four
(verified: `catplot(kind="strip")` does not call `seaborn.stripplot`).

Both methods call `_configure_legend` themselves as their last statement, so
the legend and the hue colours are both in place by the time the wrapped method
returns.

`dodge=True` is the second shape and needs its own answer: it splits each
category into one collection per hue level, so every collection is uniformly
coloured and `hue_groups` declines for a different reason. Measured — six
collections of three points, one unique colour each — and matplotlib still
writes one `<g>` per point for them, so a per-point selector remains valid.

The no-hue decomposition should not change: #426 fixed these charts to read one
layer per category and `tests/core/plot/test_scatter_own_collection.py` pins it.

jooyoungseo:
Closes #586.

## The defect

`sns.stripplot(..., hue=...)` emitted exactly the schema of the same call
**without** a `hue=`. Not a wrong grouping — no grouping at all, on a chart
whose subject is the comparison between the levels.

| call | layers | names | `z` |
|---|---|---|---|
| `stripplot(x, y)` | `point` ×3 | — | — |
| `stripplot(x, y, hue=)` | `point` ×3 | — | — |
| `swarmplot(x, y, hue=)` | `point` ×3 | — | — |
| `catplot(kind="strip", hue=)` | `point` ×3 | — | — |

The three layers were the three *categories*; the legend `['x', 'y']` seaborn
drew never reached the schema.

## Why

These charts registered at the inner `Axes.scatter` calls seaborn makes, one
per category. `hue_groups` needs two things from the collection at that
moment and **neither exists yet** — `plot_strips` writes both afterwards:

```python
points = ax.scatter(sub_data["x"], sub_data["y"], color=color, **plot_kws)
if "hue" in self.variables:
    points.set_facecolors(self._hue_map(sub_data["hue"]))
...
self._configure_legend(ax, _scatter_legend_artist, common_kws=plot_kws)
```

Measured by spying on `Axes.scatter`:

```
at each of the three returns:  legend present = False, facecolor rows = 1, unique = 1
after the call:                legend = ['x', 'y'], unique colours per collection = 2
```

`hue_groups` is not at fault — it was being asked before the answer existed.
`sns.scatterplot` escapes only because it is wrapped at the seaborn level and
read afterwards.

## The reading

New `maidr/patch/stripplot.py` wraps `_CategoricalPlotter.plot_strips` and
`plot_swarms`, the way `plot_bars` and `plot_boxens` already are — `catplot`
drives the plotter directly and imports neither public function, so a patch on
`seaborn.stripplot` would have reached three of the four rows above (verified:
`catplot(kind="strip")` does not call `seaborn.stripplot`). Both methods call
`_configure_legend` as their last statement, so drawing inside the internal
context and reading after it puts everything in place.

The grouping comes off the plotter's own `_hue_map`, not off a legend:

- a faceted `catplot` panel has **no legend** — the grid builds one at the
  figure, afterwards;
- a panel holding only one of the levels could not be named from a legend
  anyway, since one colour against a swatch is a guess.

Levels are keyed on the three colour channels rather than four, because
`alpha=` scales the drawn points and leaves the mapping alone — measured,
`(0.12, 0.47, 0.71, 0.4)` drawn against a lookup entry of `(..., 1.0)`.

A **numeric** `hue=` is declined: seaborn builds one lookup entry per distinct
value, so eighteen rows offer eighteen "levels". That is a colour scale, and
one layer per point is not a reading of it — the same chart `hue_groups`
declines one level down.

## What a panel gets

- **A hue level each**, spanning the categories, named by the level with the
  hue variable on `z`. That is the decomposition `scatterplot(hue=)` already
  emits for the same frame (#544), with the category on each point's `xLabel`
  — so the same data drawn two ways finally reads the same way.
- **A collection each, unnamed**, when there is no hue to read: one layer per
  category, exactly what #426 settled and `test_scatter_own_collection.py`
  pins. Untouched.

`ScatterPlot` now takes a **list** of collections, since a hue level spans
seaborn's one-collection-per-category drawing, and addresses each point
through its own collection's id. A layer handed a single collection — every
scatter before this — takes the one-entry path through all of it unchanged.

`GROUP_LABEL` is a new fallback for the `z` label, for the three charts that
have a grouping and no legend title to read it from: `catplot`, `legend=False`,
and a `hue=` that repeats `x` (seaborn draws that one titleless).

## Verification

| call | after |
|---|---|
| `stripplot(hue=)` | `x`(9), `y`(9), `z: hue` |
| `swarmplot(hue=)` | `x`(9), `y`(9), `z: hue` |
| `stripplot(hue=, dodge=True)` | `x`(9), `y`(9) |
| `stripplot(hue=, alpha=.4)` | `x`(9), `y`(9) |
| `stripplot(y=, x=, hue=)` | `x`(9), `y`(9), categories on `yLabel` |
| `catplot(kind="strip", col=, hue=)` | `x`(6), `y`(3) \| `x`(3), `y`(6) |
| `stripplot(hue="num")` | unchanged — 3 unnamed |
| `stripplot()` | unchanged — 3 unnamed, 6 points each |

- Every drawn point announced exactly once, no duplicates, values matching the
  offsets.
- Selectors resolved against the rendered SVG with `lxml.cssselect`: 18
  selectors, each matching exactly **one** element, 18 distinct across the
  chart — no overlap and no misses. The same check on `strip plain` and
  `scatterplot(hue=)` is unchanged.
- Eleven mutations applied one at a time against a restored baseline, ten
  killed: reading collection 0 for every selector, extracting only the first
  collection, dropping the categorical guard, matching on full RGBA, dropping
  the hue-order sort, drawing outside the internal context, reading every
  collection rather than the new ones, grouping a single-level hue, dropping
  the `z` fallback, and truncating the colour list.
- The eleventh — removing the decline when a collection's facecolor rows do not
  match its point count — **survives**, because that branch is unreachable
  today: seaborn writes one row per point on every shape measured (strip,
  swarm, dodged, faceted, translucent, unfilled marker, and the empty
  collections a faceted grid leaves, which have neither). Rather than leave
  that implicit, `test_seaborn_gives_every_point_its_own_colour` pins the
  assumption across all six, so the release that ends it turns a test red
  instead of making that branch quietly load-bearing.
- `tests/core/plot/test_categorical_scatter_hue.py` — 29 tests.
- Full suite: **2829 passed, 18 skipped**. `ruff check` clean.

---

### DIFF
--- a/maidr/core/plot/scatterplot.py
+++ b/maidr/core/plot/scatterplot.py
@@ -15,8 +15,9 @@ from maidr.exception import ExtractionError
 from maidr.util.mixin import CollectionExtractorMixin, LineExtractorMixin
 
 
-#: The keyword ``Axes.scatter`` hands its own ``PathCollection`` to this layer
-#: under. Named once and imported at both ends rather than spelled twice:
+#: The keyword a patch hands this layer the collections its own call drew
+#: under -- one ``PathCollection``, or a list of them where a layer spans
+#: several. Named once and imported at both ends rather than spelled twice:
 #: ``kwargs.get`` falls back to sweeping the axes on a mismatch, so a typo
 #: would not raise -- it would quietly restore the behaviour #426 removed.
 #:
@@ -24,11 +25,26 @@ from maidr.util.mixin import CollectionExtractorMixin, LineExtractorMixin
 #: imports ``maidr.core`` and not the other way about.
 DRAWN_POINTS = "_maidr_points"
 
-#: The keyword the scatter patch hands one hue group to this layer under: a
-#: ``(name, members)`` pair naming the group and listing the collection
-#: offsets that belong to it. Absent for an ungrouped scatter, which is the
-#: chart this layer has always read and still reads unchanged.
+#: The keyword a patch hands one hue group to this layer under: a
+#: ``(name, members)`` pair naming the group and listing, **per collection of**
+#: ``DRAWN_POINTS`` **and in the same order**, the offsets that belong to it.
+#: One collection is one entry; the list is not flattened, because an offset
+#: means nothing without the collection it indexes. Absent for an ungrouped
+#: scatter, which is the chart this layer has always read and still reads
+#: unchanged.
 HUE_GROUP = "_maidr_hue_group"
+
+#: The keyword a patch hands the grouping *variable's* name under, for the
+#: charts whose legend cannot be asked for it. Only ever a fallback: the
+#: legend's title is read first and wins, because it is what the chart shows.
+#:
+#: Three charts need it, and all three are ones the legend cannot answer for.
+#: ``sns.catplot`` builds its legend at the *figure* after the panels are
+#: drawn; ``legend=False`` suppresses it outright; and a ``hue=`` that repeats
+#: the ``x`` variable makes seaborn draw one with no title at all. Each still
+#: has a grouping, and a group named "x" with nothing saying what "x" is a
+#: kind of is half a reading.
+GROUP_LABEL = "_maidr_group_label"
 
 #: How closely two colours must agree to be the same colour. Both sides come
 #: from the same palette object -- the legend handle is built from the swatch
@@ -129,6 +145,38 @@ def _named_colours(legend, drawn: set[tuple[float, ...]]) -> dict | None:
     return named
 
 
+def _collections(given) -> list[PathCollection]:
+    """
+    The collections a patch handed this layer, in the order it drew them.
+
+    Either spelling is read: ``Axes.scatter`` has one collection to give and
+    gives it, while a patch whose layer spans several -- a hue group over
+    ``seaborn``'s one-collection-per-category strip plot (#586) -- gives the
+    list. Everything below works in the list, so a single collection becomes
+    a list of one here rather than a second path through the layer.
+
+    Filtered by type rather than trusted, for the reason the one-collection
+    form was: ``seaborn.scatterplot`` is patched through the same wrapper and
+    returns an ``Axes``, not a collection. An empty answer falls back to
+    sweeping the axes, which is right for that call and for that call alone.
+
+    Parameters
+    ----------
+    given : Any
+        Whatever arrived under :data:`DRAWN_POINTS`.
+
+    Returns
+    -------
+    list of PathCollection
+        Possibly empty, which is the caller's signal to sweep instead.
+    """
+    if isinstance(given, PathCollection):
+        return [given]
+    if isinstance(given, (list, tuple)):
+        return [part for part in given if isinstance(part, PathCollection)]
+    return []
+
+
 def hue_groups(ax: Axes, collection: PathCollection) -> list[tuple[str, list[int]]] | None:
     """
     The hue groups a scatter was drawn with, or ``None`` when it has none.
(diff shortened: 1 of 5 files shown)


jooyoungseo:
Two of the three taken in `44be102`; the third I'd push back on.

**Taken.** The second `plotter_axes(instance)` is gone — the panel list is fixed before the plotter draws (measured: one before and after for `stripplot`, two for `catplot(col=)`), so the second call was answering the first one's question. And `_hue_levels` now carries `list[tuple[str, list[list[int]]]] | None` — you're right that the shape is exactly what `ScatterPlot` indexes by, so it belongs in the signature and not only the prose.

**Not taken: the `Dict`/`Tuple` style.** The uppercase spelling is the convention for this exact signature in the modules `stripplot.py` was modelled on, not a slip:

```
maidr/patch/barplot.py:23     args: Tuple[Any, ...], kwargs: Dict[str, Any]
maidr/patch/barplot.py:229    args: Tuple[Any, ...], kwargs: Dict[str, Any]
maidr/patch/barplot.py:276    args: Tuple[Any, ...], kwargs: Dict[str, Any]
maidr/patch/candlestick.py:16 args: Tuple[Any, ...],
```

`barplot.py:276` is `sns_categorical_bars` — the `plot_bars` wrapper this one is the sibling of. `scatterplot.py` is a different module with a different (matplotlib-level) signature. Making this file the odd one out among the plotter wrappers would trade one inconsistency for another, so it stays as it is; a sweep across all of them is its own change if the project wants lowercase everywhere.

On the private-seaborn-API note: agreed, and worth being explicit that the risk is bounded. Every one of `_hue_map.map_type`, `.lookup_table` and `.levels` is read through `getattr` with a decline as the default, and a decline means the chart reads exactly as it did before this PR — one unnamed layer per category. A seaborn release that moves them costs the naming, not the chart. `test_seaborn_gives_every_point_its_own_colour` is the one that would go red loudly, which is the intent.

---

jooyoungseo:
Measured the RGB-collision case rather than leaving it as a nice-to-have, and it turns out to be reachable in two ways, not one — pinned in `39f3cd6`:

```
palette={"x": (0,0,1,.3), "y": (0,0,1,.9)}   ->  [None, None, None]   # same hue, different opacity
palette={"x": "blue",     "y": "blue"}        ->  [None, None, None]   # the same colour twice
palette={"x": "red",      "y": "blue"}        ->  [("x", 9), ("y", 9)]
```

Both collisions fall through the `len(named) == len(lookup)` guard to the ungrouped reading, with every point still announced exactly once — so the failure mode is "hue dropped", not "both levels handed to whichever name matched first". The second case is the more interesting one: it needs no rounding at all, just a palette that draws two levels alike, which is a thing people do deliberately. Worth having a test say so.

On the seaborn-version watch: agreed, and `test_seaborn_gives_every_point_its_own_colour` is deliberately written against six drawing shapes rather than one so a bump cannot pass by exercising only the easy path.

---

PR #589, opened 2026-08-22
fix(hist): name the hue groups of a filled step or poly histogram

### LINKED ISSUE #587: A filled step or poly histogram loses its hue group names
## What happens

`sns.histplot(..., hue=..., element="step"|"poly")` — the **filled** spelling —
emits one layer per group, correctly split, and names neither. The reader gets
two `hist` announcements with nothing to tell them apart, on a chart whose
whole point is the comparison between the groups. `element="bars"` on the same
frame names both.

Measured on seaborn 0.13.2, two groups of twenty rows, four bins:

| call | layer names |
|---|---|
| `element="bars"`, `fill=True` | `["b", "a"]` |
| `element="step"`, `fill=True` | `[None, None]` |
| `element="poly"`, `fill=True` | `[None, None]` |
| `element="step"`, `fill=False` | `["b", "a"]` (as of #585) |

The unfilled outline now names its groups, so the filled twin is the odd one
out — the same distribution, drawn the other way, reads with less information.

## Why

`maidr/patch/histogram.py` registers the filled outlines with the collection
alone:

```python
drew = _drew_outlines(axes, outlines)
if drew:
    for outline in drew:
        FigureManager.create_maidr(axes, PlotType.HIST, collection=outline)
    return drawn
```

No `GROUP_NAME` is computed, so there is nothing for the layer to carry. The
bar path (`deferred_names(...)` over the containers) and the unfilled-outline
path (the same over the lines) both do compute one; this branch is the only
one that does not.

`SteppedHistPlot.__init__` also calls `HistPlot.__init__(ax)` without its
kwargs, so even a name passed here would be dropped — the same defect #585
fixed on `OutlinedHistPlot`. Both halves need doing: forward the kwargs, and
give the patch a name to forward.

## What a fix looks like

Mirror the unfilled branch, matching each `PolyCollection`'s face colour
against the legend swatch that names it, via the existing `_names_for`. Worth
measuring first whether seaborn draws the filled outlines translucent while
the swatches are opaque — `_names_for`'s second pass exists for exactly that
divergence on the KDE overlay, and it may or may not apply here.

Not urgent: nothing is announced *wrongly* today, only anonymously.

jooyoungseo:
Closes #587.

## The defect

`sns.histplot(hue=..., element="step"|"poly")` split into one layer per group
and named **none** of them. Not a wrong reading — two `hist` announcements over
one axis with nothing to tell them apart, which is exactly the position
`MaidrLayer.name` was added for (xability/maidr#828).

Measured, two groups of twenty rows over four bins:

| call | layer names |
|---|---|
| `element="bars"` | `["b", "a"]` |
| `element="step"` | `[None, None]` |
| `element="poly"` | `[None, None]` |
| `element="step"`, `fill=False` | `["b", "a"]` (since #585) |

`element=` is a purely visual choice, so the filled outline was the only
spelling of this chart that cost the reader the grouping.

## Why, and both halves of it

The patch registered each outline with its collection and nothing else — **no
name was ever computed**:

```python
for outline in drew:
    FigureManager.create_maidr(axes, PlotType.HIST, collection=outline)
```

And `SteppedHistPlot.__init__` called `HistPlot.__init__(ax)` without its
kwargs, so a name handed over would have been dropped on the way in — the same
defect #585 fixed on `OutlinedHistPlot`, which is why the review there flagged
this class as latent. It was not latent; it was silent for a second reason as
well, and fixing either alone changes nothing.

## Which colour names a group

The outline's **face**.

By default the edge carries the hue too, so either would name the groups — the
face matches the legend swatch outright (both `(0.12, 0.47, 0.71, 0.25)`,
translucency included) and the edge matches through `_names_for`'s
alpha-insensitive second pass. Nothing in the ordinary chart distinguishes the
two choices, which is how a mutation swapping them survived the first pass.

`edgecolor=` is what separates them, and it is an ordinary thing to pass:

```
element="step"                       edges (1.0, .50, .05) / (.12, .47, .71)
element="step", edgecolor="black"    edges (0, 0, 0)       / (0, 0, 0)
```

One edge colour for every group names nothing — an edge-based match returns
`[None, None]` for a chart that is merely outlined in black. The face is
untouched by it. `test_an_outline_drawn_with_one_edge_colour_is_still_named`
pins that, and it is what turns the mutation from surviving to killed.

Names are deferred like the bars' and the density curves', because a
`pairplot`'s legend does not exist until every panel has been drawn (#561).

## Verification

| call | after |
|---|---|
| `element="bars"` | `["x", "y"]` |
| `element="step"` | `["x", "y"]`, counts per group correct |
| `element="poly"` | `["x", "y"]`, counts per group correct |
| `element="step"`, `y=` | `["x", "y"]`, both `horz` |
| `element="step"`, `edgecolor="black"` | `["x", "y"]` |
| `element="step"`, no hue | `[None]` — unchanged |
| `pairplot(diag_kind="hist", element="step")` | `['y', 'x', 'y', 'x']` — the deferred path |

- Each layer's counts checked against `np.histogram` over that group's rows
  alone, so the names are attached to the right distributions rather than
  merely present.
- Four mutations applied one at a time against a restored baseline, all killed:
  dropping the kwargs again, computing no names, reading the edge colour, and
  pairing the names in reverse.
- Full suite: **2838 passed, 18 skipped**. `ruff check` clean.

## Still out of scope

`displot(kind="hist", fill=False)` registers nothing at all — `displot` drives
the plotter directly and that patch reads containers only, so the unfilled
spelling never reaches a reading there. Pre-existing and wider than this;
worth its own issue rather than folding in.

---

### DIFF
--- a/maidr/core/plot/stepped_histogram.py
+++ b/maidr/core/plot/stepped_histogram.py
@@ -62,7 +62,11 @@ class SteppedHistPlot(HistPlot):
 
     def __init__(self, ax: Axes, collection: PolyCollection, **kwargs) -> None:
         self._collection = collection
-        super().__init__(ax)
+        # Forwarded, not dropped. `HistPlot.__init__` is what reads
+        # `GROUP_NAME`, so calling it with the axes alone leaves a `hue=`
+        # chart with a layer per group and no name on any of them -- the
+        # defect #585 fixed on the unfilled twin of this class, and #587 here.
+        super().__init__(ax, **kwargs)
         # One `<path>` for the whole outline, as `Axes.stairs` has: there is no
         # per-bin element for a selector to name, and one naming the outline
         # would light the whole distribution up at every bin.
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -458,8 +458,13 @@ def _face_colour(outline: PolyCollection):
     Returns
     -------
     tuple of float or None
-        The rounded RGBA, or ``None`` when the collection names no single
-        colour.
+        The rounded RGBA of the collection's first row, or ``None`` when it
+        holds no rows or that row names no colour.
+
+        The first row *is* the colour here rather than a sample of several:
+        seaborn draws one outline per series and colours it in one call, so
+        the collection carries a single face. Said rather than checked, since
+        a uniformity test over one row could not fail.
     """
     rows = np.asarray(outline.get_facecolor())
     return _rgba(rows[0]) if len(rows) else None
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -432,6 +432,39 @@ def _outlines_of(ax: Axes | None) -> list:
     ]
 
 
+def _face_colour(outline: PolyCollection):
+    """
+    The one colour a filled outline was drawn in, if it has one.
+
+    Its **face**, which is what carries the hue. By default the edge carries
+    it too -- measured, the two agree but for the translucency the face and
+    the legend swatch share -- so either would name the groups, the edge
+    through `_names_for`'s alpha-insensitive pass. `edgecolor=` is what
+    separates them: it is the caller's to set, and setting it once colours
+    *every* group's edge alike::
+
+        histplot(hue="g", element="step")                    edges (1.0, .50, .05) / (.12, .47, .71)
+        histplot(hue="g", element="step", edgecolor="black") edges (0, 0, 0) / (0, 0, 0)
+
+    Two groups drawn one colour name nothing, and a chart outlined in black
+    is an ordinary chart rather than a strange one. The face is untouched by
+    it, so it is what this reads.
+
+    Parameters
+    ----------
+    outline : PolyCollection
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Fair catch on the docstring — fixed in `6172ba1`, by correcting the prose rather than adding the check.

`kdeplot._collection_colour` needs its `len(colours) != 1` test because a KDE fill can carry several rows. A stepped histogram's `PolyCollection` cannot: seaborn draws one outline per series and colours it in a single call. Measured on a two-group stepped histogram — one facecolor row per collection, both of them. A uniformity test over one row could not fail, so it would be an unreachable branch of exactly the kind worth not adding; the docstring now says the first row *is* the colour and why, instead of implying a check.

While measuring for this PR I also found that `displot` is silent for every `element` but `bars` — filed as #590 with the matrix. It is a different patch (`displot` drives `_DistributionPlotter` directly and that registrar reads containers only) and it is a chart producing no HTML rather than a naming gap, so it stays out of here.

---

Issue #591, opened 2026-08-22
A faceted displot resolves every panel's group names from the last panel


jooyoungseo:
## What happens

`sns.displot(..., hue=..., col=...)` emits its layers correctly split — each
panel's groups, each with the right counts — and names **none** of them.
The unfaceted `histplot` spelling of the same data names them all.

Measured: three groups, `col` splitting them so panel 0 holds groups `a` and
`b` and panel 1 holds only `c`.

```
layers today:   [(None, [0, 10, 0]), (None, [10, 0, 0]), (None, [0, 0, 10])]
```

The counts are right; the names are gone. Asking each panel for its own names
directly gives:

```
panel 0 -> ['b', 'a']
panel 1 -> [None]          # one container, below names_for's two-artist floor
histplot, one panel -> ['c', 'b', 'a']
```

So two of the three names exist and are being thrown away.

## Why

`sns_distribution_hist` builds its deferred name resolver **inside** the
per-panel loop, capturing the loop variables by reference:

```python
for ax in plotter_axes(instance):
    ...
    containers = _new_bar_containers(ax, seen)
    for container, name in zip(containers, deferred_names(
        lambda: _group_names(ax, containers), len(containers)
    )):
```

`deferred_names` stores the callable and runs it at *render*, which is the
whole point of it — a `pairplot`'s legend does not exist until every panel is
drawn (#561). By then the loop has finished, so `ax` and `containers` both
hold the **last** panel's values, and every panel's names resolve against
them.

Here that means panel 1's single container answers for the whole figure, and
a single artist is exactly the case `names_for` declines (`len(colours) < 2`)
— so every layer comes back `None`. With a differently shaped grid it would
resolve to names belonging to another panel instead, which is worse: right
number of names, attached to the wrong distributions.

Only the faceted path is affected. `histplot` draws one panel, so there is
nothing for the binding to be late against.

## The fix

Bind the loop variables into the lambda at the point it is made
(`lambda ax=ax, containers=containers: ...`), so each panel's names resolve
against its own axes and its own containers, still lazily.

Worth checking the same shape wherever `deferred_names` is called from inside
a loop over panels.

jooyoungseo:
[...]

PR #592, opened 2026-08-22
feat(hist): read every element displot draws, not only bars

### LINKED ISSUE #590: displot is silent for every element but bars
## What happens

`sns.displot(..., element="step")` and `element="poly"` register **no layer at
all** and the figure falls back to a static image, filled or not. The same
chart drawn through `sns.histplot` reads fine.

Measured on seaborn 0.13.2, forty rows over four bins:

| call | layers |
|---|---|
| `histplot(x="v")` | `hist` |
| `histplot(x="v", element="step")` | `hist` |
| `histplot(x="v", element="step", fill=False)` | `hist` |
| `displot(x="v")` | `hist` |
| `displot(x="v", element="step")` | **nothing** — `UnsupportedPlotError` |
| `displot(x="v", element="step", fill=False)` | **nothing** |
| `displot(x="v", element="poly", fill=False)` | **nothing** |
| `displot(x="v", hue="g", element="step", fill=False)` | **nothing** |

`element="bars"` is the only spelling `displot` reads. Every other one is a
figure that produces no HTML rather than a wrong announcement — the #522 shape,
where one chart is readable through one spelling and silent through the other.

## Why

`displot` does not import `histplot`; it drives
`_DistributionPlotter.plot_univariate_histogram` directly, which is patched
separately from `sns_hist`. That patch registers **containers only**:

```python
for container, name in zip(containers, deferred_names(...)):
    FigureManager.create_maidr(
        ax, PlotType.HIST, **{DRAWN_BARS: container, GROUP_NAME: name}
    )
```

`element="step"`/`"poly"` draw a `PolyCollection` (filled) or a `Line2D`
(unfilled) and leave no container, so the loop has nothing to iterate and the
panel is registered as nothing at all.

Both readings already exist and are tested — `SteppedHistPlot` for the filled
outline and `OutlinedHistPlot` for the unfilled one (#583, #587). What is
missing is the branch in the plotter patch that reaches them, which is the same
shape `sns_hist` already has:

```python
drew = _drew_outlines(axes, outlines)            # -> SteppedHistPlot
outlined = _drew_outline_lines(axes, lines, kde, horizontal)   # -> OutlinedHistPlot
```

The faceted case needs the per-panel handling `plotter_panels` gives, and the
unfilled branch needs an orientation — `plotter.data_variable` says which axis
the distribution runs along, the plotter-level equivalent of the `y=` keyword
`_asked_for_horizontal` reads for `histplot`.

## Scope note

Filed rather than folded into #587, which was about *naming* a reading that
already happened. This is a reading that does not happen at all, through a
different patch, and covers four combinations rather than one.

jooyoungseo:
Closes #590 and #591 — two defects found in the same function, kept as two
commits.

## 1. `displot` read only one of its elements (#590)

`sns.displot(..., element="step")` and `element="poly"` registered **no layer
at all** and the figure fell back to a static image, filled or not. The same
chart through `sns.histplot` read fine — the asymmetry #522 and #446 were both
about.

| call | before |
|---|---|
| `histplot(x="v", element="step")` | `hist` |
| `histplot(x="v", element="step", fill=False)` | `hist` |
| `displot(x="v")` | `hist` |
| `displot(x="v", element="step")` | **nothing** |
| `displot(x="v", element="step", fill=False)` | **nothing** |
| `displot(x="v", element="poly", fill=False)` | **nothing** |

Those elements draw a `PolyCollection` when filled and a bare `Line2D` when
not, so the panel holds no `BarContainer` and the bar registrar passed over it.
Both readings already existed — `SteppedHistPlot` (#583) and `OutlinedHistPlot`
(#587). What was missing was the branch that reaches them, the same two
`sns_hist` has, factored into `_register_outlines`.

The unfilled reading needs an orientation, and `displot` forwards neither `x`
nor `y` to the plotter method. It comes from the plotter's own
`data_variable`, measured as `'x'` or `'y'` for every spelling.

## 2. A faceted panel's names came from the last panel (#591)

Found while measuring the above. `deferred_names` stores its resolver and runs
it at **render** — which is what makes a `pairplot`'s legend readable (#561) —
and the resolver was built inside the per-panel loop, closing over the loop
variables:

```python
for ax in plotter_axes(instance):
    containers = _new_bar_containers(ax, seen)
    for container, name in zip(containers, deferred_names(
        lambda: _group_names(ax, containers), len(containers)   # late
    )):
```

By the time it ran, `ax` and `containers` held the **last** panel's values.

```
displot(hue="g", col="c")  over a grid whose panels hold different groups

before:  [(None, [0,10,0]), (None, [10,0,0]), (None, [0,0,10])]
after:   [('b',  [0,10,0]), ('a',  [10,0,0]), (None, [0,0,10])]
asked panel by panel:  panel 0 -> ['b', 'a'],  panel 1 -> [None]
```

Every layer came back `None` because the last panel held a single container,
and a lone artist is exactly what `names_for` declines. With a differently
shaped grid it would have resolved to another panel's names instead — right
count, wrong distributions. Only the faceted path was affected; `histplot`
draws one panel.

## Verification

- **All twelve** element × fill × orientation combinations now emit a schema
  identical to `histplot`'s — orientation, bin bounds and counts alike. That is
  the parametrised test, and it cannot be satisfied by inventing numbers
  because `histplot`'s reading is pinned already.
- The ambiguous shape #585 fixed — a `y=` two-bin `poly` whose counts climb —
  reads `horz` through this path too, with the bins on `y` and the counts on
  `x`.
- `hue=` names its groups through both branches: the filled outline from its
  face colour, the unfilled one from the line's.
- Faceted panel names checked against each layer's **counts**, so the names are
  attached to the right distributions rather than merely present.
- Seven mutations applied one at a time against a restored baseline, **six
  killed**: restoring the late binding, never registering outlines, dropping
  the filled branch, dropping the unfilled branch, forcing the orientation
  vertical, and binding a name to the wrong panel. Two of those (orientation,
  wrong panel) survived a first pass and were what showed the test set had gaps
  — the `y=`-poly and filled-hue cases are there because of it.
- The survivor is the per-axes ownership snapshot, which mirrors the one the
  bar branch already keeps. `displot` builds its own grid, so a panel reaching
  here holds nothing and removing it changes no measurable behaviour today; it
  is kept so one ownership rule covers every artist a panel can arrive with,
  and the stale comment claiming those panels "do not exist yet" is corrected
  (measured: one panel for a plain `displot`, two for `col=`).
- Full suite: **2857 passed, 18 skipped**. `ruff check` clean.

---

### DIFF
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -677,47 +677,13 @@ def sns_hist(wrapped, instance, args, kwargs) -> Axes:
         # context -- which is what made the same figure readable through one
         # spelling and silent through the other (#522).
         axes = FigureManager.get_axes(drawn)
-        drew = _drew_outlines(axes, outlines)
-        if drew:
-            # `element="step"` / `"poly"`: the same distribution drawn as one
-            # closed outline per series instead of a row of bars.
-            #
-            # Named from the legend by colour, the way the bars and the
-            # unfilled outlines are. Deferred for the reason `_curve_names`
-            # gives: a `pairplot`'s legend does not exist until every panel
-            # has been drawn (#561).
-            names = deferred_names(
-                lambda: _names_for(axes, [_face_colour(one) for one in drew]),
-                len(drew),
-            )
-            for outline, name in zip(drew, names):
-                FigureManager.create_maidr(
-                    axes, PlotType.HIST, collection=outline, **{GROUP_NAME: name}
-                )
-            return drawn
-        horizontal = _asked_for_horizontal(kwargs)
-        outlined = _drew_outline_lines(
-            axes, lines, bool(kwargs.get("kde", False)), horizontal
-        )
-        if outlined:
-            # The same two elements drawn `fill=False`: seaborn swaps the
-            # collection for a bare `Line2D`, so the branch above finds
-            # nothing and the chart was silent (#583). Named from the legend
-            # by colour, as the filled spelling and the density already are.
-            names = deferred_names(
-                lambda: _names_for(axes, [_rgba(line.get_color()) for line in outlined]),
-                len(outlined),
-            )
-            for line, name in zip(outlined, names):
-                FigureManager.create_maidr(
-                    axes,
-                    PlotType.HIST,
-                    **{
-                        OUTLINE_LINE: line,
-                        OUTLINE_HORIZONTAL: horizontal,
-                        GROUP_NAME: name,
-                    },
-                )
+        if _register_outlines(
+            axes,
+            outlines,
+            lines,
+            bool(kwargs.get("kde", False)),
+            _asked_for_horizontal(kwargs),
+        ):
             return drawn
         if _drew_mesh(axes, meshes):
             # Named from `stat` rather than hardcoded, because it is what the
@@ -788,8 +754,8 @@ def sns_hist(wrapped, instance, args, kwargs) -> Axes:
 
 
 def _register_outlines(
-    ax: Axes, outlines: list, lines: list, kde: bool, horizontal: bool
-) -> None:
+    ax: Axes | None, outlines: list, lines: list, kde: bool, horizontal: bool
+) -> bool:
     """
     Register the panel's distribution when it was drawn as an outline.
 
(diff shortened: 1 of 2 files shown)


jooyoungseo:
All four taken, in `1ad03fc`.

**Duplication.** You're right, and it was the one worth doing now rather than soon after: that rule has already needed three edge cases (`#583`'s uneven poly, `#585`'s orientation tie, `#587`'s face colour), so a second copy would have diverged on the fourth. `_register_outlines` now returns whether it registered anything — which is exactly what the axes-level path needs before it goes looking for a mesh — and `sns_hist` calls it in place of its own copy. Net effect on the diff is that `sns_hist` gets *shorter* than it was on `main`.

**Default-arg capture.** Correct — those locals belong to a single call of `_register_outlines`, so they cannot be late the way #591's were. Worse than unnecessary: written as a guard they read as one, against a bug that cannot occur there. Removed.

**Formatting.** The call site is broken one-argument-per-line and both `names = deferred_names(...)` are pulled out above their `for`, matching the rest of the module.

**The untested snapshot.** Noted in the code rather than only in the PR description, including that a mutation emptying it passes the suite — so the status is discoverable from the file.

The mutation sweep was re-run against the refactored code: two of the seven no longer applied (the call site moved, the lambdas lost their default args), so they were re-aimed at the new text. Six killed again, F unchanged. Full suite still 2857 passed, 18 skipped.

One thing worth flagging in the other direction: because both interfaces now share the helper, mutations C and D — dropping the filled and unfilled branches — are killed through *both* paths rather than one, which is a small tightening the duplication was costing.

---

jooyoungseo:
Good instinct on `kde` — measured rather than assumed, and pinned in `a90a81d`.

It does arrive as a keyword on the inner call, for every spelling:

```
displot(element="poly", fill=False, kde=True)   plotter kwargs kde=True   ->  nothing
displot(element="step", fill=False, kde=True)   plotter kwargs kde=True   ->  hist
displot(element="poly", fill=False)             plotter kwargs kde=False  ->  hist
histplot(element="poly", fill=False, kde=True)                            ->  nothing
```

So the decline is reachable through `displot` and agrees with `histplot`'s, which is now asserted both ways. And you were right that it was worth checking rather than assuming: a mutation forcing `kde = False` **passed** the suite before these cases and is killed by them. That is the same gap the orientation and wrong-panel mutations had — the third time on this PR that the mutation pass found the tests thin rather than the code wrong, which is what it is for.

Worth being precise about what a silently-`False` `kde` would have cost, since it is a little worse than an unreachable decline: `element="poly", fill=False, kde=True` would then have *registered* — announcing the density curve as a distribution, with `gridsize=` points read as bins. Not an unreachable path, a wrong reading.

Full suite now 2860 passed, 18 skipped. Mutation tally 7/8, with the ownership snapshot the known and documented survivor.

---

Issue #558, opened 2026-08-21
A hue-grouped `histplot` announces one distribution and silently drops the rest


jooyoungseo:
`sns.histplot(hue=...)` draws one `BarContainer` per group. py-maidr registers **one layer**, reads the **first** container, and every other group vanishes from the announcement while staying on screen.

## Measured

seaborn 0.13.2, 60 observations split across two groups, `bins=5`:

```python
sns.histplot(data=df, x="a", hue="g", bins=5)
```

```
containers: 2 [5, 5]
  container 0 heights: [0, 4, 10, 9, 5]     <- group x
  container 1 heights: [1, 11, 8, 8, 4]     <- group y

  layer 0 hist y=[0.0, 4.0, 10.0, 9.0, 5.0]  <- only x is announced

  true x: [0, 4, 10, 9, 5]
  true y: [1, 11, 8, 8, 4]
```

Group `y`'s distribution is drawn and never spoken. Nothing raises; the chart reads as a complete single-distribution histogram, which is the failure mode that is worse than a crash.

It does not depend on how the groups are laid out, and it gets worse with more of them:

| call | containers drawn | layers registered |
| --- | --- | --- |
| `histplot(hue=g)` (default `multiple="layer"`) | 2 | **1** |
| `histplot(hue=g, multiple="stack")` | 2 | **1** |
| `histplot(hue=g, multiple="dodge")` | 2 | **1** |
| `histplot(hue=g, multiple="fill")` | 2 | **1** |
| `histplot(hue=g)` with three groups | 3 | **1** |
| `histplot(x=a)` — no hue | 1 | 1 |

## Where

Two sites, both taking the **first** new container:

- `maidr/patch/histogram.py:481` — `sns_hist`, for `sns.histplot(...)`
- `maidr/patch/histogram.py:551` — the `_DistributionPlotter.plot_univariate_histogram` wrapper, which is what `sns.displot(...)` and `catplot`-style figure-level calls reach

Both call `_new_bar_container(ax, before)`, which by construction answers with one container.

`Axes.hist` does **not** have this bug: `mpl_hist` already registers one layer per container (`_drawn_containers`), which is what #553 and #555 settled for a multi-dataset `ax.hist([a, b])`. So the same two distributions are read through matplotlib and dropped through seaborn.

## A comment that measurement falsifies

`_new_bar_container`'s docstring says the choice of container is unobservable:

[...]

PR #594, opened 2026-08-22
fix(box): announce every box a hue-grouped box plot draws

### LINKED ISSUE #593: A hue-grouped boxplot announces half its boxes and drops the rest
## What happens

`sns.boxplot(..., hue=...)` draws one box per (category × hue level) and
announces **only the first hue level's**. The other level's boxes are on
screen and absent from the schema entirely — no warning, no error, and a
chart that reads as a complete single-grouped box plot.

Three categories, two hue levels, six boxes drawn:

```python
df = pd.DataFrame({"cat": list("abc")*20, "grp": ["p","q"]*30, "val": rng.normal(size=60)})
sns.boxplot(df, x="cat", y="val", hue="grp")
```

```
medians actually drawn   grp=p: [-0.54,   0.068, 0.2687]
                         grp=q: [-0.1045, 0.2446, 0.3534]

emitted                  n=3   z=['a','b','c']   q2=[0.2687, -0.54, 0.068]
```

The three emitted medians are exactly **group p's**. Group q is drawn and
never spoken — the #558 shape, one level over.

The siblings all get it right, which is what makes this legible:

| call | boxes drawn | emitted | `z` |
|---|---|---|---|
| `boxplot(hue=)` | 6 | **3** | `['a','b','c']` |
| `boxenplot(hue=)` | 6 | 6 | `['a, p', 'a, q', 'b, p', …]` |
| `violinplot(hue=)` | 6 | 6 | `['a_p', 'a_q', 'b_p', …]` |
| `catplot(kind="box", hue=)` | 6 | 3 + 3 | both layers `['a','b','c']` |

`dodge=False` loses the same three. `catplot` is a second, milder defect: it
registers a layer per `bxp` call so all six survive, but the two layers carry
the same `z` values and no name, so nothing says which group either is.

There is a length mismatch alongside it: the same chart emits **6 selectors
against 3 data rows**, so the highlight indices run past the end of the data.

## Why

`BoxPlot._extract_bxp_maidr` ends with:

```python
for whisker, cap, median, outlier, level in zip(
    whiskers, caps, medians, outliers, levels
):
```

`levels` is `extract_level(self.ax, MaidrKey.X)` — the category **tick
labels**, three of them. The other four lists hold one entry per box, six of
them. `zip` stops at the shortest, so three boxes are dropped on the floor and
the survivors are paired with tick labels in order.

The plumbing above it is fine, and that is worth stating because it narrows
the fix: the container accumulates both `bxp` calls correctly, and hands the
layer everything.

```
add_bxp_context: +3 boxes, container now 3
add_bxp_context: +3 boxes, container now 6
create_maidr BOX with 6 boxes, 12 whiskers, 6 medians
```

Six boxes arrive. Three come out.

## Where the fix belongs

`levels` has to name every box rather than every tick, which means naming each
box by the distribution it summarises: its category **and** its hue level.

The geometry is there to read, and it is the same lattice
`BoxenPlot._category_of` already reads for exactly this:

```
ticks       0 -> 'a',  1 -> 'b',  2 -> 'c'
box centres -0.2667, 0.7333, 1.7333   (offset -0.267)
             0.1333, 1.1333, 2.1333   (offset +0.133)
legend      ['p', 'q']
```

Nearest tick gives the category, rank among the dodge offsets gives the level.
`boxenplot` names the result `"a, p"`; `violinplot` names it `"a_p"` — worth
settling on one while fixing this rather than adding a third.

Whatever `z` ends up saying, the count is the defect: six boxes drawn must be
six boxes announced, and the selector list must be the same length as the
data.

jooyoungseo:
Closes #593.

## The defect

`sns.boxplot(..., hue=...)` draws one box per category **per level** and
announced only the first level's. Not a wrong reading of six boxes — three of
them absent from the schema entirely, no warning, no error, on a chart that
reads as a complete single-grouped box plot.

Three categories, two levels, six boxes drawn:

```
medians actually drawn   grp=p: [-0.54,   0.068, 0.2687]
                         grp=q: [-0.1045, 0.2446, 0.3534]

emitted                  n=3   z=['a','b','c']   q2=[0.2687, -0.54, 0.068]
```

Exactly group p's. Group q is on screen and never spoken — the #558 shape, one
level over. `dodge=False` loses the same three.

There is a length mismatch alongside it: the selector list is built from the
artists rather than from the data, so the same chart emitted **6 selectors
against 3 rows** — every highlight past the third addressing a box nothing
announces.

## Why

One `zip`:

```python
for whisker, cap, median, outlier, level in zip(
    whiskers, caps, medians, outliers, levels
):
```

`levels` is `extract_level(...)` — the axis's **tick labels**, one per
category. The other four hold one entry per box. `zip` ends at the shortest.

The plumbing above it was never at fault, and that is worth saying because it
narrows the fix to one function:

```
add_bxp_context: +3 boxes, container now 3
add_bxp_context: +3 boxes, container now 6
create_maidr BOX with 6 boxes, 12 whiskers, 6 medians
```

Six boxes arrive. Three came out.

## The reading

`_named_boxes` answers per box, from the drawing:

- **Which category** — the nearest tick to the box's own position. Safe by
  construction rather than by luck: seaborn dodges a category's levels inside
  a slot narrower than the unit its ticks are spaced by, so a box never lands
  nearer another category's tick. Measured against a half-spacing of 0.5, the
  outermost box of a two-level split sits 0.267 from its own tick.
- **Which level** — the colour it was drawn in, matched against the legend
  swatch that names it. That is the match `patch/histogram` and
  `patch/stripplot` already make, and it works here where the dodge lattice
  `BoxenPlot._category_of` reads would be overkill: a box carries its level on
  its face, and a boxen *ladder* does not.

It returns exactly one name per box, so the `zip` cannot truncate again.

Engaged only where there are **more boxes than ticks**, so every chart that
was already right keeps the reading it had — `ax.boxplot`, the ungrouped
seaborn box, and any chart with custom `positions=` all take the path they
took before.

## Results

| call | before | after |
|---|---|---|
| `boxplot(hue=)` | 3 rows, `['a','b','c']` | 6 rows, `['a, p','b, p','c, p','a, q','b, q','c, q']` |
| `boxplot(hue=, dodge=False)` | 3 rows | 6 rows |
| `boxplot(y=cat, hue=)` | 3 rows | 6 rows, categories read off y |
| `boxplot(hue=, legend=False)` | 3 rows | 6 rows, category-only names |
| `boxplot()` | 3 rows, `['a','b','c']` | unchanged |
| `ax.boxplot()` | 3 rows | unchanged |

`z` reads `"a, p"`, matching what `boxenplot` already emits. (`violinplot`
spells the same thing `"a_p"` — an inconsistency worth settling separately
rather than adding a third spelling here.)

## Verification

- Each announced median checked against `np.median` over that
  (category, level)'s own rows — so the names are attached to the right
  distributions rather than merely present, in both orientations.
- Selector count now equals data count.
- Six mutations applied one at a time against a restored baseline, **all
  killed**: restoring the truncating zip, never naming the levels, reading the
  category off the wrong axis, taking the first tick instead of the nearest,
  reading the box centre from the other column, and preferring the edge colour
  to the face. The wrong-axis one survived a first pass — the horizontal test
  checked the values but not the names, which is now closed.
- `tests/core/plot/test_grouped_boxplot.py` — 10 tests.
- Full suite: **2870 passed, 18 skipped**. `ruff check` clean.

## Still open

`catplot(kind="box", hue=)` is the milder half and is **not** fixed here. It
registers one layer per `bxp` call, so all six boxes survive — but both layers
carry `z = ['a','b','c']` and no name, so nothing says which group either is.
Its legend is figure-level and does not exist when the layers register, so the
colour match returns nothing at that moment; naming it needs the deferral
`GROUP_NAME` has and `z` does not. Worth its own issue rather than folding in.

---

### DIFF
--- a/maidr/core/plot/boxplot.py
+++ b/maidr/core/plot/boxplot.py
@@ -2,16 +2,90 @@ from __future__ import annotations
 
 import uuid
 
+import numpy as np
 from matplotlib.axes import Axes
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
+from maidr.core.plot.scatterplot import _rgba
 from maidr.exception import ExtractionError
+from maidr.util.legend_names import names_for
 from maidr.util.mixin import (
     ContainerExtractorMixin,
     DictMergerMixin,
     LevelExtractorMixin,
 )
+
+
+def _box_centre(box, horizontal: bool) -> float | None:
+    """
+    Where on the category axis a box sits.
+
+    Read off the artist rather than counted, because a hue splits a category's
+    slot into one box per level and only the position says which slot a box
+    landed in. Both spellings of a box answer: ``patch_artist=True`` -- which
+    is what seaborn draws -- gives a ``PathPatch``, and the plain
+    ``ax.boxplot`` gives a ``Line2D`` outline.
+
+    Parameters
+    ----------
+    box : Any
+        One box artist.
+    horizontal : bool
+        Whether the categories run up the y axis.
+
+    Returns
+    -------
+    float or None
+        The midpoint of the box's extent along the category axis, or ``None``
+        for an artist neither spelling reads.
+    """
+    path = getattr(box, "get_path", None)
+    if path is not None:
+        vertices = np.asarray(path().vertices, dtype=float)
+    elif hasattr(box, "get_xdata"):
+        vertices = np.column_stack(
+            [np.asarray(box.get_xdata(), dtype=float),
+             np.asarray(box.get_ydata(), dtype=float)]
+        )
+    else:
+        return None
+
+    if vertices.ndim != 2 or not len(vertices):
+        return None
+    along = vertices[:, 1 if horizontal else 0]
+    if not np.all(np.isfinite(along)):
+        return None
+    return float((along.min() + along.max()) / 2)
+
+
+def _box_colour(box):
+    """
+    The one colour a box was drawn in, if it has one.
+
+    Its face where it is filled and its edge otherwise, which is the same
+    order :func:`maidr.core.plot.scatterplot._handle_colour` asks a legend
+    handle in -- a filled box carries the hue on its face, and an unfilled
+    one has only its outline to carry it.
+
+    Parameters
+    ----------
+    box : Any
+        One box artist.
+
+    Returns
+    -------
+    tuple of float or None
+        The rounded RGBA, or ``None`` when the artist names no single colour.
+    """
+    for getter in ("get_facecolor", "get_edgecolor", "get_color"):
+        read = getattr(box, getter, None)
+        if read is None:
+            continue
+        colour = _rgba(read())
+        if colour is not None:
+            return colour
+    return None
 
 
 class BoxPlotContainer(DictMergerMixin):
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Both taken in `2e58ad0`, and you were right that the `Line2D` branch is unreachable — measured rather than assumed:

```
Line2D box at position 5.0
  get_path().vertices x  [4.925, 5.075, 5.075, 4.925, 4.925]
  get_xdata()            [4.925, 5.075, 5.075, 4.925, 4.925]
```

Same data-space vertices, so the first branch had already handled it. Removed — an unreachable defensive branch is one nothing can ever fail against, so it earns its keep only as a comment, not as code. What *is* worth writing down is the thing you checked and I had only assumed: `Axes.bxp` builds the outline from explicit vertices because a notched box is not a rectangle, so `get_path()` is never a placeholder unit square. That reasoning is now in the docstring.

Line wrapped too. The two lines still over 88 in that file are the pre-existing `nth-child(... of use:not(...))` selector strings, which I have left alone.

---

PR #596, opened 2026-08-22
fix(box): name each catplot box layer for the level it holds

### LINKED ISSUE #595: catplot(kind="box", hue=) gives two layers nothing tells apart
The milder half of #593, left out of #594 deliberately.

## What happens

`sns.catplot(..., kind="box", hue=...)` announces **all** its boxes — it
registers one layer per `bxp` call, so nothing is lost — but the two layers
are indistinguishable:

```
catplot(x="cat", y="val", hue="grp", kind="box")

  layer 0   n=3   name=None   z=['a','b','c']   q2=[0.2687, -0.54,  0.068 ]
  layer 1   n=3   name=None   z=['a','b','c']   q2=[-0.1045, 0.2446, 0.3534]
```

Both carry the same three `z` values and neither carries a `name`, so a reader
moving between them hears the same three categories twice with nothing saying
that one is group `p` and the other group `q`.

The axes-level spelling of the same chart now reads `['a, p', 'b, p', 'c, p',
'a, q', 'b, q', 'c, q']` (#594), so the two interfaces disagree again — the
#522 / #446 shape.

## Why

Each `catplot` layer holds one `bxp` call's boxes, which is one level's worth,
so `len(boxes) == len(ticks)` and `_named_boxes` — the per-box naming #594
added — is not engaged. It would not help if it were: the colour match needs a
legend, and `catplot`'s is **figure-level and built after** the panels are
drawn, so at registration there is no legend on the axes and none on the
figure either.

That is the same timing `deferred_names` solves for `GROUP_NAME` (#561). The
difference is where the name has to land: `z` is a **per-point field inside
`data`**, not a layer attribute, so it cannot hold a callable resolved at
render the way a layer name can.

## What a fix needs

Either

- give the layer a `name` from the deferred legend, leaving `z` as the
  category — which tells the two apart without touching the data payload, and
  is what the scatter and histogram layers already do; or
- defer the `z` values themselves, which means `_extract_plot_data` gaining a
  resolve-at-render step it does not have today.

The first is much the smaller change and probably the right one: `z` says what
the split is *by*, and `name` says which side of it a layer is — the division
`ScatterPlot.render` already documents.

Not a data loss, unlike #593. Every box is announced; only the grouping is
unsayable.

jooyoungseo:
Closes #595 — the milder half of #593, scoped out of #594 and now done.

## The defect

`sns.catplot(..., kind="box", hue=...)` announces **all** its boxes; it
registers one layer per `bxp` call, so nothing is lost the way the axes-level
spelling lost three. But the two layers were indistinguishable:

```
layer 0   n=3   name=None   z=['a','b','c']   q2=[ 0.2687, -0.54,   0.068 ]
layer 1   n=3   name=None   z=['a','b','c']   q2=[-0.1045,  0.2446, 0.3534]
```

Same three categories twice, no name on either, so nothing said one was group
`p` and the other `q`. Since #594 the axes-level spelling reads
`['a, p', 'b, p', 'c, p', 'a, q', …]`, so the two interfaces disagreed — the
#522 / #446 shape.

## Why the name had to be deferred

The colour is right there on the boxes, but the legend is not: `catplot` builds
one at the **figure**, after every panel is drawn. At registration there is
none on the axes and none on the figure either, so a match made then finds
nothing. That is exactly the timing `deferred_names` already solves for
`GROUP_NAME` on a `pairplot` (#561), and `BoxPlot` now honours the same key —
string or callable — resolving it in `render()`.

## `name_for`

`names_for` declines a lone artist: *"a lone artist needs nothing to be told
apart from."* That is right where the artists of one layer are named against
each other, and wrong where a **layer** is the artist. A `bxp` call draws one
level's boxes across every category, so the layer holds one colour and that
colour is precisely what says which level it is.

`name_for(ax, colour)` is that one-colour half, in the same module and making
the same two passes in the same order — RGBA first, three-channel after it for
an artist drawn at a different opacity from its swatch. `_match_swatches`
still declines a colour two swatches claim.

## Where the name goes, and where it does not

`z` keeps the category; `name` says which side of the split the layer is — the
division `ScatterPlot.render` documents.

The axes-level `sns.boxplot` is deliberately untouched. Its boxes reach one
layer together, so the level belongs **per box** (`"a, p"` on each `z`, from
#594); naming that layer as well would assert the whole of it was one level.
A test pins that.

| call | before | after |
|---|---|---|
| `catplot(kind="box", hue=)` | two layers, no name | `name='p'` / `name='q'` |
| `catplot(kind="box", hue=, col=)` | no name | named per panel |
| `catplot(kind="box")` | no name | unchanged |
| `ax.boxplot()` | no name | unchanged |
| `sns.boxplot(hue=)` | levels on `z` | unchanged |

## Verification

- Layer names checked against each layer's **medians**, not registration
  order, so a name attached to the wrong layer fails rather than passes.
- The faceted grid is covered separately, since it has more panels for a
  resolver reading the wrong one to go astray in.
- Five mutations applied one at a time against a restored baseline, **four
  killed**: handing over no name, resolving it eagerly instead of at render,
  dropping it in the layer, and inverting the one-colour lookup's guard.
- The survivor is the guard against naming a call whose boxes do not share a
  colour. Nothing measured produces one — `bxp` colours a call's boxes
  together, and the two callers reaching this code (`ax.boxplot`, `catplot`)
  each draw one colour per call. It stays because the answer to "which level
  is this layer" for a layer holding two is *neither*, and an arbitrary one is
  worse than none; that is now written in the docstring rather than left to be
  rediscovered.
- `tests/core/plot/test_grouped_boxplot.py` — 15 tests (5 new).
- Full suite: **2875 passed, 18 skipped**. `ruff check` clean.

---

### DIFF
--- a/maidr/core/plot/boxplot.py
+++ b/maidr/core/plot/boxplot.py
@@ -7,6 +7,7 @@ from matplotlib.axes import Axes
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
+from maidr.core.plot.maidr_plot import GROUP_NAME
 from maidr.core.plot.scatterplot import _rgba
 from maidr.exception import ExtractionError
 from maidr.util.legend_names import names_for
@@ -239,6 +240,16 @@ class BoxPlot(
     def __init__(self, ax: Axes, **kwargs) -> None:
         super().__init__(ax, PlotType.BOX)
 
+        # The hue level this layer's boxes belong to, when the patch could
+        # name it. `None` is every chart with no hue and every one whose
+        # boxes reach the layer together -- there the level is on each box's
+        # `z` instead. A callable names it at render, which is what a
+        # `catplot` needs: its legend is built at the figure after every
+        # panel is drawn, so at registration there is none to read (#595).
+        group_name = kwargs.pop(GROUP_NAME, None)
+        self._group_name = (
+            group_name if isinstance(group_name, str) or callable(group_name) else None
+        )
         self._bxp_stats = kwargs.pop("bxp_stats", None)
         self._orientation = kwargs.pop("orientation", "vert")
         self._bxp_extractor = BoxPlotExtractor(orientation=self._orientation)
@@ -309,8 +320,20 @@ class BoxPlot(
         return selector if self._orientation == "vert" else list(reversed(selector))
 
     def render(self) -> dict:
+        """
+        Add ``orientation`` to the base schema, and the layer's group name
+        where one call drew one hue level's boxes.
+
+        ``MaidrLayer.name`` is the field xability/maidr#828 added so two
+        layers of a kind can be told apart, which is exactly where a
+        ``catplot(kind="box", hue=...)`` leaves a reader: every box is
+        announced, across two layers carrying the same three categories.
+        """
         base_schema = super().render()
         box_orientation = {MaidrKey.ORIENTATION: self._orientation}
+        name = self._group_name() if callable(self._group_name) else self._group_name
+        if name:
+            box_orientation[MaidrKey.NAME] = name
         return DictMergerMixin.merge_dict(base_schema, box_orientation)
 
     def _named_boxes(self, boxes: list, labels: list[str], key: MaidrKey) -> list[str]:
--- a/maidr/core/plot/boxplot.py
+++ b/maidr/core/plot/boxplot.py
@@ -7,7 +7,7 @@ from matplotlib.axes import Axes
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
-from maidr.core.plot.maidr_plot import GROUP_NAME
+from maidr.core.plot.maidr_plot import group_name_of
 from maidr.core.plot.scatterplot import _rgba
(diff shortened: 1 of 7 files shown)


jooyoungseo:
All four taken in `4a1b730`.

**The unexercised decline.** Taken up, and it was worth it — asserted directly on `_level_of` rather than through a chart, since no chart draws such a call. The mutation that named a mixed call anyway now fails, so the sweep is 6/6 rather than 5/6 with a documented survivor. The test also asserts that one colour on the *same* axes and legend does resolve, so the decline is the mixture and not the setup.

**The duplicated line.** Factored to `group_name_of` in `maidr_plot.py`, now used by all three. Your "and presumably `SmoothPlot`" was right, and it had already drifted: `SmoothPlot` read the key with `get` where `HistPlot` and `BoxPlot` used `pop`. Harmless there — nothing reads it again and the base constructor is called without the keyword arguments — but that is exactly the shape drift takes, so the helper's docstring records it rather than quietly erasing it.

**Return annotation** added (`Callable[[], str | None] | None`), and the **stale `GROUP_NAME` comment** now names `BoxPlot` alongside the other two.

Full suite: 2876 passed, 18 skipped.

---

PR #598, opened 2026-08-22
feat(rug): read the hue a rug plot was grouped by

### LINKED ISSUE #597: A hue-grouped rugplot reads as one ungrouped strip of ticks
## What happens

`sns.rugplot(..., hue=...)` emits exactly the schema of the same call without
`hue=`: one layer, every observation in it, and nothing anywhere saying which
group a tick belongs to.

```
rugplot(x="val")            -> point  name='val'  n=12
rugplot(x="val", hue="grp") -> point  name='val'  n=12
```

Identical. The `name` is the *variable* the rug marks, not a level, and there
is no `z` — so a chart drawn to compare two distributions' raw observations
reads as one undifferentiated strip.

## Why it is readable

The grouping is right there in what seaborn drew. It draws the whole rug as
**one `LineCollection` with a colour per segment**, and builds the legend that
names those colours inside the call:

```
rugplot(x="val")            1 collection, 12 segments, colour rows=1,  unique=1, legend None
rugplot(x="val", hue="grp") 1 collection, 12 segments, colour rows=12, unique=2, legend ['p','q']

names_for(ax, per-segment colours)
  -> ['p','p','p','p','p','p','q','q','q','q','q','q']
```

Every segment is named, correctly, by machinery already in the tree.

That is the same shape `scatterplot(hue=)` has — one collection, a colour per
element, a legend naming them — which #544 splits into one layer per group,
and the same shape #586 fixed for `stripplot`/`swarmplot`. A rug is the one
member of that family still folding its groups together.

## Where the fix belongs

`maidr/patch/rugplot.py` registers one `RugPlot` per new `LineCollection`, and
`RugPlot` reads every segment of the collection it is handed. Neither knows
about groups: there is no `HUE_GROUP` equivalent, and `RUG_LABEL` carries the
variable's name rather than a level's.

The split has to happen in the patch, as it does for scatter — a layer *is*
one entry in the schema, and one collection has to become several layers.
`RugPlot` then needs to read its own subset of the segments and address them
through a per-segment selector, the way `ScatterPlot` does under a hue.

`legend=False` is the decline: the colours are still there but nothing names
them, and groups called "1" and "2" are not an improvement on one strip
(measured — `names_for` returns all `None` there).

## Two neighbours checked and found correct

Measured at the same time, and **not** defects — recorded so they are not
re-investigated:

- `pointplot(hue=)` emits only an `error_bar` layer, which looked like a
  missing estimates layer. It is not: the estimate rides on each entry's `y`,
  with the interval on `yMin`/`yMax`, and `errorbar=None` gives a `line`
  layer instead. Coherent either way.
- `clustermap` appeared to emit 3 rows for a 5×4 frame. It emits a `heat`
  layer whose `data` is a **dict** of `points`/`x`/`y`; the "3" was its key
  count. The points are the full 5×4, reordered to the clustering with `y`
  reordered to match (`['Y','X','V','W','Z']`), which is correct.

jooyoungseo:
Closes #597.

## The defect

`sns.rugplot(..., hue=...)` emitted exactly the schema of the same call
**without** `hue=`:

```
rugplot(x="val")            -> point  name='val'  n=12
rugplot(x="val", hue="grp") -> point  name='val'  n=12
```

Identical. The `name` is the *variable* the rug marks, not a level, and there
is no `z` — so a chart drawn to compare two distributions' raw observations
read as one undifferentiated strip. The grouping was not unnamed; it was
absent.

## Why it was readable all along

Seaborn draws the whole rug as **one `LineCollection` with a colour per
tick**, and builds the legend that names those colours inside the call:

```
rugplot(x="v")           colour rows=1,  unique=1, legend None
rugplot(x="v", hue="g")  colour rows=12, unique=2, legend ['p','q']

names_for(ax, per-tick colours)
  -> ['p','p','p','p','p','p','q','q','q','q','q','q']
```

That is the shape `scatterplot.hue_groups` reads point by point (#544) and
`patch/stripplot` reads for a categorical scatter (#586). A rug was the last
member of the family folding its groups together.

The split is made in the **patch**, for the reason `scatterplot.scatter`
gives: a layer *is* one entry in the schema, and this is one artist.

## What a grouped layer does differently

- Keeps to **its own ticks** — and remembers where they sat among the drawn
  ones, because the selector numbers by segment. A layer that filtered its
  positions without that would highlight the first *N* ticks of the rug
  rather than its own.
- Addresses them **per segment** (`path:nth-of-type(n)`). The whole-collection
  selector an ungrouped rug uses would light the entire rug up for a layer
  announcing half of it.
- Names the grouping variable on `z` and the level on `name`.

An ungrouped rug is untouched: one layer, one whole-collection selector,
named after its variable.

## The one thing the colours cannot say

Whether the hue is a grouping at all. A **numeric** `hue=` is a colour scale,
and on a small frame seaborn's legend samples every value — so every tick
matches a swatch and `names_for` names all of them. Measured on four
observations, the first draft of this split them into **four layers of one
tick each**, which is not a reading of a scale.

`_DistributionPlotter.plot_rug` is wrapped to record the plotter's
`map_type` — the same discriminator `patch/stripplot` uses, and the same
decline `hue_groups` makes one artist type over. Wrapped only to record it;
the drawing is untouched and the registration is still made by `rug()`.

## Results

| call | before | after |
|---|---|---|
| `rugplot(x=, hue=)` | one layer, all ticks | `p`(6) / `q`(6), `z='grp'` |
| `rugplot(y=, hue=)` | one layer | split the same way |
| `hue_order=` reversed | — | layers in legend order, not draw order |
| `hue=` numeric | one layer | unchanged — declined |
| `legend=False` | one layer | unchanged — declined |
| `rugplot(x=)` | one layer | unchanged |

## Verification

- Each layer holds its own level's observations, checked **by value**: a
  split that got the membership backwards would still give two layers of two.
- 6 selectors resolved against the rendered SVG with `lxml.cssselect`, each
  matching exactly one element, 6 distinct — no overlap, no misses.
- Eight mutations applied one at a time against a restored baseline, **all
  killed**. Three needed tests that did not exist and now do: the
  legend-order tie, and the two guards — a colour list that does not
  correspond to the ticks, and a tick no swatch names. Both had to be
  asserted directly on `_hue_groups`, because no chart draws a rug they
  decline; the first version of those tests passed either way because the
  axes they built had no legend, which is exactly what the mutation pass
  caught.
- Two existing tests pinned the old shape. Their intent — that a hue split
  loses no observation — is kept and asked of the reading it has now, rather
  than deleted.
- `tests/core/plot/test_rugplot.py` — 26 tests (10 new).
- Full suite: **2886 passed, 18 skipped**. `ruff check` clean.

## Two neighbours measured and found correct

Checked in the same sweep and **not** defects, recorded so they are not
re-investigated:

- `pointplot(hue=)` emits only an `error_bar` layer, which looked like a
  missing estimates layer. The estimate rides on each entry's `y` with the
  interval on `yMin`/`yMax`, and `errorbar=None` gives a `line` layer.
- `clustermap` appeared to emit 3 rows for a 5×4 frame. Its `data` is a dict
  of `points`/`x`/`y` — the "3" was the key count. The points are the full
  5×4, reordered to the clustering with `y` reordered to match.

---

### DIFF
--- a/maidr/core/plot/rugplot.py
+++ b/maidr/core/plot/rugplot.py
@@ -8,6 +8,7 @@ from matplotlib.collections import LineCollection
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
+from maidr.util.legend_names import title_of
 
 #: Key the patch hands this layer the ``LineCollection`` its own call drew
 #: under. Named like ``eventplot.DRAWN_EVENTS`` and read the same way: the
@@ -225,8 +226,16 @@ class RugPlot(MaidrPlot):
         # A grouped layer names the grouping *variable* on z, the way the
         # scatter and line layers do: `z` says what the split is by, `name`
         # says which side of it this layer is.
+        #
+        # Read off whichever legend named the groups, which is not always
+        # this axes' own: the patch names them through `legend_of`, and that
+        # falls back to a lone figure-level legend for a panel carrying
+        # none. `_legend_title` asks `ax.get_legend()` only, so on that path
+        # the layers came out named `a` and `b` with no `z` -- each saying
+        # which side of a grouping it was without the chart ever saying what
+        # the grouping was.
         if self._group_members is not None:
-            z_label = self._legend_title()
+            z_label = title_of(self.ax)
             if z_label:
                 axes_data[MaidrKey.Z] = self._axis_config(label=z_label)
 
--- a/maidr/core/plot/rugplot.py
+++ b/maidr/core/plot/rugplot.py
@@ -18,6 +18,17 @@ DRAWN_RUG = "_maidr_rug"
 #: Key carrying the name to announce the layer under, when there is one.
 RUG_LABEL = "_maidr_rug_label"
 
+#: Key the patch hands one hue group over under: a ``(name, members)`` pair
+#: naming the group and listing the segments of :data:`DRAWN_RUG` that belong
+#: to it. Absent for an ungrouped rug, which is the chart this layer has
+#: always read and still reads unchanged.
+#:
+#: The split is made in the patch rather than here for the reason
+#: :func:`maidr.patch.scatterplot.scatter` gives: a layer *is* one entry in
+#: the schema, and seaborn draws a hue-grouped rug as one ``LineCollection``
+#: carrying a colour per tick, so one artist has to become several layers.
+RUG_GROUP = "_maidr_rug_group"
+
 #: What the layer's constant axis is called when the chart does not say.
 #: A rug's ticks sit in a strip against the frame rather than at a measured
 #: height, so the coordinate across the tick names the strip, not a value.
@@ -126,6 +137,13 @@ class RugPlot(MaidrPlot):
         )
         self._label = kwargs.get(RUG_LABEL, None)
 
+        # The hue group this layer reads, when the patch found one. `None`
+        # means the layer reads every tick the collection holds, which is
+        # what an ungrouped rug has always done.
+        group = kwargs.get(RUG_GROUP, None)
+        self._group_name = group[0] if group else None
+        self._group_members = list(group[1]) if group else None
+
         # Read once, here, rather than again per render. Two reasons, and
         # the first is correctness: `MaidrPlot.render()` builds the *axes*
         # payload before the data, so an orientation settled in
@@ -137,6 +155,18 @@ class RugPlot(MaidrPlot):
         # revalidating every segment each time buys nothing.
         read = read_rug(self._collection)
         self._positions, self._along_x = read if read is not None else ([], True)
+
+        # Kept to the group's own ticks, and their places among the drawn
+        # ones kept with them: the selector numbers by segment, so a layer
+        # that filtered its positions without remembering where they came
+        # from would highlight the first N ticks of the rug rather than its
+        # own -- the failure nothing announced can see (xability/maidr#814).
+        self._drawn = list(range(len(self._positions)))
+        if self._group_members is not None:
+            members = set(self._group_members)
+            kept = [index for index in self._drawn if index in members]
+            self._positions = [self._positions[index] for index in kept]
+            self._drawn = kept
 
         # Assigned here rather than relied upon, for the reason `EventPlot`
         # and `HexbinPlot` give: matplotlib stamps a gid at *draw* time and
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Thanks — the "worth a second look" item was a real defect, and it is fixed in `860ddd3`. Taking the four points in turn.

## The `z` label and the figure-level legend — fixed

Reproduced exactly as described. A rug drawn `legend=False` beside one `fig.legend(title="g")` naming the same palette:

```
ax.get_legend()  None
fig.legends      1

before:  name='a' z=None
         name='b' z=None
```

Two layers each saying *which side* of a grouping they are on, and nothing anywhere saying *what* the grouping is. Worse than the unsplit reading it replaced, in the one respect that matters.

The cause is the one you named: grouping detection goes through `legend_of`, which falls back to a lone figure-level legend, and `_legend_title` asks `ax.get_legend()` alone. Two halves of one decision reading two different sources.

Fixed with `title_of(ax)` in `maidr/util/legend_names.py` — the title off whichever legend `legend_of` chose — and `RugPlot._extract_axes_data` now calls it. Scoped to rug, as you suggested: `PointPlot`/`MultiLinePlot` reach the fallback through their own paths and changing `_legend_title` itself would move all three at once.

Pinned by `test_a_split_named_by_a_figure_legend_still_says_what_it_split_by`, which fails on the old sourcing (`1 failed, 26 passed`) and passes now.

## Formatting — done, though CI would not have caught it

You were right about the code and right to flag it; the mechanism is slightly different from what you expected. The step named "Check format with ruff" runs `ruff check --diff`, not `ruff format --check`:

```yaml
- name: Check format with ruff
  uses: astral-sh/ruff-action@v3
  with:
    args: "check --diff"
```

Format is not enforced anywhere — `ruff format --check` reports **110 files would be reformatted** across the repo — which is why CI was green. `ruff format` applied to the two files this PR touches all the same; it produced exactly your suggested `_named`, plus collapsing one comprehension in `_drawn_by_this_call` and removing the stray blank line you spotted.

## `LineCollection` imports — hoisted

All three, including the one that predates this PR at what was line 194, so the file has one convention rather than two.

## The shared "hue groups from a legend" helper

Agreed, and deliberately not done here. Scatter, strip and rug reach the same conclusion by three routes because each has a different thing to read it off — the plotter's `lookup_table` for strip, per-point face colours for scatter, per-segment colours here — and the shared part is only the tail: colours → names → grouped indices, decline on any ambiguity. Worth extracting, but it is a refactor across three call sites with its own falsification to do, not a rider on a fix. Not filed yet; I will open it separately so the decline rules and the #502 ordering have one place to live.

## Verification

- Full suite: **2887 passed, 18 skipped**.
- `ruff check` and `ruff format --check` clean on every file this PR touches.
- The new test falsified by reverting `title_of(self.ax)` to `self._legend_title()`: fails, as it should.

---

Issue #600, opened 2026-08-22
A hand-drawn confidence band is dropped, while seaborn's identical one is announced


jooyoungseo:
## The chart

The most common way to draw a confidence interval in bare matplotlib — a line for the estimate, a translucent fill for the interval:

```python
ax.plot(x, mean)
ax.fill_between(x, lo, hi, alpha=0.3)
```

Measured:

```
line   n=1  point 0: {'x': 0.0, 'y': -0.92}
```

One layer, the estimate, nothing else. **The band is dropped in silence** — no layer, no warning, no `yMin`/`yMax`. A reader is told the mean and never told the uncertainty around it, which is the one thing that chart was drawn to show.

## Why this is now a defect rather than a limitation

`maidr/patch/fillbetween.py` declines the two-curve form deliberately, and says why:

> The two-curve form draws an interval, whose content is the gap rather than either edge, and MAIDR has no reading for it that would not invent an estimate the chart never drew — so it is left unregistered (#339).

That reasoning was right when it was written. It is no longer true of this codebase. The reading exists, and two other layers already use it:

```
sns.regplot(df, x="x", y="y")
  smooth  {'x': 0.0, 'y': 1.14, 'yMin': 0.29, 'yMax': 1.80}

sns.lineplot(long, x="x", y="y")     # errorbar band
  line    {'x': 0.0, 'y': -0.92, 'yMin': -2.28, 'yMax': 0.16}
```

`yMin`/`yMax` on the estimate's own points. Exactly the shape r-maidr settled `geom_smooth(se=TRUE)` on.

And the objection does not apply to *this* call. Nothing has to be invented: the `plot` supplies `y`, the `fill_between` supplies `yMin` and `yMax`. All three numbers were drawn. The chart matplotlib draws here is the same chart seaborn draws — the user just assembled it from two calls instead of one.

So the same figure is announced two different ways depending on which library composed it, which is the divergence #556 and #583 were both about.

## What should and should not change

**Should**: a two-curve `fill_between` over the same positions as a line on the same axes attaches to it as `yMin`/`yMax`, the way `lineplot`'s own band does.

**Should not**: the standalone case. `ax.fill_between(x, lo, hi)` with no line has no estimate anywhere — only the interval — and the original argument holds unchanged. It stays declined.

That split is the whole of the fix, and it is what makes it safe: the band is read only where the thing it is a band *around* was also drawn.

## Questions to settle by measurement, not assumption

[...]

Issue #562, opened 2026-08-21
A line with a confidence band announces the line and drops the band — `sns.lineplot` and `plot` + `fill_between`


jooyoungseo:
Found by sweeping every `Axes` plotting method and every axes-level seaborn function for what py-maidr reads. Coverage is good; this is the gap that costs a reader something a sighted one gets for free.

## Measured

seaborn 0.13.2, four repeats per x so seaborn draws its default 95% CI band:

| chart | layers | what each point carries |
| --- | --- | --- |
| `sns.regplot(x, y)` | `point`, `smooth` | `x`, `y`, **`yMin`, `yMax`** |
| `sns.lineplot(data, x, y)` — repeated x | `line` | `x`, `y` |
| `ax.plot(x, y)` + `ax.fill_between(x, lo, hi)` | `line` | `x`, `y` |

The band is drawn in all three. `regplot`'s is announced; the other two are not.

Confirming the band really is there for `lineplot`:

```
lines: 1  collections: ['PolyCollection']
   band vertices: 43   first 3: [[0.0, 1.4996], [0.0, 1.2071], [0.5263, 1.7293]]
line points: 20
```

One 20-point line, one 43-vertex band around it, and a reader hears only the line — with no indication that the chart shows any uncertainty at all.

## Why this one matters

`sns.lineplot` aggregates and draws a confidence band **by default** whenever the data has repeated x values. It is the seaborn idiom for "a trend with its uncertainty", and the band is usually the point of drawing it that way. A reader told only the mean has been given the half of the chart that is least in doubt.

`ax.plot(...)` + `ax.fill_between(x, lo, hi)` is the same chart written by hand, and it is what matplotlib's own documentation shows for plotting with error bands.

## The shape already exists, and is already used

`SmoothPoint.yMin` / `yMax` — one estimate with the interval drawn around it. `SmoothPlot._confidence_band_at` already recovers exactly this from a `PolyCollection` for `regplot` (#451), and validates its own reading rather than trusting position: a shaded region is this fit's band only if it **brackets every fitted sample**, which is the property a confidence band has and an unrelated `fill_between` on the same axes does not.

So the work is to give `LinePlot` the same treatment, not to invent a mechanism.

r-maidr reached this first: `geom_smooth(se = TRUE)` used to announce the fitted line and drop its ribbon, and #135's remainder moved the band onto the points as `yMin`/`yMax`. The Python binding still has the gap for the line spelling.

## What is *not* proposed

[...]

PR #603, opened 2026-08-22
fix(line): decline a band shaded across the page rather than up it

### LINKED ISSUE #601: A horizontal confidence band is announced as a vertical interval the chart never drew
## The chart

A horizontal line chart with its interval — position up `y`, the measured value along `x`, and the band shaded along `x` to match:

```python
pos = np.array([0., 1., 2., 3., 4.])
val = np.array([1., 2., 3., 2., 4.])

ax.plot(val, pos)                                  # the estimate
ax.fill_betweenx(pos, val - 0.5, val + 0.5, alpha=.3)   # its interval
```

Measured:

```
line
  {'x': 1.0, 'y': 0.0, 'yMin': 0.0, 'yMax': 1.5}
  {'x': 2.0, 'y': 1.0, 'yMin': 0.5, 'yMax': 3.0}
  {'x': 3.0, 'y': 2.0, 'yMin': 1.5, 'yMax': 3.5}
  {'x': 2.0, 'y': 3.0, 'yMin': 0.5, 'yMax': 3.0}
  {'x': 4.0, 'y': 4.0, 'yMin': 3.0, 'yMax': 4.0}

true interval, on x:  (0.5,1.5) (1.5,2.5) (2.5,3.5) (1.5,2.5) (3.5,4.5)
```

Every one of those bounds is wrong, and wrong in three separate ways at once.

**Wrong axis.** The interval this chart draws runs along `x` — it is uncertainty about the *value*. It is emitted as `yMin`/`yMax`, a claim about `y`, and `y` here is the position. The chart states no uncertainty about position at all; positions are exact by construction.

**Wrong numbers.** They are not the interval transposed either. They are the polygon's *vertical extent* at each x — an artefact of the band being sloped, not a measurement of anything. Point 5 is announced as 3.0–4.0 against a true 3.5–4.5: the announced interval does not even contain the true one, and its `yMax` equals `y` itself.

**Confidently.** Nothing warns. A reader is told an estimate and a precision, and the precision is invented.

Silence would be better than this, which is the standard every decline in `maidr/util/confidence_band.py` and `maidr/patch/fillbetween.py` is already held to.

## Why it happens

`band_edges_at` collects, for each distinct x in the region's vertices, the lowest and highest y — then checks the line's samples are bracketed. Its module docstring says this "assumes nothing about orientation."

It does. Reading the extent *at each x* is exactly an assumption that the interval runs vertically. For a `fill_betweenx` region it holds no meaning, and the bracketing test does not catch it: a horizontal band around a horizontal line brackets that line vertically too, because the band surrounds it.

The docstring anticipates the artist but draws the wrong conclusion from it:

> seaborn draws a violin body with `fill_betweenx`, so a violin is the same class as a band

Right — and the bracketing test does reject a violin, because a violin body does not surround the line of some other series. It rejects it by luck of geometry rather than by knowing what the region is, and a transposed band is the case where that luck runs out.

## The fix

The patch already knows the answer with certainty and does not pass it on. `maidr/patch/fillbetween.py` wraps `fill_between` and `fill_betweenx` separately — `fill_betweenx` **is** the horizontal spelling, no inference required. Tag the collection as it is drawn and have `band_edges_at` skip a region tagged horizontal.

Handed over rather than searched for, which is the convention `OUTLINE_HORIZONTAL` in `maidr/core/plot/outlined_histogram.py` already sets for the same class of question, and for the same reason: the drawing cannot always answer it and the caller always can. Counting distinct coordinates *nearly* works — a real band has one position per column and two edge values, so the value axis carries about twice the distinct values — but measurement shows it is not safe:

```
fill_between  band     uniqX=  5  uniqY=  5      <- ties on a small frame
fill_betweenx band     uniqX=  5  uniqY=  5
regplot       band     uniqX=100  uniqY=200
lineplot      band     uniqX=  5  uniqY= 10
violin (y=)          uniqX=200  uniqY=100
```

Values colliding across positions is enough to tie it, and a tie would have to be declined — silently dropping a band that works today. The tag has no such failure mode.

## Scope

- Tag the collection in `maidr/patch/fillbetween.py`; both spellings, so the absence of a tag never has to mean anything.
- Skip a horizontally-drawn region in `band_edges_at`.
- Leave every vertical band exactly as it is. `regplot`, `lineplot`, `plot` + `fill_between`, and the violin's existing rejection must all be unchanged — pinned, not assumed.

## Deliberately not in scope

**Reading the horizontal band correctly.** `LinePoint` in the maidr grammar carries `yMin`/`yMax` and no `xMin`/`xMax`, so there is no field for an interval that runs along `x`; a horizontal band cannot be announced until the grammar has one. That is an upstream change, and one to make on evidence of demand rather than pre-emptively. Until then the band is dropped — which is what this chart got before #562 and is honest, where the current reading is not.

jooyoungseo:
Closes #601.

## The defect

A horizontal line chart with a horizontal interval — position up `y`, the measured value along `x`:

```python
ax.plot(val, pos)
ax.fill_betweenx(pos, val - 0.5, val + 0.5, alpha=.3)
```

```
before                                     true interval, on x
{'x': 1.0, 'y': 0.0, 'yMin': 0.0, 'yMax': 1.5}    (0.5, 1.5)
{'x': 2.0, 'y': 1.0, 'yMin': 0.5, 'yMax': 3.0}    (1.5, 2.5)
{'x': 3.0, 'y': 2.0, 'yMin': 1.5, 'yMax': 3.5}    (2.5, 3.5)
{'x': 2.0, 'y': 3.0, 'yMin': 0.5, 'yMax': 3.0}    (1.5, 2.5)
{'x': 4.0, 'y': 4.0, 'yMin': 3.0, 'yMax': 4.0}    (3.5, 4.5)
```

Wrong three ways at once, and confidently:

- **Wrong axis.** The interval is about the *value*, which is on `x`. It is emitted on `y` — the positions, which the chart states no uncertainty about at all.
- **Wrong numbers.** Not the interval transposed either. They are the polygon's vertical extent at each x, an artefact of the band being sloped. The last point's announced interval does not contain the true one, and its `yMax` is its own `y`.
- **No warning.** A reader is told an estimate and a precision, and the precision is invented.

## Why bracketing did not catch it

`maidr/util/confidence_band.py` identifies a band by whether it brackets the line's samples, and its docstring says reading the lowest and highest vertex at each x "assumes nothing about orientation."

It does — that *is* the assumption that the interval runs vertically. The docstring anticipates the artist and draws the wrong conclusion from it:

> seaborn draws a violin body with `fill_betweenx`, so a violin is the same class as a band

Right, and bracketing does reject a violin — because a violin body does not surround some other series' line. It rejects it by luck of geometry rather than by knowing what the region is, and a horizontal band around a horizontal line is where that luck runs out: the band surrounds the line, so of course it brackets it vertically too.

## The fix

The patch knows the answer outright and was not passing it on. `fill_betweenx` **is** the horizontal spelling — no inference. Every region either spelling draws is tagged with which way the fill runs, and `band_edges_at` skips a horizontal one.

Handed over rather than searched for, which is what `OUTLINE_HORIZONTAL` already does for the same class of question.

Tagged on **every** region, including draws made inside another patch (`stackplot`'s bands) and the ones the existing declines leave unregistered — a region carries no layer of its own and is still read as some line's band later, so the tag has to outlive the decision not to register it. That also lets an absent tag mean one thing only: this patch did not draw the region, so it reads as it always did.

### Why not count the vertices

It nearly works — a band has one position per column and two edge values, so the value axis generally carries about twice the distinct values. Measured, it ties:

```
fill_between  band   uniqX=  5  uniqY=  5     <- both spellings tie
fill_betweenx band   uniqX=  5  uniqY=  5
regplot       band   uniqX=100  uniqY=200
lineplot      band   uniqX=  5  uniqY= 10
violin (y=)          uniqX=200  uniqY=100
```

Values colliding across positions is enough, and a tie would have to be declined — silently dropping bands that read correctly today. The tag has no such failure mode.

## Deliberately not fixed here

**Reading the horizontal band correctly.** `LinePoint` carries `yMin`/`yMax` and no `xMin`/`xMax`, so the grammar has no field for an interval running along `x`. Until it does, the band is dropped — which is what this chart got before #562, and honest where the current reading is not. An upstream change to make on evidence of demand, not pre-emptively.

## Results

| chart | before | after |
|---|---|---|
| `plot` + `fill_betweenx` band | invented `yMin`/`yMax` | no band |
| `plot` + `fill_between` band | correct band | unchanged |
| `sns.regplot` | correct band | unchanged |
| `sns.lineplot` | correct band | unchanged |
| violin body nearby | correctly rejected | unchanged, now by the tag |
| `fill_between` area under its own line | correctly rejected | unchanged |

## Verification

- Nine mutations applied one at a time against a restored baseline, **all killed** — guard disabled, predicate inverted, predicate forced constant either way, default flipped either way, tag inverted, tag forced constant, internal-context draw left untagged. Each of the three narrow ones is killed by exactly one test, so each has coverage of its own rather than riding on another's.
- One mutation initially survived: `bool(...)` in place of `is True`. Investigated rather than left — the patch is the only writer and writes a bool, so the two agree on every reachable value and no test could tell them apart. Resolved by taking the simpler spelling, not by adding a test for a value nothing produces.
- `tests/core/plot/test_line_confidence_band.py` — 16 tests (5 new), including the two guards asserted directly on `shaded_along_y` because no chart reaches them: an untagged region, and a region drawn under the internal context.
- Full suite: **2892 passed, 18 skipped**. `ruff check --diff` (what CI runs) clean, exit 0.

## Related, and closed as invalid

I first filed #600 claiming the *vertical* hand-drawn band was dropped entirely. It is not — #562 reads it, by bracketing, in either call order. I had filed that off the `fillbetween.py` docstring's "left unregistered" without measuring the emitted schema; that docstring describes only its own patch. Closed with the measurement that disproves it. This PR is the part of that sweep that survived contact with the numbers.

---

### DIFF
--- a/maidr/patch/fillbetween.py
+++ b/maidr/patch/fillbetween.py
@@ -143,7 +143,7 @@ def _magnitudes(wrapped, args: tuple, kwargs: dict, position: str, value: str):
     return positions, values
 
 
-def _tagged(collection, transposed: bool):
+def _tagged(collection: Collection, transposed: bool) -> Collection:
     """
     Record which way a region was shaded, on the region itself.
 
--- a/maidr/patch/fillbetween.py
+++ b/maidr/patch/fillbetween.py
@@ -8,6 +8,7 @@ from matplotlib.collections import Collection
 from maidr.core.context_manager import ContextManager
 from maidr.core.enum import PlotType
 from maidr.patch.common import _argument, _draw_quietly
+from maidr.util.confidence_band import DRAWN_ALONG_Y
 
 
 def _baseline_is_zero(wrapped, args: tuple, kwargs: dict, name: str) -> bool:
@@ -142,6 +143,48 @@ def _magnitudes(wrapped, args: tuple, kwargs: dict, position: str, value: str):
     return positions, values
 
 
+def _tagged(collection, transposed: bool):
+    """
+    Record which way a region was shaded, on the region itself.
+
+    Asked later by :func:`maidr.util.confidence_band.band_edges_at`, which
+    reads the lowest and highest vertex **at each x** and so answers for a
+    vertical interval only. A region shaded the other way about has no such
+    interval to read, and bracketing does not catch it: a horizontal band
+    around a horizontal line surrounds that line vertically too, because it
+    surrounds it. Measured, ``plot(val, pos)`` under
+    ``fill_betweenx(pos, lo, hi)`` came out announcing the polygon's vertical
+    extent as ``yMin``/``yMax`` -- on the axis carrying the *positions*, which
+    the chart states no uncertainty about at all (#601).
+
+    Recorded here rather than worked out there because ``fill_betweenx`` *is*
+    the horizontal spelling: no inference is involved, which is the argument
+    :data:`maidr.core.plot.outlined_histogram.OUTLINE_HORIZONTAL` makes for
+    the same class of question.
+
+    Set on **every** region either spelling draws, including the ones drawn
+    inside another patch and the ones the declines below leave unregistered.
+    A region carries no layer of its own and is still read as some line's
+    band, so the tag has to outlive the decision not to register it. Setting
+    it unconditionally is also what lets its absence mean one thing only:
+    that this patch did not draw the region.
+
+    Parameters
+    ----------
+    collection : matplotlib.collections.Collection
+        The region the wrapped call drew.
+    transposed : bool
+        True for ``fill_betweenx``.
+
+    Returns
+    -------
+    matplotlib.collections.Collection
+        The same object, tagged.
+    """
+    setattr(collection, DRAWN_ALONG_Y, transposed)
+    return collection
+
+
 def _fill(
     wrapped,
     instance,
@@ -197,10 +240,10 @@ def _fill(
     # is a property of matplotlib's `stackplot` rather than a rule about
     # nested draws.
     if ContextManager.is_internal_context():
-        return _draw_quietly(wrapped, args, kwargs)
+        return _tagged(_draw_quietly(wrapped, args, kwargs), transposed)
 
     with ContextManager.set_internal_context():
-        collection = _draw_quietly(wrapped, args, kwargs)
+        collection = _tagged(_draw_quietly(wrapped, args, kwargs), transposed)
 
     if not _baseline_is_zero(wrapped, args, kwargs, other):
         return collection
--- a/maidr/util/confidence_band.py
+++ b/maidr/util/confidence_band.py
@@ -23,6 +23,25 @@ validates itself: a region is a line's band only if it **brackets every one of
 that line's samples**, which is the property a band has by construction and an
 unrelated shaded region does not.
 
+One thing it *does* assume, and has to. Reading the lowest and highest vertex
+**at each x** is a reading of a vertical interval, so a region shaded the other
+way about is not one this can answer for. Bracketing does not catch it: a
+horizontal band around a horizontal line surrounds that line vertically too,
+because it surrounds it. Measured, ``plot(val, pos)`` under
+``fill_betweenx(pos, lo, hi)`` came out carrying the polygon's vertical extent
+as ``yMin``/``yMax`` -- an artefact of the band being sloped, on the axis that
+holds the *positions*, which the chart states no uncertainty about at all
+(#601).
+
+Which way a region was shaded is therefore asked of
+:data:`maidr.patch.fillbetween.DRAWN_ALONG_Y` rather than inferred, for the
+reason :data:`maidr.core.plot.outlined_histogram.OUTLINE_HORIZONTAL` gives:
+the drawing cannot always say and the caller always can. Counting distinct
+coordinates nearly works -- a band has one position per column and two edge
+values, so the value axis generally carries twice the distinct values -- and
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Both nits applied in `02c66e7`.

**`_tagged` annotations** — right, and it was the only helper in the file without them, against its two siblings. Now `def _tagged(collection: Collection, transposed: bool) -> Collection:`.

**Local imports in the tests** — not deliberate, just carried over from how I wrote them. Hoisted to the module level with the rest, so the file has one convention.

Thanks for checking the two things the description asserted rather than taking them from it — that the tag is set before the declines and in both branches of `_fill`, and that both consumers of `edges_of` go through `band_edges_at` so there is no bypass. Your point about matplotlib's `violinplot` building its `PolyCollection` bodies directly rather than through `fill_between`/`fill_betweenx` is the one I had checked from the other side (they were already rejected by bracketing) — it is worth having both halves on the record, because it means the violin case is now belt-and-braces rather than resting on either alone.

The `SPANS_ALONG_X` precedent in `spanplot.py` I had not spotted; `OUTLINE_HORIZONTAL` was the one I reasoned from. Three instances of the same pattern is a better argument for it than the one I made.

---

PR #604, opened 2026-08-22
refactor(hue): give the shared hue-group tail one home

### LINKED ISSUE #599: Three patches reach "hue groups from a legend" three different ways
Raised in review on #598 and filed rather than folded in, because it is a refactor across three live call sites with its own verification to do.

## What is duplicated

`scatterplot.hue_groups`, `patch/stripplot._hue_levels` and `patch/rugplot._hue_groups` all answer the same question — *which observations belong to which hue level* — and all three answer it by the same route: read a colour per drawn thing, match each colour to the level that owns it, group the indices, and decline outright on any ambiguity rather than emit a partial split.

They do not share a line of code.

| | scatter (#544) | strip/swarm (#586) | rug (#597) |
|---|---|---|---|
| colour per | point (`get_facecolor` rows) | point, across N collections | segment (`get_colors` rows) |
| names from | `_named_colours` (legend swatches) | `plotter._hue_map.lookup_table` | `names_for` (legend swatches) |
| level order | legend order, by dict insertion | `hue_map.levels`, by `sorted` key | legend order, by explicit sort |
| numeric-hue decline | in `hue_groups` | `map_type` on the plotter | `map_type` via a `ContextVar` |

## Why it is worth doing

The tail is genuinely common — colours → names → grouped indices → decline — and it now carries two conventions that have to stay in step across all three:

- **#502's ordering**: layers come out in the *legend's* order, not the draw order. Three implementations, three mechanisms (insertion order, a `sorted` key, an explicit sort). A fourth artist type would have to rediscover which one is the convention.
- **The decline rules**: a colour no swatch names, a colour list that does not correspond to the drawn things, a numeric hue. Each of the three enumerates these itself, and #598's review pass found that two of rug's were only *exercised* once a test asserted them directly — no chart draws a rug that reaches them.

Drift here is silent by construction: each patch has its own tests, all three would stay green, and the divergence would only show as two charts of the same frame announcing their groups in different orders.

## Why the heads must stay separate

Not a case for one function. Each patch reads a different artist, and that half is irreducibly per-artist:

- scatter has face-colour rows on one collection;
- strip has them across N collections and a plotter to consult directly, which is the point — a `catplot` panel has no legend of its own to read;
- rug has segment colours and needs the plotter's `map_type` fetched through a `ContextVar`, because the registration happens in `rug()` rather than under the wrapped method.

So the extraction is of the **tail only**: something taking `(ax, colours, level_order)` and returning grouped indices or `None`. The three heads keep producing colours in their own way.

## Scope

- Extract the shared tail into `maidr/util/legend_names.py`, beside `names_for`/`name_for`/`legend_of`, which is already where "which legend names this axes" lives.
- Move the three declines into it, so they are stated once.
- Have all three patches call it.
- Verify by mutation on the shared helper: each decline and the ordering must be killed by tests belonging to *each* of the three call sites, not just one — otherwise the extraction has quietly made two of them untested.

No behaviour change intended. Every existing test across the three should pass untouched; if one needs editing, that is the signal the three had already drifted and the difference needs recording rather than smoothing over.

jooyoungseo:
Closes #599.

## What moved

`scatterplot.hue_groups` and `patch/rugplot._hue_groups` ended in the same three decisions and made them two different ways:

| | scatter, before | rug, before |
|---|---|---|
| decline on an unnamed thing | per-point `if name is None` | `any(name is None …)` |
| decline on fewer than two groups | `len(groups) > 1` on non-empty groups | `len(members) < 2` |
| legend order | a dict built in legend order | `sorted` against the legend's texts |

`maidr/util/hue_groups.grouped_by_name(names, order)` now takes that tail. Order is a **parameter** rather than an assumption, because it is the half most likely to drift — the two callers reached the same convention by two mechanisms, and neither said so where the other could see it.

## What did not move, and why

**What a chart is read *from* stays per artist**, because it genuinely is: a face colour per point, a colour per segment, a plotter's `lookup_table` consulted outright. Only the tail is common.

**Strip and swarm keep their own**, and this is a correction to what #599 assumed. Two differences, both by design:

- Their members are a **list per collection**, not flat positions — seaborn draws a categorical scatter as one collection per category, so a hue level spans several. Routing that through a flat helper and re-splitting adds work rather than removing it.
- A panel holding **one** level is deliberately *kept* and named — the opposite of the rule extracted here. A faceted `catplot` puts each level in whichever panels its rows fall in, and knowing which one a panel holds is the whole of what #586 was missing.

So this is an extraction across two callers, not three. Two is the threshold, and the ordering drift is the reason to reach it.

**Scatter still reads `ax.get_legend()`, not `legend_of`.** The figure-level fallback is rug's alone today; giving it to scatter would be a behaviour change and does not belong in a refactor.

## The mutation sweep found what it is for

#599 asked that each rule be killed by tests belonging to *each* call site, "otherwise the extraction has quietly made two of them untested." It had:

```
                            direct      scatter     rug
m1 unnamed decline removed   killed      killed      killed
m2 one-group decline removed killed      SURVIVED    SURVIVED
m3 ordering dropped          killed      SURVIVED    killed
```

Both survivors were real. Neither rule was tested at the scatter call site before this PR either — the extraction revealed the hole rather than opening it.

- **Legend ordering, scatter.** Added a chart where legend order and draw order disagree: the frame's first two rows are level `b`, `hue_order` puts `a` first, and the layers come out `a, b`. Before, no scatter test could tell the two orders apart.
- **Fewer than two groups, rug.** A `hue=` on a one-level column: measured, it reads as one layer named `value` — the variable, which is what an ungrouped rug announces.
- **Fewer than two groups, scatter — genuinely unreachable.** Measured: a one-level `hue=` gives two face-colour rows in one colour and one legend swatch, so scatter's own `len(named) < 2` declines it first and `grouped_by_name` is never reached. Recorded in a test that says so, rather than left as an unexplained hole or papered over with a test of something else.

After: m1, m2 and m3 are each killed at every call site where they are reachable.

One mutation is deliberately left to the direct tests: reversing the positions **within** a group. Both callers consume members as a `set` (`scatterplot.py`'s `[set(members) …]`, `rugplot.py`'s `set(self._group_members)`), so that order is unobservable through either chart. Pinned directly because a future caller could observe it.

## Tests

- `tests/util/test_hue_groups.py` — 9 direct tests, including the two behaviours no caller can reach: an omitted `order` (both callers always pass their legend's), and a name the order does not mention (both draw their names from the legend they order by). The latter is what `order.index` would raise on, so it is defined rather than left to chance.
- Two call-site tests as above.

## Verification

- Full suite: **2904 passed, 18 skipped** — 2892 before, plus the 12 added here. No existing test needed editing, which is the signal a no-behaviour-change extraction wants.
- `ruff check --diff` (what CI runs) clean, exit 0.

---

### DIFF
--- a/maidr/core/plot/scatterplot.py
+++ b/maidr/core/plot/scatterplot.py
@@ -12,6 +12,7 @@ from matplotlib.colors import to_rgba
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
 from maidr.exception import ExtractionError
+from maidr.util.hue_groups import grouped_by_name
 from maidr.util.mixin import CollectionExtractorMixin, LineExtractorMixin
 
 
@@ -203,12 +204,11 @@ def hue_groups(ax: Axes, collection: PathCollection) -> list[tuple[str, list[int
       read as one layer, even though the same axes would split if asked a
       line later. seaborn builds its legend inside the call, so its charts
       are unaffected.
-    - **A point no swatch claims.** A *continuous* hue is the case: measured,
-      ``hue='v'`` on a numeric column gives ten distinct colours for ten
-      points against five legend levels sampled at round numbers, so most
-      points match nothing. That is a colour *scale*, not a grouping, and
-      splitting it into one layer per point would be nonsense.
-    - **Fewer than two groups.** Nothing to tell apart.
+    Two more declines are :func:`~maidr.util.hue_groups.grouped_by_name`'s
+    rather than this function's, and its docstring carries the charts behind
+    them: a point no swatch claims, and fewer than two groups. They live
+    there because the rug split reaches the same two from a different artist,
+    and three implementations of one rule had begun to drift (#599).
 
     Parameters
     ----------
@@ -236,15 +236,13 @@ def hue_groups(ax: Axes, collection: PathCollection) -> list[tuple[str, list[int
     if named is None or len(named) < 2:
         return None
 
-    members: dict[str, list[int]] = {name: [] for name in named.values()}
-    for index, colour in enumerate(colours):
-        name = named.get(colour)
-        if name is None:
-            return None
-        members[name].append(index)
-
-    groups = [(name, found) for name, found in members.items() if found]
-    return groups if len(groups) > 1 else None
+    # `named` was built in legend order, so its values are the order #502
+    # settled a grouped layer's layers on. The grouping itself, and the two
+    # declines that go with it, are `grouped_by_name`'s -- shared with the rug
+    # split, which reaches the same three decisions from a different artist.
+    return grouped_by_name(
+        [named.get(colour) for colour in colours], list(named.values())
+    )
 
 
 class ScatterPlot(MaidrPlot, CollectionExtractorMixin, LineExtractorMixin):
--- a/maidr/patch/rugplot.py
+++ b/maidr/patch/rugplot.py
@@ -19,6 +19,7 @@ from maidr.core.plot.rugplot import (
 )
 from maidr.core.plot.scatterplot import _rgba
 from maidr.patch.common import _draw_quietly, prospective_axes, wrap_seaborn
+from maidr.util.hue_groups import grouped_by_name
 from maidr.util.legend_names import legend_of, names_for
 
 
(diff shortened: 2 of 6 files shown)


jooyoungseo:
Both minor points applied in `f5f5e1b`.

**Type hints** — right, and they were the only loose ones in a function whose return type is written out exhaustively. Now `names: list[str | None]` and `order: list[str] | None`, with the numpydoc parameter block matching.

**The double scan** — replaced with a single lookup. Worth noting what that turned up: the obvious `{name: place for place, name in enumerate(order)}` is *not* equivalent, because a repeated name in the order would take the **last** place while `list.index` gives the first. Built first-occurrence-wins, and pinned with a test — unreachable from either caller, since a legend does not list one name twice, but it is the only case where the two spellings differ, so it is the one that says the lookup replaced `.index` rather than merely resembling it. Mutating `setdefault` to plain assignment fails that test and nothing else.

Thanks for tracing the two equivalences rather than taking them from the description — particularly the empty-`order` one. I had written the `if not order` branch as a simplification without checking what the old expression did in that case; you are right that `sorted` with a constant key preserved first-appearance order there, so the branch reproduces the old behaviour exactly rather than changing it. That is a better justification than the one I gave, and I would not have known it was true.

Your reading of `named.values()` as legend order is also the load-bearing assumption of the scatter side, and it is now pinned by `test_the_layers_come_out_in_the_legend_s_order_not_the_drawing_s` rather than resting on `_named_colours`'s construction alone.

Full suite after the change: **2905 passed, 18 skipped**. `ruff check --diff` clean.

---

PR #605, opened 2026-08-22
feat(rug): give a rug the axis bounds its braille is built from

jooyoungseo:
Refs xability/maidr#1132, where this was measured. The r-maidr half follows separately.

## The gap

A point layer renders braille **only in grid mode**, and grid mode is built from `axes.{x,y}.{min,max,tickStep}`. A rug emitted labels alone. Measured against maidr's `ScatterTrace`:

```
axes as emitted before          braille { empty: true }
    x: { label: 'v' }, y: { label: 'Rug' }

with the bounds this adds       braille { values: [[2, 1, 0, 1]], min: 0, max: 2 }
    x: { label: 'v', min: 0, max: 10, tickStep: 2.5 }
    y: { label: 'Rug', min: 0, max: 1, tickStep: 1 }
```

So a rug was the one chart with **no braille surface reachable by any keystroke**. An ordinary scatter can enter grid mode and get one; a rug could not, because it never emitted what grid mode is built from.

## Why it matters more here than for a scatter

`values: [[2, 1, 0, 1]]` is the observation count per cell along the axis the rug marks — two observations in the first quarter, one in the second, none in the third, one in the fourth. That is the clustering a rug exists to show.

And it is the one thing a rug's **audio cannot carry**. Every tick sits at the same place on the axis pitch is mapped from, so `MathUtil.interpolate` hits its `fromMin === fromMax` guard and every tick plays at `minFrequency`. Position reaches the reader through stereo pan and text, one tick at a time; the shape of the distribution reached them nowhere. Braille is the modality that can hold a whole distribution at once.

## What is emitted

- **The observation axis** takes the chart's own limits and tick step — the grid the frontend builds is of the *chart*, so a reader feeling it is feeling the plotting area they would see. Declined on the same grounds `ScatterPlot._extract_axes_data` declines them: a non-linear scale, ticks that are not evenly spaced, or bounds enclosing no whole cell.
- **The axis across the ticks** is supplied whole as `{min: 0, max: 1, tickStep: 1}`. A rug is one row deep by construction and no reading of the panel could say so. One row is the right shape — measured, `tickStep` 0.5 gives `[[2, 1, 0, 1], [0, 0, 0, 0]]`, a row of the surface spent saying nothing.

## Additive only

Grid mode is entered deliberately, so the ordinary reading must not move. Verified rather than assumed: walking all four ticks with and without the bounds gives **byte-identical** `freq`, `panning` and `text` at every step, and the layer's name, data and selector shape are pinned unchanged.

Every case measured:

| chart | x | y |
|---|---|---|
| `rugplot(x=)` | `v`, 0.6–9.4, step 1 | `Rug`, 0–1, step 1 |
| `rugplot(y=)` | `Rug`, 0–1, step 1 | `v`, 0.6–9.4, step 1 |
| `rugplot(hue=)`, both layers | bounds | bounds, `z` preserved |
| rug over a `kdeplot` | bounds; the kde layer untouched | |
| log axis | label only | label only |

## Verification

- Five mutations, one at a time against a restored baseline: bounds never emitted, the wrong axis read, the strip two rows deep, and the two guards. **Four killed outright.**
- **Two survived and masked each other**, which took two rounds to untangle:
  - The grid-validity check and the log-scale check both declined the one log chart I had tested, so dropping either left the suite green. Fixed for the first by a chart that isolates it — a *linear* axis with `set_xticks([0, 1, 5, 10])`, where the scale guard passes and the tick step is `None`. That mutation now dies.
  - The log-scale check survives even so, and the reason is worth the note it now carries: matplotlib answers `get_xlim` in **log space** and `get_xticks` in **data space**, so a log chart's tick step is measured against a span it does not belong to and the `step > (high - low)` clause declines it anyway. Measured — ticks `[1, 2, 3]` against limits `(0.163, 0.914)`.

    Kept as a guard and **documented as currently unexercised**, with a test pinning the two spaces rather than the decline. It is a fact about matplotlib, not about this layer; if a release ever makes the two agree, that test goes red and the guard starts earning its keep. Same treatment as `_point_colours`'s count check.
- `tests/core/plot/test_rugplot.py` — 36 tests (9 new). `tests/core/test_axes_schema.py` passes, so the canonical `axes` contract still holds.
- Full suite: **2913 passed, 18 skipped**. `ruff check --diff` clean, exit 0.

## Not in scope

This does not make a rug's *audio* informative — pitch still carries nothing in the default mode, and xability/maidr#1132 lists four options for that, none of which I have taken. It means the information is not unavailable, only unavailable in one modality.

---

### DIFF
--- a/maidr/core/plot/rugplot.py
+++ b/maidr/core/plot/rugplot.py
@@ -8,7 +8,7 @@ from matplotlib.collections import LineCollection
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
-from maidr.core.plot.scatterplot import ScatterPlot
+from maidr.util.grid_axes import tick_step
 from maidr.util.legend_names import title_of
 
 #: Key the patch hands this layer the ``LineCollection`` its own call drew
@@ -218,9 +218,9 @@ class RugPlot(MaidrPlot):
         frontend builds is of the *chart*, and a reader feeling it is feeling
         the plotting area they would see. Taken from the same three places
         :meth:`~maidr.core.plot.scatterplot.ScatterPlot._extract_axes_data`
-        takes them, and declined on the same grounds -- a log scale, ticks
-        that are not evenly spaced, or bounds that do not enclose at least one
-        cell. A grid built on any of those would be a surface whose cells do
+        takes them -- the step through the helper both share -- and declined
+        on the same grounds: a log scale, ticks that are not evenly spaced, or
+        bounds that do not enclose at least one cell. A grid built on any of those would be a surface whose cells do
         not correspond to the axis a reader is told about.
 
         Only this axis is asked. The one across the ticks is supplied whole by
@@ -249,7 +249,7 @@ class RugPlot(MaidrPlot):
 
         low, high = self.ax.get_xlim() if self._along_x else self.ax.get_ylim()
         ticks = self.ax.get_xticks() if self._along_x else self.ax.get_yticks()
-        step = ScatterPlot._compute_tick_step(ticks)
+        step = tick_step(ticks)
 
         if step is None or low >= high or step <= 0 or step > (high - low):
             return None
@@ -268,8 +268,6 @@ class RugPlot(MaidrPlot):
         axes_data = super()._extract_axes_data()
         along = MaidrKey.X if self._along_x else MaidrKey.Y
         across = MaidrKey.Y if self._along_x else MaidrKey.X
-        axes_data[across] = self._axis_config(label=RUG_AXIS_LABEL)
-
         # Bounds on both axes, which is what makes the layer reachable in
         # grid mode -- and grid mode is the only mode where a point layer
         # renders braille at all. Measured against maidr's `ScatterTrace`:
@@ -301,6 +299,8 @@ class RugPlot(MaidrPlot):
             axes_data[across] = self._axis_config(
                 label=RUG_AXIS_LABEL, min=0, max=1, tick_step=1
             )
+        else:
+            axes_data[across] = self._axis_config(label=RUG_AXIS_LABEL)
 
         # A grouped layer names the grouping *variable* on z, the way the
         # scatter and line layers do: `z` says what the split is by, `name`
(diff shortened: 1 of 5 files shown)


jooyoungseo:
You are right about the coupling, and it was inconsistent of me — I made exactly the opposite argument in #604 one change earlier. Fixed in `862eca2`, along with both nits.

## The coupling

`maidr/util/grid_axes.tick_step` now holds the reading; `ScatterPlot._compute_tick_step` stays as a one-line delegate so its own callers are untouched, and `RugPlot` imports the helper instead of reaching across.

**Only the step moved**, and the reason is worth stating because your suggestion went further ("maybe the whole linear/uniform-ticks/valid-bounds check"). I looked at that and did not do it: the two callers do not ask the same question. `ScatterPlot._is_valid_grid_config` requires **both** axes linear and declines them together; `RugPlot` asks only of the axis its observations lie along. Folding them into one helper would change what a scatter emits for a chart with one transformed axis — a log `y` against a linear `x` currently loses both, and a per-axis helper would give it `x` bounds. That is a real behaviour change and it does not belong in this PR. So the shared piece is the part that genuinely is shared, and each caller keeps its own validity rule. The helper's docstring says so, so the next person does not have to rediscover it.

Four direct tests, including one pinning that a descending axis yields a *negative* step rather than being corrected — the callers reject a non-positive step, and which of them should is theirs to decide.

## Both nits

- **The doubled `across` write** — restructured into an `else`, so the label-only config is built only when there are no bounds.
- **`z` on the grouped rug** — added. You are right that the description claimed it and the test did not check it; the code path is untouched, but that is exactly the kind of claim that should be pinned rather than asserted in prose.

## `EventPlot`

Good catch, and it is the same shape — one point layer, braille only in grid mode. Not tracked anywhere yet. I have not measured whether an event plot's axes carry bounds already (its "Row" axis is a genuine integer axis rather than a synthetic one-row strip, so it may differ in kind from the rug), and I would rather measure before filing than file on the analogy. Noting it here so it is on the record either way; I will look at it after this and xability/r-maidr#224 land.

## Verification

- Full suite: **2917 passed, 18 skipped** (2913 before, plus the 4 helper tests).
- `ruff check --diff` clean, exit 0.

---

PR #607, opened 2026-08-22
feat(eventplot): give a raster the axis bounds its braille is built from

### LINKED ISSUE #606: An eventplot has no braille either, for the same reason a rug had none
Raised in review on #605 and filed after measuring rather than on the analogy, because the shape turns out to differ from the rug's in one way that matters.

## The gap

`ax.eventplot([[1, 2, 3, 9], [2, 4, 6]])` emits:

```
point  name=None  axes={ x: {label: 'X'}, y: {label: 'Row'} }
       data=[{x: 1.0, y: 0.0}, {x: 2.0, y: 0.0}, {x: 3.0, y: 0.0}, …]

point  name=None  axes={ x: {label: 'X'}, y: {label: 'Row'} }
       data=[{x: 2.0, y: 1.0}, {x: 4.0, y: 1.0}, {x: 6.0, y: 1.0}]
```

Labels only. A point layer renders braille **only in grid mode**, and grid mode is built from `axes.{x,y}.{min,max,tickStep}` — so an event plot, like a rug before #605, has no braille surface reachable by any keystroke.

It is the same class of loss for the same reason: the events' positions along `x` are announced one at a time and their *distribution* — where they bunch up, where the gaps are — reaches the reader nowhere. On an event plot that is often the whole point, since the chart is usually a timeline of occurrences.

## Where it differs from the rug, and why that decides the fix

A rug's across axis is synthetic: every entry sits at 0, so #605 supplies `{min: 0, max: 1, tickStep: 1}` whole and calls it one row deep.

An event plot's is **real**. Each layer sits at its own integer row — layer 1 entirely at `y = 0`, layer 2 entirely at `y = 1` — and "Row" is a genuine axis across the layers rather than a placeholder. So the bounds cannot be copied from the rug:

- Each layer is still one row deep *in its own row*, so the across bounds want to span that layer's row rather than 0 to 1 — something like `{min: row - 0.5, max: row + 0.5, tickStep: 1}`.
- Whether the grid should instead span **all** rows, so a reader can feel one row against another, is a design question the rug never had to answer. A per-layer grid gives a 1×N density per row; a whole-chart grid would give an M×N surface, which is arguably the better reading of an event plot and is not what any other point layer does.

That second point is why this is not a copy of #605. It needs deciding before it is written.

## The `x` half is uncontroversial

Whatever is decided about rows, the observation axis takes the chart's own bounds exactly as the rug's does, through `maidr.util.grid_axes.tick_step` and the same validity rule. That half could land on its own.

## Related

- #605 — the rug half, where this was measured from.
- xability/maidr#1132 — the finding that a point layer's braille lives only in grid mode, and that a constant-axis layer's audio carries nothing.
- xability/r-maidr#224 — the same change for `geom_rug()`. r-maidr has no event-plot reading, so this is py-maidr's alone.

jooyoungseo:
Closes #606.

## What was wrong

A point layer renders braille only in grid mode, and grid mode is built from `axes.{x,y}.{min,max,tickStep}`. An event plot emitted labels alone, so `ScatterTrace` came back `{empty: true}` — a raster was the second chart, after a rug in #605, with no braille surface reachable by any keystroke.

Measured against `ScatterTrace` with the layer this branch emits:

| layer payload | `braille` |
|---|---|
| labels only (before) | `{empty: true}` |
| row 0, `x 0.7–7.3 step 1`, `y -0.5–0.5 step 1` | `values [[1, 0, 0, 1, 0, 0, 1]]` |
| row 1, `x 0.7–7.3 step 1`, `y 0.5–1.5 step 1` | `values [[0, 1, 0, 0, 1, 0, 0]]` |

That spacing is the one thing a raster's audio cannot carry. Every event in a row is the same pitch — they all sit at the same place on the axis pitch is mapped from — so the pattern the chart is drawn to show is exactly what the reader was missing.

## One grid per row

#606 left open whether the grid should be per-row or one surface for the whole chart. That is settled by the architecture rather than by preference: `ScatterTrace` holds `gridCells` as instance state and builds them from its own layer's `data`, so it never sees a sibling row's points. A whole-chart surface, one row felt against another, is not something a producer can ask for — each row gets its own.

## Each row gets the cell it was drawn in

`one_row_around(get_lineoffset())`, not one centred on zero. That is the whole difference between a raster and a rug, which always sits at zero: `eventplot` stacks its rows at 0, 1, 2 …, and `lineoffsets=` moves them further.

Handed the zero-centred cell instead, the second row's points all fall outside the surface and it comes back `[[0, 0, 0, 0, 0, 0, 0]]` — a grid that renders, reports nothing wrong, and says the row is empty. Every coordinate in the payload stays right while that happens, which is the kind of gap no assertion about the data would catch.

## The extraction

`RugPlot._observation_bounds` becomes `maidr/util/grid_axes.bounds_along`, lifted whole — three declines and the note on why the log-scale guard is currently redundant included — beside the `tick_step` #605 already extracted. Second caller, so it moves; the same threshold #599 set for the hue split.

`ScatterPlot` deliberately keeps its own rule and this PR does not touch it: a scatter needs **both** axes linear and declines them together, while a rug and a raster ask only of the axis their entries lie along. Folding those into one helper would change what a scatter emits for a chart with one transformed axis, which is not this change's business.

`one_row_around(position)` centres the single cell on the entries rather than starting at them. A rug's strip moves from `0..1` to `-0.5..0.5` as a result. Measured against `ScatterTrace` before the pinned test values were touched, both spellings give `values [[2, 1, 0, 1]]` — but `0..1` puts the entries on a cell *edge*, and which side of an edge a value falls on is the frontend's tie-break rather than something the producer states. Centring also generalises to a layer whose row is not zero, which is what the raster needs.

## Declined together

An axis that cannot carry a grid — non-linear, unevenly spaced ticks, or bounds enclosing no whole cell — takes the row cell with it. Bounds on one axis alone build nothing, and emitting half a grid reads as a chart that meant to offer one.

## Additive

Points, layer names, selectors and axis labels are unchanged; `test_the_bounds_change_nothing_the_layer_already_said` pins that against the values the layer emitted before this branch.

## Verification

- `tests/core/plot/test_eventplot.py` → 22 passed (8 new)
- `tests/core/plot/test_rugplot.py` → 36 passed
- Full suite: **2925 passed, 18 skipped**
- `ruff check` / `ruff format --check` clean on every touched file

Each new test was checked by mutation, applied alone against a restored baseline — all six killed:

| mutation | result |
|---|---|
| row cell always at zero | 3 failed |
| no bounds on the row axis | 4 failed |
| no bounds on the event axis | 2 failed |
| bounds read off the wrong axis | 3 failed |
| row label recomputed as `"Row"` | 2 failed |
| `along` key swapped | 2 failed |

---

### DIFF
--- a/maidr/core/plot/eventplot.py
+++ b/maidr/core/plot/eventplot.py
@@ -9,6 +9,7 @@ from matplotlib.collections import EventCollection
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
+from maidr.util.grid_axes import bounds_along, one_row_around
 
 #: The keyword the event patch hands one row's collection to this layer under.
 #: A chart is one collection per row and one layer per collection, so the
@@ -190,9 +191,7 @@ class EventPlot(MaidrPlot):
         self._marks = [index for index, _ in drawn]
 
         if horizontal:
-            return [
-                {MaidrKey.X: position, MaidrKey.Y: offset} for _, position in drawn
-            ]
+            return [{MaidrKey.X: position, MaidrKey.Y: offset} for _, position in drawn]
         return [{MaidrKey.X: offset, MaidrKey.Y: position} for _, position in drawn]
 
     def _extract_axes_data(self) -> dict:
@@ -219,6 +218,45 @@ class EventPlot(MaidrPlot):
         if not named:
             axes_data[stacked] = self._axis_config(label="Row")
 
+        # Bounds on both axes, which is what makes the layer reachable in
+        # grid mode -- the only mode where a point layer renders braille at
+        # all. Measured against maidr's `ScatterTrace`, an event plot came
+        # back with an empty braille state, the same gap a rug had before
+        # #605 (#606).
+        #
+        # The grid a trace builds is of its **own** layer: `ScatterTrace`
+        # holds `gridCells` as instance state and never sees a sibling's
+        # points. So a whole-chart surface, one row against another, is not
+        # something a producer can ask for -- each row gets its own, and that
+        # settles the question #606 left open rather than choosing between
+        # the two.
+        along = MaidrKey.X if horizontal else MaidrKey.Y
+        bounds = bounds_along(self.ax, horizontal)
+        if bounds is not None:
+            low, high, step = bounds
+            axes_data[along] = self._axis_config(
+                label=axes_data[along].get(MaidrKey.LABEL),
+                min=low,
+                max=high,
+                tick_step=step,
+            )
+            # This row's own line offset, not zero: an event plot stacks its
+            # rows at 0, 1, 2 ... and each layer holds exactly one of them,
+            # so the cell has to be the one its events actually sit in. A rug
+            # is the same shape with the offset always zero.
+            row_low, row_high, row_step = one_row_around(
+                float(self._collection.get_lineoffset())
+            )
+            # Read back rather than recomputed, so the caller's label and the
+            # "Row" fallback above are the one answer to what this axis is
+            # called and a grid never renames it.
+            axes_data[stacked] = self._axis_config(
+                label=axes_data[stacked].get(MaidrKey.LABEL),
+                min=row_low,
+                max=row_high,
+                tick_step=row_step,
+            )
+
         return axes_data
 
     def _get_selector(self) -> str | list[str]:
@@ -239,6 +277,5 @@ class EventPlot(MaidrPlot):
             return []
 
         return [
-            f"g[id='{gid}'] > path:nth-of-type({index + 1})"
-            for index in self._marks
+            f"g[id='{gid}'] > path:nth-of-type({index + 1})" for index in self._marks
         ]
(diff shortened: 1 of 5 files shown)


jooyoungseo:
Both points taken; ba946c8 addresses them.

**The formatting churn — right finding, opposite cause.** The pinned ruff is `0.3.4` and my local `uv run ruff --version` is `0.3.4` too, so those hunks are not a newer ruff leaking in. They are 0.3.4 *reverting* a newer one: three spots in these files were checked in with a later ruff's wrapping, and running `ruff format` over the whole file rewrapped them back. CI would never have caught it either way — `ci.yml` runs `check --diff` and `check --select FA102`, never `format --check`, so nothing enforces the formatter's output at all. Restored to what was checked in, so the diff is the change and nothing else.

**The unit cell.** Added the reasoning to `one_row_around` itself rather than to the caller, since that is where the next reader will ask. Row spacing is not reflected because there is no neighbour to measure against: `ScatterTrace` builds `gridCells` from the points of the layer it holds and never sees a sibling's, so `lineoffsets=2` moves which cell a row sits in, not how wide it is.

**The rug's schema change** is a fair thing to flag and worth being precise about: `RugPlot` only started emitting `min`/`max`/`tickStep` on that axis in #605, one PR ago, so the `0..1` spelling has never been in a release. Nothing downstream can have snapshotted it yet.

---

PR #609, opened 2026-08-22
fix(hue): name a faceted panel's lone group from the legend that names it

### LINKED ISSUE #608: A faceted panel holding one hue level leaves that level unnamed
Follow-up to #591, which is about the same figure but a different half. The late-binding defect #591 describes is fixed — each panel now resolves against its own axes and containers — and the measurement in that issue records what is still missing:

```
panel 1 -> [None]          # one container, below names_for's two-artist floor
```

That is this issue.

## What happens

`displot(hue=..., col=...)` over a grid whose panels hold different numbers of groups names every panel except the ones holding exactly one. The same data drawn unfaceted names all of them.

Measured on seaborn 0.13.2, three groups with `col` putting `a` and `b` in one panel and `c` alone in the other:

```
displot(x="v", hue="g", col="c", bins=3)                  ['b', 'a', None]
displot(x="v", hue="g", col="c", bins=3, element="step")  ['b', 'a', None]
displot(x="v", hue="g", col="c", kind="kde")              ['b', 'a', None]
displot(x="v", hue="g", bins=3)      # unfaceted control  ['c', 'b', 'a']
```

Bars, step outlines and KDE curves alike. A reader walking the layers hears two named distributions and then one that will not say which group it is, on a chart whose legend names it plainly.

## Why

`names_for` declines a lone artist, and says so:

> A single container is left unnamed whatever the legend says: one distribution needs nothing to tell it apart from, and a name there would read as though the chart held more.

That is right about one axes and wrong about this figure. The question the floor asks is "is this artist alone on its axes"; the question that decides whether a name is worth having is "is this layer alone in the chart". A faceted grid separates the two: the figure holds three distributions, the panel holds one.

The name is recoverable — it is the figure legend, which the grid draws for all the panels at once:

```
panel 1 colours   [(0.173, 0.627, 0.173, 0.5)]
names_for   -> [None]
name_for    -> ['c']
```

`name_for` is the helper #595 added for exactly this shape: "`names_for` declines a lone artist … which is right where the artists of one layer are named against each other, and wrong where a *layer* is the artist."

## Scope

Five resolvers reach the floor from the per-panel loops, so the rule wants stating once rather than five times:

| path | resolver |
|---|---|
| bars | `_group_names(ax, containers)` |
| `element="step"` / `"poly"`, filled | `_names_for(ax, [_face_colour(...)])` |
| `element="step"` / `"poly"`, `fill=False` | `_names_for(ax, [_rgba(...)])` |
| `kind="kde"` curves | `_curve_names(ax, curves)` |
| `kind="kde"` fills | `_fill_names(ax, fills)` |

## What must not change

The floor is right for the unfaceted case and has to stay right there. `histplot(x="v", hue="g")` where `g` has a single level draws one distribution and names it `None` today; a fix that names it would be reading a legend as though the figure held more than it does — and would put the axes-level and figure-level spellings of one chart out of step, which is the asymmetry #522, #446 and #590 were each about. Both spellings emit `None` today and both should keep doing so:

```
displot(one_level, x="v", hue="g")   None    # figure legend, one panel
histplot(one_level, x="v", hue="g")  None    # axes legend
```

So the predicate is the panel count, not where the legend was hung. `plotter_axes(instance)` already answers it in both loops (`histogram.py:885`, `kdeplot.py:326`).

jooyoungseo:
Closes #608. Follow-up to #591, whose binding half landed in #592 and whose remaining `[None]` is this.

## What was wrong

`names_for` declines a lone artist, and says so:

> A single container is left unnamed whatever the legend says: one distribution needs nothing to tell it apart from, and a name there would read as though the chart held more.

That asks whether an artist is alone **on its axes**. It is right about one chart and wrong about one panel of a grid, where the two questions come apart: the figure holds several distributions and the panel holds one of them.

Measured on seaborn 0.13.2, three groups with `col` putting `a` and `b` in one panel and `c` alone in the other:

| chart | before | after |
|---|---|---|
| `displot(x=, hue=, col=, bins=3)` | `['b', 'a', None]` | `['b', 'a', 'c']` |
| `displot(…, element="step")` | `['b', 'a', None]` | `['b', 'a', 'c']` |
| `displot(…, kind="kde")` | `['b', 'a', None]` | `['b', 'a', 'c']` |
| `displot(x=, hue=, bins=3)` — unfaceted control | `['c', 'b', 'a']` | unchanged |

A reader walking the layers heard two named distributions and then one that would not say which group it was — on a chart whose legend names it plainly, and whose unfaceted spelling names all three. The name was there for the asking:

```
panel 1 colours   [(0.173, 0.627, 0.173, 0.5)]
names_for   -> [None]
name_for    -> ['c']
```

`name_for` is the helper #595 added for exactly this shape: "wrong where a *layer* is the artist."

## The rule, stated once

`maidr/util/legend_names.names_for_panel(ax, colours, faceted)` sits beside the two helpers it chooses between. The five resolvers that reach the floor route through it:

| path | resolver |
|---|---|
| bars | `_group_names` |
| `element="step"`/`"poly"`, filled | `_names_for_panel(ax, [_face_colour(...)])` |
| `element="step"`/`"poly"`, `fill=False` | `_names_for_panel(ax, [_rgba(...)])` |
| `kind="kde"` curves | `_curve_names` |
| `kind="kde"` fills | `_fill_names` |

`faceted` is `len(plotter_axes(instance)) > 1`, read once per draw in each of the two per-panel loops (`histogram.py`, `kdeplot.py`) — a property of the grid, not of what any panel drew. Both axes-level entry points keep the default `False`, so `histplot` and `kdeplot` are untouched.

## Why the panel count and not the legend's placement

The tempting shortcut is "a *figure* legend names more than this panel, so a lone artist under one can be named." It is wrong: seaborn's `FacetGrid` builds a figure legend for a single panel too. Taking it would leave

```
displot(one_level, x="v", hue="g")   'a'     # figure legend, one panel
histplot(one_level, x="v", hue="g")  None    # axes legend
```

— the same chart, two spellings, two answers, which is the asymmetry #522, #446 and #590 were each about. Both emit `None` today and `test_a_single_panel_holding_one_group_is_still_unnamed` pins that they still do.

## Only the floor moves

Every other decline is unchanged: no legend, a colour no swatch claims, a colour two swatches claim. `test_the_panels_holding_two_groups_are_unchanged` asserts the two-group panel in order rather than sorted, so a name landing on the wrong layer — the thing #591 was about — still fails.

`TestAFacetedHistogramNamesEachPanelFromItself` (the #591 test) changes one expected value from `None` to `"c"`, and its docstring records that the failure it guards against is now three layers reading `'c', 'c', 'c'` rather than three `None`s. Either way the names would not be each panel's own, so it still pins the binding.

## Verification

- `tests/core/test_displot_distribution.py` → 41 passed (5 new)
- Full suite: **2930 passed, 18 skipped**
- `ruff check` clean on every touched file

Mutations, each applied alone against a restored baseline — all six killed:

| mutation | result |
|---|---|
| `names_for_panel` ignores `faceted` (floor always stands) | 5 failed |
| floor never stands (name any lone artist) | 1 failed |
| `faceted = len(panels) >= 1` | 1 failed |
| bars resolver drops `faceted` | 3 failed |
| outline registrar drops `faceted` | 1 failed |
| KDE registrar drops `faceted` | 1 failed |

---

### DIFF
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -460,7 +460,7 @@ def _face_colour(outline: PolyCollection):
     Its **face**, which is what carries the hue. By default the edge carries
     it too -- measured, the two agree but for the translucency the face and
     the legend swatch share -- so either would name the groups, the edge
-    through `names_for`'s alpha-insensitive pass. `edgecolor=` is what
+    through `_names_for`'s alpha-insensitive pass. `edgecolor=` is what
     separates them: it is the caller's to set, and setting it once colours
     *every* group's edge alike::
 
--- a/maidr/patch/histogram.py
+++ b/maidr/patch/histogram.py
@@ -22,8 +22,18 @@ from maidr.core.plot.outlined_histogram import reads as outline_reads
 from maidr.core.plot.scatterplot import _rgba
 from maidr.core.plot.step_histogram import STEP_COUNTS, STEP_EDGES, STEP_ORIENTATION
 from maidr.core.plot.stepped_histogram import reads as _reads_outline
-from maidr.patch.common import _draw_quietly, common, plotter_axes, prospective_axes, wrap_seaborn
-from maidr.patch.kdeplot import _curve_names, _names_for, deferred_names
+from maidr.patch.common import (
+    _draw_quietly,
+    common,
+    plotter_axes,
+    prospective_axes,
+    wrap_seaborn,
+)
+from maidr.patch.kdeplot import (
+    _curve_names,
+    _names_for_panel,
+    deferred_names,
+)
 
 
 @wrapt.patch_function_wrapper(Axes, "hist")
@@ -209,7 +219,9 @@ def _new_bar_containers(ax: Axes | None, before: list) -> list[BarContainer]:
     ]
 
 
-def _group_names(ax: Axes | None, containers: list) -> list[str | None]:
+def _group_names(
+    ax: Axes | None, containers: list, faceted: bool = False
+) -> list[str | None]:
     """
     Name each container from the legend swatch drawn in its colour.
 
@@ -241,7 +253,14 @@ def _group_names(ax: Axes | None, containers: list) -> list[str | None]:
     distribution needs nothing to tell it apart from, and a name there would
     read as though the chart held more.
 
-    The match itself is ``kdeplot._names_for``, which is the same one this
+    Unless it is one panel of a grid, where those two stop being the same
+    thing -- the figure holds several distributions and this panel holds one
+    of them, so the name is worth having and the legend has it. `faceted`
+    says which case this is; see :func:`~maidr.util.legend_names.names_for_panel`
+    for why the panel count decides it rather than where the legend was hung
+    (#608).
+
+    The match itself is ``kdeplot._names_for_panel``, which is the same one this
     used to make inline against ``_named_colours`` plus a second pass this
     lacked. That pass is not a nicety here: a ``pairplot(hue=...)`` draws its
     bars translucent and builds its figure legend from **opaque** swatches --
@@ -256,6 +275,8 @@ def _group_names(ax: Axes | None, containers: list) -> list[str | None]:
         The axes drawn on, for its legend.
     containers : list
         The containers this call drew, in draw order.
+    faceted : bool
+        Whether this panel is one of several.
 
     Returns
     -------
@@ -265,7 +286,7 @@ def _group_names(ax: Axes | None, containers: list) -> list[str | None]:
     if ax is None:
         return [None] * len(containers)
 
-    return _names_for(ax, [_container_colour(c) for c in containers])
+    return _names_for_panel(ax, [_container_colour(c) for c in containers], faceted)
 
 
 def _container_colour(container) -> tuple[float, ...] | None:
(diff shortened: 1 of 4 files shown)


jooyoungseo:
All three addressed in 3025600.

**1. The `_names_for` re-export is not dead** — it reads that way, which is why the note is now at the import. Three tests reach the colour match through this module, which is where it lived before it moved to `maidr/util/legend_names.py`:

```
tests/core/plot/test_hue_kde_naming.py:164,173,184   from maidr.patch.kdeplot import _names_for
tests/core/plot/test_pairplot_group_names.py:246     from maidr.patch.kdeplot import _handle_colour, legend_of
```

Right instinct about the blanket `noqa` hiding it, though, so the comment now says what keeps both names alive.

**2. The degenerate case is real, and I measured it.** A `col=` grid whose `hue=` has one level across the whole frame:

```
displot(x="v", hue="g", col="c", bins=3)   before ['None', 'None']   after ['only', 'only']
  ax0 containers=1 legend=False
  ax1 containers=1 legend=False
  figure legends: 1
```

Kept rather than guarded against. The name is **true**: the figure legend says `g: only`, a sighted reader sees exactly that, and the panels are told apart by their `col` level rather than by hue — so the redundancy costs nothing a reader had. The case the floor is right about is the other one, where a name on a single-panel chart implies a companion layer that does not exist, and that still declines.

So `faceted` is stated as a proxy in `names_for_panel`'s docstring, with the divergence and the measurement, and `test_a_grid_whose_hue_has_one_level_names_every_panel_alike` pins it — so a later decision to decline it is a deliberate change to a test rather than a silent drift.

**3.** `_face_colour`'s docstring goes back to `_names_for`, matching the rest of the file.

---

Issue #345, opened 2026-08-10
feat: plot coverage roadmap — unpatched matplotlib/seaborn/plotly plots


jooyoungseo:
## Description

Tracking issue for plots py-maidr cannot read. Complements #255, which covers the seaborn categorical/distribution plots; this one covers the matplotlib, plotly and figure-level gaps.

## Parity gaps — supported elsewhere, missing here

These are not new chart types. The MAIDR runtime already handles them; the Python binding does not emit them, so the same figure is accessible or not depending on which library the user reached for.

- [ ] #337 — `pcolormesh` / `pcolor` are not patched, so a very common way to draw a matplotlib heatmap registers nothing. `QuadMesh.draw` is already patched for highlighting, so only the registration and extraction are missing. **Cheapest item here.**
- [ ] #338 — `stacked_normalized_bar` exists in the JS core but has no `PlotType` member, so 100% stacked bars are unreachable from Python
- [ ] #343 — the plotly path lacks violin, candlestick and smooth, all three of which the matplotlib path and the browser-side plotly adapter already support

## New plot types

- [ ] #339 — area plots (`stackplot`, `fill_between`) — blocked on xability/maidr#788
- [ ] #340 — error bars (`Axes.errorbar`) — blocked on xability/maidr#789
- [ ] #341 — `hist2d` and `hexbin` (the `hist2d` half needs no core work)
- [ ] #342 — `pairplot` / `jointplot`

## Figure-level seaborn

- [ ] #344 — `catplot` / `relplot` / `displot`. These are what the seaborn documentation steers users toward, so someone following the official tutorial has a good chance of landing on an unsupported path.

## Suggested order

#337 first — smallest change, and it closes a silent failure on a mainstream matplotlib call. Then #338 and #343, which are parity work with no upstream dependency. #339 and #340 follow the core traces.

## Related

- #255 — missing seaborn cartesian plots (#246–#254)
- Core roadmap: xability/maidr#814

jooyoungseo:
Ran every unchecked box on this roadmap against `main` (`af0ae1e`). **All of them pass except one**, so this issue is tracking work that is largely already done.

Verified by plotting and reading the emitted schema, not by looking for patch modules — a file existing is not the same as a chart being described.

## Parity gaps

| box | status |
|---|---|
| #337 `pcolormesh` | `['HEAT']` |
| #337 `pcolor` | `['HEAT']` |
[...]

PR #611, opened 2026-08-22
fix(hue): read a shared-axis panel's legend, so a jointplot's marginals are named

### LINKED ISSUE #610: A hue-split jointplot's marginals are two anonymous curves, because the legend is on the joint axes
Found while measuring `pairplot`/`jointplot` against the roadmap in #345 (#342). Almost everything there reads correctly — the sweep is at the bottom — and this is the one gap.

## What happens

`sns.jointplot(data=..., x=..., y=..., hue=...)` splits all three panels by hue, names the joint panel's layers, and names neither marginal's.

Measured on seaborn 0.13.2, two levels `p` and `q`:

```
[0][0] smooth name=None  data 30 points     # top marginal, level 1
[0][0] smooth name=None  data 30 points     # top marginal, level 2
[1][0] point  name='p'   data 30 points     # joint
[1][0] point  name='q'   data 30 points     # joint
[1][1] smooth name=None  data 30 points     # right marginal, level 1
[1][1] smooth name=None  data 30 points     # right marginal, level 2
```

The data is right in every panel. A reader who moves onto a marginal finds two density curves with identical announcements and nothing to tell them apart — which is the defect #558 named, on the two panels of the chart where it was not yet fixed.

## Why

Not the lone-artist floor (#608): each marginal holds **two** curves. The marginals simply have no legend to read.

```
ax0 (joint)      own legend=True    legend_of -> Legend
ax1 (marg x)     own legend=False   legend_of -> None
ax2 (marg y)     own legend=False   legend_of -> None
figure legends: 0
```

`legend_of` falls back to a **figure** legend and only to that, which is what makes a `pairplot`'s diagonal readable (#561). A `JointGrid` hangs its one legend on `ax_joint` instead, so the marginals find nothing and every resolver downstream declines correctly on the input it was given.

## Why the obvious widening is wrong

"Read any sibling axes' legend" walks straight into the hazard `legend_of` already documents:

> One figure legend is read as naming *every* axes, and nothing in the artists can say otherwise: several panels with independent hues draw the same default colour cycle, so a legend built for one of them matches all of them. Measured, two `kdeplot(hue=...)` panels drawn `legend=False` with one `fig.legend()` for the first come out with the first panel's names on both.

Two unrelated panels in a hand-built figure would then take each other's names.

## A signal that says the panels are one chart

`JointGrid` builds its marginals **sharing an axis with the joint axes** — that is how the marginal lines up with the scatter it sits beside — and matplotlib records it:

```
jointplot                    ax0 legend=True   shares_with=[1, 2]
                             ax1 legend=False  shares_with=[0]
                             ax2 legend=False  shares_with=[0]
plt.subplots(1, 2)                             shares_with=[]        <- the hazard case
plt.subplots(1, 2, sharex=True)                shares_with=[1]
```

So: **when an axes has no legend of its own and exactly one axes it shares an axis with has one, that legend names its colours.** From either marginal the answer is `ax_joint`, uniquely.

This is narrower than the fallback already in place, not wider. The two-panel `kdeplot` figure the docstring measures is `plt.subplots(1, 2)` — unshared — so it keeps returning `None`, and `tests/core/plot/test_pairplot_group_names.py` keeps pinning it. What is newly in scope is `plt.subplots(1, 2, sharex=True)` with a legend on one panel only, which is a figure whose panels were declared to be on one scale.

The axes' own legend must still win, for the reason already written down.

## The rest of the sweep, for the record

Everything below reads correctly today and is listed so #342 can be closed against measurements rather than assumption. Layers are `(type, name, points)` per subplot cell.

| chart | reading |
|---|---|
| `pairplot` | diagonal `hist`, off-diagonal `point` |
| `pairplot(hue=)` | diagonal `smooth` ×2 named, off-diagonal `point` ×2 named |
| `pairplot(kind="reg")` | off-diagonal `point` + `smooth` |
| `pairplot(kind="kde")` | diagonal `smooth`, off-diagonal `contour` |
| `pairplot(kind="hist")` | diagonal `hist`, off-diagonal `heat` |
| `pairplot(corner=True)` | lower triangle read, upper cells `layers: []` |
| `pairplot(hue=, diag_kind="hist")` | diagonal `hist` ×2 named, off-diagonal `point` ×2 named |
| `jointplot` | marginal `hist` ×2, joint `point` |
| `jointplot(kind="hex")` | joint `hexbin` |
| `jointplot(kind="kde")` | marginal `smooth`, joint `contour` |
| `jointplot(kind="reg")` | joint `point` + `smooth`, marginals `hist` + `smooth` |
| `jointplot(kind="hist")` | joint `heat` |
| `jointplot(kind="resid")` | joint `point`, marginals `hist` |
| `residplot` | `point` |

Two details worth noting rather than filing:

- The **unoccupied corner** of a `JointGrid`, and the upper triangle of `pairplot(corner=True)`, emit `{id, layers: []}`. That is the shape `maidr`'s `Subplot` explicitly supports — "producers legitimately emit `layers: []` for an unoccupied grid cell" — so it is correct, not a phantom.
- The **right marginal** is read transposed correctly: `orientation: "horz"`, `x: "Count"`, `y: <column>`, with `yMin`/`yMax` carrying the bin edges. Not the #480 shape.

jooyoungseo:
Closes #610.

## What was wrong

A `JointGrid` draws three panels off one hue mapping and hangs the one legend that names it on `ax_joint`. `legend_of` reads an axes' own legend and, failing that, a **figure** legend — which is where a `PairGrid` puts one (#561). A `JointGrid` puts it in neither place as far as a marginal is concerned.

Measured on seaborn 0.13.2, `jointplot(hue=...)` over two levels:

| panel | before | after |
|---|---|---|
| top marginal, curve 1 | `None` | `q` |
| top marginal, curve 2 | `None` | `p` |
| joint, point layer 1 | `p` | `p` |
| joint, point layer 2 | `q` | `q` |
| right marginal, curve 1 | `None` | `q` |
| right marginal, curve 2 | `None` | `p` |

The data was right in every panel. A reader who moved onto a marginal found two density curves with identical announcements and nothing to tell them apart — the defect #558 named, on the two panels where it was not yet fixed.

Not the lone-artist floor #608 lifted: each marginal holds **two** curves. There was simply no legend for them to be matched against.

```
ax0 (joint)      own legend=True    legend_of -> Legend
ax1 (marg x)     own legend=False   legend_of -> None
ax2 (marg y)     own legend=False   legend_of -> None
figure legends: 0
```

## The signal, and why it is narrower than what is already there

The tempting widening — "read any sibling's legend" — walks into the hazard `legend_of` already documents and `test_pairplot_group_names.py` already pins: two unrelated panels drawing the same default colour cycle would take each other's names.

A `JointGrid` builds its marginals **sharing an axis** with the joint axes, because that is how a marginal lines up with the scatter beside it, and matplotlib records it:

```
jointplot                        ax0 legend=True   shares=[1, 2]
                                 ax1 legend=False  shares=[0]
                                 ax2 legend=False  shares=[0]
plt.subplots(1, 2)                                 shares=[]        <- the hazard case
plt.subplots(1, 2, sharex=True)                    shares=[1]
```

So: **when an axes has no legend of its own and no figure legend, and exactly one axes it shares an axis with has one, that legend names its colours.** From either marginal the answer is `ax_joint`, uniquely.

The hazard case is `plt.subplots(1, 2)` — unshared — so it is untouched and still answers `None`; `test_an_unshared_panels_legend_is_not_read` holds that line from the other side. What is newly in scope is `plt.subplots(1, 2, sharex=True)` with a legend on one panel only, which is a figure whose panels were declared to be on one scale, and `test_a_shared_panels_legend_is_read` pins it so the widening is visible rather than implied.

The two existing rules are unchanged: the axes' own legend wins, and two candidates decline.

## Found while measuring the roadmap

#610 came out of sweeping `pairplot`/`jointplot` against #345/#342. Everything else there reads correctly today — 14 chart/kind combinations, listed in the issue — and this was the one gap. Two things worth knowing that are **not** defects and are recorded there so they are not re-investigated:

- The unoccupied corner of a `JointGrid`, and the upper triangle of `pairplot(corner=True)`, emit `{id, layers: []}`. That is the shape maidr's `Subplot` explicitly supports for an unoccupied grid cell.
- The right marginal is read transposed correctly: `orientation: "horz"`, `x: "Count"`, `y: <column>`, with `yMin`/`yMax` carrying the bin edges. Not the #480 shape.

## Verification

- `tests/core/plot/test_jointplot_group_names.py` → 7 passed (new file)
- With `test_pairplot_group_names.py` → 21 passed
- Full suite: **2938 passed, 18 skipped**
- `ruff check` clean on both touched files

Mutations, each applied alone against a restored baseline — all five killed:

| mutation | result |
|---|---|
| sharer fallback removed | 2 failed |
| no uniqueness guard (first sharer wins) | 1 failed |
| every axes on the figure counts as a sharer | 1 failed |
| sharer consulted before the axes' own legend | 1 failed |
| only x-sharing counts (drops the right marginal) | 1 failed |

---

### DIFF
--- a/maidr/util/legend_names.py
+++ b/maidr/util/legend_names.py
@@ -54,7 +54,10 @@ def legend_of(ax: Axes | None):
     which is a figure whose panels were declared to be on one scale.
 
     Exactly one sharer, for the reason the figure fallback gives: two of them
-    cannot say which names this axes' colours.
+    cannot say which names this axes' colours. And **only** where the figure
+    carries none: two figure legends are ambiguous already, and a neighbour
+    having one of its own does not resolve them -- it would name the axes off
+    a third legend while two candidates for it stood unexamined.
 
     Parameters
     ----------
@@ -73,12 +76,15 @@ def legend_of(ax: Axes | None):
         return own
     figure = getattr(ax, "figure", None)
     legends = list(getattr(figure, "legends", ()) or ())
-    if len(legends) == 1:
-        return legends[0]
+    if legends:
+        # Two of them decline here rather than falling through to a sharer:
+        # a figure legend this axes could not be matched to does not stop
+        # being ambiguous because a neighbour has one of its own.
+        return legends[0] if len(legends) == 1 else None
     shared = [
-        sibling.get_legend()
+        legend
         for sibling in _sharing_an_axis_with(ax)
-        if sibling.get_legend() is not None
+        if (legend := sibling.get_legend()) is not None
     ]
     return shared[0] if len(shared) == 1 else None
 
--- a/maidr/util/legend_names.py
+++ b/maidr/util/legend_names.py
@@ -36,6 +36,26 @@ def legend_of(ax: Axes | None):
     written down here rather than left to be discovered, and pinned in
     `tests/core/plot/test_pairplot_group_names.py`.
 
+    A ``JointGrid`` puts the same one legend somewhere neither of those finds
+    it: on ``ax_joint``, an *axes* legend, leaving the two marginals with none
+    of their own and the figure with none either. Measured on a
+    ``jointplot(hue=...)``, the joint panel's two point layers are named and
+    each marginal's two density curves are not -- two identical
+    announcements, which is the defect #558 named (#610).
+
+    So a third place is read, and it is **narrower** than the figure fallback
+    rather than wider: an axes an axis is *shared* with. A ``JointGrid``
+    builds its marginals sharing one with the joint axes, because that is how
+    a marginal lines up with the scatter beside it, and matplotlib records
+    it. Two panels of a hand-built figure do not share by default, so the
+    hazard measured above -- ``plt.subplots(1, 2)`` with one ``fig.legend()``
+    -- is untouched and still answers ``None``. What is newly in scope is
+    ``plt.subplots(1, 2, sharex=True)`` with a legend on one panel only,
+    which is a figure whose panels were declared to be on one scale.
+
+    Exactly one sharer, for the reason the figure fallback gives: two of them
+    cannot say which names this axes' colours.
+
     Parameters
     ----------
     ax : Axes or None
@@ -53,7 +73,41 @@ def legend_of(ax: Axes | None):
         return own
     figure = getattr(ax, "figure", None)
     legends = list(getattr(figure, "legends", ()) or ())
-    return legends[0] if len(legends) == 1 else None
+    if len(legends) == 1:
+        return legends[0]
+    shared = [
+        sibling.get_legend()
+        for sibling in _sharing_an_axis_with(ax)
+        if sibling.get_legend() is not None
+    ]
+    return shared[0] if len(shared) == 1 else None
+
+
+def _sharing_an_axis_with(ax: Axes) -> list:
+    """
+    Every other axes this one shares an x or a y axis with.
+
+    Asked of matplotlib's own groupers rather than inferred from position or
+    from the figure's axes list, because sharing is the thing being tested and
+    it is a fact the axes carries.
+
+    Parameters
+    ----------
+    ax : Axes
+        The axes drawn on.
+
+    Returns
+    -------
+    list
+        The sharers, ``ax`` itself excluded. Empty when it shares nothing --
+        which is the default for a figure built with ``plt.subplots``.
+    """
+    sharers: list = []
+    for grouper in (ax.get_shared_x_axes(), ax.get_shared_y_axes()):
+        for sibling in grouper.get_siblings(ax):
+            if sibling is not ax and not any(one is sibling for one in sharers):
+                sharers.append(sibling)
+    return sharers
 
 
 def names_for(ax: Axes, colours: list) -> list:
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Good catch — that was a real gap between the code and the docstring I wrote for it, and I had it wrong. Fixed in af6cac7.

`if len(legends) == 1: return legends[0]` fell through to the sharer whenever the figure did not carry *exactly* one legend, which quietly includes **two** — the case `test_two_figure_legends_name_nothing` pins as irreducibly ambiguous. Naming the axes off a neighbour there would answer the question while two candidates for it stood unexamined, which is the opposite of the rule every other decline in the module follows.

Guarded on `not legends`, and pinned from the other side by `test_two_figure_legends_still_decline_even_with_a_sharer` — two figure legends, one legend-bearing sharer, still `None`. Verified the new test actually bites: restoring the old spelling turns it red and leaves the other 21 green, which is also why it did not show up before (the existing pin uses a single, non-sharing axes, so `shared` was always empty there).

The walrus nit is in too, since it was the same expression.

---

Issue #582, opened 2026-08-22
A hue-grouped ECDF announces each curve under the other group's name


jooyoungseo:
`sns.ecdfplot(hue=...)` draws one curve per group. py-maidr names them from the legend by position, and seaborn draws an ECDF's hue levels in the **reverse** of its legend order, so with two groups every point of both curves is announced under the wrong name.

Measured with groups that cannot be confused — `low` is 1–5, `high` is 100–500, no overlap:

```python
df = pd.DataFrame({
    "x": [1, 2, 3, 4, 5] + [100, 200, 300, 400, 500],
    "grp": ["low"] * 5 + ["high"] * 5,
})
sns.ecdfplot(df, x="x", hue="grp", ax=ax)
```

```
legend texts : ['low', 'high']
line 0       : x in [100.0, 500.0]     <- the "high" data
line 1       : x in [1.0, 5.0]         <- the "low" data

announced z='low'   x in [100.0, 500.0]   <- wrong
announced z='high'  x in [1.0, 5.0]       <- wrong
```

Nothing errors and no value is altered. Only the names are exchanged, which is the reading a user is least able to catch: the numbers look right, and the curve they belong to is the one thing the chart was drawn to compare.

## Why this one and not the others

The rest of the hue family is correct, and the contrast is what localises the bug. Measured on the same data:

| call | result |
| --- | --- |
| `ecdfplot(hue=)` | **names swapped** |
| `kdeplot(hue=)` | correct — `name: "low"` on the curve peaking at x=2.8, `"high"` on the one peaking at x=281 |
| `lineplot(hue=)` | correct |

`kdeplot` was given per-curve names in #560 and pairs them with the right curve; `lineplot`'s labels line up with the draw order to begin with. The ECDF is the case where seaborn's draw order and legend order disagree, and it is the one that pairs them by index.

This is #502's defect ("hue group named by legend order") surviving in a call that #502 did not cover.

## Expected

The curve holding 1–5 is announced `low`, and the curve holding 100–500 is announced `high` — matched to the group whose data each actually holds, rather than to a position in the legend.

Issue #255, opened 2025-10-13
feat: Missing Seaborn Cartesian Plots to be supported


Bhoomika2905:
# Feature Request

## Description

Currently, our visualization module supports only a limited subset of Seaborn Cartesian plots.
To enhance analytical and visualization capabilities, we should add support for the following additional plot types:

## Plots

- [ ] #254 – Add `sns.violinplot()` support
- [ ] #253 – Add `sns.boxenplot()` support
- [ ] #252 – Add `sns.stripplot()` support
- [ ] #251 – Add `sns.swarmplot()` support
- [ ] #250 – Add `sns.rugplot()` support
- [ ] #249 – Add `sns.ecdfplot()` support
- [ ] #248 – Add `sns.lmplot()` support
- [ ] #247 – Add `sns.residplot()` support
- [ ] #246 – Add `sns.pointplot()` support

## Checklist

- [ x ] I have searched for similar feature requests and confirmed that this is a new request.
- [ x ] I have provided a clear and concise description of the feature.
- [ x ] I have explained the motivation behind this feature request.
- [ x ] I have outlined a proposed solution or ideas for implementing this feature.
- [ x ] I have provided any additional context or screenshots if applicable.

jooyoungseo:
Status against `main` (`af0ae1e`), measured by plotting each one and reading the emitted schema rather than by looking for patch modules.

**Seven of the eight sub-issues pass.**

| issue | call | emitted layers |
|---|---|---|
| #247 | `sns.residplot` | `['SCATTER']` |
| #248 | `sns.regplot` | `['SCATTER', 'SMOOTH']` |
| #248 | `sns.lmplot` (figure-level) | `['SCATTER', 'SMOOTH']` |
| #249 | `sns.ecdfplot` | `['STEP']` |
| #251 | `sns.swarmplot` | `['SCATTER']` |
| #252 | `sns.stripplot` | `['SCATTER']` |
| #254 | `sns.violinplot` | `['VIOLIN_BOX', 'VIOLIN_KDE']` |
| — | `sns.kdeplot` / `histplot` / `scatterplot` | `['SMOOTH']` / `['HIST']` / `['SCATTER']` |
| **#250** | **`sns.rugplot`** | **NOT REGISTERED** |

#248 was worth checking twice: `lmplot` is figure-level and so bypasses the axes-level patches, which is exactly the shape that made `catplot` need its own work in #450/#458. It registers correctly either way.

## #250 is real, and blocked on the core

[...]

Issue #254, opened 2025-10-13
feat: Support ViolinPlot


Bhoomika2905:
# Feature Request

## Description

Currently, our visualization module supports only a limited subset of Seaborn Cartesian plots.  
To enhance analytical and visualization capabilities, we should add support for the `sns.violinplot()` function.  
A **violin plot** visualizes the distribution of numerical data for one or more categories, combining aspects of boxplots and KDE (kernel density estimate) plots. It is useful for understanding data spread and skewness across different groups.

## Proposed Solution

Use Seaborn’s built-in `sns.violinplot()` as the plotting layer. It should handle arguments like `x`, `y`, `hue`, and `data` and integrate seamlessly with existing plot theming.

## Additional Context

Below is a simple example demonstrating how a violin plot works:

```python
import seaborn as sns
import matplotlib.pyplot as plt

# Load sample dataset
tips = sns.load_dataset("tips")

# Create a violin plot
sns.violinplot(x="day", y="total_bill", hue="sex", data=tips, split=True)

plt.title("Seaborn Violin Plot Example")
plt.xlabel("Day of the Week")
plt.ylabel("Total Bill ($)")
plt.tight_layout()
plt.show()
```

## Checklist

- [ x ] I have searched for similar feature requests and confirmed that this is a new request.
- [ x ] I have provided a clear and concise description of the feature.
- [ x ] I have explained the motivation behind this feature request.
- [ x ] I have outlined a proposed solution or ideas for implementing this feature.
- [ x ] I have provided any additional context or screenshots if applicable.

jooyoungseo:
Done. Measured on the current `main`:

```
sns.violinplot   layers=['violin_box', 'violin_kde']   renders with a payload
```

**Two** layers rather than one, because a violin is both: the box summarises and the KDE shows the shape, and a reader moves between them. #398 settled how the box statistics are carried (they were being rounded to four decimals, which flattened small data), and the boxplot/violin binding work settled reading the caller's spelling rather than re-parsing it.

Closed along with the rest of #255, where the whole list is measured together.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #613, opened 2026-08-22
fix(hue): name an lmplot's hue groups from the legend that names their colours

### LINKED ISSUE #612: An lmplot's hue groups are split correctly and named nothing
Found while sweeping the seaborn coverage issues in #255 against measurements. Most of them read; this is a gap.

## What happens

`sns.lmplot(data=..., x=..., y=..., hue=...)` splits into one scatter and one fitted curve **per hue level** — the data proves the split is right — and names none of the four layers.

Measured on seaborn 0.13.2, two levels `p` and `q` of thirty rows each:

```
lmplot(x="x", y="y", hue="g")
    point   name=None  n=30   first x=0.126
    smooth  name=None  n=1
    point   name=None  n=30   first x=-1.010     <- a different thirty
    smooth  name=None  n=1

scatterplot(x="x", y="y", hue="g")   # the control
    point   name='p'   n=30
    point   name='q'   n=30
```

Each pair really is one group: the two point layers hold disjoint halves of the frame. So a reader is handed four layers over one axes — point, curve, point, curve — with nothing saying which pair is `p` and which is `q`, while the same data through `scatterplot` names both. That is the defect #558 named, on the regression path.

The name is available. The figure legend resolves from the panel:

```
ax0  own legend=False   figure legends=1   legend_of -> Legend
```

## Why

`maidr/patch/regplot.py` never passes `GROUP_NAME`. `lmplot(hue=...)` is implemented as one `regplot` call per level, so each call registers its scatter and its curve anonymously:

```python
FigureManager.create_maidr(axes, PlotType.SCATTER, **{DRAWN_POINTS: points})
...
_register_curves(axes, described, instance, args, kwargs)
```

Both destinations already honour the key. `SmoothPlot.__init__` reads it through `group_name_of` and `render()` emits `MaidrKey.NAME`; `ScatterPlot` takes it the same way, which is what names the control above.

## Shape of the fix

This is #595's shape — "wrong where a *layer* is the artist" — so `name_for(ax, colour)` rather than `names_for`: each `regplot` call draws one group, and the colour of its collection and its curve is what says which level.

It has to be **deferred**. `lmplot` builds its one figure legend through `FacetGrid.add_legend()` after every panel is drawn, so at registration there is no legend anywhere — the same timing #561 hit with `pairplot`, and the reason `GROUP_NAME` accepts a callable.

## What must not change

- A bare `regplot` with no hue draws one group against no legend, so `name_for` returns `None` and the layer reads exactly as it does today.
- `lmplot` without a hue is the same: a `FacetGrid` with no legend.
- `regplot(x_estimator=...)` takes the `ERRORBAR` branch, which is a different registration and out of scope here.

## The rest of the #255 sweep, for the record

Measured so the issues can be closed against evidence rather than assumption. Layers are `(type, name)`.

| issue | chart | reading |
|---|---|---|
| #247 | `residplot` | `point` |
| #248 | `lmplot` | `point` + `smooth` |
| #249 | `ecdfplot` | `step` |
| #249 | `ecdfplot(hue=)` | one `step` holding **two series**, each named by `z` — the multi-series convention #582 settled, not a fold |
| #251 | `swarmplot` | one `point` per category |
| #252 | `stripplot` | one `point` per category |
| #254 | `violinplot` | `violin_box` + `violin_kde` |
| #254 | `violinplot(hue=, split=True)` | `violin_box` + `violin_kde` |

jooyoungseo:
Closes #612.

## What was wrong

`lmplot(hue=...)` is one `regplot` call per level, and each call registers a scatter and a fitted curve. The **split was right** — the two point layers hold disjoint halves of the frame — and all four layers came out anonymous.

Measured on seaborn 0.13.2, two levels of thirty rows each:

| chart | before | after |
|---|---|---|
| `lmplot(x, y, hue=)` | `point None`, `smooth None`, `point None`, `smooth None` | `point p`, `smooth p`, `point q`, `smooth q` |
| `scatterplot(x, y, hue=)` — control | `point p`, `point q` | unchanged |

A reader was handed point, curve, point, curve over one axes with nothing saying which pair was which, while the same data through `scatterplot` named both. That is the defect #558 named, on the regression path.

The name was there for the asking:

```
ax0  own legend=False   figure legends=1   legend_of -> Legend
```

## Two things were needed, and neither works alone

**The colour has to name one _layer_, not one artist among several.** This is #595's shape — "wrong where a *layer* is the artist" — so `name_for` rather than `names_for`: a `regplot` call draws one group, and the colour of its curve and its collection is what says which level.

The curve is asked first: a `Line2D` carries exactly one colour, so there is nothing to resolve. The scatter's collection answers for `fit_reg=False`, which draws points and no curve at all — `test_a_hue_split_lmplot_drawn_without_its_fit_is_still_named` covers that branch, and dropping it leaves that chart exactly as it was.

**The match has to be deferred.** `FacetGrid.add_legend()` runs after every panel is drawn, so at registration there is no legend anywhere — the timing #561 hit with `pairplot`, and the reason `GROUP_NAME` accepts a callable. Resolving eagerly puts every layer back to `None`; the mutation table below shows it.

## `ScatterPlot` reads `GROUP_NAME`

It took its name only from `HUE_GROUP`, which also carries *which points* belong to the group, because the scatter patch splits one collection between levels. A `regplot`'s scatter is its level entire, so there is nothing to filter and only the name is in question — which is exactly what `GROUP_NAME` is for, and what `SmoothPlot`, `HistPlot` and the others already read. `render` resolves the callable there, as `SmoothPlot` already did.

## What is unchanged

- A bare `regplot`: one group, no legend, `name_for` declines. `test_a_regplot_without_a_hue_is_unnamed`.
- `lmplot` without a hue: a `FacetGrid` with no legend to build. `test_an_lmplot_without_a_hue_is_unnamed`.
- `regplot(x_estimator=...)` takes the `ERRORBAR` branch, which is a different registration and untouched.

## Found while sweeping #255

The other seaborn coverage issues were measured at the same time and read correctly today, recorded in #612 so they can be closed against evidence: `residplot` (#247) → `point`; `ecdfplot` (#249) → one `step` holding two **series** each named by `z`, which is the multi-series convention #582 settled rather than a fold; `swarmplot`/`stripplot` (#251/#252) → one `point` per category; `violinplot` (#254), including `split=True` → `violin_box` + `violin_kde`.

## Verification

- `tests/core/plot/test_lmplot_group_names.py` → 6 passed (new file)
- Full suite: **2945 passed, 18 skipped**
- `ruff check` clean on all three touched files

Mutations, each applied alone against a restored baseline — all six killed:

| mutation | result |
|---|---|
| name resolved eagerly, not deferred | 4 failed |
| scatter registration drops the name | 4 failed |
| curve registration drops the name | 2 failed |
| colour taken only from the curve (no `fit_reg=False` fallback) | 1 failed |
| `ScatterPlot` back to `HUE_GROUP` only | 4 failed |
| `ScatterPlot.render` does not resolve a callable | 6 failed |

---

### DIFF
--- a/maidr/core/plot/errorbar.py
+++ b/maidr/core/plot/errorbar.py
@@ -11,9 +11,9 @@ from matplotlib.container import ErrorbarContainer
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
+from maidr.core.plot.maidr_plot import group_name_of
 from maidr.exception import ExtractionError
 from maidr.util.mixin import DictMergerMixin
-
 
 
 def _is_drawn(value: object) -> bool:
@@ -83,6 +83,14 @@ class ErrorBarPlot(MaidrPlot, DictMergerMixin):
     """
 
     def __init__(self, ax: Axes, **kwargs) -> None:
+        # The hue group these estimates belong to, when the patch could name
+        # it. `lmplot(x_estimator=..., hue=...)` draws one estimate-and-band
+        # per level and the curve beside it is already named, so leaving this
+        # unnamed announced a group and its own uncertainty as unrelated
+        # layers (#612). Opted into here rather than read by `MaidrPlot`; see
+        # `GROUP_NAME` for why that is per class.
+        self._group_name = group_name_of(kwargs)
+
         super().__init__(ax, PlotType.ERRORBAR)
 
         # The patch hands over the exact container its call produced. Looking
@@ -101,6 +109,12 @@ class ErrorBarPlot(MaidrPlot, DictMergerMixin):
         """
         Build the layer schema, adding the orientation the bars were drawn in.
 
+        Adds the group's name too, when the chart has one -- the same field,
+        and for the same reason, as ``SmoothPlot``: a hue-split ``lmplot``
+        draws one band per level, and two of them over one axes with nothing
+        to tell them apart leave a reader hearing the identical announcement
+        twice.
+
         Returns
         -------
         dict
@@ -110,9 +124,17 @@ class ErrorBarPlot(MaidrPlot, DictMergerMixin):
         # `self._orientation` -- so the read below has to come after it, not
         # alongside it.
         base_schema = super().render()
-        return self.merge_dict(
-            base_schema, {MaidrKey.ORIENTATION: self._orientation}
-        )
+        added: dict = {MaidrKey.ORIENTATION: self._orientation}
+
+        # A callable is resolved here rather than at registration, which is
+        # what an `lmplot` needs: `FacetGrid.add_legend()` runs after every
+        # panel is drawn, so the legend that names the colours does not exist
+        # when the layer registers (#561, #612).
+        name = self._group_name() if callable(self._group_name) else self._group_name
+        if name:
+            added[MaidrKey.NAME] = name
+
+        return self.merge_dict(base_schema, added)
 
     def _extract_plot_data(self) -> list[dict]:
         container = self._resolve_container()
--- a/maidr/core/plot/maidr_plot.py
+++ b/maidr/core/plot/maidr_plot.py
@@ -24,13 +24,19 @@ import uuid
 #: ``super().__init__(ax, PlotType.X)`` without forwarding their keyword
 #: arguments, so a key read here would arrive for some layer types and be
 #: swallowed by the rest -- a promise kept by accident of how each
-#: constructor happens to be written. ``HistPlot``, ``SmoothPlot`` and
-#: ``BoxPlot`` opt in today, each in one visible line --
-#: :func:`group_name_of`.
+#: constructor happens to be written. ``HistPlot``, ``SmoothPlot``,
+#: ``BoxPlot``, ``ScatterPlot`` and ``ErrorBarPlot`` opt in today, each in one
+#: visible line -- :func:`group_name_of`.
 #:
-#: ``ScatterPlot`` and ``EventPlot`` name their layers by other means -- a hue
-#: split and row labels -- and are unaffected, being passed neither this key
-#: nor anything that collides with it.
+#: ``ScatterPlot`` reads it **second**, after its own ``HUE_GROUP``: that key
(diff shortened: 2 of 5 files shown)


jooyoungseo:
Both addressed in 630392f.

**The stale docstring** was the right thing to catch — that comment is the source of truth for which class honours which key, and this branch made it say the opposite of what the code does. It now lists `ScatterPlot` among the opt-ins and says why it reads `HUE_GROUP` first: that key carries which *points* belong to a group, and `GROUP_NAME` carries only the name.

**The `x_estimator` end state** is real and reachable, and I've fixed it rather than shipping the inconsistency now that it's been named. Measured on `lmplot(x_estimator=np.mean, hue="g")`:

| | before | after |
|---|---|---|
| `error_bar` | `None` | `p` |
| `smooth` | `p` | `p` |
| `error_bar` | `None` | `q` |
| `smooth` | `q` | `q` |

A group and its own uncertainty announced as unrelated layers is the same reading #451 was about from the other side, where the intervals were their own fitted curves. `ErrorBarPlot` opts into `GROUP_NAME` and resolves the callable in `render` the way `SmoothPlot` does — one line per side, the same shape as everything else here. `test_a_binned_lmplot_names_its_estimates_and_its_fit_alike` pins it, `test_a_binned_regplot_without_a_hue_is_unnamed` pins that the branch stays additive, and three more mutations are killed:

| mutation | result |
|---|---|
| `ErrorBarPlot` back to not reading `GROUP_NAME` | 1 failed |
| errorbar registration drops the name | 1 failed |
| `ErrorBarPlot.render` does not resolve a callable | 2 failed |

**The closure allocation** I've left as is, and it is worth saying why rather than just agreeing: the deferral is not an optimisation that could be skipped when there is no hue, it is the only correct order. At registration a `FacetGrid` has no legend yet, so there is nothing to check that would tell an ungrouped call from a grouped one — asking `legend_of` early would answer `None` for both. The cost is one closure per `regplot` call, resolved once at render.

Full suite after the change: **2947 passed, 18 skipped**.

---

PR #614, opened 2026-08-23
fix(plotly): read a bar chart in the order categoryorder draws it

### LINKED ISSUE #495: A plotly bar chart with categoryorder is read in the data&#39;s order, not the drawn one
The bar half of what #489 fixed for the heatmap, and the Python side of maidr#987 / maidr#989.

## What

`PlotlyBarPlot._extract_plot_data` zips the trace's own arrays:

```python
def _extract_plot_data(self) -> list[dict]:
    x, y = paired_axes(self._trace)
    return [
        {MaidrKey.X: self._to_native(xv), MaidrKey.Y: self._to_native(yv)}
        for xv, yv in zip(x, y)
    ]
```

`categoryorder` sorts the category axis and leaves `x` and `y` exactly as the author wrote them, so the layer is emitted in an order the chart does not show. `PlotlyGroupedBarPlot` and the stacked/normalized paths walk the same arrays and have the same gap.

## Measured

```python
X = ['charlie', 'alpha', 'bravo']
fig = go.Figure(data=[go.Bar(x=X, y=[3, 1, 2])])
fig.update_layout(xaxis={'categoryorder': 'category ascending'})
```

emits

```
[{'x': 'charlie', 'y': 3}, {'x': 'alpha', 'y': 1}, {'x': 'bravo', 'y': 2}]
```

while plotly draws them **alpha, bravo, charlie** left to right. Stacked is the same — every series comes out `charlie, alpha, bravo`:

```json
"type": "stacked_bar",
"data": [[{"x":"charlie","z":"A","y":3},{"x":"alpha","z":"A","y":1},{"x":"bravo","z":"A","y":2}], …],
"selectors": ".subplot.xy .trace.bars .point > path"
```

Every label still carries its own value and the highlight still lands on the right bar, so nothing reads as broken. What is wrong is everything that treats the index as a **position**: which way arrowing travels, the stereo pan, the braille line, the autoplay sweep, and — for a stacked layer — which cells the summary row sums together.

## What the fix looks like here

`_drawn_category_order` in `maidr/plotly/heatmap.py` already resolves this from the declared layout, and its limits carry over unchanged: only `"array"` with a `categoryarray` and the two `"category"` sorts can be answered offline; the aggregate orders (`total`, `sum`, `mean`, `min`, `max`, `median`) need plotly's own aggregation and tie-breaking and are declined. The axis-type guard from #493 applies too — a numeric or ISO-date axis resolves as *linear* or *date* and ignores `categoryorder` outright, so applying a declared order to one would reorder a chart plotly did not.

That helper is worth lifting to `PlotlyPlot` rather than copied: `_axis_runs_backwards`, `_looks_numeric`, `_looks_like_date` and `_drawn_category_order` are all about an axis, not about a heatmap.

## The part that needs a decision

Reordering the data alone would break a highlight that is currently right — plotly draws bars in the trace's own order, so the shared selector resolves in that order too. maidr#988 and maidr#992 solved that on the JS side by narrowing the selector alongside the data: a flat `string[]` for a plain bar, a `string[][]` grid for a segmented one, both permuted the same way.

Emitting those shapes from here couples the schema to a bundle new enough to read them. Measured against an older bundle the failure is a **lost** highlight rather than a wrong one — `Svg.selectAllElements` guards on `typeof query === 'string'` and answers `[]` — so it degrades rather than breaks, and the layer keeps the corrected order. That is the better of the two failures, but it is a real trade and belongs in the PR description rather than being made silently. `warn_if_bundle_cannot_render` checks trace *types*, not selector shapes, so it will not catch this on its own.

## Depends on

- maidr#991 (merged) — a `string[]` reaching a bar layer's highlight at all
- maidr#992 — a `string[][]` reaching a segmented layer's

Both are in the JS core, so a py-maidr fix should land after a maidr release carrying them.

jooyoungseo:
Closes #495 — both halves: the plain `go.Bar` and the grouped/stacked/normalized paths.

## What was wrong

`categoryorder` sorts the category axis and leaves the trace's own `x` and `y` exactly as the author wrote them, so the arrays alone do not say what the chart shows.

```python
go.Bar(x=['charlie', 'alpha', 'bravo'], y=[3, 1, 2])
fig.update_layout(xaxis={'categoryorder': 'category ascending'})
```

| | order |
|---|---|
| emitted (before) | `charlie, alpha, bravo` — the trace's |
| drawn | `alpha, bravo, charlie` — left to right |

Every label still carried its own value and the highlight still landed on the right bar, so nothing read as *broken*. What was wrong is everything that treats the index as a **position**: which way arrowing travels, the stereo pan, the braille line, the autoplay sweep, and — for a stacked layer — which cells the summary row sums together.

## The helpers move rather than get copied

`_drawn_category_order`, `_looks_numeric`, `_looks_like_date` and `_axis_runs_backwards` go from `PlotlyHeatmapPlot` to `PlotlyPlot`. Each is about an *axis* rather than about a heatmap, and #495 says as much. Unchanged in the move, limits included:

- only `"array"` with a `categoryarray` and the two `"category"` sorts are resolved offline; the aggregate orders need plotly's own aggregation and tie-breaking and are declined
- a numeric or ISO-date axis resolves as *linear* or *date* and ignores `categoryorder` outright, so a declared order is not applied to one

`_to_native` is reached through `self`, so the heatmap's own override still applies to a heatmap. The 929 existing plotly tests pass unchanged across the move.

## The selectors move with the data, because they have to

This is the part #495 flagged as a real trade, and I measured it rather than reasoning about it.

**Plain bar** — in Chromium, on the chart above:

```
.point groups, document order:   x = 773, 37, 405      <- charlie, alpha, bravo
chart draws:                     alpha, bravo, charlie
```

So the single selector this layer emitted resolves in the order the points *used to be in*. Reordering the points alone would leave every highlight one place out — a reading that is right and a highlight that is wrong, the blind spot xability/maidr#814 names. `.point:nth-of-type(k)` matches exactly one element, the kth in document order, for every k. Verified end to end against the emitted list:

```
drawn 0  'alpha'    matched 1  left  37
drawn 1  'bravo'    matched 1  left 405
drawn 2  'charlie'  matched 1  left 773
```

**Grouped and stacked** — a **grid**, `selectors[group][category]`, which is the shape `data` already has and what a segmented layer addresses its cells by:

```
grouped   traceGroups 2   perGroup [[767, 37, 402], [913, 183, 548]]
stacked   traceGroups 2   perGroup [[767, 37, 402], [767, 37, 402]]
```

Two sibling `.trace.bars` groups under `g.barlayer.mlayer`, one per trace and in the traces' own order, each holding its categories in that trace's order. `.trace.bars:nth-of-type(t) .point:nth-of-type(c)` matched exactly one element for all six pairs, and the coordinates say which: dodged puts the two traces side by side at one category, stacked one above the other.

**A chart that declares no order, or one that cannot be resolved, keeps the single string it has always had** — so nothing that reads correctly today changes shape.

### The bundle trade, stated rather than made silently

A `string[]` reaches a bar layer's highlight as of xability/maidr#991, and a `string[][]` a segmented layer's as of xability/maidr#992. Against an older bundle `Svg.selectAllElements` guards on `typeof query === 'string'` and answers `[]`, so the highlight is **lost** rather than wrong, while the announced order stays corrected. That is the better of the two failures, and it arises only on a chart that declares an order. `warn_if_bundle_cannot_render` checks trace *types*, not selector shapes, so it will not catch this on its own.

## Two things worth calling out

**One order for every group.** Resolved once from the first trace's category axis: the traces of a group share that axis by construction, which is what makes them one chart rather than several.

**The normalising pass is reordered with the rest.** `stack_shares` matches by category rather than by index, so the two are independent — but reordering one and not the other would put `alpha`'s share on `charlie` while both looked plausible. `test_a_normalized_stack_keeps_each_share_with_its_own_category` pins the pairing rather than the count.

**A group whose traces carry different numbers of categories is declined whole.** Reordering some groups and not others is worse than reordering none, and indexing a shorter trace by the first's positions would raise.

## Verification

- `tests/plotly/test_plotly_bar_category_order.py` → 20 passed (new file)
- `tests/plotly` → 949 passed
- Full suite: **2967 passed, 18 skipped**
- `ruff check` clean across `maidr/plotly/` and `tests/plotly/`

Mutations, each applied alone against a restored baseline — all nine killed:

| mutation | result |
|---|---|
| plain: order resolved but not applied to the data | 10 failed |
| plain: data reordered, selector left a single string | 2 failed |
| plain: category axis always `x` (breaks horizontal) | 2 failed |
| plain: selector indices not permuted (identity) | 2 failed |
| grouped: order resolved, groups not reordered | 3 failed |
| grouped: grid selector left a single string | 1 failed |
| grouped: normalized path skips the reorder | 1 failed |
| grouped: no length guard across traces | 1 failed |
| grouped: trace group not scoped in the grid | 1 failed |


jooyoungseo:
Both reviews landed real findings. Fixed in c11fcba.

## The grouped-path bug is real, and it was mine

The second review is right and the diagnosis is exact: `_category_order` resolved indices into the **first** trace's arrays and `_in_drawn_order` applied them positionally to every group. Reproduced on the suggested case — A written `charlie, alpha, bravo`, B written `alpha, bravo, charlie`, same length, same set:

| | before | after |
|---|---|---|
| group A | `alpha 1, bravo 2, charlie 3` | unchanged |
| group B | `bravo 20, charlie 30, alpha 10` | `alpha 10, bravo 20, charlie 30` |

Every point still carried its own label while the *column* it landed in belonged to another category — this issue's own defect one level up, and it desynced the selector grid alongside it. The point about `stack_shares` deliberately matching by category and my reintroducing the index assumption is the right framing.

Resolved **per trace** now, each against its own labels, and declined whole when two traces come back with different sequences of names — a grid's column has to mean one category in every group. The explicit length guard goes with it: a short trace declines through the same check rather than through a special case, and `test_a_group_whose_traces_disagree_about_length_is_declined` still passes.

The selector grid follows: A permutes (`2, 3, 1`), B is already sorted and does not (`1, 2, 3`). `test_traces_written_in_different_orders_each_get_their_own_permutation` pins both halves, and restoring the one-order-for-the-group spelling turns it red on its own.

## The horizontal y-axis question — measured, no flip needed

The first review was right not to accept the alphabetical test as evidence, so I measured it in Chromium rather than reasoning by analogy with the heatmap:

```
horizontal, unsorted     y = ['charlie', 'alpha', 'bravo']
  tick labels, top to bottom:   bravo, alpha, charlie
  → bottom to top:              charlie, alpha, bravo   = the trace's order

horizontal, category ascending
  tick labels, top to bottom:   charlie, bravo, alpha
  → bottom to top:              alpha, bravo, charlie   = what this branch emits
```

So **bottom-up is the order py-maidr already ships** for a horizontal bar: the trace order it has always emitted for an unsorted chart runs from the bottom lane. A sorted chart matching that is consistent; flipping it would put two spellings of one chart in opposite orders. The heatmap flips because a matrix has a reading order; a bar chart's lanes do not.

Pinned by `test_a_horizontal_bar_is_announced_from_the_bottom_lane_up`, using a declared array (`bravo, charlie, alpha`) that is neither sorted nor reversed — so it catches an inverted mapping where an alphabetical sort cannot.

## The rest

- **Stale docstrings** on the moved helpers: fixed. `_axis_runs_backwards` no longer says "a heatmap's rows" and now records that it is unwired for bars; `_drawn_category_order` drops the heatmap-only aside and carries the bottom-up measurement above.
- **Base `_get_selector` annotation**: widened to `str | list | dict` with a docstring naming the three shapes and who emits each.
- **Invalid escape sequence**: the grid docstring is now a raw string. Good catch — `W605` is not selected here, so nothing would have flagged it.
- **Selector assertions per sort path**: `test_every_resolved_sort_permutes_the_selectors_too` covers descending and the declared array, not only ascending.
- **Reversed bar axis**: confirmed out of scope and now said so in `_axis_runs_backwards` itself rather than left implicit. #495 is `categoryorder`.
- **Release sequencing**: xability/maidr#991 and #992 are both in the JS core — `bar.ts` accepts `string | string[] | (string | null)[][]` and `segmented.ts` branches on arrays — so the shapes this emits are readable by a current bundle. Against an older one the highlight degrades to lost rather than wrong, as stated.

## Verification

- `tests/plotly/test_plotly_bar_category_order.py` → 25 passed (5 new)
- `tests/plotly` → 954 passed
- Full suite: **2972 passed, 18 skipped**

Three further mutations, each applied alone against a restored baseline — all killed:

| mutation | result |
|---|---|
| one order applied to the whole group (the reviewed bug) | 1 failed |
| selector grid built from the first trace's order | 1 failed |
| differing name sequences not declined | 2 failed |

---

PR #616, opened 2026-08-23
feat(seaborn): read the marks of seaborn.objects, which registered nothing

### LINKED ISSUE #615: feat: every `seaborn.objects` chart registers nothing — the modern seaborn interface has no patch behind it
`seaborn.objects` is seaborn's declarative interface, added in 0.12 and the one the documentation's own tutorial now opens with. py-maidr reads none of it. Every mark, on every path, registers nothing at all, and `maidr.render()` raises `UnsupportedPlotError`.

Measured on seaborn 0.13.2 / matplotlib 3.9.4, one `.add()` per row:

```
so.Plot(frame, x=, y=).add(so.Dot())      NOTHING REGISTERED (UnsupportedPlotError)
                      .add(so.Dots())     NOTHING REGISTERED (UnsupportedPlotError)
                      .add(so.Line())     NOTHING REGISTERED (UnsupportedPlotError)
                      .add(so.Bar())      NOTHING REGISTERED (UnsupportedPlotError)
                      .add(so.Area())     NOTHING REGISTERED (UnsupportedPlotError)
   ... and with color= NOTHING REGISTERED (UnsupportedPlotError)
```

The classic spelling of the same three charts reads today:

```
sns.scatterplot(...)   ['point']
sns.lineplot(...)      ['line']
sns.barplot(...)       ['bar']
```

So this is not one unsupported chart type. It is the same charts, written the way seaborn now teaches, going silent.

## Why nothing fires

`maidr/patch/` wraps the *user-facing* drawing calls — `Axes.scatter`, `Axes.plot`, `Axes.bar`, and seaborn's `_CategoricalPlotter` / `_DistributionPlotter` methods. A `Mark` does not call any of them. It draws through the **artist** API instead: `ax.add_collection`, `ax.add_line`, `ax.add_patch`. None of those is a patched name, so the whole interface passes through untouched.

It is a second front door with no patch behind it, which is why the failure is total rather than per plot type.

What each mark leaves on the axes, measured:

| mark | artist |
|---|---|
| `Dot`, `Dots` | `PathCollection` |
| `Line`, `Path` | `Line2D` |
| `Lines`, `Paths`, `Dash`, `Range` | `LineCollection` |
| `Bar` | `Rectangle` ×n **plus a `BarContainer`** |
| `Bars` | `PatchCollection` |
| `Area`, `Band` | `Polygon` |
| `Text` | `Text` ×n |

Three of those the existing plot classes can already read unchanged: `ScatterPlot` takes a collection through `DRAWN_POINTS`, `LinePlot` takes `lines=`, and `BarPlot` takes containers through `DRAWN_BARS`. So `Dot` / `Dots` / `Line` / `Path` / `Bar` need no new extraction at all — only a patch that hands the artist over.

## Where the hook goes

`Plotter._plot_layer(p, layer)` runs once per `.add()` and draws that layer across every subplot. Wrapping it and diffing each axes' artists across the call names exactly which artists a layer drew, per axes, without having to predict how many a mark makes — which varies with grouping, since `so.Line(color=)` draws one `Line2D` per level.

`Plot.plot()` is the wrong hook: from there the layer→artist mapping has to be reconstructed from add order, and that is guesswork the diff does not need.

## Reaching the figure

Worth recording, because it shapes what a user has to write:

```
so.Plot(...).plot()               plt.get_fignums() == []     # no pyplot registration
so.Plot(...).plot(pyplot=True)    plt.get_fignums() == [1]    # plt.gcf() is it
so.Plot(...).on(fig).plot()       the caller already holds it
so.Plot(...)._repr_png_()         plt.get_fignums() == []
```

`.on(fig)` and `plot(pyplot=True)` both hand back a figure `maidr.render()` accepts today, so the registration half is useful on its own. Letting `maidr.render()` take a `so.Plot` directly is a separate and smaller question — it belongs in a follow-up rather than folded in here.

## Suggested order

1. `Dot`, `Dots` → `point`; `Line`, `Path` → `line`; `Bar` → `bar`. No new extraction, three existing classes, and it covers what the tutorial opens with.
2. `Bars` (`PatchCollection`) and `Lines` / `Paths` (`LineCollection`) — the same charts drawn as one collection rather than n artists.
3. `Area` / `Band` → `area`, `Range` → `error_bar`, `Text` → the labelled scatter of xability/maidr#1106.

Marks left unclaimed must stay unclaimed rather than be guessed at, and a chart that mixes a claimed layer with an unclaimed one should still read the claimed one — the whole-chart-to-a-picture failure of xability/r-maidr#225 is what to avoid here.

jooyoungseo:
Closes #615.

`seaborn.objects` is seaborn's declarative interface, and py-maidr read none of it — not one plot type missing, the whole front door. Measured before, each row one `.add()`:

```
so.Plot(frame, x=, y=).add(so.Dot())    NOTHING REGISTERED
                      .add(so.Line())   NOTHING REGISTERED
                      .add(so.Bar())    NOTHING REGISTERED
sns.scatterplot(...)                    ['point']
sns.lineplot(...)                       ['line']
sns.barplot(...)                        ['bar']
```

The same charts, written the way seaborn now teaches, going silent.

## Why nothing fired

`maidr/patch/` wraps the *user-facing* drawing calls — `Axes.scatter`, `Axes.plot`, `Axes.bar`, and seaborn's plotter methods. A `Mark` calls none of them. It draws through the **artist** API instead: `ax.add_collection`, `ax.add_line`, `ax.add_patch`. None of those is a patched name.

## Where the hook goes

`Plotter._plot_layer` runs once per `.add()` and draws that layer across every panel, which makes it the one place where "which artists belong to which layer" is still answerable. Taking each axes' artists **before and after** answers it exactly, without predicting how many a mark makes — `so.Line(color=)` draws one `Line2D` per level, and a faceted layer draws on some panels and not others.

A panel a layer never drew on registers nothing, so a `col`/`row` grid's empty combinations do not become layers a reader can walk into and find nothing in (#421). That falls straight out of the diff rather than needing its own rule.

## What is read

| mark | artist | as | handed over through |
|---|---|---|---|
| `Dot`, `Dots` | `PathCollection` | `point` | `DRAWN_POINTS` |
| `Line`, `Path` | `Line2D` | `line` | `lines=` |
| `Bar` | `BarContainer` | `bar` | `DRAWN_BARS` |

No new extraction: three existing plot classes already take their artists this way, so a `so.Bar` reaches the same `BarPlot` as `Axes.bar` and its orientation comes off the container seaborn built. Measured, horizontal reads magnitude-on-x / label-on-y exactly as `Axes.barh` does, and every emitted selector resolves in the page it was built from — same shapes the classic spelling produces.

`Bars`, `Lines`, `Paths`, `Area`, `Band`, `Range`, `Dash` and `Text` are left for the follow-ups #615 lists.

## Matched by name, not by ancestry

seaborn's mark hierarchy does not track what a mark draws:

```
Line  < Path  < Mark      Line2D
Dash  < Paths < Mark      LineCollection
Range < Paths < Mark      LineCollection
```

So dispatching on ancestry would claim `Dash` and `Range` as lines and hand a `LineCollection` to `LinePlot`, which cannot see it. A name misses a mark seaborn renames; ancestry would claim one that cannot be read — and a wrong reading is worse than a missing one. The test pins the hierarchy itself, so the release that changes it fails there.

## Additive by construction

A mark outside the table is drawn and left **entirely** alone — not drawn inside the internal context — so it registers exactly what it registered before: nothing. Pinned for `Area`, `Bars`, `Lines`, `Paths`, `Dash`, `Text` and `Range`, and a chart mixing a read layer with an unread one still reads the read one (the whole-chart-to-a-picture failure of xability/r-maidr#225, from the side where declining is right).

## The wrap is guarded, unlike the other seaborn patches

`seaborn._core.plot.Plotter._plot_layer` is a private method of a private module. `maidr/patch/_seaborn_version.py` states a floor for `_CategoricalPlotter` / `_DistributionPlotter`, and there is no such floor to state here — so a rename is turned into a warning naming what stops reading, rather than taking `import maidr` down and breaking every *classic* seaborn chart over a mark nobody in that process drew.

## Verification

`uv run pytest` — **3002 passed, 18 skipped**, up from 2996; `ruff check --diff` clean; `uv lock --check` clean.

30 new tests. Mutation sweep, each change applied alone against a restored baseline:

| mutation | outcome |
|---|---|
| drop the empty-panel guard | killed (1) |
| `Bar` not `singular` | killed (4) |
| ignore the before-snapshot | killed (1) |
| claim every mark | killed (5) |
| read only the first panel | killed (2) |
| drop the artist-class filter | killed (1) |

Two of those — *claim every mark* and *drop the artist-class filter* — **survived the first sweep**, because each is masked by the other: every unread mark measured draws into a holder no reading reads, *or* draws a class the filter removes. Dropping both together would read a `so.Dash`'s `LineCollection` through `ScatterPlot`, which silently falls back to sweeping the whole axes and announces a neighbouring layer's points as this one's. They are now asked of `_reading_for` and `_held` directly, with the masking written down in the test file, so removing either is a failing test rather than a silent widening.

---

### DIFF
--- a/maidr/patch/__init__.py
+++ b/maidr/patch/__init__.py
@@ -38,5 +38,6 @@ from . import (  # noqa: E402, F401
     candlestick,
     mplfinance,
     violinplot,
+    seaborn_objects,
     seaborn_probe,
 )
--- a/maidr/patch/seaborn_objects.py
+++ b/maidr/patch/seaborn_objects.py
@@ -245,7 +245,7 @@ def _wrap() -> None:
     except ImportError:  # pragma: no cover - seaborn without the objects API
         Plotter = None  # type: ignore[assignment]
 
-    if Plotter is None or not hasattr(Plotter, "_plot_layer"):  # pragma: no cover
+    if Plotter is None or not hasattr(Plotter, "_plot_layer"):
         warnings.warn(
             "maidr: seaborn._core.plot.Plotter._plot_layer is not there to "
             "wrap, so seaborn.objects charts are not read. Every mark -- "
--- a/maidr/patch/seaborn_objects.py
+++ b/maidr/patch/seaborn_objects.py
@@ -1,0 +1,263 @@
+"""Read the marks of ``seaborn.objects``, seaborn's declarative interface."""
+
+from __future__ import annotations
+
+import warnings
+from typing import Any, NamedTuple
+
+import wrapt
+from matplotlib.axes import Axes
+from matplotlib.collections import PathCollection
+from matplotlib.container import BarContainer
+from matplotlib.lines import Line2D
+
+from maidr.core.context_manager import ContextManager
+from maidr.core.enum import PlotType
+from maidr.core.figure_manager import FigureManager
+from maidr.core.plot.barplot import DRAWN_BARS
+from maidr.core.plot.scatterplot import DRAWN_POINTS
+from maidr.patch.common import _draw_quietly
+
+
+class _Reading(NamedTuple):
+    """
+    How one mark is read: what it is, where it lands, and how it is handed on.
+
+    Attributes
+    ----------
+    plot_type : PlotType
+        What the layer is registered as.
+    holder : str
+        The ``Axes`` attribute the mark's artists arrive in -- ``collections``,
+        ``lines`` or ``containers``. Read by name because that is what makes
+        the before-and-after diff possible without knowing how many artists a
+        mark makes.
+    artist : type
+        The artist class to keep. The holder is filtered by it rather than
+        taken whole, so a layer that lands beside an unrelated artist of
+        another class in the same list describes only its own.
+    binding : str
+        The keyword the artists are handed over under, which is the existing
+        plot class' own: nothing new is extracted for any of these marks.
+    singular : bool
+        Whether that keyword names **one** artist rather than the list.
+        ``DRAWN_BARS`` does, because ``Axes.bar`` draws exactly one container
+        and ``BarPlot`` wraps what it is given in a list of one. So a reading
+        that finds several becomes several layers rather than one truncated
+        to the first -- which is also how a hue-grouped bar is already read
+        elsewhere (#593, #595), one layer per group. Measured, no ``so.Bar``
+        spelling reaches that branch: plain, horizontal, ``color=``,
+        ``Dodge()``, ``Stack()``, ``Hist()`` and faceted are each a single
+        container of n bars.
+    """
+
+    plot_type: PlotType
+    holder: str
+    artist: type
+    binding: str
+    singular: bool = False
+
+
+#: The marks read today, keyed by class name.
+#:
+#: By **name** rather than by ``isinstance``, and that is load-bearing here
+#: rather than a style choice. seaborn's mark hierarchy does not track what a
+#: mark draws::
+#:
+#:     Line  < Path  < Mark      Line2D
+#:     Dash  < Paths < Mark      LineCollection
+#:     Range < Paths < Mark      LineCollection
+#:
+#: ``Dash`` and ``Range`` are ``Paths`` subclasses that draw a
+#: ``LineCollection``, so dispatching on ancestry would claim two marks whose
+#: artists this cannot read -- the shape xability/r-maidr#225 hit from the
+#: other side, where ``class(geom)[1]`` declined a subclass that *was*
+#: readable. A name misses a mark seaborn renames; ancestry would claim one it
+#: cannot read, and a wrong reading is worse than a missing one.
+#:
+#: A caller's own ``Mark`` subclass is declined for the same reason: it is not
+#: in this table, and what it draws is not knowable from the class it came
+#: from.
+_READINGS: dict[str, _Reading] = {
+    "Dot": _Reading(PlotType.SCATTER, "collections", PathCollection, DRAWN_POINTS),
+    "Dots": _Reading(PlotType.SCATTER, "collections", PathCollection, DRAWN_POINTS),
+    "Line": _Reading(PlotType.LINE, "lines", Line2D, "lines"),
+    "Path": _Reading(PlotType.LINE, "lines", Line2D, "lines"),
+    "Bar": _Reading(PlotType.BAR, "containers", BarContainer, DRAWN_BARS, True),
+}
+
+
+def _reading_for(layer: Any) -> _Reading | None:
+    """
+    How to read one layer, or ``None`` when its mark is not one of these.
+
+    Parameters
+    ----------
+    layer : Any
+        The layer ``Plotter._plot_layer`` was handed. A ``TypedDict`` at type
+        level and a plain dict at runtime; anything else declines.
+
+    Returns
+    -------
+    _Reading or None
+        ``None`` leaves the layer exactly as it was, which is unregistered.
+    """
+    if not isinstance(layer, dict):
+        return None
+    return _READINGS.get(type(layer.get("mark")).__name__)
+
+
+def _held(ax: Axes, reading: _Reading) -> list:
+    """
+    The artists of this reading's kind currently on one axes.
+
+    Parameters
+    ----------
+    ax : Axes
+        The panel to look at.
+    reading : _Reading
+        Which holder to read and which class to keep.
+
+    Returns
+    -------
+    list
+        In the order the axes holds them, which is the order they were drawn.
+    """
+    return [
+        artist
+        for artist in getattr(ax, reading.holder, [])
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Three of the four taken, in `102b72c`. The fourth is filed as #617.

**The wrap guard is now tested.** It was the one thing in the file marked `# pragma: no cover` and, as the review says, the branch the whole "guarded, unlike the other seaborn patches" argument rests on. `test_a_seaborn_that_moved_the_hook_warns_rather_than_failing_the_import` monkeypatches `Plotter` to a class without `_plot_layer`, asserts the `UserWarning` fires, and asserts nothing got wrapped — the pragma is gone.

**The call convention is pinned.** `test_the_hook_still_takes_the_layer_as_its_second_argument` asserts `inspect.signature(Plotter._plot_layer).parameters` is `["self", "p", "layer"]`. The review's reasoning for why this matters is exactly right and worth recording: a reorder degrades to `layer is None` → `_reading_for` declines → every chart registers nothing, which is *also* what an unnamed mark does, so nothing else in the file would go red.

**The container count is asserted rather than measured.** `test_a_colour_split_bar_draws_one_container_and_reads_as_one_layer` is parametrized over plain / `Dodge()` / `Stack()` and asserts `len(ax.containers) == 1` and one `bar` layer. This is worth stating precisely, because it corrects something in the review's premise as well:

```
so.Plot(..., color="g").add(so.Bar())                1 BarContainer
                       .add(so.Bar(), so.Dodge())    1 BarContainer
                       .add(so.Bar(), so.Stack())    1 BarContainer
```

A `color=` split does **not** produce several containers — unlike `seaborn.barplot(hue=)`, which draws one per level. So the `singular=True` branch is real but unreached today, and the test now says which of the two is true rather than leaving it to the description.

**On the hue gap — #617.** The same correction applies, and it makes the issue sharper rather than softer. `so.Dot(color=)` draws **one** `PathCollection` carrying a colour per point, not several:

```
so.Plot(..., color="g").add(so.Dot())     1 PathCollection   point  name=None  n=4
                       .add(so.Line())    2 Line2D           line   name=None  2 series
```

That single-collection shape is exactly what `seaborn.scatterplot(hue=)` produces and what `hue_groups()` already inverts, so the fix is likely wiring rather than new extraction. And the naming half has a clear route: `so.Plot` hangs its legend on the **figure** (`ax.get_legend() is None`, `len(fig.legends) == 1`), which is the second tier `legend_of` already reads for `PairGrid` (#561). #617 records all of that, plus a third item the review did not reach — `so.Stack()` and `so.Norm()` read as a plain `bar` rather than `stacked_bar` / `normalized_bar`, and `so` states the position transform explicitly on `layer["move"]`, which is a cleaner signal than the classic path gets.

Full suite after the change: **3007 passed, 18 skipped**.

---

jooyoungseo:
Both of the second review's gaps are closed in `a91ca07`.

**Faceting now covers each mark.** The point the review makes is right and worth restating: the panel diff is shared, but what each mark draws into it is not. `Line` draws one `Line2D` per group and `Bar` one container per panel, so a facet case for those two is the only place those handovers meet the per-panel branch — not a repeat of the `Dot` one. `test_every_mark_reads_one_layer_per_panel_when_faceted` is parametrized over all three.

**The several-containers branch is now reached.** This was the last "trust me" in the file: `singular=True` turning several containers into several layers is unreachable through any `so.Bar` spelling measured, so it was forward-looking correctness with nothing holding it — and, as the review says, it would be exercised for the first time *silently* by a seaborn release that changed how `so.Bar` draws.

It is driven through `_layer` directly, because a chart cannot reach it:

```python
def draw(*args, **kwargs):
    axes.bar(["a"], [4.0])
    axes.bar(["b"], [9.0])

_layer(draw, _Plotter(), (None, {"mark": so.Bar()}), {})
assert len(axes.containers) == 2
assert _kinds(figure) == ["bar", "bar"]
```

The two `Axes.bar` calls run inside the internal context `_layer` already sets, so they register nothing of their own and what comes back is this layer's reading alone. Verified as discriminating rather than assumed — truncating the handover to `own[0]` turns exactly this test red:

```
M8-truncate-to-first    1 failed, 38 passed
```

Full suite: **3011 passed, 18 skipped**.

On the `ImportError` branch staying `# pragma: no cover` — agreed with the review's reasoning, and recording it so it is a decision rather than an oversight: `seaborn._core.plot` has existed since 0.12 and `_seaborn_version.py` already floors seaborn at 0.13, so that branch cannot be reached by any seaborn this package will import. The `hasattr` branch beside it *can* be — a rename inside a module that exists — which is why that one is tested and this one is not.

---

Issue #617, opened 2026-08-23
A `seaborn.objects` mark split by `color=` reads as one unnamed layer


jooyoungseo:
Follow-up from the review on #616, which reads `so.Dot` / `so.Dots` / `so.Line` / `so.Path` / `so.Bar`. Those charts now register, and a `color=` split within one `.add()` is not told apart.

Measured on seaborn 0.13.2, four rows over two levels `p` and `q`:

```
so.Plot(frame, x=, y=, color="g").add(so.Dot())              1 PathCollection
                                                             point  name=None  n=4
                                    .add(so.Bar())           1 BarContainer
                                                             bar    name=None  n=4
                                    .add(so.Bar(), so.Dodge())    bar name=None n=4
                                    .add(so.Bar(), so.Stack())    bar name=None n=4

so.Plot(frame, x=, y=, color="g").add(so.Line())             2 Line2D
                                                             line   name=None  2 series
```

Not a regression — none of these registered at all before #616 — but it is the shape #558, #593, #595 and #612 each fixed on the classic side, arriving unfixed on the new one.

## Two separate things

**The name.** `so.Line(color=)` already splits correctly into a series per level; what it lacks is the level's *name*, so a reader gets two anonymous series. The machinery is close at hand: `so.Plot` hangs its legend on the **figure**, and `legend_of` already reads exactly one figure legend as its second tier (the `PairGrid` path of #561). Measured, a `color=`-split `so.Plot` has `ax.get_legend() is None` and `len(fig.legends) == 1`.

**The split.** `so.Dot(color=)` draws **one** `PathCollection` carrying a colour per point, exactly like `seaborn.scatterplot(hue=)` — which is what `hue_groups()` in `maidr/core/plot/scatterplot.py` already inverts, reading the colours against the legend that names them. So this half may be mostly a matter of wiring `HUE_GROUP` through, rather than new extraction.

A `so.Bar(color=)` is different again: it draws one container for every level together, where `seaborn.barplot(hue=)` draws one per level. So the classic reading's assumption — a container is a group — does not transfer, and the split would have to come from the bars' colours.

## And a third thing, separate again

[...]

PR #619, opened 2026-08-23
fix(seaborn): split a colour-grouped seaborn.objects bar, and get its categories back

jooyoungseo:
Part of #617, following #618 which did the scatter.

Measuring the bar half turned up something worse than the issue recorded. A colour-split `so.Bar` does not read as "a plain `bar`" — it reads as a bar chart **whose categories are coordinates**. Measured on two categories and two levels:

```
                    bars   ticks        announced x
color= (no move)      4    ['a','b']    ['0', '0', '1', '1']
color= + Stack()      4    ['a','b']    ['0', '0', '1', '1']
color= + Dodge()      4    ['a','b']    ['-0.2', '0.2', '0.8', '1.2']
```

The `Dodge()` row is the clearest: a reader is told the categories are `-0.2`, `0.2`, `0.8`, `1.2`. Those are the dodge offsets.

After:

```
color= + Dodge()      bar 'p'  [('a', 1.0), ('b', 3.0)]
                      bar 'q'  [('a', 2.0), ('b', 4.0)]
```

## Why it happened

`BarPlot._labels_for` announces bar **positions** whenever the tick labels do not number the bars. That rule is right, and #382 is why it exists — a numeric x has a locator that picks its own breaks, and pairing three bars with five ticks used to raise and take the whole figure's HTML with it.

A colour split walks straight into it, because every level's bars land on one axes against one tick per *category*. `seaborn.barplot(hue=)` never reaches it: it draws one container **per level**, so each layer already holds one bar per tick.

So the split is what fixes the labels, and the two bar items on #617 turn out to be one piece of work.

## What changed

**`bar_groups()`** does for a `BarContainer` what `hue_groups()` does for a `PathCollection` — one artist carries every level, so the grouping survives only in the bars' colours and the legend naming them.

**Both now end in one shared `groups_from_colours()`**, the part that is about the legend rather than about the artist. Three implementations of that rule had already drifted once (#599); this is where the second caller arrived, so it was extracted rather than copied.

**The handover is a synthetic `BarContainer` per group**, not a filter inside `BarPlot`. The patches are the ones on the axes, so the selectors resolve unchanged and the two layers address their own bars — the collision review caught on xability/r-maidr#226, asserted here.

**`BarPlot` opts into `GROUP_NAME`**, following `HistPlot`, `SmoothPlot`, `BoxPlot`, `ScatterPlot` and `ErrorBarPlot`.

## Still open, deliberately

What a colour-split bar should be **typed** as. All three spellings read as `bar`, where the classic path answers `dodged_bar`. `layer["move"]` states the transform explicitly — cleaner than anything the classic side gets — and that is the remaining item on #617.

## Verification

`uv run pytest` — **3031 passed, 18 skipped**, up from 3021. `ruff check --diff` clean.

8 new tests. One existing test changed its **premise** rather than its expectation: `test_a_colour_split_bar_draws_one_container_and_reads_as_one_layer` asserted the old single-layer reading, and now asserts the split while keeping the container-count claim that was its real content (that count is why `DRAWN_BARS`' singular branch stays unreached).

| mutation | outcome |
|---|---|
| `_handovers` skips the bar split | killed (7) |
| `BarPlot` stops reading `GROUP_NAME` | killed (4) |
| `render` stops emitting the name | killed (4) |
| the synthetic container takes every patch, not the group's | killed (1) |

---

### DIFF
--- a/maidr/core/plot/barplot.py
+++ b/maidr/core/plot/barplot.py
@@ -8,6 +8,7 @@ from matplotlib.patches import Rectangle
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
+from maidr.core.plot.maidr_plot import group_name_of
 from maidr.exception import ExtractionError
 from maidr.util.mixin import (
     BarPositionMixin,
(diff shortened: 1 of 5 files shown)


jooyoungseo:
All three taken, in `7efc77f`. The third turned out not to be polish.

**3 — coverage.** The horizontal case was a live hole. Dropping `orientation=own[0].orientation` from the synthetic container and running the whole file passed 20/20 before this commit; with the new test it fails, alone:

```
M1 drop orientation:        1 failed, 21 passed
M2 reverse member order:    3 failed, 19 passed
restored:                  22 passed
```

A split horizontal bar would have defaulted to vertical and put its categories in the magnitude field — the reading #950 exists to warn about — and nothing would have caught it. The three-group case is the weaker of the two but still earns its place: with two levels the bars pair up, so an ordering bug and a correct ordering can coincide; with three (`p q r p q r`) each level's bars are non-contiguous, which is what M2 kills.

**2 — docstring drift.** Moved the two declines to `groups_from_colours`, where they now run, and left `hue_groups` with the artist half — reading colours off a `PathCollection` — pointing at the shared function for every reason to decline. Also dropped a sentence that had ended up in both docstrings after the extraction.

**1 — unused parameter.** Gone; every call site was `_bars(None)`.

`uv run pytest` — **3033 passed, 18 skipped**. `ruff check --diff maidr/ tests/` clean.

On the typing question: agreed that it is #617's, and `layer["move"]` is where the answer comes from. That is next.

---

Issue #620, opened 2026-08-23
A 100% stacked bar drawn with `so.Norm(func="sum", by=["x"])` reads as a plain stack


jooyoungseo:
Found while doing #617's third item, which reads `so.Dodge()` and `so.Stack()` off `layer["move"]`. `so.Norm()` was left out deliberately, and this records what it would take.

## `so.Norm()` is not the 100% stack transform

The obvious move — put `Norm` beside `Stack` and `Dodge` and call it `stacked_normalized_bar` — is wrong, because `so.Norm()`'s default is `func="max"` with no `by`, which rescales *within each colour group*, not within each category. Measured on two categories over two levels, `(x, y0, height)` per bar:

```
Stack()                     (-0.4, 0.0, 1.0)  (-0.4, 1.0, 2.0)  (0.6, 0.0, 3.0)  (0.6, 3.0, 4.0)
Stack(), Norm()             (-0.4, 0.0, 0.333) (0.6, 0.0, 1.0) (-0.4, 1.0, -0.571) (0.6, 3.0, -2.0)
Norm(), Stack()             (-0.4, 0.0, 0.333) (-0.4, 0.333, 0.5) (0.6, 0.0, 1.0) (0.6, 1.0, 1.0)
```

`Stack()` then `Norm()` draws **negative heights**. `Norm()` then `Stack()` leaves the categories summing to 0.833 and 2.0 — not to 1. Neither is a 100% stack, so typing either as one would announce proportions that are not proportions.

## The spelling that *is* one

```
Norm(func="sum", by=["x"]), Stack()          (-0.4, 0, 0.333) (-0.4, .333, .667) (0.6, 0, 0.429) (0.6, .429, .571)
Norm(func="sum", by=["x"], percent=True), Stack()   33.333 / 66.667 and 42.857 / 57.143
```

Each category sums to exactly 1 (or 100). That is `position="fill"`'s equivalent, and today it reads as `stacked_bar` — the segments are announced, the fact that they are shares of a whole is not.

## What it needs

`PlotType.NORMALIZED` exists in the enum and is **plotly-only** in this package: `MaidrPlotFactory` routes `DODGED` and `STACKED` to `GroupedBarPlot` and has no `NORMALIZED` branch at all, and the only emitters are `maidr/plotly/grouped_bar.py` and `maidr/plotly/grouped_histogram.py`. So this is two pieces:

[...]

Issue #527, opened 2026-08-20
Two heatmaps on one axes are both read from the first one's mesh


jooyoungseo:
`HeatPlot` finds its artist by searching the axes rather than by being told which one it was registered for, so a second heatmap on the same axes is announced with the first one's values.

## Measured

matplotlib 3.9.4, unmodified `main`:

```python
fig, ax = plt.subplots()
ax.pcolormesh(np.arange(6).reshape(2, 3))          # 0..5
ax.pcolormesh(np.arange(100, 106).reshape(2, 3))   # 100..105
```

```
heat [[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]
heat [[0.0, 1.0, 2.0], [3.0, 4.0, 5.0]]
```

Two layers, both the first mesh. The second chart's numbers are nowhere in the output, and nothing raises — the reader navigates two layers, hears the same values in each, and has no way to tell that one of them is a copy.

## Why

`HeatPlot._extract_plot_data` calls `self.extract_scalar_mappable(self.ax)`, which resolves *per axes*:

```python
return next(sm for sm in ax.get_children() if isinstance(sm, ScalarMappable))
```

Every `HeatPlot` on one axes therefore resolves to the same artist. The mixin's own comment has always said so — *"We assume only one ScalarMappable is present to avoid plot clutter, even though multiples are technically possible"* — this issue is that admission written down as a defect, with the output it produces.

## Scope

The fix is not in the extractor's selection rule but in *when* the artist is chosen: a layer should be bound to the mesh that was on the axes when it was registered, rather than looking one up at extraction time. That is a change to how `MaidrPlot` subclasses receive their artist, so it reaches beyond the heatmap.

Worth checking whether the same shape applies elsewhere while it is being done. Anything that resolves its artist from `self.ax` at extraction time has the same exposure whenever two of a kind share an axes.

## Related

- #522 fixes a *different* symptom of the same "resolve from the axes" habit: `extract_scalar_mappable` picked a scatter's `PathCollection` over the mesh beside it, because a `PathCollection` is a `ScalarMappable` too. That one is fixable by preferring a mesh, and is fixed there. This one is not — both candidates really are meshes.
- #526 — the same family of charts is labelled with its axis ticks rather than its cells.

PR #621, opened 2026-08-23
feat(seaborn): type a colour-split seaborn.objects bar from its position transform

### LINKED ISSUE #617: A `seaborn.objects` mark split by `color=` reads as one unnamed layer

jooyoungseo:
The third and last item on #617, after #616 (registration), #618 (scatter split) and #619 (bar split).

A colour-split `so.Bar` with `so.Dodge()` or `so.Stack()` read as plain `bar` layers, where the same chart written `seaborn.barplot(hue=)` answers `dodged_bar`. A reader lost the grouping structure, the `z` axis naming it, and cross-group navigation between levels at one category.

`so` states the transform outright as a `Move` on the layer, which is a **cleaner signal than anything the classic path gets**: `seaborn.barplot(hue=)` is read as dodged by *counting containers*, and a stacked bar has to be declared through `maidr.stacked()`.

Measured, two categories over two levels:

| spelling | before | after |
|---|---|---|
| `color=` | `bar`, `bar` | `bar`, `bar` (unchanged) |
| `color=` + `Dodge()` | `bar`, `bar` | `dodged_bar` |
| `color=` + `Stack()` | `bar`, `bar` | `stacked_bar` |
| `color=` + `Dodge(), Stack()` | `bar`, `bar` | `stacked_bar` |

The grouped payload is the one the classic path already emits, so the two spellings of one chart now read alike:

```
dodged_bar   axes.z.label 'g'
             [[{'x':'a','z':'p','y':1.0}, {'x':'b','z':'p','y':3.0}],
              [{'x':'a','z':'q','y':2.0}, {'x':'b','z':'q','y':4.0}]]
```

## Plain `color=` keeps the split

It carries no transform and **overplots** the levels at the same position — measured, four bars at two x values. That is neither a dodge nor a stack, so a grouped reading would claim a structure the chart does not have. #619's named layers stay the honest reading of overplotted bars.

## Two supporting changes

**`GroupedBarPlot` can be handed its containers** instead of sweeping the axes. The sweep is right for the classic spelling, where the levels *are* separate containers, and cannot serve an `so.Bar(color=)`, whose levels arrive in one container and are split by colour after the fact — the sweep finds the one undivided container and reads every level as one group. Same shape `BarPlot._own_containers` has had since #527.

**Its two legend reads go through `legend_of`.** `so.Plot` hangs its legend on the **figure** (`ax.get_legend() is None`, `len(fig.legends) == 1`), so without this the `z` axis had no label and the groups fell back to the containers' matplotlib labels — a reader told the levels apart by nothing they can name. Same tier #561 added for `PairGrid`.

## `so.Norm()` is deliberately not read as a 100% stack

It looks like the missing symmetric case and is not one. `Norm()` defaults to `func="max"` with no `by`, which rescales within each *colour group*:

```
Stack(), Norm()    heights  0.333,  1.0,  -0.571,  -2.0     <- negative
Norm(), Stack()    categories sum to 0.833 and 2.0          <- not 1
```

The spelling that **is** a 100% stack is `Norm(func="sum", by=["x"])` before `Stack()`, which does sum to exactly 1 per category. Reading it needs a `stacked_normalized_bar` emitter this side does not have — `PlotType.NORMALIZED` is plotly-only in this package, with no `MaidrPlotFactory` branch. Filed as #620 with both measurements and the two pieces it needs, and documented at `_MOVES` so the next reader does not add it as the obvious omission.

## Verification

`uv run pytest` — **3037 passed, 18 skipped**, up from 3033. `ruff check --diff maidr/ tests/` clean.

Four tests changed their **premise** rather than their expectation, each because `Dodge()`/`Stack()` now group where they used to split; each was rewritten to keep the claim that was its real content:

- `test_a_colour_split_bar_announces_its_categories_not_its_coordinates` — moved to the plain `color=` spelling, which stays split; the same claim is made against the grouped shape below it.
- `test_every_position_transform_splits_the_same_way` → `test_a_position_transform_types_the_layer`, now covering `Dodge(), Stack()` together as well.
- `test_three_groups_split_three_ways_in_legend_order` — kept on the split path, joined by a grouped three-group twin. Three levels still matter for the same reason: with `p q r p q r` no level's bars are contiguous.
- `test_a_horizontal_colour_split_bar_keeps_its_orientation` — parametrized over both readings; the synthetic container's `orientation` now feeds `GroupedBarPlot._extract_orientation` too.
- `test_a_colour_split_bar_draws_one_container_and_splits_into_its_groups` — kept its container-count claim, which holds for every spelling, and dropped the type claim that now belongs elsewhere.
- `test_a_layer_that_drew_several_containers_becomes_several_bar_layers` — mechanical: `_handovers` now returns `(PlotType, kwargs)` pairs.

One new test, `test_the_grouped_reading_names_its_z_axis_from_the_figure_legend`, pins both legend reads.

| mutation | outcome |
|---|---|
| `_grouped_type` never types from the move | killed (4) |
| `Dodge` ordered before `Stack` in `_MOVES` | killed (1) |
| `GroupedBarPlot` sweeps the axes instead of taking its containers | killed (2) |
| both legend reads back to `ax.get_legend()` | killed (2) |

The last one killed only 1 before the `z`-label test was added — that test is why it now kills 2.

Closes #617.

---

### DIFF
--- a/maidr/core/plot/grouped_barplot.py
+++ b/maidr/core/plot/grouped_barplot.py
@@ -6,12 +6,20 @@ from matplotlib.container import BarContainer
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
 from maidr.exception import ExtractionError
+from maidr.util.legend_names import legend_of
 from maidr.util.mixin import (
     BarPositionMixin,
     ContainerExtractorMixin,
     DictMergerMixin,
     LevelExtractorMixin,
 )
+
+#: Key a patch uses to hand this layer the containers it drew, one per group,
+#: instead of letting it sweep the axes. A `seaborn.objects` bar needs it for
+#: a reason the classic path never has: `so.Bar(color=)` draws every level
+#: into *one* container, so the groups are synthesised from the bars' colours
+#: and are not on the axes to be found (#617).
+DRAWN_GROUPS = "_maidr_bar_groups"
 
 
 class GroupedBarPlot(
@@ -24,6 +32,7 @@ class GroupedBarPlot(
     def __init__(self, ax: Axes, plot_type: PlotType, **kwargs) -> None:
         super().__init__(ax, plot_type)
         self._orientation = "vert"
+        self._own_groups = kwargs.get(DRAWN_GROUPS, None)
 
     @property
     def _is_horizontal(self) -> bool:
@@ -85,8 +94,16 @@ class GroupedBarPlot(
         return axes_data
 
     def _extract_z_label_from_legend(self) -> str:
-        """Return the legend title text (trimmed) or an empty string."""
-        legend = self.ax.get_legend()
+        """
+        Return the legend title text (trimmed) or an empty string.
+
+        Through :func:`~maidr.util.legend_names.legend_of` rather than
+        ``ax.get_legend()``, which is what this read before. The two answer
+        the same question -- which legend names this axes' groups -- and the
+        wider answer also reads a lone *figure* legend, which is where
+        ``so.Plot`` puts the only legend a colour-split bar has (#617).
+        """
+        legend = legend_of(self.ax)
         if legend is None:
             return ""
         title = legend.get_title()
@@ -94,8 +111,33 @@ class GroupedBarPlot(
             return ""
         return title.get_text().strip()
 
+    def _own_containers(self) -> list[BarContainer] | None:
+        """
+        The containers this layer holds: the ones handed to it, or the axes'.
+
+        Sweeping the axes is what every grouped bar read before, and it is
+        right for the classic spelling -- ``seaborn.barplot(hue=)`` draws one
+        container per level and nothing else onto that axes. It cannot serve
+        a ``seaborn.objects`` bar, whose levels arrive in a single container
+        and are split by colour after the fact (#617): the sweep would find
+        the one undivided container and read every level as one group.
+
+        The same shape :meth:`maidr.core.plot.barplot.BarPlot._own_containers`
+        already has, for the reason #527 gave it -- a layer reads the artists
+        its own call drew.
+
+        Returns
+        -------
+        list of BarContainer, optional
+            The handed-over containers when there are any, otherwise every
+            container on the axes.
+        """
+        if self._own_groups is not None:
+            return self._own_groups
+        return self.extract_container(self.ax, BarContainer, include_all=True)
+
     def _extract_plot_data(self) -> list[list[dict]]:
-        plot = self.extract_container(self.ax, BarContainer, include_all=True)
+        plot = self._own_containers()
         data = self._extract_grouped_bar_data(plot)
 
         if data is None:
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Both actionable nits taken, in `18a03b0`.

**Faceted coverage.** Measured before writing the assertion — a `col=` grid with a colour-split `Dodge()` bar gives one `dodged_bar` per panel, each holding both groups with the right values and `axes.z.label` on both:

```
panel L   [[('p','a',1.0), ('p','b',3.0)], [('q','a',2.0), ('q','b',4.0)]]
panel R   [[('p','a',5.0), ('p','b',7.0)], [('q','a',6.0), ('q','b',8.0)]]
```

The part worth pinning turned out to be the `z` label on the **second** panel: `so.Plot` builds one figure legend for the whole grid, and both panels have to be named by it. That is `legend_of`'s figure tier doing the same job #561 gave it.

**`move` typing.** Now `list[Any] | None`, with a note on why it is not `seaborn._core.moves.Move` — naming that private class is precisely the dependency this module avoids by matching on the class name, so the annotation would undo the reason the code is written that way.

On the third note — `test_a_colour_split_bar_draws_one_container_and_splits_into_its_groups` losing its `_kinds` assertion: agreed the two files now have to be read together, and I left a pointer in that test's docstring saying where the type claim went. The split is deliberate: that test's subject is *how many containers seaborn draws*, which is a fact about seaborn and holds for every spelling, while the type now varies by spelling and belongs with the transform tests.

`uv run pytest` — **3038 passed, 18 skipped**. `ruff check --diff maidr/ tests/` clean.

---

PR #622, opened 2026-08-23
feat(seaborn): read a seaborn.objects 100% stacked bar as one

### LINKED ISSUE #620: A 100% stacked bar drawn with `so.Norm(func="sum", by=["x"])` reads as a plain stack

jooyoungseo:
Closes #620, which #621 filed rather than guessed at.

`so.Norm(func="sum", by=["x"])` before `so.Stack()` draws a 100% stacked bar — each category's segments sum to exactly 1 — and it read as `stacked_bar`. The segments were announced; that they are *shares of a whole* was not.

```
                                          before          after
Norm(func="sum", by=["x"]), Stack()       stacked_bar     stacked_normalized_bar
same, percent=True                        stacked_bar     stacked_normalized_bar
```

`PlotType.NORMALIZED` was **plotly-only** in this package, with no `MaidrPlotFactory` branch at all. `GroupedBarPlot` now serves all three grouped types: they differ in how the groups relate — side by side, piled up, piled up to a whole — not in where the numbers come from. Nothing is computed here, unlike the plotly path: `layout.barnorm` leaves the raw values in the trace so `barnorm.py` has to derive the shares (#338), while seaborn has already written them into the bars.

## Why the spelling cannot decide it

This is the part worth reading. `Norm` takes a `func` and a `by`, and only some combinations normalise *within the category*. Measured, two categories over two levels, totalling each category's stack:

| moves | stack total |
|---|---|
| `Norm(func="sum", by=["x"])` then `Stack()` | **1.0** |
| `Norm(func="sum", by=["x", "color"])` then `Stack()` | 2.0 |
| `Norm(func="sum")` then `Stack()` | 0.583 / 1.417 |
| `Norm()` (default `func="max"`) then `Stack()` | 0.833 / 2.0 |
| `Stack()` then `Norm(func="sum", by=["x"])` | 0.429 / 1.0 |

The second row is the trap. It names the category axis, so the obvious rule — "is the orientation axis in `by`?" — claims it, and it normalises **each level** to 1 so the stacks reach twice the whole. A reader would be told shares that sum to 200%.

So the **drawn bars** are asked instead, and both halves are required:

- a **sum-normalisation among the moves**, which no plain `Stack()` has, and
- **every category landing on a whole** (1 or 100, matching the two conventions `barnorm_scale` already emits).

An ordinary stack whose categories happen to total alike keeps reading as `stacked_bar`; a normalisation that did not land is refused.

## Order is not checked, and the totals are why

`Stack()` then `Norm()` normalises the already-stacked tops and can draw segments of negative height, so an order check looks obviously right. It earns nothing. After that order the tops are each category's `t_i / sum(t_i)` over the **cumulative** tops, which are all wholes only when one top carries the whole sum — every level but the last at zero — and **seaborn draws no rectangle for a zero**, so those levels leave no colour behind, the split finds one group, and no grouped layer is made at all.

Measured both ways: three levels with one at zero totals 0.8 and 0.667 and is turned away on the totals; two levels with one at zero draws **two** patches of one colour and reads as a plain `bar`. An order check would only refuse charts that never reach the function. It was written, survived its mutation, and was removed — the reasoning is at the code so it is not re-added.

## Verification

`uv run pytest` — **3049 passed, 18 skipped**, up from 3038. `ruff check --diff maidr/ tests/` clean.

11 new tests: an 8-way parametrization over the spellings above, the horizontal shares payload, the stack-normalised-afterwards pair, and one unit test on `_normalises_to_a_whole` directly.

That last one exists because of a surviving mutant. `func == "sum"` looked like decoration — dropping it passed everything — since a `func="max"` stack totals `sum / max`, which with two drawn levels is always above 1 and below 100, and a level at zero draws nothing. A hundred equal levels would land it on exactly 100.0 and announce each segment as 1 *percent* when it is the whole of its level. Stubbing the moves and the bars says that in three lines instead of a hundred-level chart.

| mutation | outcome |
|---|---|
| never promotes to `NORMALIZED` | killed (3) |
| any `Norm` counts, not just `func="sum"` | killed (1) |
| skips the whole check entirely | killed (4) |
| forgets `100.0` as a whole | killed (1) |
| reads a horizontal stack as vertical | killed (1) |
| factory drops the `NORMALIZED` branch | killed (3) |

---

### DIFF
--- a/maidr/core/plot/maidr_plot_factory.py
+++ b/maidr/core/plot/maidr_plot_factory.py
@@ -155,7 +155,14 @@ class MaidrPlotFactory:
             if kwargs.get(DRAWN_RUG) is not None:
                 return RugPlot(single_ax, **kwargs)
             return ScatterPlot(single_ax, **kwargs)
-        elif PlotType.DODGED == plot_type or PlotType.STACKED == plot_type:
+        elif plot_type in (PlotType.DODGED, PlotType.STACKED, PlotType.NORMALIZED):
+            # One class, three types. They differ in how the groups relate --
+            # side by side, piled up, piled up to a whole -- not in where the
+            # numbers come from, which is one magnitude per bar either way.
+            # A `NORMALIZED` layer's bars are already shares, because the
+            # library drew them that way; the plotly path computes its own
+            # (`maidr/plotly/barnorm.py`) only because plotly draws the raw
+            # values and normalises them in the view (#338, #620).
             return GroupedBarPlot(single_ax, plot_type, **kwargs)
         elif PlotType.SMOOTH == plot_type:
             return SmoothPlot(single_ax, **kwargs)
--- a/maidr/patch/seaborn_objects.py
+++ b/maidr/patch/seaborn_objects.py
@@ -172,18 +172,113 @@ def _panels(plotter: Any) -> list[Axes]:
 #: and then stacks within each dodged slot, so the segments a reader steps
 #: through are the stack's.
 #:
-#: ``Norm`` is deliberately absent. It looks like the 100% stack's transform
-#: and is not one: `so.Norm()` divides by a *per-group* maximum, so measured
-#: on two categories over two levels, `Norm()` after `Stack()` draws negative
-#: heights and `Norm()` before it leaves the categories summing to 0.833 and
-#: 2.0 rather than to 1. The spelling that is a 100% stack is
-#: `so.Norm(func="sum", by=["x"])` before `so.Stack()`, and reading it needs
-#: a `stacked_normalized_bar` emitter this side does not have -- `NORMALIZED`
-#: is plotly-only in this package. Reported separately (#617).
+#: ``Norm`` is not here, because whether it makes a 100% stack cannot be
+#: answered from its class name. See :func:`_normalises_to_a_whole`.
 _MOVES: tuple[tuple[str, PlotType], ...] = (
     ("Stack", PlotType.STACKED),
     ("Dodge", PlotType.DODGED),
 )
+
+
+#: What a 100% stack's segments sum to, per category. ``so.Norm`` writes
+#: shares as fractions by default and as percentages under ``percent=True``,
+#: and the schema carries whichever the chart drew -- the plotly side does
+#: the same, through ``barnorm_scale`` (#338).
+_WHOLES = (1.0, 100.0)
+
+#: How far a stack's total may sit from a whole and still be one. Segments
+#: come off the drawn rectangles, so they carry the rounding of every share
+#: seaborn computed; measured on the spellings below, the worst total was
+#: within 1e-9 of its whole, so this is loose by a wide margin and still far
+#: tighter than any total that is not a whole (the nearest measured miss is
+#: 2.0).
+_WHOLE_TOLERANCE = 1e-6
+
+
+def _normalises_to_a_whole(move: list[Any] | None, container: BarContainer) -> bool:
+    """
+    Whether a stacked layer's segments are *shares of a whole*.
+
+    Asked of the drawn bars rather than of the spelling, because the spelling
+    cannot answer it. ``so.Norm`` takes a ``func`` and a ``by``, and only some
+    combinations normalise within the category. Measured, two categories over
+    two levels, totalling each category's stack:
+
+    ======================================  =======
+    ``Norm(...)`` before ``Stack()``        total
+    ======================================  =======
+    ``func="sum", by=["x"]``                1.0
+    ``func="sum", by=["x", "color"]``       2.0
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Both findings taken, in `5bc09cf`. Both were real.

**Callable `func`.** Fixed rather than documented — `Norm(func=numpy.sum)` draws exactly what `Norm(func="sum")` does, so reading the two differently is a difference in the reader's experience with no difference in the chart. Matched on `__name__`, which covers `numpy.sum` and the builtin. A lambda still declines: what this reads is the layer's *stated* intent, and a lambda states none, so declining keeps the chart reading as the stack it is rather than claiming a whole on a guess. All four spellings are asserted.

**`Dodge()` + `Stack()` + `Norm`.** Measured before deciding, four levels dodged and stacked under `Norm(func="sum", by=["x"])`:

```
tops by drawn x:  -0.4:0.1  -0.2:0.2  -0.0:0.3  0.2:0.4
                   0.6:0.192 0.8:0.231 1.0:0.269 1.2:0.308
reads as:         stacked_bar
```

Your reading of the mechanism is right — the dodge splits the keys along with the bars — and the outcome is the safe one, for a reason worth stating: each drawn column is **one dodge slot's share of its category**, not the category's, so none of them is a whole and none should be. It is the rule working, not a gap in it. Now asserted (the test checks all eight keys and that none is within 1e-6 of a whole) and written at the code.

On the two nits I left alone: the docstring length is the file's convention, as you note, and `move or ()` twice over a list of at most a handful of `Move`s is not worth an extra binding.

`uv run pytest` — **3051 passed, 18 skipped**, up from 3049. `ruff check --diff maidr/ tests/` clean.

Mutations on the new code, each alone: `_sums` matches strings only → killed (1); `_sums` accepts any non-`None` func → killed (2).

---

Issue #627, opened 2026-08-23
Fifteen plotly trace types register nothing although the core already draws them


jooyoungseo:
## What I measured

I drove every `plotly.graph_objects` trace type through `PlotlyMaidr` and read `_flatten_maidr()` back, with eight known-good controls in the same run to prove the harness works. plotly 6.7.0.

Controls, all read:

```
bar          bar(4)
scatter      point(4)
scattergl    point(4)
heatmap      heat(dict)
pie          pie(4)
box          box(1)
histogram    hist(3)
violin       violin_box(1) violin_kde(1x26)
candlestick  candlestick(4)
ohlc         candlestick(4)
```

Everything else in the catalogue produced **no layer at all** — `PlotlyPlotFactory.create` returns `None` for any type outside `bar` / `scatter` / `scattergl` / `box` / `candlestick` / `ohlc` / `heatmap` / `histogram` / `pie`, and `_extract_plots` drops it.

## The part that is a gap rather than a decision

Fifteen of those declined types name a chart the MAIDR core **already knows how to draw**. Their `TraceType` is in `src/type/grammar.ts`, and the bundle shipped in this wheel mentions every one of them:

| plotly trace | grammar `TraceType` | in the shipped bundle |
|---|---|---|
| `waterfall` | `waterfall` | yes |
| `funnel` | `funnel` | yes |
| `funnelarea` | `funnel` | yes |
| `indicator` (gauge mode) | `gauge` | yes |
| `treemap` | `treemap` | yes |
| `sunburst` | `sunburst` | yes |
| `icicle` | `icicle` | yes |
| `sankey` | `sankey` | yes |
| `parcats` | `alluvial` | yes |
| `parcoords` | `parallel_coordinates` | yes |
| `contour` | `contour` | yes |
| `histogram2dcontour` | `contour` | yes |
| `histogram2d` | `heat` | yes |
| `barpolar` | `polar_area` | yes |
| `scatterpolar` | `radar` | yes |
| `choropleth` | `choropleth` | yes |

(Checked with `maidr.util.bundle_capability.bundle_trace_types()`, so "yes" means the installed bundle's own factory switches on that string — not that a table says so.)

So this is not "MAIDR cannot read a waterfall chart". MAIDR reads waterfall charts. A waterfall chart drawn in plotly, from Python, is silent — and it is silent in the quiet way, with no layer, no warning and no fallback image, because `_extract_plots` treats an unrecognised trace as nothing to say rather than as something it could not say.

[...]

PR #635, opened 2026-08-23
feat(plotly): read the two polar traces

jooyoungseo:
Seventh tranche of #627. Eleven of the fifteen now read.

## What was wrong

```python
go.Scatterpolar(r=[3, 9, 5, 1], theta=["N", "E", "S", "W"])   ->   layers: []
go.Barpolar(r=[3, 9, 5, 1], theta=["N", "E", "S", "W"])       ->   layers: []
```

Both draw spokes around a circle, and the core builds both on `RadarTrace` — a radar joins the spokes into an outline, a polar area fills the wedge between them, and a reader navigates the same spokes either way. Drawn since xability/maidr#833.

## The pair a polar trace states

`theta` is the angle and `r` the radius. There is no `x` or `y` on the trace at all, so reading the cartesian pair would give an empty layer and reading them the wrong way round would put the angle where the magnitude belongs.

A spoke with no radius is dropped. Plotly draws nothing for it, and `RadarTrace` places its spokes at an equal share of the circle **by count** — so keeping a gap would rotate every later spoke, and the stereo panning with it.

The payload is a line's, a list of series, because that is what `RadarTrace` extends.

## Why the barpolar gets no selector

So is the *selector* contract, and that is what decides how far this goes: `LineTrace.mapToSvgElements` compares `selectors.length` against the **series** count, not the point count.

Measured in Chromium:

| trace | drawn |
|---|---|
| `scatterpolar` | one `path.js-line` per trace, inside its own `.polarlayer .scatterlayer .trace` |
| `barpolar` | no per-series path at all — four bars for four spokes, under `.polarlayer .barlayer` |

For a radar that is exactly one element for one series. For a polar area there is nothing that fits: four selectors would be read as four *series*, and the one selector its single series is allowed would have to name the whole `.trace.bars` group, which outlines every bar at once and so highlights the same thing at every step of the walk. Neither answers "where am I", so the layer ships without a highlight and keeps its audio, braille and text — the outcome #145 established for a layer with nothing to point at.

Verified on `[radar, bars, radar]`, which also pins that a barpolar does not shift a radar's position (they are numbered among the polar *scatter* traces, since that is what `scatterlayer` holds):

```
OK  RADAR       series=1 resolved=1   .polarlayer .scatterlayer .trace:nth-child(1) path.js-line
--  POLAR_AREA  no selector (as designed)
OK  RADAR       series=1 resolved=1   .polarlayer .scatterlayer .trace:nth-child(2) path.js-line
```

## Two things I got wrong first, and how

**The angular axis has no title.** I wrote `_extract_axes_data` to read `layout.polar.angularaxis.title` alongside the radial one. Plotly rejects it outright — *"Invalid property specified for object of type plotly.graph_objs.layout.polar.AngularAxis: 'title'"* — and `'title' in AngularAxis._valid_props` is False while the radial axis's is True. The test I wrote for the assumption failed rather than the code, which is the order I want it in. The angle now always takes the generic word, with the measurement in the docstring.

**`POLAR_AREA` had no `display_name`.** The existing guard in `test_areaplot.py` caught it on the full-suite run: without the override a user who wrote `go.Barpolar` would be told **"polar_area"** — a schema token — in the same sentence as "stacked bar". That test exists for exactly this class of mistake, and this is the first multi-word type added since it was written.

## Tests

`tests/plotly/test_plotly_polar.py` — 12 tests: each trace read as its own type, `theta`/`r` parametrised over both, the list-of-series wrapping, the dropped spoke, the radar selector, the polar area read-but-unaddressed, the barpolar not shifting a radar's position, the radial title, the angular axis having no title to read, and the generic fallback.

Mutation-swept, eight mutations one at a time against a restored baseline, all caught:

```
M1  x/y read instead of theta/r        5 failed, 7 passed
M2  angle and radius swapped           4 failed, 8 passed
M3  series wrapping dropped            5 failed, 7 passed
M4  empty radius kept                  1 failed, 11 passed
M5  barpolar gets a selector           1 failed, 11 passed
M6  radial title ignored               1 failed, 11 passed
M7  barpolar shifts the radar position 1 failed, 11 passed
M8  _extract_plots block dropped       1 failed, 11 passed
```

Full suite in two batches (the whole run OOMs in this container): `tests/core` 1931 passed, 5 skipped; the rest 1235 passed, 13 skipped.

## Remaining on #627

`parcats`, `parcoords`, `contour`, `histogram2dcontour`, `histogram2d`, `choropleth`.

---

### DIFF
--- a/maidr/core/enum/plot_type.py
+++ b/maidr/core/enum/plot_type.py
@@ -64,6 +64,12 @@ class PlotType(str, Enum):
     CANDLESTICK = "candlestick"
     VIOLIN_KDE = "violin_kde"
     VIOLIN_BOX = "violin_box"
+    #: Spokes around a circle, one radius per angle. `RADAR` joins them into
+    #: an outline and `POLAR_AREA` fills the wedge between them; a reader
+    #: navigates the same spokes either way, which is why the core builds
+    #: both on one trace.
+    RADAR = "radar"
+    POLAR_AREA = "polar_area"
     #: Weighted flow between nodes, drawn as ribbons whose width is the
     #: magnitude. Stated as one point per link -- the two nodes it joins and
     #: how much moves -- rather than as a series, because the chart is a
@@ -128,6 +134,7 @@ _DISPLAY_NAMES = {
     PlotType.STACKED: "stacked bar",
     PlotType.STACKED_AREA: "stacked area",
     PlotType.NORMALIZED_AREA: "100% stacked area",
+    PlotType.POLAR_AREA: "polar area",
     PlotType.VIOLIN_BOX: "violin",
     PlotType.VIOLIN_KDE: "violin",
 }
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -858,6 +858,39 @@ class PlotlyMaidr:
                     self._plots.append(plot)
                 merged.update(id(t) for t in funnelarea_traces)
 
+            # Polar traces, one layer each. A `scatterpolar` is numbered
+            # among the polar *scatter* traces, because that is what the
+            # `.polarlayer .scatterlayer` groups hold -- a `barpolar` draws
+            # into `.polarlayer .barlayer` and shifts nothing there.
+            polar_traces = [
+                t for t in group_traces if t.get("type") in ("scatterpolar", "barpolar")
+            ]
+            if polar_traces:
+                # `PlotType` is imported here rather than at module level
+                # because `_extract_plots` already imports it locally
+                # further down, which makes the name local to the whole
+                # function -- a module-level reference above that point is
+                # an unbound local, not the module's.
+                from maidr.core.enum.plot_type import PlotType
+                from maidr.plotly.polar import PlotlyPolarPlot
+
+                scatter_position = 0
+                for polar_trace in polar_traces:
+                    is_scatter = polar_trace.get("type") == "scatterpolar"
+                    plot = PlotlyPolarPlot(
+                        polar_trace,
+                        layout,
+                        PlotType.RADAR if is_scatter else PlotType.POLAR_AREA,
+                        trace_position=scatter_position,
+                        **axis_kwargs,
+                    )
+                    if is_scatter:
+                        scatter_position += 1
+                    plot.row_index = row
+                    plot.col_index = col
+                    self._plots.append(plot)
+                merged.update(id(t) for t in polar_traces)
+
             # Sankeys, one layer each. Only this loop knows whether the
             # figure holds a second one, which is what decides whether
             # either can be addressed -- see `PlotlySankeyPlot._get_selector`.
--- a/maidr/plotly/plotly_maidr.py
+++ b/maidr/plotly/plotly_maidr.py
@@ -239,6 +239,18 @@ class PlotlyMaidr:
             if key.startswith("yaxis") and isinstance(val, dict):
                 y_starts.add(_domain_start(val, "domain"))
 
+            # A polar subplot is placed like a cartesian one -- by a
+            # rectangle `make_subplots` wrote into the layout -- but under
+            # its own key rather than an axis pair, so it needs its own
+            # branch. Without it a `[bar, polar]` grid collected one x start
+            # and collapsed to a single cell holding both layers, and two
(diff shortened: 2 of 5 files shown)


jooyoungseo:
The multiple-polar-subplots concern is right, and measuring it turned up more than the review described. All three points are fixed in 193a9b3.

## The subplot bug, measured

I built a 1×2 polar grid and read the DOM in Chromium:

```
polarlayer > g:  ["polar", "polar2"]
scatterlayers:   2      tracesPerScatterlayer: [1, 1]

.polarlayer .scatterlayer .trace:nth-child(1) path.js-line  ->  2 elements
                                                                 (one in polar, one in polar2)
.polarlayer .scatterlayer .trace:nth-child(2) path.js-line  ->  0 elements
```

So it was wrong in both directions at once: the first layer's selector
outlined **both charts** on one keypress, and the second layer's selector
resolved to **nothing**.

The fix scopes to the trace's own subplot — `subplot` is `"polar"`/`"polar2"`,
and that string is both the `layout` key and the class plotly puts on the
`<g>` — and counts the position among that subplot's own scatterpolar
traces. Re-measured, every emitted selector resolves to exactly one element,
and different layers to different elements:

| figure | selectors | resolved |
|---|---|---|
| single polar | `g.polar` nth-1 | 1 |
| two radars, one subplot | `g.polar` nth-1, nth-2 | 1, 1 — distinct |
| 1×2 polar grid | `g.polar` nth-1, `g.polar2` nth-1 | 1, 1 — distinct |
| 1×2 with a barpolar between two radars in `polar2` | `g.polar` nth-1, `g.polar2` nth-1, nth-2 | 1, 1, 1 — distinct |

## Two more, from following the same thread

**The grid was worse than described, and it hits the ordinary mixed case.**
A polar subplot's rectangle is in `layout.polar.domain`, which
`_subplot_domain_starts` collected from neither its axis branch nor its
domain-trace branch. So it is not only multi-polar figures:

```
make_subplots(1, 2, specs=[[{"type":"xy"}, {"type":"polar"}]])
  before:  grid 1x[1]   cell(0,0) [BAR, RADAR]
  after:   grid 1x[2]   cell(0,0) [BAR]   cell(0,1) [RADAR]
```

A bar beside a radar was read as one chart. Fixed by giving
`_subplot_domain_starts` a `polar*` branch, symmetric with the `xaxis`/`yaxis`
one, and placing each trace with a new `_polar_domain_start()` — the polar
analogue of `_trace_domain_start`, which you correctly identified as the
right shape but the wrong source.

**The axis titles had the same figure-wide assumption.**
`_extract_axes_data` read `self._layout.get("polar")` for every trace, so a
second subplot announced the *first* one's radial title — a wrong word rather
than a missing one, which a reader has no way to catch. Now reads
`layout[self._subplot]`.

## The two minor points

**`trace_position` default** — removed, for the reason you gave. It now
matches `PlotlyLinePlot`.

**NaN radius** — you were right, and the two spellings do both occur.
Measured: `go.Scatter(y=np.array([4, np.nan, 6])).to_dict()` gives
`{"dtype": "f8", "bdata": ...}`, and `as_list` decodes that to `nan`; the
same list written with `None` stays `None`. So the same chart read correctly
or rotated depending on how its author happened to hold the data. `r` entries
are now dropped when non-finite as well as when `None`.

Worth noting the difference from the rest of the codebase: an ordinary
scatter emits `{"y": NaN}` into the schema and always has. That is a
figure-wide convention I have not touched here. It matters more for a radar
because `RadarTrace` places spokes by *count*, so a kept gap rotates every
later spoke — which is exactly what the existing `None` test guards.

## Verification

Tests: 18 in `test_plotly_polar.py`, 5 of them new. Full suite green in both
batches — `tests/core` 1931 passed / 5 skipped, the rest 1241 passed / 13
skipped.

Mutation sweep, each reverted alone against a restored baseline:

| mutation | failures |
|---|---|
| selector back to unscoped `.polarlayer .scatterlayer` | 3 |
| position back to a running counter across subplots | 2 |
| axes back to `layout.polar` for every trace | 1 |
| `_subplot_domain_starts` polar branch removed | 2 |
| placement back to the group's row/col | 2 |
| gap test back to `radius is not None` | 1 |
| all restored | 18/18 pass |

---

Issue #642, opened 2026-08-23
Read a plotly contour trace — what is measured, and what is left


jooyoungseo:
The last substantial piece of #627. `go.Contour` still produces a figure
with no layers. Filing the measurements rather than a half-implementation,
because the hard part turned out to be a real matching problem and the next
attempt should start from data.

## The shape is right

The core has `TraceType.CONTOUR`, whose `ContourTrace` extends `LineTrace`:
one curve per level, each point carrying the level it runs at. Measured in
Chromium, plotly draws **exactly that**:

```
levelCount: 4
per level:  {paths: 1, subpaths: [1], level: 0.2}
            {paths: 2, subpaths: [1, 1], level: 0.4}
            {paths: 2, subpaths: [1, 1], level: 0.6}
            {paths: 2, subpaths: [1, 1], level: 0.8}
```

(Two separated Gaussian peaks. At 0.2 the curves merge into one; above it
they are two.)

So:

- one `g.contourlevel` per level, carrying `__data__.level` — the level value
  itself, not an index;
- one `<path>` per **disjoint curve** inside it, each a single subpath.

That is one addressable element per series, which is what `LineTrace`
resolves. `.contourlayer .contourlevel:nth-of-type(k) path:nth-of-type(j)`
is the natural selector — *if* the emitted curves come out in the same
order.

## What is missing, and what py-maidr already has

py-maidr builds its schema offline, and plotly computes the curves in the
browser. The Python side holds only `z` (plus optional `x`/`y`).

That is not fatal — it is the situation r-maidr already solved for base R
`contour()`, whose NEWS entry reads: *"The curves come from
`grDevices::contourLines()`, which is the computation `contour()` itself does
and takes the same defaults … so the announced levels are the drawn ones
rather than a plausible set."* The Python equivalent is `contourpy`, which
ships with matplotlib and is therefore already a dependency.

## The part that needs a campaign: plotly's auto levels

`contours.start`/`end`/`size` are exact when the author writes them, and
that case is easy. The **auto** case is the common one, and its rule is not
what it first looks like. Measured, one `go.Contour(z=...)` per z range:

| z range | start | end | size | levels |
|---|---|---|---|---|
| 0 … 20 | 2 | 18 | 2 | 9 |
| 0 … 1 | 0.1 | 0.9 | 0.1 | 9 |
| 0 … 100 | 10 | 90 | 10 | 9 |
| -5 … 5 | -4 | 4 | 1 | 9 |
[...]

Issue #636, opened 2026-08-23
An empty plotly domain or polar trace still emits a layer


jooyoungseo:
## What happens

A plotly trace that draws nothing still reaches the schema as a layer with an
empty `data` array — for every family except the cartesian line one.

Measured, layers per figure:

```
empty pie        [[1]]     go.Pie(labels=[], values=[])
empty sankey     [[1]]     go.Sankey(node={'label':[]}, link={'source':[],'target':[],'value':[]})
empty scatterpolar [[1]]   go.Scatterpolar(r=[], theta=[])
empty parcoords  [[1]]     go.Parcoords(dimensions=[])
empty scatter    [[0]]     go.Scatter(x=[], y=[])   <- the one that is filtered
```

## Why

This is #421's ghost, in the families #421 did not reach. That fix works by
excluding undrawn traces from the *line and area groupings* in
`_extract_plots`:

```python
area_traces = [t for t in group_traces if is_area_trace(t) and draws_marks(t)]
connected_traces = [
    t for t in group_traces
    if is_connected_line_trace(t) and not is_area_trace(t) and draws_marks(t)
]
```

`draws_marks()` is never asked about a pie, a sankey, a hierarchy, a
scatterpolar or a parcoords. Each of those has its own build block appending
a plot unconditionally, so an empty one becomes an empty layer.

## Why it matters

A layer with no points is not a harmless no-op. It is a cell the reader can
tab into and find nothing in — #421's own words — and the core does worse
than ignore it for some types: `LineTrace.text` dereferences an undefined
point on an empty series and throws, which propagates out of `Figure` and
takes the whole render down (xability/maidr#905). `ParallelTrace` extends
`LineTrace`, so a parcoords with no dimensions is on exactly that path.

## Fix

One guard rather than five: drop a plot whose extracted payload is empty,
after the build blocks have run, so every trace type reaches the same answer
and a new one added later inherits it. The line and area families keep their
own `draws_marks()` exclusion, which does more than this — it also keeps the
*positions* of the surviving series contiguous, which a later filter cannot
recover.

Needs measuring per type before landing, since "empty" is not one shape: a
pie with no slices, a sankey with no links, a hierarchy with no nodes and a
parcoords with no dimensions all reach it differently.

## Scope

Predates this: the pie block is long-standing, and the sankey, polar and
parcoords blocks are from #627's tranches. Filing rather than patching
inside a feature PR so all of them are fixed in one place.

PR #643, opened 2026-08-23
feat(plotly): read a contour at the levels its author declared

jooyoungseo:
Closes part of #627; implements step 1 of #642.

`go.Contour` produced a figure with no layers. The core has had
`TraceType.CONTOUR` since the matplotlib side was read (#539) — one curve per
level, each point carrying the level it runs at.

## The curves are not in the trace

Plotly ships a grid and a level spacing and traces the curves in the browser,
so reading this chart means running the same marching squares here. `contourpy`
does it, and is what matplotlib traces its own contours with — so the two
py-maidr contour readings now describe curves computed the same way. It is
declared in `pyproject.toml` rather than relied upon to arrive with matplotlib,
which is exactly the accident the `matplotlib` entry beside it already exists to
stop repeating.

Which contourpy algorithm is load-bearing, not incidental: `mpl2014` and
`serial` disagree about the order they return curves in. On a level holding two
open curves and one closed one, plotly writes the open ones first and `mpl2014`
returns them in exactly that order, while `serial` returns the closed one first.

## Only an explicit spec is read

`contours.start`/`end`/`size`, all three, `size > 0`, and `autocontour` not
`True` — which overrides an otherwise complete spec (measured). Plotly's rule
for picking levels itself lives in plotly.js and did not reduce to a formula:
across the nine z ranges tabulated in #642, eight fit "round `(max - min) / 15`
up to a 1/2/5×10ⁿ step" and a field spanning `0 … 3` does not. Reading a chart at
levels it does not draw would announce curves that are not there, so the auto
case stays with #642 and the layer declines — leaving the figure with no contour
layer rather than a wrong one (#636).

A `contours.type: "constraint"` declines too, and not only for want of a spec:
it draws **one** curve at `value` and means "the region beyond it", so a
`constraint` written *with* a full `start`/`end`/`size` (which plotly ignores —
measured, one group, at 0.5) would otherwise be read as five levels the chart
does not draw. The test writes it that way on purpose.

## The level list, measured

Plotly steps `start`, `start + size`, … while the level is **below
`end + size / 10`**. Neither obvious rule fits:

| spec | `<= end` | `end + size/2` | `end + size/10` | plotly draws |
|---|---|---|---|---|
| `start=0.2, end=0.8, size=0.05` | 12 | 13 | **13** | 13 |
| `start=0, end=0.9, size=0.5` | 2 | 3 | **2** | 2 |

Accumulated rather than computed as `start + k * size`, so the level announced,
the level handed to the tracer and the level plotly drew are one number —
stepping 0.05 from 0.2 lands on `0.8000000000000002`, and that is the
thirteenth group's `__data__.level`.

`start > end` is swapped, which is what plotly does with it.

## The highlight, and where it stops

Plotly gives every **declared** level a `g.contourlevel` — including the ones
the field never reaches, which get the group and no `path` at all — and one
`<path>` per disjoint curve inside it. So a level is addressable: the groups run
in declared order on every figure measured. A **curve within a level** is not.

A sweep of 33 fields and 207 levels — random sums of gaussians, a saddle, a
monkey saddle, ripples, a staircase, noise:

- **count disagreements: 0.** Plotly and contourpy always found the same number
  of curves in a level.
- **order disagreements: 18**, and *not only on contrived fields* — five were
  ordinary two-peaked gaussians.

So a positional per-curve selector would resolve to a real element and to the
wrong one. The core parses the resolved path to place the per-point highlights
(`LineTrace.mapViaPathParsing`), so every point of that series would land on an
island the reader is not on — worse than none, which is the outcome #145
settled.

A layer whose every drawn level draws a single curve therefore has one *forced*
mapping rather than a chosen one, and keeps its highlight; a layer with an
island anywhere ships without one and keeps its audio, braille and text.

The other case with nothing to point at is `coloring: "fill"` (the default) with
`showlines: False`: plotly then writes **no** `g.contourlevel` and draws only
the bands. The layer still reads — a band's boundary is exactly where the level
curve runs — and claims no highlight. `showlines` is honoured only under `fill`;
under `heatmap` the groups are written whatever it says (measured), so reading
it as "no lines" there would drop a highlight the chart does have.

### Verified end to end

Every emitted selector was resolved in Chromium against the rendered figure and
checked to match one element whose geometry is the curve it was emitted for —
across a two-peak field, a saddle, a mixed open/closed field, levels above the
field, two traces on one subplot, two subplots, a third subplot, a contour after
a `histogram2dcontour`, and a filled contour. 0 failures.

## Placement and scoping

A contour is a cartesian trace, so it groups by axis pair and needs no domain
work. Its selector is scoped by its position among the subplot's traces that
draw into the `contourlayer` — which includes `histogram2dcontour`, measured: a
figure holding one of each gave the layer two `g.contour` groups in declaration
order, while a `heatmap` sat elsewhere in a `heatmaplayer`. Counting only
contour traces would hand the contour the histogram's group. The position map is
keyed by identity rather than found with `.index`, which compares by value: two
traces drawing the same field are equal dicts, so both would be handed position
0.

## Grid handling

`x`/`y` when they describe the grid, `x0`/`dx` otherwise at plotly's defaults;
`transpose` turns `z` over and leaves the coordinates (measured on an asymmetric
field, where the two readings put the peak in different places and plotly drew
the transposed one); a missing cell is a hole and the curves stop at it, which
plotly does too. A **descending** coordinate array needs no special handling —
measured, plotly and contourpy both map column index → `x[index]` and put the
curve in the same place.

Declined: a grid with no cell in it, a ragged grid, a coordinate array that
misses the grid's width, and a **categorical** axis — a contour crosses
*between* columns, so a curve sits a fraction of the way from one category to
the next, a position a category name cannot express.

A runaway `size` (`dict(start=0, end=1, size=1e-9)` is writable) is declined
rather than traced: building a schema must never cost the render (#421, #636).

## Tests

36 new tests in `tests/plotly/test_plotly_contour.py`. Two assert against
`_grid` and `declared_levels` directly, because those decisions have no other
window — contourpy's own error masks the first, and reaching the second through
a figure would be the hang it guards against.

## Mutation sweep

23 mutations, each reverting one decision alone against a restored baseline.
**22 killed, 1 survivor**, and the sweep changed the shipped code three times:

- The `len(curve) < 2` guard I had copied from the matplotlib reading **survived
  and was removed** — contourpy's `SeparateCode` output never returns a curve
  shorter than two vertices (probed across five degenerate fields), unlike a
  matplotlib compound path, so it was dead code.
- The `nan=True` parameter on the number reader **survived and was removed** —
  numpy already writes NaN for a `None` in a float array, so it changed nothing.
- The bare `constraint` test **survived** and was sharpened to carry a full
  spec, which is the only form that reaches the guard.

**The one documented survivor** is `np.ma.masked_invalid`: passing the NaN array
straight through gives the same curves today, so removing the mask breaks no
test. That is `mpl2014` treating NaN as missing rather than anything contourpy
promises — the mask is what its documentation specifies for missing points — so
it stays, and the docstring says exactly this.

## Checks

- `ruff check` clean on the touched files
- `uv lock --check` clean
- `tests/core`: 1931 passed, 5 skipped
- `tests/plotly`: 1156 passed
- `tests/altair tests/patch tests/util tests/widget tests/test_backend.py`: 185 passed

## Still open under #627

`histogram2d` and `histogram2dcontour`, which carry raw samples and need
plotly's 2-D autobinning matched first; and #642's auto-level rule, which this
PR's decline is waiting on.

---

### DIFF
--- a/maidr/plotly/candlestick.py
+++ b/maidr/plotly/candlestick.py
@@ -20,7 +20,25 @@ _MARK_OF = {"candlestick": "path.box", "ohlc": "> path"}
 #: Trace types that share ``g.boxlayer`` and so count towards a candlestick's
 #: position within it. A ``go.Violin`` draws into ``g.violinlayer`` and a
 #: scatter into ``g.scatterlayer``, so neither shifts the count.
-_BOXLAYER_TRACE_TYPES = frozenset({"box", "candlestick"})
+#: Groups of trace types that plotly draws into a *single* DOM layer, so each
+#: member shifts the others' group index there.
+#:
+#: ``box``/``candlestick`` share ``g.boxlayer``. Counting only same-typed
+#: traces was symmetric-looking and wrong in one direction: a candlestick
+#: counted the boxes beside it, while a box ignored the candlestick and
+#: claimed the group the candlestick had already taken (#395).
+#:
+#: ``contour``/``histogram2dcontour`` share ``g.contourlayer`` -- measured in
+#: Chromium on a figure holding one of each plus a heatmap: the contour layer
+#: held two ``g.contour`` groups in declaration order, while the heatmap sat
+#: elsewhere in a ``heatmaplayer`` of its own.
+#:
+#: A type absent from every group is numbered among its own kind, which is
+#: what a layer with one occupant needs.
+_SHARED_LAYERS = (
+    frozenset({"box", "candlestick"}),
+    frozenset({"contour", "histogram2dcontour"}),
+)
 
 
 def is_ohlc_trace(trace: dict) -> bool:
@@ -65,16 +83,11 @@ def layer_position(traces: list[dict], trace: dict) -> int:
         The zero-based position among traces drawing into the same layer.
     """
     kind = trace.get("type")
-    if kind in _BOXLAYER_TRACE_TYPES:
-        # Both directions, not just the candlestick's. `go.Box` and
-        # `go.Candlestick` draw into the same `g.boxlayer`, so each shifts
-        # the other's group index. Counting only same-typed traces was
-        # symmetric-looking and wrong in one direction: a candlestick
-        # counted the boxes beside it, while a box ignored the candlestick
-        # and claimed the group the candlestick had already taken (#395).
-        mates = _BOXLAYER_TRACE_TYPES
-    else:
-        mates = frozenset({kind}) if kind else frozenset()
+    mates = frozenset({kind}) if kind else frozenset()
+    for shared in _SHARED_LAYERS:
+        if kind in shared:
+            mates = shared
+            break
 
     position = 0
     for candidate in traces:
--- a/maidr/plotly/contour.py
+++ b/maidr/plotly/contour.py
@@ -1,0 +1,368 @@
+from __future__ import annotations
+
+import math
+from typing import Any
+
+import numpy as np
+
+from maidr.core.enum.maidr_key import MaidrKey
+from maidr.core.enum.plot_type import PlotType
+from maidr.plotly.plotly_plot import PlotlyPlot, as_list
+
+#: The trace types that draw into a subplot's ``contourlayer``. Both take a
+#: ``g.contour`` group there, so a contour's position among *these* is what
+#: scopes its selector -- measured in Chromium on a figure holding a
+#: ``histogram2dcontour``, a ``contour`` and a ``heatmap``: the contour layer
+#: held two groups, in declaration order, and the heatmap was elsewhere in a
+#: ``heatmaplayer`` of its own.
+_CONTOUR_LAYER_TYPES = ("contour", "histogram2dcontour")
+
+#: Which marching-squares implementation to trace the curves with.
+#:
+#: Not a detail: ``mpl2014`` and ``serial`` disagree about the **order** they
+#: return curves in, and the order is what pairs a series with the ``<path>``
+#: that draws it. Measured on a level carrying two open curves and one closed
+#: one, plotly writes the open ones first and ``mpl2014`` returns them in
+#: exactly that order, while ``serial`` returns the closed one first. It is
+#: also what matplotlib traces its own contours with, so the two py-maidr
+#: contour readings describe curves computed the same way.
+_CURVE_ALGORITHM = "mpl2014"
+
+#: How many levels a trace may declare before it is declined.
+#:
+#: A runaway guard, not a judgement about how many levels are useful:
+#: ``size`` is the author's, so ``dict(start=0, end=1, size=1e-9)`` asks for a
+#: billion of them and would hang the export -- and building a schema must
+#: never cost the render (#421, #636). No figure anyone draws comes near this.
+_LEVEL_LIMIT = 1000
+
+#: How far past ``end`` plotly's level loop still steps, as a fraction of
+#: ``size``. Measured rather than assumed, across seven specs: a run at
+#: ``start=0.2, end=0.8, size=0.05`` draws a level at ``0.8000000000000002``
+#: -- past ``end``, so the loop is not ``<= end`` -- while ``start=0, end=0.9,
+#: size=0.5`` stops at ``0.5`` and does *not* draw ``1.0``, so it is not
+#: ``end + size / 2`` either. ``end + size / 10`` fits all seven.
+_END_TOLERANCE = 10
+
+
+class PlotlyContourPlot(PlotlyPlot):
+    """Extract data from a Plotly ``contour`` trace.
+
+    A scalar field drawn as the curves along which it is constant. The core
+    reads it as `CONTOUR`, whose point carries a `level` as well as an x and a
+    y, and whose series is **one curve** rather than one level -- a field with
+    two peaks crosses a level twice, and joining the two islands into one
+    series would announce a curve running across ground the field never took.
+
+    **The curves are not in the trace.** Plotly ships a grid and a level
+    spacing and computes the curves in the browser, so reading this chart
+    means tracing the same marching squares here. That is what `contourpy`
+    does -- it is what matplotlib traces its own contours with, and it arrives
+    with matplotlib, which py-maidr already requires.
+
+    Two independent implementations agreeing is not a contract, so what they
+    agree about was measured rather than assumed. Across 33 fields and 207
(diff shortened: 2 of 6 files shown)


jooyoungseo:
Thanks — the first point was a genuine miss, and I've taken it. Pushed as `646567a`.

## `layer_position` reuse — taken

You're right that this is the same question `layer_position` already answers, and that `_BOXLAYER_TRACE_TYPES` is already the "two types share one layer" case. I've folded both into one table:

```python
_SHARED_LAYERS = (
    frozenset({"box", "candlestick"}),                # g.boxlayer (#395)
    frozenset({"contour", "histogram2dcontour"}),     # g.contourlayer
)
```

`draws_into_the_contour_layer` and the identity-keyed dict are gone, and the contour build block now reads like the candlestick and waterfall blocks beside it. `layer_position` already matches with `is`, which is exactly what my own map was written for — two traces drawing the same field are equal dicts — so nothing was lost in the swap.

Verified rather than assumed:

- All 1156 `tests/plotly` tests still pass, including the two position tests.
- The browser end-to-end check re-run: every emitted selector still resolves to one element whose geometry is the curve it was emitted for, across all nine figures. 0 failures.
- Four mutations against the refactored helper, each reverted alone — dropping the contour group, dropping the box/candlestick group, comparing by value instead of identity, counting every trace instead of layer-mates. **All four killed.**

## Logging — taken

Added `_logger.debug(..., exc_info=True)` on the contourpy failure, matching `maidr/util/cdn.py`'s convention. The comment says why it is worth saying there and nowhere else in this file: every reason a *chart* declines is guarded above that point, so reaching it means the environment is not what it was measured to be.

## `uv.lock` — checked, and it is pure churn

Fair thing to want checked rather than eyeballed, so I parsed both revisions and compared:

```
packages added:   none
packages removed: none
versions changed: none
total packages:   206 → 206
```

`contourpy` was already in the lock as a matplotlib transitive, so declaring it added no package and moved nothing — which is the point of declaring it. The whole diff is the broadened `marker = "python_full_version..."` annotations you spotted, from a newer `uv` writing the same resolution differently.

---

jooyoungseo:
Note 2 was worth acting on — pushed as `c2b1890`.

You read the project's own convention correctly: the comment above `matplotlib` says the base list is what `import maidr` itself needs, and `contourpy` is imported lazily by `maidr/plotly/contour.py` and by nothing else. Moved into the `plotly` extra, where it says something true rather than something merely harmless. Functionally a no-op either way — matplotlib already brings it, and every CI job installs `--all-extras`, plus the `test` extra lists `matplotlib` itself — but that is the reason to *declare* it and not a reason to put it anywhere in particular. `uv lock --check` clean, 1156 `tests/plotly` still passing.

Notes 1 and 3 I'm leaving as they are, and agreeing with your reading of both:

- **The broad `except`.** You've named the real residual risk exactly: a future edit that bypasses `_grid`'s validation would be swallowed rather than surfaced. I'd rather that than a schema build taking a render down (#421, #636), and the `_logger.debug(..., exc_info=True)` at least makes it findable. Worth a maintainer knowing the trade is deliberate.
- **`layer_position`'s O(n·m).** Same shape as the `funnel`/`waterfall`/`ohlc` loops beside it, so making the contour the one exception would be the inconsistency. If it ever matters it should change for all four at once.

---

jooyoungseo:
The `generator.lines()` finding is a real gap and you're right about the consequence. Fixed in `12bba1d`.

## The guard did not cover the tracing

Exactly as you describe: the `try/except` wrapped construction and stopped there, while `.lines()` — where the work actually happens, and so where an unforeseen grid or a stricter future contourpy would raise — sat outside it. An exception there leaves `_extract_plot_data`, leaves `render()`, and takes the whole figure's schema with it. That is the one outcome every other guard in this module exists to prevent, and it was the one path out.

All the tracing is inside the guard now:

```python
generator = contour_generator(...)
traced = [generator.lines(level)[0] for level in levels]
```

and `_series_levels` is built afterwards from `traced`, so a failure part way through cannot leave a half-filled list for `_get_selector` to read — which would have put every later series on the wrong level's group.

**Pinned by a test rather than by the diff.** It renders a bar beside the contour with a tracer whose `.lines()` raises, and asserts the bar still arrives and the contour does not — the bar is the witness that the figure survived. Narrowing the guard back to construction alone fails it (verified: `1 failed, 36 passed`).

## `:nth-of-type` — you named the right footgun, and here is the measurement

You're right that `:nth-of-type` counts by tag among *all* siblings, not among the ones matching the class beside it, so it is only exact where the siblings are homogeneous. That deserved to be a measurement rather than an inference, so I measured it directly across seven configurations — every `coloring` mode (`lines`, `fill`, `heatmap`, `none`), `showlabels: True`, a `histogram2dcontour` sibling, and two contours on one subplot:

| container | children, in every configuration |
|---|---|
| `.contourlayer` | `g.contour` only |
| `g.contourlines` | `g.contourlevel` only |
| `g.contourlevel` | `path` only |

So the selector is exact rather than lucky. Recorded in `_get_selector`'s docstring, since it is what the selector rests on — and your wider point stands: nothing in the repo would catch plotly.js regrouping these in a version bump. The end-to-end resolution check I run against Chromium is a script, not a test, because the plotly path has no browser test harness. Worth its own issue if that gap matters beyond this layer.

## Re-verified after the change

- 1157 `tests/plotly` tests pass.
- The browser end-to-end check: 10 figures, every emitted selector resolving to one element whose geometry is the curve it was emitted for. 0 failures.

The two smaller notes I'm leaving: the `uv.lock` churn is confirmed marker-only (parsed both revisions — no package added, removed, or moved version, 206 → 206), and the `contourpy` placement is the one the previous round moved into the `plotly` extra.

---

jooyoungseo:
Correcting one thing I said above: I wrote that "the plotly path has no browser test harness". That's wrong — `tests/browser/` exists, drives Chromium behind `--run-browser`, and CI runs it as the "Accessibility tests in a browser" job that is green on this PR. What is missing is narrower and more actionable: no test in it resolves a plotly layer's selectors against the chart.

Filed as #644, with what the test would need to do and the three things that make it fiddly — the layers that deliberately emit no selectors, the choropleth needing network, and the per-shape difference between "one element per point" and "one path per series".

The rest of that comment stands.

---

PR #645, opened 2026-08-23
feat(plotly): read a 2-D histogram as the heatmap it draws

jooyoungseo:
Closes part of #627, and answers the 2-D binning half of #642's closing note.

`go.Histogram2d` produced a figure with no layers.

## It is a heatmap

A rectangular grid of cells each carrying a number is exactly that, and plotly
agrees: measured in Chromium, a `histogram2d` draws a single `<image>` into its
subplot's `heatmaplayer` — the same element a `go.Heatmap` draws. So the layer
extends `PlotlyHeatmapPlot` and shares its selector, its row/column ordering
and its axis titles.

What differs is where the grid comes from. A heatmap is handed one; a
`histogram2d` carries raw samples and lets plotly bin them in the browser.

## The 2-D binning rule is the 1-D one, minus one exponent

This is the part #642 flagged as needing "plotly's 2-D autobinning matched
first". It turned out to be a single character of difference.

`autoBin` computes its rough width as `2 * stdev(data) / n ** (is2d ? 0.25 : 0.4)`
and rounds it through the same `autoTicks`. py-maidr already matches everything
but the exponent. Measured against the browser on eight axes across four
figures — gaussian, uniform, a six-value sample, a normalised one — `0.4`
matched **none** and `0.25` matched **all eight**.

The test pins it on a deterministic sample rather than a random one. The same
thirty values `0 … 29`:

| trace | plotly's width |
|---|---|
| `go.Histogram(x=v)` | **5** |
| `go.Histogram2d(x=v, y=v)` | **10** |

Reusing the 1-D exponent would announce six columns where the chart draws
three. `nbinsx`/`nbinsy` hints and explicit `xbins`/`ybins` needed no change at
all — the difference reaches exactly the automatic branch, which is why
`compute_bin_edges` grew one keyword and nothing else.

## Reading a cell

- **Named by the range it covers**, not its index or its centre — "a count of 4"
  says nothing without "between 0 and 2", and the range is what a sighted reader
  takes off the axis. r-maidr settled the same question the same way for
  `geom_bin_2d` (xability/r-maidr#136).
- **`histfunc`** reduces `z` rather than counting rows; an aggregating
  `histfunc` with no `z` falls back to counting, which is what plotly does with
  it (measured).
- **A cell nothing landed in** is a `0` under `count`/`sum` and `None` under
  `avg`/`min`/`max` — measured, plotly paints the first and leaves the second
  unpainted. An average of nothing is not zero.
- **`histnorm`** is the 1-D rule with the cell's **area** where it uses the bin's
  width. All four forms verified on a grid of unequal x and y widths.
- **The third axis is named for what the cells hold** — `Count`, or the
  normalisation, or the aggregate. A heatmap's numbers are the author's, so the
  parent emits a `z` only when the colour bar is titled; these are computed, so
  the name is known here, and a titled colour bar still wins.

## Verified against the browser

Every emitted cell and every bin label compared with `gd.calcdata` across **18
figures**: auto widths, `nbins` hints, explicit bins, all four `histnorm`
forms, three `histfunc` aggregates, samples sitting on the bin edges, a
reversed x axis, a reversed y axis, a single-cell grid, and a subplot beside a
bar. **0 disagreements.**

## Mutation sweep — and the defect it found

18 mutations, each reverting one decision alone against a restored baseline.
**All 18 killed, no survivors** — after the sweep changed the shipped code
three times:

- **A real defect.** I had folded a sample sitting exactly on the closing edge
  into the bin below it — the obliging thing to do, and wrong. Measured: with
  `xbins` running to 4, plotly **drops** a sample at 4, while one at an interior
  edge of 2 counts in `[2, 4)` and one at the opening edge of 0 counts in
  `[0, 2)`. Every bin is half-open, including the last. The fold would have made
  that cell one fuller than the one drawn. Removed, with the three edge cases
  now parametrised.
- **A redundant guard removed.** The explicit "no samples" check was already
  answered by "nothing usable to bin" — an empty array and an array of nothing
  usable are the same chart.
- **A weak test sharpened.** The `histfunc="avg"` case had one sample per cell,
  where a sum passes for an average. It now puts two in one cell.

## Checks

- `ruff check` clean on the touched files
- `tests/core`: 1931 passed, 5 skipped
- `tests/plotly` + altair/patch/util/widget/backend: 1371 passed

## Still open under #627

`histogram2dcontour`, which bins the same way this does and then contours the
result — so it now needs only #642's auto-level rule, which is what a plotly
contour still cannot read.

---

### DIFF
--- a/maidr/plotly/histogram.py
+++ b/maidr/plotly/histogram.py
@@ -29,7 +29,7 @@ def _plotly_round_up(val: float, array: list[float], reverse: bool = False) -> f
     return array[lo]
 
 
-def _plotly_default_size0(arr: np.ndarray) -> float:
+def _plotly_default_size0(arr: np.ndarray, *, is_2d: bool = False) -> float:
     """Compute the default rough bin size when ``nbinsx`` is not given.
 
     Mirrors the logic in Plotly.js ``axes.autoBin`` (the ``else`` branch
@@ -38,7 +38,20 @@ def _plotly_default_size0(arr: np.ndarray) -> float:
         distinctData = Lib.distinctVals(data)
         msexp = 10 ** floor(log(minDiff) / LN10)
         minSize = msexp * roundUp(minDiff / msexp, [0.9, 1.9, 4.9, 9.9], true)
-        size0 = max(minSize, 2 * stdev(data) / n^0.4)
+        size0 = max(minSize, 2 * stdev(data) / n^(is2d ? 0.25 : 0.4))
+
+    Parameters
+    ----------
+    arr : np.ndarray
+        The raw data values.
+    is_2d : bool, default False
+        Bin an axis of a **two-dimensional** histogram, which plotly bins
+        more coarsely: the sample-size exponent is ``0.25`` rather than
+        ``0.4``, which is `autoBin`'s own ``is2d`` flag and the only thing
+        that differs between the two. Measured against the browser on eight
+        axes across four figures -- gaussian, uniform, a six-value sample and
+        a normalised one -- ``0.4`` matched none of them and ``0.25`` matched
+        all eight.
     """
     sorted_arr = np.sort(arr)
     n = len(sorted_arr)
@@ -61,9 +74,9 @@ def _plotly_default_size0(arr: np.ndarray) -> float:
         min_diff / msexp, [0.9, 1.9, 4.9, 9.9], reverse=True
     )
 
-    # size0 = max(minSize, 2 * stdev / n^0.4)
+    # size0 = max(minSize, 2 * stdev / n^(is2d ? 0.25 : 0.4))
     stdev = float(np.std(arr, ddof=0))
-    size0 = max(min_size, 2 * stdev / (n ** 0.4))
+    size0 = max(min_size, 2 * stdev / (n ** (0.25 if is_2d else 0.4)))
 
     if not np.isfinite(size0) or size0 <= 0:
         return 1.0
@@ -725,7 +738,11 @@ class PlotlyHistogramPlot(PlotlyPlot):
 
 
 def compute_bin_edges(
-    arr: np.ndarray, bins: dict | None = None, nbins: int | None = None
+    arr: np.ndarray,
+    bins: dict | None = None,
+    nbins: int | None = None,
+    *,
+    is_2d: bool = False,
 ) -> np.ndarray:
     """Compute bin edges that match Plotly's autobinning algorithm.
 
@@ -747,6 +764,11 @@ def compute_bin_edges(
     ----------
     arr : np.ndarray
         The raw data values.
+    is_2d : bool, default False
+        Bin one axis of a ``histogram2d``, which changes only the sample-size
+        exponent of the automatic width -- see :func:`_plotly_default_size0`.
+        An explicit ``size`` or ``nbins`` is honoured identically either way,
+        which is why this reaches no further than that one branch.
 
     Returns
     -------
@@ -797,7 +819,7 @@ def compute_bin_edges(
     if nbins is not None:
         size0 = data_range / max(1, nbins)
     else:
-        size0 = _plotly_default_size0(arr)
+        size0 = _plotly_default_size0(arr, is_2d=is_2d)
     dtick = _plotly_dtick(size0)
 
     # 2. Initial bin start: one tick below the first tick >= data_min
--- a/maidr/plotly/histogram2d.py
+++ b/maidr/plotly/histogram2d.py
@@ -333,6 +333,12 @@ def _normalised(
 ) -> list[list[float | None]]:
     """Rescale the cells the way ``histnorm`` does.
 
+    Only the numbers. A cell holding ``None`` is one nothing landed in, and
+    what plotly shows *there* depends on the ``histnorm`` as well as the
+    ``histfunc`` -- so it is left as it is and :func:`_as_payload`, which
+    knows both, decides. Splitting it that way keeps the two questions apart:
+    what a cell's number becomes, and what a cell with no number becomes.
+
     The one-dimensional rule with the bin's **area** where it uses the bin's
     width -- measured against the browser for all four forms on a grid of
     unequal x and y widths: ``percent`` and ``probability`` divide by the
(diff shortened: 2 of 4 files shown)


jooyoungseo:
The `z` pairing bug is real and you traced it exactly right. Fixed in `14d020c`.

## Reproduced first

```
ValueError: operands could not be broadcast together with shapes (4,) (2,)
```

out of the extraction, so the **whole figure's** schema goes, not just this layer — and your diagnosis of why is precise: `values[:paired]` only caps, `values[: placed.size]` only caps again, and `placed & isfinite(values)` then broadcasts a mask of four against an array of two. Your pointer to `paired_arrays`'s docstring is the right one: this is the same hazard it was written to close for the 1-D path, arrived at again in two dimensions.

## What plotly actually does, measured both ways

Before folding `z` into `paired` I checked whether it should shorten the pairing unconditionally, and it should not:

| trace | plotly draws (bottom-first) |
|---|---|
| `histfunc="avg"`, `z=[10, 30]`, four x/y | `[[20, null], [null, null]]` — **two** samples used |
| `histfunc="sum"`, `z=[10, 30]`, four x/y | `[[40, 0], [0, 0]]` — same two |
| default `count`, `z=[10, 30]`, four x/y | `[[2, 1], [1, 0]]` — **all four**, exactly as with no `z` |

So `z` shortens the pairing only where it is read. Folding it in unconditionally would drop two samples the chart draws — which is a quieter bug than the crash, and worth a test of its own. Both directions are now pinned, and both mutations are killed:

```
KILLED  Z1 z does not shorten the pairing
KILLED  Z2 z shortens the pairing unconditionally
```

The truncation happens up front now, beside the x/y pairing, and `_aggregate`'s defensive `[: placed.size]` re-slice is gone — it was papering over exactly this.

## The other three, all taken

- **91 characters.** You're right, and right that `ruff check` was silent: ruff's default `select` does not carry `E501`, so the configured `line-length = 88` only binds the formatter. `ruff check --select E501` does flag it. Rewrapped.
- **`_values` computed twice.** Now one predicate, `_reduces_values(trace)`, used by both `_cell_name` and `_binned_grid` — so "does an aggregating histfunc have a z to reduce" is one rule rather than two that could drift.
- **The single-cell grid.** Fair catch on the gap between what was measured and what is regression-tested. It has a test now (`[[2.0]]`, one label on each axis) — the smallest chart there is, and the one an off-by-one shows up in.

All 1189 `tests/plotly` tests pass.

---

jooyoungseo:
The `histnorm` + aggregate question was the right one to press on — plotly does paint that cell, and I had it wrong. Fixed in `90daa04`.

## Measured, and you were right

Twenty combinations in Chromium — `avg`/`min`/`max`/`sum` against no norm and all four:

| histfunc | no histnorm | any histnorm |
|---|---|---|
| `avg` | `[[20, 40], [20, null]]` | `[[25, 50], [25, **0**]]` |
| `min` | `[[10, 40], [20, null]]` | `[[14.29, 57.14], [28.57, **0**]]` |
| `max` | `[[30, 40], [20, null]]` | `[[33.33, 44.44], [22.22, **0**]]` |

An empty cell is `NaN` without a norm and **0** with any of them. Rescaling runs over the whole grid without carrying the "no answer" marker through, so the two rules do not compose — exactly what `PlotlyHistogramPlot._extract_plot_data` documents for one dimension, and I had not carried it across. Deciding from `histfunc` alone announced "no value" on a cell the chart paints as zero.

`_as_payload` reads both now, and it is pinned by a 12-case parametrisation plus a control for the no-norm side. Two mutations against it, both killed:

```
KILLED  N1 histnorm ignored for the empty marker
KILLED  N2 an empty cell is always zero
```

The *values* were already right — `_normalised` counts a `None` cell as 0 toward the total, which is what gives 25/50/25 above. Only the marker was wrong.

## The other three, all taken

- **Counting loop.** Now `np.bincount` over the flattened cell index, so the work scales with the grid rather than the sample count — the shape `aggregate_bins` already has. `_aggregate` keeps its grouping, since `min`/`max`/`avg` need the values themselves. A mutation that transposes the flat index is killed.
- **Mismatched x/y without a `z`.** Added — the analogue of the short-`z` pair on the counting path, where `z` is not read at all.
- **The unindexed selector.** Stated in the class docstring where it is inherited, and filed as #647 with the `_SHARED_LAYERS` route that #643 established for `contour`/`histogram2dcontour`. You're right that it predates this class and that reading a second trace type into the layer is what makes it reachable a second way.

## The en dash

Worth confirming rather than assuming, so I checked what reaches the page: the schema goes through `json.dumps` with the default `ensure_ascii`, so the label is emitted as `–` — valid JSON, and it parses back to the en dash in the browser. No literal non-ASCII byte in the HTML.

## Re-verified

- 1203 `tests/plotly` tests pass (46 in this file).
- The browser end-to-end check re-run across all 18 figures: every cell and every bin label still matches `gd.calcdata`. 0 disagreements.

---

jooyoungseo:
Both actionable notes taken, in `a9a3ed8`.

**The test gap you identified is the right one.** The `histnorm` × aggregate pair was pinned only for what happens to an *empty* cell, while the populated ones — precisely the `_normalised` / `_aggregate`-`None`-padding interaction you name — were verified against the browser and left unguarded. Two of them now assert the scaled value:

| combination | expected | why it discriminates |
|---|---|---|
| `avg` + `percent` | `[[25, 0], [25, 50]]` | of the averages' **own** total of 80, not of the four samples |
| `avg` + `density` | `[[5, 0], [5, 10]]` | per unit of the 2 × 2 cell |

The percent case is the one that would catch a regression where `_normalised` took its denominator from the sample count instead of the values — the distinction `apply_histnorm` documents and which `avg` makes visible, since 20 + 20 + 40 ≠ 4.

**And the forward comment.** Your read of `_normalised` as "losing" the zero-fill is exactly the reading the code invited, so it now says what it does and does not do: it scales numbers, and leaves an empty cell to `_as_payload`, which knows the `histnorm` as well as the `histfunc`. Two questions kept apart — what a cell's number becomes, and what a cell with no number becomes.

The two notes I'm not acting on, and agreeing with your framing of both:

- **`_aggregate`'s Python grouping.** Right that it is asymmetric with the vectorised `_count`. `min`/`max`/`avg` need the values themselves rather than a tally, so the vectorised form would be a sort-and-segment — worth it if it ever shows up in a profile, not before.
- **`_edge` and scientific notation.** Real, and bounded the way you say: plotly's automatic widths are 1/2/5×10ⁿ, so it takes an explicit `xbins` of about `1e-10` to reach it. Worth remembering rather than guarding speculatively.

1205 `tests/plotly` tests pass.

---

Issue #646, opened 2026-08-23
A histogram whose rough bin width lands exactly on 2, 5 or 10 is read at bins plotly does not draw


jooyoungseo:
## What happens

`_plotly_dtick` in `maidr/plotly/histogram.py` rounds a rough bin width up to a
"nice" 1/2/5×10ⁿ value, mirroring plotly's `Axes.autoTicks`:

```python
base = 10 ** math.floor(math.log10(size0))
ratio = size0 / base
for nice in (2, 5, 10):
    if nice >= ratio * (1 - 1e-9):  # small tolerance for FP
        return base * nice
```

Plotly's `Lib.roundUp` is **strictly greater**, not `>=`. Its binary search
advances on `arrayIn[mid] <= val`, so it returns the first element *above* the
value. Where the two differ — `ratio` landing exactly on 2, 5 or 10 — py-maidr
announces a bin grid the chart does not draw.

## Measured

`nbinsx` makes the rough width exactly `range / nbins`, so an exact ratio is
reachable on demand. In Chromium:

| trace | rough width | ratio | plotly draws | py-maidr computes |
|---|---|---|---|---|
| `go.Histogram(x=linspace(0, 30, 61), nbinsx=15)` | 2 | **2.0** | **5** | 2 |
| `go.Histogram(x=linspace(0, 75, 76), nbinsx=15)` | 5 | **5.0** | **10** | 5 |
| `go.Histogram(x=linspace(0, 28.5, 58), nbinsx=15)` | 1.9 | 1.9 | 2 | 2 |
| `go.Histogram(x=linspace(0, 31.5, 64), nbinsx=15)` | 2.1 | 2.1 | 5 | 5 |

The two rules agree everywhere except the boundary, and on the boundary
py-maidr picks the width below. A reader is then told the chart has bins twice
as many and half as wide as the ones drawn, with every count wrong to match.

The `* (1 - 1e-9)` tolerance makes it worse rather than better: it widens the
band in which the loose rule fires, so a ratio a hair *below* 2 also takes the
wrong branch.

## The fix

```python
for nice in (2, 5, 10):
    if nice > ratio:
        return base * nice
```

All 1186 `tests/plotly` tests pass with the change — no existing test encoded
the loose rule, which is why it survived this long.

## Why it matters beyond the histogram

This is the whole of what #642 was blocked on. That issue tabulated nine z
ranges for `go.Contour`'s automatic levels and found that eight fit "round
[...]

Issue #644, opened 2026-08-23
No test resolves a plotly layer's selectors against the chart it describes


jooyoungseo:
## What is missing

Every plotly layer emits CSS selectors that are supposed to resolve, in the
browser, to the element each series or point describes. Nothing in the suite
checks that. `tests/plotly/` asserts on the emitted *strings*:

```python
assert "g.contour:nth-of-type(2)" in layer["selectors"][0]
```

which is a check that the code built the string it meant to build, not that the
string finds anything. A selector that silently stops matching — plotly.js
regrouping its DOM in a version bump, a class rename, a wrapper element
appearing — passes every test in the repo and reaches users as a chart whose
highlight has quietly stopped working.

## The matplotlib path does not have this hole

There the exported SVG is available in Python, and the `test` extra carries
`cssselect` for exactly this reason:

> "Resolves an emitted CSS selector against the exported SVG, which is the only
> way to check that a selector addresses the element its point describes."

A plotly figure has no Python-side SVG — plotly.js builds the DOM in the
browser — so the same check needs a browser.

## The harness already exists

`tests/browser/` runs a Shiny app in Chromium behind `--run-browser`, and CI
runs it as "Accessibility tests in a browser". What it does not have is a
plotly figure whose selectors get resolved. That is the gap, and it is a
smaller one than it looks: the fixtures, the browser discovery, the skip
behaviour and the CI job are all there.

## What the test would do

For a figure holding several plotly layer types:

1. render it through maidr and read the emitted schema;
2. for each layer, `document.querySelectorAll(selector)` each entry;
3. assert each resolves to **exactly one** element;
4. where the layer is positional (a bar per point, a curve per series), assert
   the element's geometry is the one the payload describes — unproject its
   position through the subplot's axes and compare with the point.

Step 4 is what catches the failure that matters. A selector resolving to *an*
element is easy; resolving to the *right* one is the property every one of
these layers claims. It is exactly the check that found the contour's
curve-ordering problem in #643 — plotly and `contourpy` order a level's islands
differently, so a selector list built in the tracer's order resolved to real
elements and the wrong ones. That was caught by a script run by hand, and
nothing in the repo would have caught it.

## Worth knowing before starting

- Several layers deliberately emit **no** selectors, each for a different
  measured reason: barpolar (#635), parcoords (#637), parcats (#639),
  choropleth (#640), and a contour whose level holds islands (#643). The test
  has to treat "no selectors" as a valid answer rather than a failure.
[...]

PR #649, opened 2026-08-23
feat(plotly): read a contour at the levels plotly picks for itself

### LINKED ISSUE #642: Read a plotly contour trace — what is measured, and what is left

jooyoungseo:
Closes #642.

`go.Contour` ships a grid, not curves, and — unless the author names both `start` and `end` — not levels either. #643 read the traces that name their own levels and declined the rest, because the rule plotly chooses levels by lives in plotly.js and nine measured fields would not fit one formula. That left every default `go.Contour(z=...)` unread.

## What the rule turned out to be

One comparison, not a missing formula. Measured step by step against the browser:

1. A rough step of `(zmax - zmin) / (ncontours or 15)`, rounded up to a 1/2/5×10ⁿ value by the same `roundUp` a histogram's bin width goes through — **strictly** greater, which is the whole of what made a field spanning `0 .. 3` look unreproducible. That was #646, fixed in #648; with it read correctly the fields stopped disagreeing.
2. `start`: the first multiple at or above `zmin`, moved up one when it lands *on* it.
3. `end`: the last multiple at or below `zmax`, moved down one when it lands on it.
4. If those cross — an `ncontours` too small to fit a multiple inside — both become their own midpoint and the field gets a single level.

Two consequences fall out and are handled rather than special-cased:

- **A trace naming both ends but no width is no longer declined.** Plotly keeps the ends and derives a width for them through the same round-up: measured on `start=0.2, end=0.8`, a missing `size` and a zero `size` both draw 0.05, and `ncontours=4` draws 0.2.
- **A constant field needs no guard of its own.** Its range is zero, so the round-up answers 1 and step 4 puts one level at the constant value — which is what plotly draws for it. That level *is* the whole field, so it holds no curve, and #636's guard drops the empty layer. The explicit guard this replaced was removed as redundant.

A `constraint` contour still declines: it draws one curve at `value` and means "the region beyond it", so reading its `start`/`end`/`size` would announce levels the chart does not draw.

## The four roundings

Steps 2 and 3 each round twice, and the four roundings pull in opposite directions on purpose:

- The multiple itself is taken **with a tolerance**, because a range that divides evenly rarely says so in binary. `-0.3 / 0.05` is `-5.999999999999999`; rounding that at face value starts the list a whole level late.
- Whether the result then lands **on** the floor or ceiling is tested **exactly**, because the value the tolerance recovered is a hair off the bound rather than on it. `0.009 / 0.0001` rounds up to a top level of `0.009000000000000001` — 1.7e-18 from the ceiling — which plotly keeps and a 1e-9 comparison would drop.

Measured against plotly's drawn levels on **49 figures**: 26 z ranges from `0 .. 0.07` to `0 .. 1000` (positive, negative and straddling), 8 explicit `ncontours` from 1 to 30, and 15 ranges picked for landing on exactly those ties. All 49 agree level for level, and **each of the four roundings is the difference on at least one of them** — none is defensive padding.

## Level groups that hold no curve still count

Plotly gives every level in its list a `g.contourlevel` of its own, drawn or not, and the selectors index that list. A field ramping `-0.3 .. 0` gets six groups, the first at `-0.30000000000000004` — below the field's own floor, holding no path — so the five curves live in groups 2–6. Resolved in Chromium: every emitted selector lands on the group at its own level.

## Testing

- `tests/plotly/test_plotly_contour.py`: 43 tests (was 31). New coverage for the five ways plotly takes the levels over, the derived width, `ncontours`, the midpoint fallback, a field with no range, and both binary ties.
- A 17-mutation sweep over the level rule, each mutation reverted alone against a restored baseline: all 17 killed. The two ties above are what kill the four rounding mutations — they were the sweep's only survivors before these tests, and two of the four were not being mutated at all until this round.
- `tests/core` 1931 passed / 5 skipped; `tests/plotly` 1216 passed; the rest 185 passed / 13 skipped. `ruff check` clean.

Follow-up already filed: #644 (no browser test resolves plotly selectors in CI — every selector in this PR was resolved by hand in Chromium instead).

---

### DIFF
--- a/maidr/plotly/contour.py
+++ b/maidr/plotly/contour.py
@@ -275,6 +275,14 @@ def draws_its_lines(trace: dict) -> bool:
 def levels_of(trace: dict, z: Any = None) -> list[float] | None:
     """Return the levels plotly draws: the author's, or the ones it picks.
 
+    Which route applies is asked first, rather than read off a None coming
+    back from the first one. :func:`declared_levels` answers None for two
+    unrelated reasons -- the author did not name their own levels, and the
+    author named a *billion* of them -- and only the first is a reason to
+    pick levels instead. Falling through on both would read a runaway spec at
+    nine levels of someone else's choosing, which is a chart the author did
+    not write and plotly does not draw.
+
     Parameters
     ----------
     trace : dict
@@ -286,9 +294,31 @@ def levels_of(trace: dict, z: Any = None) -> list[float] | None:
     Returns
     -------
     list of float or None
-        The levels, or None when the trace is not a set of levels at all.
+        The levels, or None when there are none to read.
     """
-    return declared_levels(trace) or automatic_levels(trace, z)
+    if declares_levels(trace):
+        return declared_levels(trace)
+    return automatic_levels(trace, z)
+
+
+def declares_levels(trace: dict) -> bool:
+    """Whether the levels are the author's rather than plotly's to pick.
+
+    **Both ends** is what makes them the author's. Plotly coerces
+    ``autocontour`` to true when either ``start`` or ``end`` is missing, so a
+    trace naming one of them, or only a ``size``, has its levels picked for
+    it -- measured, all three draw the same nine levels as a trace naming
+    nothing at all. An explicit ``autocontour: True`` overrides even a
+    complete spec, and a ``constraint`` contour is not a set of levels at
+    all (see :func:`_levels_spec`).
+    """
+    contours = _levels_spec(trace)
+    if contours is None or trace.get("autocontour") is True:
+        return False
+    return (
+        _a_number(contours.get("start")) is not None
+        and _a_number(contours.get("end")) is not None
+    )
 
 
 def declared_levels(trace: dict) -> list[float] | None:
@@ -301,12 +331,10 @@ def declared_levels(trace: dict) -> list[float] | None:
     ``start=0.2, size=0.2`` both routes agree on 0.4 and disagree on the
     third level, and the level is also what is handed to the curve tracer.
 
-    **Both ends** is what makes a spec the author's. Plotly coerces
-    ``autocontour`` to true when either ``start`` or ``end`` is missing, so a
-    trace naming one of them, or only a ``size``, has its levels picked for it
-    -- measured, all three draw the same nine levels as a trace naming nothing
-    at all. :func:`automatic_levels` takes those, and takes an explicit
-    ``autocontour: True``, which overrides a complete spec.
+    What makes a spec the author's is :func:`declares_levels`; None here
+    means either that it is not theirs or that theirs runs away (see
+    :func:`_stepped`), which is why :func:`levels_of` asks which case it is
+    before reading this.
 
     With both ends named, a missing or zero ``size`` is **not** a decline:
     plotly keeps the ends and derives a width for them, through the same
@@ -314,13 +342,13 @@ def declared_levels(trace: dict) -> list[float] | None:
     no size and a zero size both draw a width of 0.05, which is
     ``(0.8 - 0.2) / 15`` rounded up, and ``ncontours=4`` draws 0.2.
     """
-    contours = _levels_spec(trace)
-    if contours is None or trace.get("autocontour") is True:
+    if not declares_levels(trace):
         return None
 
+    contours = _levels_spec(trace) or {}
     start = _a_number(contours.get("start"))
     end = _a_number(contours.get("end"))
-    if start is None or end is None:
+    if start is None or end is None:  # pragma: no cover - `declares_levels`
         return None
 
     low, high = (start, end) if start <= end else (end, start)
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Thanks — all three addressed in f4031b3, and the second nit turned out to be hiding a real defect rather than a style question.

**The masked-array coupling.** You're right, and right about why: `np.asarray` on a masked array gives the buffer *under* the mask, so the `isfinite` filter worked because the NaN that caused the masking is still sitting there. `automatic_levels` now reads `np.ma.filled(np.ma.asarray(z, dtype=float), np.nan)` before filtering, which says "only the finite, unmasked values count" rather than depending on `_grid`'s masking strategy. `test_a_masked_field_picks_its_levels_from_what_is_left` masks a *finite* peak by hand — the case where the two readings differ — and fails on the old line.

**The `or` in `levels_of` — this one was a bug, not a nit.** You noted it relies on `_stepped` never returning an empty list. It also relies on `declared_levels` returning None for only one reason, and it returns None for two: "the author did not name their own levels", and "the author named more than a thousand of them", which is the runaway guard `_stepped` exists to enforce. So `go.Contour(z=..., contours=dict(start=0, end=1, size=1e-9))` — the exact spec `test_a_spacing_that_asks_for_a_billion_levels_is_declined` guards — fell through to `automatic_levels` and came back as a nine-level contour. That test called `declared_levels` directly, so it never saw it; the new `test_a_runaway_spec_declines_rather_than_falling_back_to_automatic` goes through `_layers` and asserts the figure gets no contour layer at all. `levels_of` now asks `declares_levels(trace)` first and commits to that route, so a decline on the author's route is a decline.

**`ncontours` of 0 or negative.** Measured rather than reasoned about: plotly's own validator rejects both — `go.Contour(z=..., ncontours=0)` raises `ValueError`, as does `-3` — so a trace built through `graph_objects` cannot carry one and the falsy-`or` branch is unreachable by contract. Noted in the docstring as exactly that, rather than left to read as a verified data point.

Both fixes are mutation-checked: reverting either line alone turns its test red. `tests/core` 1931 passed / 5 skipped, `tests/plotly` 1218 passed, the rest 185 passed / 13 skipped.

---

PR #652, opened 2026-08-23
fix(plotly): honour a bin spec that names a start or an end but no size

### LINKED ISSUE #650: A plotly bin spec that names a start or an end but no size is ignored outright
## What I measured

`compute_bin_edges` in `maidr/plotly/histogram.py` honours `xbins.start` and `xbins.end` **only inside the `"size" in bins` branch**. A spec that names a start (or an end) without a size falls through to the automatic path, which never looks at either, so the author's bin boundaries are discarded.

plotly 6.7.0, Chromium, `go.Histogram(x=[1,1,2,2,2,3,3,4,5,5,5,5,6,6,7,8,8,9,9,10], xbins=...)`, comparing plotly's resolved `_fullData[0].xbins` against `compute_bin_edges`:

| `xbins` | plotly's edges | py-maidr's edges |
|---|---|---|
| *(none)* | −0.5, 1.5, 3.5, 5.5, 7.5, 9.5, 11.5 | same ✓ |
| `{"size": 2}` | −0.5, 1.5, … | same ✓ |
| `{"start": 0, "size": 2}` | 0, 2, 4, 6, 8, 10, 12 | same ✓ |
| **`{"start": 0}`** | **0, 2, 4, 6, 8, 10, 12** | −0.5, 1.5, 3.5, 5.5, 7.5, 9.5, 11.5 |
| **`{"start": 0, "end": 12}`** | **0, 2, 4, 6, 8, 10, 12** | −0.5, 1.5, … |
| `{"end": 12}` | −0.5, 1.5, … | same ✓ |

The `end`-only row agrees by luck on this data — plotly's automatic end already landed inside it — rather than because the branch reads `end`.

## What the reader gets

Every bin is announced by the range it covers, so a shifted edge set misreports every bin's label. It also misreports the **counts** whenever the author's start is not half a bin from the automatic one. Same data, `xbins={"start": 0.5}`:

```
plotly   start=0.5 size=2 end=11.5   counts [5, 3, 6, 3, 3]      (5 bars)
py-maidr edges -0.5 … 11.5           counts [2, 5, 5, 3, 4, 1]   (6 bins)
```

Six bins announced for a chart that draws five, and not one of the numbers is a number on the chart.

## Where it reaches

`compute_bin_edges` is the shared binning rule, so this is not only `go.Histogram`:

- `histogram` (1-D), both orientations — `xbins` / `ybins`.
- `histogram2d` (#645) — both axes, via `binned_cells`.
- `histogram2dcontour`, once it is read, where it also decides the grid's **coordinates**: I hit this while reading that trace, where an `xbins={"start": 0}` put every traced curve half a bin off plotly's.

## What plotly does

The automatic path computes a size and then a start; an explicit `start` simply replaces the computed one, and an explicit `end` replaces the computed end, both independently of whether the size was given. That is the same precedence the `"size" in bins` branch already implements — it is just not reached when the size is absent.

## Suggested shape

Lift the `start`/`end` handling out of the explicit-size branch so it applies to the computed size too: work out the size (explicit, `nbins` hint, or automatic), then apply the author's `start`/`end` over the top. The anti-clustering shift (`_auto_shift_bins`) should stay skipped when the author gave a `start`, which is what the existing branch already does and what the measurements above confirm.

jooyoungseo:
Closes #650.

`compute_bin_edges` read `xbins.start` and `xbins.end` only inside its explicit-size branch. A spec that named a window without a width fell through to the automatic path, which never looked at either — so the author's bin boundaries were discarded outright.

What that cost, measured in Chromium on twenty integers from 1 to 10 with `xbins=dict(start=0.5)`:

```
plotly draws     5 bars   [5, 3, 6, 3, 3]   from 0.5 in steps of 2
py-maidr said    6 bins   [2, 5, 5, 3, 4, 1] from -0.5
```

Not one of the six numbers is a number on the chart, and not one of the six labels names a bar.

## The shape of the fix

The two paths become one: work out the width (the author's, an `nbins` hint rounded up, or the automatic one), then apply the author's `start` and `end` over it. That fold is only safe because the two paths seeded `_auto_shift_bins` differently and the seeds converge — which `TestBothCallersSeedTheShiftEquivalently` was already pinning, for exactly this reason. Its docstring now says what it is guarding.

## Four things measured on the way, each its own test

- **A window that is not a whole number of bins keeps its part-bin.** Plotly steps while the *bin's own* start is below `end`, so `start=0.5, end=9, size=2` draws five bars, not the four that rounding the span down gives — and the four would lose the samples at 9, 9 and 10. The old `end + size/2` convention was right only when the span's fractional part was zero or at least a half.
- **The last bin is half-open like the rest.** With a window ending at 6, plotly drops the samples sitting exactly on it; numpy's closed final interval folded them in and announced `[4, 6)` holding seven of a chart's five. The two-dimensional reading of the same binning settled this the same way in #645, measured there on its own grid. Costs nothing where the window is plotly's own — a derived end is a whole bin past the last value, so nothing can sit on it.
- **A `size` of 0 is an absence, and it takes an `nbins` hint with it.** Measured: `xbins=dict(size=0)` draws the automatic bins, and `nbinsx` of 4 and of 12 both draw those same six bars — writing `size` at all discards the hint. Reading the hint anyway would announce twelve. A zero size also used to reach a division by it.
- **A sample with no spread is one bin wide.** Not the two that rounding one up gives: a run of 3s is binned 2.5 – 3.5, and with `start=0` from 3 to 4.

A **negative** width now declines rather than answering with one bin holding everything. Plotly draws ten bars for `size=-2` over those integers, one per distinct value, by a route worth neither guessing at nor reproducing for a spec that cannot be meant — and a single bin would misreport a chart drawing ten, which is what #636 settled against.

## Verification

- **168/168 emitted bins match the drawn bars.** Four samples (clustered integers, gaussian, wide uniform, constant) × 14 `xbins` shapes × 3 `nbinsx` settings, comparing py-maidr's emitted bin centres and counts against plotly's `calcdata` in Chromium. Every case, not just the edges.
- **44/44 on the 2-D path**, which shares this function through `histogram2d`.
- A 9-mutation sweep over the new code, each mutation reverted alone against a restored baseline: all 9 killed. Two survived the first pass and both were acted on — one got the test it was missing (a span landing a hair over a whole number, which needs data past the window to be observable at all, since an empty extra bin is trimmed before anyone hears it), and the other was a redundant guard `np.arange` already handled, now removed.
- `tests/core` 1931 passed / 5 skipped; `tests/plotly` 1228 passed; the rest 185 passed / 13 skipped. `ruff check` clean.

Found while reading `histogram2dcontour` for #627, where the same gap put every traced curve half a bin off the chart.

---

### DIFF
--- a/maidr/plotly/grouped_histogram.py
+++ b/maidr/plotly/grouped_histogram.py
@@ -185,7 +185,7 @@ class PlotlyGroupedHistogramPlot(PlotlyPlot):
 
         data: list[list[dict]] = []
         for trace, sample, other in zip(self._traces, samples, values):
-            counts, _ = np.histogram(sample, bins=edges)
+            counts = _bin_counts(sample, edges)
             first, last = _occupied_span(counts)
             if first is None:
                 # A trace with nothing in the grid still holds a place in the
@@ -262,3 +262,19 @@ def _bin_assignment(arr: np.ndarray, edges: np.ndarray) -> np.ndarray:
     from maidr.plotly.histogram import PlotlyHistogramPlot
 
     return PlotlyHistogramPlot._bin_assignment(arr, edges)
+
+
+def _bin_counts(arr: np.ndarray, edges: np.ndarray) -> np.ndarray:
+    """How many observations landed in each bin.
+
+    Counted off the assignment above rather than ``np.histogram``, for the
+    reason that function exists: ``np.histogram`` closes its final bin and the
+    assignment does not, so a value sitting exactly on a shared window's
+    ``end`` was dropped by one path and folded into the last bin by the other
+    -- in the same layer, with which one applied decided by the ``histfunc``.
+    Measured on two traces sharing ``xbins=dict(start=0, end=6, size=2)``:
+    plotly draws the top bin holding two and the count path announced three.
+    """
+    from maidr.plotly.histogram import PlotlyHistogramPlot
+
+    return PlotlyHistogramPlot._bin_counts(arr, edges)
--- a/maidr/plotly/histogram.py
+++ b/maidr/plotly/histogram.py
@@ -137,6 +137,14 @@ def _plotly_dtick(size0: float) -> float:
     return base * _plotly_round_up(size0 / base, [2, 5, 10])
 
 
+#: How far a bin count may overshoot a whole number and still be read as one.
+#: ``(end - start) / size`` decides how many bins a spec asks for, and the
+#: division rarely comes out even in binary -- ``(8 - 0) / 2`` is exact but a
+#: hair over on other spellings, and rounding that up would add a bin the
+#: chart does not draw.
+_BIN_COUNT_TOLERANCE = 1e-9
+
+
 def _auto_shift_bins(
     bin_start: float,
     data: np.ndarray,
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Both correctness points confirmed against the browser and fixed in 5e9523a — the second one turned out to be reachable, not just theoretical.

**The grouped-histogram gap is real, and your repro is exactly right.** Measured on it in Chromium: plotly draws the first trace as `[0, 2) = 1`, `[2, 4) = 2`, `[4, 6) = 2` — the sample at 6 sits on the window's `end` and is in no bin plotly made. py-maidr announced `1, 2, 3`. So the file was internally consistent and wrong before this PR, and internally *inconsistent* after it, with the `histfunc` deciding which reading a reader got.

`grouped_histogram.py` now counts through `_bin_counts`, the same `bincount`-over-`_bin_assignment` the single-trace path uses. Two tests, both failing on the old line: one asserting the counts `[1, 2, 2]` / `[1, 1]` plotly draws, and one putting a 99 on the boundary sample under `histfunc="avg"` so the aggregating path's agreement is asserted beside the counting path's rather than left to be re-derived — an average that folded the sample in would answer 50 where it answers 1.

**The `np.bincount` invariant was already violable, and by an ordinary input.** You flagged that a NaN compares `False` to both outside checks; it does, and `searchsorted` sorts it *past* the final edge, so it came back as a bin one past the end:

```
edges [0, 2, 4, 6], values [1.0, nan, 3.0, 5.0]
  assignment  [0, 3, 1, 2]        <- 3 is not a bin
  _bin_counts [1, 1, 1, 1]        <- four counts for a three-bin grid
```

That is a counts array longer than the grid it belongs to, and a NaN reaches here from any `x` carrying a non-numeric entry. `_bin_assignment` now marks non-finite values outside every bin, which is also the right reading independently — a value that is not a number is not an observation rather than an observation of zero, as #405 settled. Fixed at the assignment so the aggregate path gets it too; it was silently writing into the same out-of-range index.

**The `"size" not in named` presence check** now says at the line itself that it is deliberately a presence check and not a truthiness one, and why the same key is asked about two different ways.

`tests/core` 1931 passed / 5 skipped; `tests/plotly` 1252 passed; the rest 185 passed / 13 skipped. Both new fixes are mutation-checked: reverting either line alone turns its test red.

---

Issue #651, opened 2026-08-23
A hole in a plotly contour's field is read as a hole, but plotly fills it before tracing


jooyoungseo:
## What I measured

`maidr/plotly/contour.py`'s `_grid` masks a missing cell, on the reading that a hole stops the curves the way plotly's do. Plotly does not stop there — it **interpolates the missing point and traces through it**.

Straight from `calcdata`, plotly 6.7.0 in Chromium, on a 5×5 field with its centre cell set to `None`:

```
contour, centre cell None      -> calcdata z[2][2] = 0.6      _emptypoints = 1
heatmap, centre cell None      -> calcdata z[2][2] = None     _emptypoints = null
heatmap, connectgaps=True      -> calcdata z[2][2] = 0.6      _emptypoints = 1
```

So the fill is not a property of the data, it is a step plotly runs for contour traces (and for a heatmap that asks for it): `findEmpties` + `interp2d`, an iterative average of each missing point's orthogonal neighbours.

## What it costs the reading

The curve **counts** can agree while the curves themselves move. On a 9×9 gaussian with one cell punched on the flank of the peak, `contours.coloring="lines"`, comparing every emitted series against the geometry of the path it is announced with (80 samples per path, point-to-segment distance):

```
whole field (control)   9 levels   worst gap = 0.16 data units
one cell punched        9 levels   worst gap = 0.91 data units
```

The grid step is 0.5, so the control's 0.16 is polyline-versus-bezier sampling noise and the punched field's 0.91 is most of two grid cells. Same nine levels, same one curve each — and a reader tracing a curve near the hole is told about ground the chart does not draw a curve through.

## Why the existing test did not catch it

`test_a_hole_in_the_grid_stops_the_curves_the_way_plotly_does` was measured on a field with the *peak* punched out. Both readings drop the level the hole ate: masking drops it because the region is gone, and interpolation drops it because the filled value is the average of the neighbours, which is below the peak. The measurement was consistent with either behaviour and did not distinguish them — the docstring's "stops the curves the way plotly does" overclaims what it checked.

## Scope

- `contour` — today, as above.
[...]

PR #653, opened 2026-08-23
feat(plotly): read a histogram2dcontour as the contour of its binned counts

### LINKED ISSUE #627: Fifteen plotly trace types register nothing although the core already draws them

jooyoungseo:
Closes #627 — the fifteenth and last of its declined trace types.

`histogram2dcontour` is the only one that is two readings at once: it bins samples the way a `histogram2d` does and then draws the curves along which those *counts* are constant, the way a `contour` does. Both halves already exist (#645 and #643/#649), so this is where they meet. `PlotlyHistogram2dContourPlot` extends `PlotlyContourPlot` and overrides exactly two things — which grid the curves run through, and which values the levels come from.

## What differs from either half alone

**The grid is one bin wider at each automatic edge**, so the curves have somewhere to close. Which edges move is per-side, and not simply "the automatic ones" — it reads that way at first and it is wrong. From the plotly bundle this wheel ships, where `l` is what the author wrote:

```js
l.size || (P.start = tickIncrement(P.start, P.size, true)),
l.end === void 0 && (P.end = tickIncrement(P.end, P.size, false))
```

So the **low** edge moves unless a `size` was named — not unless a `start` was — and the **high** edge moves unless an `end` was. Confirmed across nine spellings of `xbins` on five samples: with only a `size`, the low edge stays and the high edge moves; with only an `end`, the reverse.

**The curves run through the bin centres**, not the edges. Plotly's own `calcdata` for this trace carries five coordinates for five bins where a `histogram2d` carries four edges for three.

**A cell nothing landed in traces as zero.** The heatmap reading of the same binning announces it as `None` — there is no average of nothing — but the contour reading cannot: plotly hands the grid to a tracer written in JavaScript, where a `null` compares as below every level. Not a small difference. A sparse `histfunc="avg"` grid whose four filled cells run 3.5 .. 19 draws ten curves in the browser; read as a field with holes in it, `contourpy` finds five and misses the top three levels entirely. With the gaps at zero the two agree curve for curve — on that grid, on it negated, and on a `min` of negative values where zero is *above* every level.

The **levels**, though, still come from the values that are there: plotly reports that grid's range as 3.5 to 19 and draws levels 4 .. 18. Handing the zero-filled grid to the level rule instead draws 2 .. 16, none of which is a level on the chart.

(Worth stating because it looks like it should be otherwise: plotly *does* interpolate a hole in a `go.Contour`'s own `z` before tracing it — measured, a punched cell comes back filled in `calcdata` — but a `histogram2dcontour`'s grid never goes through that step. That divergence for the plain contour is filed as #651.)

## Verification

Read end to end against the browser on 13 figures — automatic bins, an `nbins` hint, four spellings of `xbins`, three `histfunc`s, a `histnorm`, an `ncontours` and an explicit level spec. **Eleven agree curve for curve.** The two that do not are levels whose curves reach the grid's own edge, where plotly's `<path>` elements are the outlines of the filled regions rather than the curves themselves. Every one of those levels has more than one curve, so the layer declines its selectors there anyway — and measured separately, a level whose *single* curve ends on the grid edge does match its drawn path exactly, in all three `coloring` modes. So the selector guard needs no new case.

A 9-mutation sweep over the new module — no widening, widening both sides, each edge rule inverted, edges instead of centres, gaps as holes, levels from the filled grid, the third axis unnamed, the colour bar title ignored: all 9 killed.

`tests/core` 1931 passed / 5 skipped; the rest 1437 passed / 13 skipped, including 21 new tests. `ruff check` clean.

## Two changes that reach the plain contour

- **A curve that spans nothing is no longer a series.** Where the field grazes a level at exactly one grid point, `contourpy` returns a closed curve whose every vertex is that point — and the level that grazes it is rarely the cell's value written out, so those vertices differ in the last bits rather than being identical. The span is now compared against the narrowest grid cell, at a fraction no real crossing comes near. Plotly draws no path there either.
- **The colour bar title is read in one place.** Three layers now name their third axis from it — a heatmap, a choropleth and this — and a third spelling of the same three cases is how they drift apart.

---

### DIFF
--- a/maidr/plotly/choropleth.py
+++ b/maidr/plotly/choropleth.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.enum.plot_type import PlotType
-from maidr.plotly.plotly_plot import PlotlyPlot, as_list
+from maidr.plotly.plotly_plot import PlotlyPlot, as_list, colorbar_title
 
 #: What the two halves of a region's reading are, when the author named
 #: neither. A choropleth has no cartesian axes to read titles off, and these
@@ -68,7 +68,7 @@ class PlotlyChoroplethPlot(PlotlyPlot):
         region, value = _AXIS_FALLBACKS
         return {
             MaidrKey.X: self._axis_config(label=region),
-            MaidrKey.Y: self._axis_config(label=self._colorbar_title() or value),
+            MaidrKey.Y: self._axis_config(label=colorbar_title(self._trace) or value),
         }
 
     def _extract_plot_data(self) -> list[dict]:
@@ -90,16 +90,6 @@ class PlotlyChoroplethPlot(PlotlyPlot):
             for location, value in zip(locations, values)
             if value is not None
         ]
-
-    def _colorbar_title(self) -> str:
-        """The title the author gave the colour bar, or an empty string."""
-        colorbar = self._trace.get("colorbar")
-        if not isinstance(colorbar, dict):
-            return ""
-        title = colorbar.get("title", "")
-        if isinstance(title, dict):
-            title = title.get("text", "")
-        return str(title) if title else ""
 
 
 def is_choropleth_trace(trace: dict) -> bool:
--- a/maidr/plotly/contour.py
+++ b/maidr/plotly/contour.py
@@ -588,8 +588,12 @@ def _has_extent(curve: Any, cell: float) -> bool:
 def _smallest_step(x: list[float], y: list[float]) -> float:
     """The narrowest cell in the grid, as the scale a curve is measured on.
 
-    The narrowest rather than the average, so an unevenly spaced grid is
-    judged by its own finest detail rather than by a figure no cell has.
+    The narrowest rather than the widest or the average, and deliberately so
+    even though no measured grid tells them apart: the fraction below is
+    small enough that any of the three separates float noise from a crossing.
+    The narrowest gives the *lowest* floor, so it is the one choice that
+    cannot drop a real curve on the fine part of an unevenly spaced grid --
+    which is the failure that would be silent.
     """
     steps = [
         abs(b - a)
--- a/maidr/plotly/contour.py
+++ b/maidr/plotly/contour.py
@@ -44,6 +44,13 @@ _DEFAULT_LEVEL_COUNT = 15
 #: the difference decides a whole level (see :func:`automatic_levels`).
 _MULTIPLE_TOLERANCE = 1e-9
 
+#: What fraction of the narrowest grid cell a traced curve must span to be a
+#: curve rather than the point where the field grazes a level. Far above the
+#: last bits that separate five copies of one vertex (1.6e-15 on a grid of 5s,
+#: measured) and far below any crossing, which is interpolated along a cell
+#: edge and so spans a real part of one.
+_POINT_FRACTION = 1e-9
+
 #: How far past ``end`` plotly's level loop still steps, as a fraction of
 #: ``size``. Measured rather than assumed, across seven specs: a run at
 #: ``start=0.2, end=0.8, size=0.05`` draws a level at ``0.8000000000000002``
@@ -124,14 +131,16 @@ class PlotlyContourPlot(PlotlyPlot):
         """
         self._series_levels = []
 
-        grid = _grid(self._trace)
+        grid = self._field()
         if grid is None:
             return []
 
         x, y, z = grid
         # The field, not just the trace: an automatic contour chooses its
-        # levels from the range of what it draws.
-        levels = levels_of(self._trace, z)
+        # levels from the range of what it draws. Asked for separately from
+        # the grid because the two are not always the same array -- see
+        # `_level_field`.
+        levels = levels_of(self._trace, self._level_field(z))
         if not levels:
             return []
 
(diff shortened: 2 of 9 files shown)


jooyoungseo:
All three taken in 4b3573a — the first was the right thing to insist on.

**The extent filter now has its own test in `test_plotly_contour.py`.** You're right that it changes the already-shipped reading and had nothing pinning it there; my own mutation sweep for this PR covered the new module and never touched `_has_extent`. The fixture is a field with a dimple in it — one interior cell alone, everything around it above — and plotly gives the grazed level a `g.contourlevel` with **no path in it** while drawing the ring at the level above. So the assertion is that the layer's one curve is at the upper level *and* points at the second group: the empty one is still counted.

Sharpening it exposed the more interesting half. A test that only asserted the degenerate curve is dropped would have passed with `span > 0` — the dimple's grazing curve spans exactly zero. The case that needs the grid-relative comparison is a level that arrives as `0.39999999999999997` grazing a cell holding exactly 0.4: the five vertices then differ in their last bits and span 1.6e-15 on a grid whose cells are 5 wide. That is the `histnorm` figure, so its test now asserts twelve curves as well as nine levels, and reverting the comparison turns it red.

The remaining two mutations there — the narrowest cell versus the widest, and the shared binning — survive, and each for its own reason. The narrowest/widest choice has no measured case that separates them at a fraction of 1e-9; the docstring now says so, and says why the narrowest is still the one to pick (it gives the lowest floor, so it is the choice that cannot silently drop a real curve on the fine part of an uneven grid). The binning cache is an optimisation with no observable behaviour by construction.

**Binned once.** `_binned()` computes it lazily and both halves read it, so it no longer depends on which of `_field`/`_level_field` runs first either.

**`extended_edges` moved to `histogram2d.py`**, where it is called from. The deferred import is gone and `histogram2dcontour.py` is a one-directional consumer, matching the layering everywhere else here. The rule is still about how a `histogram2dcontour`'s grid differs, and its docstring says so — but it lives with its caller.

`tests/core` 1931 passed / 5 skipped; the rest 1438 passed / 13 skipped. `ruff check` clean.

---

Issue #670, opened 2026-08-25
Eight `seaborn.objects` marks are not read, and five of them have an exact home


jooyoungseo:
> **Corrected.** The first version of this issue said the marks that read do so "by coincidence", because py-maidr patches `Axes` methods and `seaborn.objects` calls none of them. That is wrong: `maidr/patch/seaborn_objects.py` wraps `Plotter._plot_layer` and reads the marks **deliberately**, dispatching on the mark's class name, and `tests/core/plot/test_seaborn_objects.py` asserts each unread mark registers nothing on purpose. The measurement below stands; the explanation of it did not.

`seaborn.objects` has thirteen marks. Five are read; eight are not.

## Measured

Every mark through `FigureManager.get_maidr(plotter._figure)`, `seaborn 0.13.2`. (The figure matters: `so.Plot.plot()` builds its **own** `Figure` rather than drawing into pyplot's current one.)

| mark | artist it leaves | read as |
| --- | --- | --- |
| `Dot`, `Dots` | `PathCollection` | `scatter` |
| `Line`, `Path` | `Line2D` | `line` |
| `Bar` | `Rectangle` + `BarContainer` | `bar` |
| `Lines`, `Paths` | `LineCollection` | — |
| `Bars` | `PatchCollection` | — |
| `Area`, `Band` | `Polygon` | — |
| `Range`, `Dash` | `LineCollection` | — |
| `Text` | none of `lines` / `collections` / `patches` | — |

An unread mark registers nothing rather than being read wrongly, which `_READINGS` is explicit about: dispatch is by class **name** because seaborn's hierarchy does not track what a mark draws (`Dash < Paths` and `Range < Paths`, and neither draws what `Paths` does), so ancestry would claim marks it cannot read. That posture is right and this issue does not propose changing it — only extending the table, one measured mark at a time.

## The geometry each unread mark leaves

Measured on `x = 0..3`, `y = 1, 3, 2, 4`, `g = a a b b`:

**`Area`** — one `Polygon` per group. The baseline run forward, then the values backward:

```
[[0,0], [1,0], [2,0], [3,0], [3,4], [2,2], [1,3], [0,1], [0,0]]
 └──── baseline, n points ────┘ └──── values, reversed ────┘
```

so the series is the second half reversed — `(0,1) (1,3) (2,2) (3,4)`, exactly the data. Split by `color=`, one polygon per group, each self-contained.

[...]

PR #673, opened 2026-08-25
fix(seaborn-objects): name a colour split from the legend wherever it was put

### LINKED ISSUE #672: A colour-split `so.Line` gives two series a reader cannot tell apart
`so.Plot(..., color="g").add(so.Line())` draws two lines and reads as two series carrying **no names at all** — no per-point `z`, and no `axes.z` naming the variable they are split by. The identical chart written with the classic function names both.

## Measured

`seaborn 0.13.2`, same frame, `x = 0..3` twice, `g = a a b b`. The emitted layer, everything but `data`:

**`sns.lineplot(data=..., x="x", y="y", hue="g")`**

```json
{"type": "line",
 "axes": {"x": {"label": "x"}, "y": {"label": "y"}, "z": {"label": "g"}},
 "selectors": ["g[id='maidr-4c8c…'] path", "g[id='maidr-4264…'] path"]}
first point: {"x": 0.0, "y": 1.0, "z": "a"}
```

**`so.Plot(..., color="g").add(so.Line())`**

```json
{"type": "line",
 "axes": {"x": {"label": "x"}, "y": {"label": "y"}},
 "selectors": ["g[id='maidr-0f61…'] path", "g[id='maidr-72b1…'] path"]}
first point: {"x": 0.0, "y": 1.0}
```

Same two series, same two selectors. One announces "a" and "b" against a `g` axis; the other announces two unnamed lines. That is the shape xability/maidr#828 exists to prevent, reached from a chart that *has* the names.

## Root cause

**A `so.Plot`'s legend is the figure's, never the axes'.** Measured on every colour-split mark: `ax.legend_` is `None` and `fig.legends` holds exactly one, whose title is the grouping variable (`'g'`), whose texts are the group names, and whose handles carry the same colours as the drawn artists.

`maidr/util/legend_names.py::legend_of` already knows this — it falls back to the figure's legend when the axes has none, and `scatterplot.py` and `grouped_barplot.py` both carry comments about having moved onto it (#617). Two readers never did:

- `maidr/core/plot/lineplot.py:297` — `if self.ax.legend_ is not None:`, so `legend_labels` is empty and every series falls back to its own label, which seaborn sets to the `_child0` sentinel. Hence no `z`.
- `maidr/core/plot/maidr_plot.py:203` — `legend = self.ax.get_legend()` in `_legend_title`, so `axes.z` is omitted. Hence no variable name.

That is why `so.Dot(color=)` is fine and `so.Line(color=)` is not: the scatter path goes through `legend_of`, the line path does not.

```
so.Dot(color="g")   ->  2 layers, name "a" / "b"     (via hue_groups -> legend_of)
so.Bar(color="g")   ->  2 layers, name "a" / "b"     (via bar_groups -> legend_of)
so.Line(color="g")  ->  1 layer,  2 unnamed series   (via ax.legend_)
```

## Not only `seaborn.objects`

`legend_of`'s docstring already names the classic cases the axes-only read misses — a `PairGrid`, whose `add_legend()` builds one figure-level legend for the whole grid, and a `JointGrid`. A line layer in either is in the same position as the `so` one. Those are not measured here; `so.Line` is.

## Why it blocks the rest of #670

`_area_handover` declines a colour-split `so.Area` outright, and says why: *"The legend holding those is the figure's, built after every layer is drawn… Two unnamed series is the shape xability/maidr#828 exists to prevent, so this waits for the naming rather than shipping without it."* The same wall stands in front of `Lines` / `Paths`, whose one `LineCollection` carries a segment per group. Fixing the lookup unblocks all three and improves `Line` / `Path` today.

Related: #670 (the eight unread marks), #617 (the same move for the scatter and bar paths), #502 (naming a hue group by legend order), #582 (pairing by colour rather than by position), xability/maidr#828 (why two of a kind must be tellable apart).

jooyoungseo:
Closes #672.

A colour-split `so.Line` drew two lines and read as two series carrying **no names at all** — no per-point `z`, and no `axes.z` naming the variable they were split by. The identical chart written with the classic function names both.

## Measured

`seaborn 0.13.2`, same frame, two groups over the same positions. The emitted layer, everything but `data`:

| | `sns.lineplot(hue="g")` | `so.Plot(color="g").add(so.Line())` |
|---|---|---|
| series | 2 | 2 |
| selectors | 2 | 2 |
| `axes.z` | `{"label": "g"}` | **absent** |
| first point | `{"x": 0.0, "y": 1.0, "z": "a"}` | `{"x": 0.0, "y": 1.0}` |

Same two series, same two selectors. One announces "a" and "b" against a `g` axis; the other announces two unnamed lines — the shape xability/maidr#828 exists to prevent, reached from a chart that *has* the names.

## Root cause

**A `so.Plot`'s legend is the figure's, never the axes'.** Measured on every colour-split mark: `ax.legend_` is `None` and `fig.legends` holds exactly one, whose title is the grouping variable, whose texts are the group names, and whose handles carry the drawn artists' own colours.

`maidr/util/legend_names.py::legend_of` already knows that — it falls back to the figure's legend when the axes has none, and `scatterplot.py` and `grouped_barplot.py` both carry comments about having moved onto it (#617). Two readers never did:

- `maidr/core/plot/lineplot.py` — `if self.ax.legend_ is not None:`, so `legend_labels` was empty and every series fell back to its own label, which seaborn sets to the `_child0` sentinel. Hence no `z`.
- `maidr/core/plot/maidr_plot.py` — `legend = self.ax.get_legend()` in `_legend_title`. Hence no `axes.z`.

That is exactly why `so.Dot(color=)` was named and `so.Line(color=)` was not: the scatter path goes through `legend_of` (via `hue_groups`), the line path did not.

## The change

Two reads, routed through the helper that already exists for this. **Nothing else moves** — every rule about *which* legend answers stays `legend_of`'s: the axes' own wins, one figure legend names the axes, two figure legends are ambiguous and decline. The `_legend_title` import is function-local because `legend_names` reaches `scatterplot`, which reaches `maidr_plot`, so a module-level import would close the cycle.

## After

```
so.Line(color="g")  ->  axes.z {"label": "g"},  first point z "a"   (was: neither)
so.Dot(color="g")   ->  axes.z {"label": "g"}                        (was: named groups, unnamed variable)
sns.lineplot(hue=)  ->  unchanged
```

## Verification

- **9 tests** in `tests/core/plot/test_figure_legend_naming.py`: the reproduction, the variable's name, that the names go with the values they were drawn with (a swap is invisible in labels alone), the classic spelling reading the same way, `so.Dot`'s `z` label, a chart with no legend naming nothing rather than something, the axes' own legend still winning over the figure's, and two figure legends declining.
- **4 mutations swept, all killed.** Two survived the first pass: one was a refactor-equivalent rewrite rather than a behaviour change and was sharpened into a real one; the other was `.strip()` on the title, pre-existing but now on a path many more charts reach, so it is pinned by a test.
- `ruff check` clean on every changed file. The 6 re-export errors `ruff check maidr tests` reports are pre-existing — they are on `origin/main` too.
- **3462 passed, 72 skipped** (`uv run pytest -q`), up from 3454 by exactly the 8 tests added at the time of the run; the ninth followed the mutation sweep.

## Why this comes before the rest of #670

`_area_handover` declines a colour-split `so.Area` outright and says why: *"The legend holding those is the figure's, built after every layer is drawn… Two unnamed series is the shape xability/maidr#828 exists to prevent, so this waits for the naming rather than shipping without it."* The same wall stands in front of `Lines` / `Paths`, whose one `LineCollection` carries a segment per group. This is that naming.

---

### DIFF
--- a/maidr/core/plot/lineplot.py
+++ b/maidr/core/plot/lineplot.py
@@ -6,7 +6,7 @@ from matplotlib.lines import Line2D
 from maidr.core.enum.maidr_key import MaidrKey
 from maidr.core.plot.scatterplot import _rgba
 from maidr.util.artist_label import series_name
-from maidr.util.legend_names import names_for
+from maidr.util.legend_names import legend_of, names_for
 from maidr.util.confidence_band import band_edges_at
 from maidr.core.enum.plot_type import PlotType
 from maidr.core.plot.maidr_plot import MaidrPlot
@@ -294,8 +294,21 @@ class MultiLinePlot(MaidrPlot, LineExtractorMixin):
         # Try to get series names from legend
         ax_legend_source = self.ax
         legend_labels = []
-        if self.ax.legend_ is not None:
-            legend_labels = [text.get_text() for text in self.ax.legend_.get_texts()]
+        # `legend_of` rather than `self.ax.legend_`, which is what this read
+        # before. The two answer differently wherever the legend was not put
+        # on the axes, and a `seaborn.objects` chart is *always* that case:
+        # measured, `Plotter._make_legend` builds one **figure** legend and
+        # leaves every panel's own at `None`, so a colour-split `so.Line`
+        # found no labels, fell back to each line's own -- seaborn's
+        # `_child0` sentinel -- and announced two series a reader could not
+        # tell apart, where `sns.lineplot(hue=)` names both (#672).
+        #
+        # The same move the scatter and bar paths already made (#617), and
+        # `names_for` below has been going through it all along, which is why
+        # only this read was left behind.
+        legend = legend_of(self.ax)
+        if legend is not None:
+            legend_labels = [text.get_text() for text in legend.get_texts()]
 
         # Which line each legend entry belongs to. A legend is not always as
         # long as the series list, and pairing the two by position when it is
--- a/maidr/core/plot/maidr_plot.py
+++ b/maidr/core/plot/maidr_plot.py
@@ -183,6 +183,16 @@ class MaidrPlot(ABC, FormatExtractorMixin):
         Shared rather than restated: ``MultiLinePlot`` and ``PointPlot`` both
         need it and would otherwise drift apart if the convention changed.
 
+        Delegates to :func:`maidr.util.legend_names.title_of`, which answers
+        the same question and is what ``RugPlot`` already calls. The two used
+        to differ, and that difference *was* #672: this asked
+        ``ax.get_legend()`` while the names were read through
+        :func:`~maidr.util.legend_names.legend_of`, so a chart whose legend
+        sits on the figure -- every ``seaborn.objects`` chart, and a
+        ``PairGrid`` -- had its groups named and the variable they split by
+        dropped. Now that both go through the same lookup, keeping two copies
+        of it would only invite them apart again.
+
         Read **live**, when the schema is built, while the group *names* are
         captured once as the plotting call is patched. A caller who retitles
         or relabels the legend in between can therefore make the two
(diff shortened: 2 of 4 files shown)


jooyoungseo:
All three findings acted on; the first two are in `4c52c7b`.

**1. `_legend_title` duplicating `title_of`** — fixed by delegating. You are right that they are now byte-identical, and right about why that is dangerous: the difference between them *was* #672, so keeping two copies of the settled version only invites them apart again. `_legend_title` now returns `title_of(self.ax)`, keeping the function-local import for the same cycle reason.

**2. Stale comment in `rugplot.py`** — fixed. It said `_legend_title` "asks `ax.get_legend()` only", which this PR made untrue. Rewritten to state the defect in the past tense and note the two now agree, with `RugPlot` keeping the direct call because it is the one that states which lookup is meant.

**3. Box / boxen** — filed as #674, and it turned out to be sharper than "not reachable today". Measured, it needs no `so.Box` mark at all:

```python
sns.boxplot(data=frame, x="cat", y="val", hue="g", ax=ax)
#   -> type=box   z={'label': 'g'}

sns.boxplot(..., ax=ax, legend=False); fig.legend(handles, ["p", "q"], title="g")
#   -> type=box   z=None
```

That second form is the `PairGrid` / `JointGrid` shape `legend_of` exists for, so the same class of bug is live in `boxplot.py:274` and `boxenplot.py:179`/`:398` right now. I kept it out of this PR because `boxenplot` reads the legend twice and whether both want the same lookup needs checking against what each names — a small piece of reading rather than a mechanical substitution.

Full suite after the follow-up: 3463 passed, 72 skipped.

---

PR #686, opened 2026-08-27
docs: tell deck authors how to keep off-slide charts out of the tab order

jooyoungseo:
## Description

Adds a `## Quarto revealjs Slides` section to `docs/index.qmd`, between the keyboard shortcuts table and the braille reference.

A chart on a slide is already keyboard reachable on its own — Tab moves into it, the arrows explore it, Shift+Tab hands focus back to the slide, and Space advances the deck. That half needs no configuration. What does need attention is a reveal.js behaviour that has nothing to do with MAIDR: reveal.js keeps the slides on either side of the current one rendered so transitions stay smooth, and marking them `hidden` does not take them out of the tab order, because reveal's own inline style overrides the attribute. On a deck with a chart on every slide, one Tab therefore lands on an off-screen slide's chart rather than the one in front of the reader.

Upstream this is hakimel/reveal.js#1587, open since 2016. The fix has since landed on reveal.js `master`, which marks every slide but the current one `inert`. It is not in a published release yet, and Quarto carries its own copy — Quarto 1.10 ships reveal.js 5.1.0 — so it will reach decks in some Quarto release after reveal.js cuts one. The section says so, and says that nothing in the deck will need to change when it does.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Documentation update

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas — not applicable, no code changed

## Changes Made

`docs/index.qmd`, 48 lines added, no other file touched:

- what the keyboard already does in a deck, and that it needs no configuration
- why one Tab can land on an off-screen slide's chart, with the upstream issue
- that reveal.js `master` fixes it and the deck will not need to change when Quarto ships it
- until then, `quarto add mcanouil/quarto-revealjs-a11y` plus the YAML to enable it
- that 0.2.3 or newer is required: earlier versions parked off-slide elements at `tabindex="-1"` and could not find them again to restore them, which left the chart on the *current* slide unreachable too

## Additional Notes

Checked in Chromium against a Quarto `revealjs` deck of matplotlib charts rendered by this package, not reasoned about. Without the extension, Tab from slide 2 or 3 lands on slide 1's chart:

```
slide 2 -> iframe @slide-1-bar
slide 3 -> iframe @slide-1-bar
```

With quarto-revealjs-a11y 0.2.3, each slide gives one path in and out, and Space still advances:

```
slide 2: chart after 2 Tab(s) -> iframe @slide-2-line*    | Shift+Tab -> skip-link | Space 2->3
slide 3: chart after 2 Tab(s) -> iframe @slide-3-scatter* | Shift+Tab -> skip-link | Space 3->4
```

The 0.2.2 warning is from a measured regression, not caution. Its restore query was `[tabindex]:not([tabindex="-1"])`, which cannot match anything it had itself set to `-1`; 0.2.3 replaced that mechanism with `inert` on the slide.

The section was rendered on its own with `quarto render` to confirm the code blocks and links come out right. No executable cell was added — a chunk shown inside a fenced block is still executed by the engine, so the section sticks to YAML and shell blocks.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Eadkcqpkod61zhhbfy8Wj8)_

### DIFF
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -470,7 +470,7 @@ A chart on a `revealjs` slide is keyboard reachable on its own: **Tab** moves
 into it, the arrow keys explore it, and **Shift+Tab** hands focus back to the
 slide, so **Space** advances the deck again. None of that needs configuring.
 
-What does need attention is a reveal.js behaviour that has nothing to do with
+What does need attention is a reveal.js behavior that has nothing to do with
 MAIDR. reveal.js keeps the slides on either side of the current one rendered so
 that transitions stay smooth, and marking them `hidden` does not take them out
 of the tab order — reveal's own inline style overrides the attribute. On a
@@ -493,13 +493,13 @@ Until then,
 [quarto-revealjs-a11y](https://github.com/mcanouil/quarto-revealjs-a11y) does
 the same thing for a Quarto deck. Add it once per project:
 
-``` bash
+```{.bash}
 quarto add mcanouil/quarto-revealjs-a11y
 ```
 
 and enable it in the deck's front matter:
 
-``` yaml
+```{.yaml}
 format:
   revealjs:
     revealjs-plugins:
--- a/docs/index.qmd
+++ b/docs/index.qmd
@@ -464,7 +464,7 @@ Below is a detailed list of keyboard shortcuts for various functions:
 
 Table: Keyboard Shortcuts {#tbl-shortcuts}
 
-## Quarto revealjs Slides
+## Quarto Reveal.js Slides
 
 A chart on a `revealjs` slide is keyboard reachable on its own: **Tab** moves
 into it, the arrow keys explore it, and **Shift+Tab** hands focus back to the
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Both nits taken in `0ca95d5`.

1. Trimmed to one blank line before `## Braille Generation`.
2. Wrapped the two skimmable points in callouts, following the pattern the rest of the file uses (`::: {.callout-x}`, then a `##` title, then the body): a `callout-warning` titled "Use quarto-revealjs-a11y 0.2.3 or newer" for the version floor, and a `callout-note` titled "The workaround below is temporary" for the upstream fix. Good call on the first one especially — an older extension version silently reintroduces the bug it is there to fix.

Also switched the section's two `---` to the literal `—` the rest of the file uses, and put the heading in Title Case (`## Quarto Reveal.js Slides`) to match its neighbours. Re-rendered the section standalone with `quarto render`: both callouts come out as callouts and the links and code blocks are intact.

---

jooyoungseo:
Both style nits taken in `08c391c`.

- **Spelling.** Confirmed before changing: the file has one `behavior` already (line 239, on `set_backend()`) and four `color`, no British variants — so `behaviour` was mine and is now `behavior`.
- **Fence form.** The two new blocks are now `{.bash}` and `{.yaml}`, matching the `{.bash}` blocks in `## Installation`. Re-rendered the section: both callouts and all code blocks still come out right.

On the heading level: leaving it at `##`. The section is mostly third-party extension setup and upstream state rather than key bindings, and nesting it under the shortcuts table would put install instructions inside a section that is otherwise a key reference.

On the external facts — all four were checked, and all four within the last hour rather than from memory:

- `hakimel/reveal.js#1587` is open, filed 2016-05-09, three comments, still no fix committed to a release.
- The fix is on `master`: `js/reveal.js` has `slide.toggleAttribute('inert', !allowAllSlides && slide !== currentSlide && !slide.contains(currentSlide))`. Not in any release — I pulled the published tarballs for `reveal.js@6.0.1` and `reveal.js@5.2.1` from npm and grepped `dist/reveal.js`: zero occurrences of `inert` in either.
- Quarto 1.10 ships reveal.js 5.1.0: `quarto --version` reports 1.10.18, and its bundled `share/formats/revealjs/reveal/dist/reveal.js` opens with `reveal.js 5.1.0 … Copyright (C) 2011-2024`. No `inert` there either.
- `quarto-revealjs-a11y` 0.2.3 is the current release, published 2026-08-24, and its `_extension.yml` reads `version: 0.2.3`. I installed it and drove a deck with it in Chromium, which is where the 2-Tab-in, Shift+Tab-out numbers in the PR description come from.

You are right that these are the details that go stale. The one with a real shelf life is the reveal.js sentence, and it is written so that it stops being a caveat rather than becoming wrong: once Quarto ships a reveal.js that carries the fix, the extension simply becomes unnecessary and the deck needs no change.

---

jooyoungseo:
This pass ran against `dccbba3`; the fence nit was already fixed in `08c391c`, pushed just before it started. On the current head the two new blocks are `{.bash}` (line 496) and `{.yaml}` (line 502), matching the `{.bash}` blocks at 164 and 169. Nothing left to do there.

The version-specific claims were all re-checked against primary sources within the last hour, not from memory — the comment above has the details, including the npm tarballs for `reveal.js@6.0.1` and `@5.2.1` (no `inert` in either) and the installed Quarto 1.10.18 bundle reporting `reveal.js 5.1.0`.

---

PR #687, opened 2026-08-27
docs: mark the roadmap's new plot types as experimental

jooyoungseo:
## What

`PlotType` defined **15** members when the plot coverage roadmap (#345) was filed. It defines **37** now. The 22 that arrived with the roadmap are not on the same footing as the 15 that predate it — most landed inside about two weeks, and **none has been through a user study** — but nothing in the enum or the docs said so.

This adds `docs/stability.qmd`, in the navbar, splitting the enum and stating what the experimental half does not promise.

## The boundary is measured, not editorial

It is the diff of `PlotType` against `d9f7aee`, the last commit on `main` before #345 was filed:

```bash
git show d9f7aee:maidr/core/enum/plot_type.py | grep -oE '^\s+[A-Z_0-9]+ = "[a-z_0-9]+"'
```

The stable set ends at `VIOLIN_KDE` / `VIOLIN_BOX`, which matches where the pre-roadmap work actually stopped.

## What the experimental set is told not to promise

- **Under active development** — changed as they are used, not maintained against a settled specification.
- **Unstable** — field names, announcement wording and navigation semantics may change without a deprecation period, including in a patch release. Code depending on the exact shape one emits is pinned to the version it was written against.
- **Not validated** — each was measured against the chart it reads, which is what the issues and tests record. Measuring that a reading is *faithful to the drawing* is a different claim from establishing that it is *useful to a reader*, and only the first has been done.
- **Not a support commitment.**

## The drift guard

`tests/core/test_plot_type_stability.py` fails if a member lands in **neither** table or in **both**. At 22 additions in two weeks, a new type reaching the enum and not the page is the expected outcome rather than an unlikely one, so the classification has to be a deliberate step.

It also checks each row's **second column** against the enum, since that string is what reaches the MAIDR schema — a wrong one would send a reader looking for a type that never appears in the output. The prose claims are pinned too, with whitespace normalised so a reflow is not what breaks them.

## Two things caught while writing it

- The `::: {.callout-warning}` block first used a `##` heading for its title, which silently truncated the section the test parses — the Experimental table came back empty and the test said so. Switched to `title="Prototypes"`.
- The cross-link to the core's schema pointed at `/SCHEMA.html`. Building the core's docs showed it renders to `/docs/SCHEMA.html`; corrected against the built site's own canonical URL and its `id="trace-type-stability"` anchor.

## Verified

```
ruff check tests/core/test_plot_type_stability.py   All checks passed
uv run pytest -q                                    3576 passed, 72 skipped
```

(`ruff check --diff` reports 37 fixable findings repo-wide, all in `example/*.ipynb` and all pre-existing — nothing in this diff.)

## Companions

The same split, same boundary, same reasoning, in the core and the R binding: xability/maidr#1208 and xability/r-maidr#287.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,11 +42,26 @@ The `maidr/patch/` modules use `wrapt` to intercept matplotlib/seaborn plot call
 
 ### Supported Plot Types
 
-Defined in `maidr/core/enum/plot_type.py`: BAR, BOX, BOXEN, CONTOUR, COUNT, DODGED, ERRORBAR, GANTT, HEAT, HEXBIN, HIST, LINE, LOLLIPOP, AREA, STACKED_AREA, NORMALIZED_AREA, PIE, SCATTER, STACKED, NORMALIZED, STEP, SMOOTH, CANDLESTICK, VIOLIN_KDE, VIOLIN_BOX.
+Defined in `maidr/core/enum/plot_type.py`, which has **37** members. They do
+not all carry the same promise -- see `docs/stability.qmd`, and keep that page
+in step when adding one (`tests/core/test_plot_type_stability.py` enforces it).
+
+**Stable** (15) -- predate the plot coverage roadmap (#345) and have been
+exercised by real readers:
+
+BAR, BOX, CANDLESTICK, COUNT, DODGED, HEAT, HIST, LINE, PIE, SCATTER,
+SMOOTH, STACKED, STEP, VIOLIN_BOX, VIOLIN_KDE
+
+**Experimental** (22) -- added by that roadmap, none validated with a reader,
+and subject to change without a deprecation period:
+
+ALLUVIAL, AREA, BOXEN, CHOROPLETH, CONTOUR, ERRORBAR, FUNNEL, GANTT, GAUGE,
+HEXBIN, ICICLE, LOLLIPOP, NORMALIZED, NORMALIZED_AREA, PARALLEL, POLAR_AREA,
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Both observations taken — pushed in `acde92a`.

**2. `CLAUDE.md` naming the pre-roadmap 25.** You're right, and it is the same drift this PR exists to stop, sitting in the file an agent reads first. Fixed here rather than deferred, since a follow-up would have left the two docs disagreeing from the moment this merged. It now names all 37, split the same way as the page and pointing at it, with the note that the test enforces the page.

The lists were generated from the enum rather than transcribed, so the inventory can't be wrong in the way the old one was.

**1. Only two members pinned to the boundary.** You called this a reasonable trade-off; on reflection it isn't quite, and the reason is in your own sentence — a future contributor could misclassify into "the wrong (but still self-consistent) table without the test catching it". The partition check is satisfied by *any* split that covers the enum.

The fix costs nothing and needs no `git show` at runtime: `STABLE_AT_D9F7AEE` is written out in full and the experimental set is asserted as its complement. That is what the R binding's test already did, so the three repos now agree on the shape as well as the boundary.

Verified by making the mistake:

```
# `AREA` moved into the Stable table
FAILED test_the_stable_set_is_exactly_the_one_that_predates_the_roadmap
FAILED test_everything_else_is_experimental
2 failed, 4 passed
# restored
6 passed
```

The partition check passes throughout — which is exactly the gap you described.

**3.** The callout catch was the test doing its job on its first run, which was a useful sign the parser was worth having.

Re-verified after the change: `pytest` 3576 passed, 72 skipped; `ruff check` clean on the changed files.

---

PR #688, opened 2026-08-28
feat: read a word cloud as its terms and their weights

jooyoungseo:
## What

The core has read a `word_cloud` trace since #796. Neither binding could emit one. This teaches py-maidr to.

`wordcloud.WordCloud` is displayed with `ax.imshow(wc)`, so it arrives through the same entry point a heatmap does — but it is not one. The cloud rasterises to an `(M, N, 3)` colour array, and `maidr.patch.heatmap` declines exactly that shape (#564) because there is no number per cell to announce.

**Measured before this change:** a figure holding only a cloud registered no layer at all, and `FigureManager.get_maidr` raised `UnsupportedPlotError`. So the cloud was **unread rather than misread**, and this is purely additive.

That shapes the tests: most of their weight is on what the new wrapper must *not* disturb.

```
imshow(2D numbers)  -> ['HEAT']              unchanged
imshow(RGB photo)   -> UnsupportedPlotError  unchanged
bar + cloud         -> ['BAR', 'WORD_CLOUD'] the bar keeps its reading
```

## Two decisions made by measurement

**The weights are relative, and the axis label says so.**

`WordCloud` divides every frequency by the largest and keeps only the ratio. Measured:

```
generate_from_frequencies({machine: 412, learning: 300, data: 250})
  -> words_ = {machine: 1.0, learning: 0.728, data: 0.607}
```

The raw counts are on **no attribute of the object** — I searched every public attribute for the value 412 and it is not there. So the axis is `Relative frequency`, not the core example's `Occurrences`, which would hand a reader "machine, 1.0" for a term that occurred 412 times.

This is not a consolation prize: the ratio is what the chart *draws*, since glyph size is proportional to it. A reader hearing 1.0 and 0.728 hears what a sighted reader sees.

**Read from `words_`, not `layout_`.**

`layout_` is the placement list, and with `repeat=True` it lists a term once per placement. Measured on a two-term cloud:

```
layout_: [(alpha, 1.0), (beta, 0.667), (alpha, 0.667), (beta, 0.444)]
words_ : {alpha: 1.0, beta: 0.667}
```

Reading `layout_` would announce alpha **twice, at two different weights**, for a repetition that is the packer filling space rather than anything in the data. `words_` is keyed by term so it cannot repeat one, and it already honours `max_words`.

## No selectors

`imshow` rasterises the whole cloud into one element; there is no per-term element to point at. Highlighting is switched **off** rather than left to the base class — its generic `g[maidr='true'] > path` would be resolved by the core and paired positionally with the terms, lighting up whatever else on the figure happened to draw that many paths.

`wc.to_array()` and `wc.to_image()` are deliberately not recognised: both hand `imshow` a plain array and the terms are not in it. A cloud shown that way stays a picture, which is the honest answer, and there is a test for it.

## Experimental, and the guard said so

`WORD_CLOUD` is listed in `docs/stability.qmd`'s experimental table. The drift guard added in #687 **required** the classification — it fails on a member that is in neither table — which is the guard doing its job on the first new type after it landed.

`test_no_plot_type_is_named_to_a_user_with_an_underscore_in_it` also caught the missing `_DISPLAY_NAMES` entry, so users see "word cloud" rather than the schema token.

## Verified

```
pytest                       3586 passed, 72 skipped
ruff check <changed files>   All checks passed
uv lock --check              clean
```

`wordcloud` is added as an optional extra and a dev dependency. Nothing in maidr imports it — the patch recognises a cloud by its `words_` mapping, structurally — so a user without it is unaffected; it is listed so `pip install maidr[wordcloud]` gets a working pair and so CI installs what the tests need in order not to skip.

## Companion

The same chart in the R binding, where the counts survive and the axis can honestly say `Occurrences`: xability/r-maidr#288.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -42,7 +42,7 @@ The `maidr/patch/` modules use `wrapt` to intercept matplotlib/seaborn plot call
 
 ### Supported Plot Types
 
-Defined in `maidr/core/enum/plot_type.py`, which has **37** members. They do
+Defined in `maidr/core/enum/plot_type.py`, which has **38** members. They do
 not all carry the same promise -- see `docs/stability.qmd`, and keep that page
 in step when adding one (`tests/core/test_plot_type_stability.py` enforces it).
 
@@ -52,12 +52,12 @@ exercised by real readers:
 BAR, BOX, CANDLESTICK, COUNT, DODGED, HEAT, HIST, LINE, PIE, SCATTER,
 SMOOTH, STACKED, STEP, VIOLIN_BOX, VIOLIN_KDE
 
-**Experimental** (22) -- added by that roadmap, none validated with a reader,
+**Experimental** (23) -- added by that roadmap, none validated with a reader,
 and subject to change without a deprecation period:
 
 ALLUVIAL, AREA, BOXEN, CHOROPLETH, CONTOUR, ERRORBAR, FUNNEL, GANTT, GAUGE,
 HEXBIN, ICICLE, LOLLIPOP, NORMALIZED, NORMALIZED_AREA, PARALLEL, POLAR_AREA,
-RADAR, SANKEY, STACKED_AREA, SUNBURST, TREEMAP, WATERFALL
+RADAR, SANKEY, STACKED_AREA, SUNBURST, TREEMAP, WATERFALL, WORD_CLOUD
 
 Note that `maidr/plotly/` builds its MAIDR schema in Python and therefore needs its own
 handling per plot type, whereas `maidr/altair/` delegates entirely to the upstream
--- a/docs/stability.qmd
+++ b/docs/stability.qmd
@@ -6,13 +6,13 @@ html-table-processing: none
 
 ## Two sets, two different promises
 
-`PlotType` in `maidr/core/enum/plot_type.py` names **37** plot types. They do
+`PlotType` in `maidr/core/enum/plot_type.py` names **38** plot types. They do
 not all carry the same promise, and nothing in the enum says so — which is what
 this page is for.
 
 Fifteen of them predate the plot coverage roadmap
-([#345](https://github.com/xability/py-maidr/issues/345)). The other twenty-two
-were added by it, most inside about two weeks, and **none of the twenty-two has
+([#345](https://github.com/xability/py-maidr/issues/345)). The other twenty-three
+were added by it, most inside about two weeks, and **none of the twenty-three has
 been through a user study**.
 
 ## Stable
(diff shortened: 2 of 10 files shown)


jooyoungseo:
All four points acted on in `14a59e7`.

## The ordering-blind test — you were right, it could not fail

`COUNTS` was already written heaviest-first, so asserting the emitted terms equalled `list(COUNTS)` proved nothing about `words_` re-sorting.

Measured which hypothesis is actually true before fixing it:

```
passed in : ['data', 'machine', 'model', 'learning']
words_    : ['machine', 'learning', 'data', 'model']
weights   : [1.0, 0.728, 0.607, 0.291]
```

`words_` does re-sort. The fixture is now written in that shuffled order, the expected order is named separately as `BY_WEIGHT`, and the test also asserts `list(COUNTS) != BY_WEIGHT` — so if the re-sorting ever stops, or if someone "tidies" the fixture back into sorted order, the test says so instead of quietly going vacuous again.

## The import-order dependency — real, and worse than I'd assumed

You called this "worth double-checking", so I checked it by building the reversed stack rather than reasoning about it. Re-applying `heat` on top makes it the outer wrapper, which is exactly what a reshuffled `patch/__init__.py` would produce:

```
as shipped (wordcloud outermost): ['word_cloud']
reversed    (heat outermost)    : UnsupportedPlotError
```

So under reversal it is not a subtle misread — nothing registers at all, and the figure raises exactly as it did before this patch existed. `heat` runs first, sets the internal context for its own draw, the call reaches this wrapper with `is_internal_context()` already true, and then `heat` declines the colour image too.

Documented in `wordcloud.py` with that measurement, and I've named the thing that actually enforces it: `test_a_cloud_is_read_rather_than_refused` fails outright under the reversed stack, because there's no layer to read. So the ordering isn't held up by the tuple alone — but nothing said so, which was your point.

## The `ExtractionError` path — covered

Defensive, as the docstring says, but worth a test for a reason I should have stated: the alternative to raising is emitting an empty layer, and a word cloud announcing no terms is indistinguishable to a reader from one whose terms were read.

## The shared-label fallback — an oversight, not a scope cut

Thanks for flagging it as possibly either. It was an oversight. One cloud per group on a facet grid carries its label on the shared outer axes rather than its own, so asking only `self.ax` handed back the generic `Term` for a figure that *had* been labelled.

Added, matching `PiePlot`. I verified the new test guards it rather than merely passing — with the fallback removed it fails, with it restored the file is green:

```
--- with the fallback removed ---
FAILED tests/patch/test_wordcloud.py::test_a_facet_grid_label_names_the_terms
1 failed, 11 passed
--- restored ---
12 passed
```

Full suite: **3588 passed, 72 skipped**. `ruff check` clean on all three files.

---

PR #689, opened 2026-08-28
docs: add a word cloud example to the gallery

jooyoungseo:
## What this adds

A worked word cloud example in the docs gallery, plus the standalone sample under `example/wordcloud/matplotlib/`.

## Why

`word_cloud` merged in #688 with only a row in `docs/stability.qmd` naming it. A reader could learn the type exists but not how to produce one.

That was an inconsistency rather than a deliberate scope cut — four other **experimental** types already carry example sections in this same file:

| Experimental type | Example in `examples.qmd` |
|---|---|
| Hexbin | yes |
| Boxen | yes |
| Error Bar | yes |
| Area | yes |
| **Word Cloud** | **was missing** |

## What the section covers

Placed after Pie Chart — the other chart here with no positional scale — and it covers the three things that are not obvious from the API:

- **The weights are relative, and the default axis label says so.** `generate_from_frequencies` divides every frequency by the largest and keeps only the ratio, so 412 is announced as `1.0` and the y axis is named "Relative frequency". Naming it after counts would hand a reader "accessibility, 1.0" for a term that occurred 412 times. The note also says the R binding *is* handed the raw counts and can honestly say "Occurrences" — same chart, two different honest readings.
- **Show the object, not the picture.** `wc.to_array()` and `wc.to_image()` hand `imshow` a plain RGB array with the terms nowhere in it, so such a figure stays an unread picture.
- **No highlighting**, because `imshow` rasterises the cloud into a single element and there is no per-term element to outline.

A `callout-warning` marks it as a prototype and links to `stability.qmd`, matching how that page describes the experimental tier.

## Verification

The example code was **run before being written down**, not composed and hoped for:

```
type      : PlotType.WORD_CLOUD
axes      : {x: 'Topic', y: 'Share of mentions'}
n points  : 8
first 3   : accessibility 1.0 | sonification 0.728 | braille 0.607
has selector: False
```

- `ruff check` clean on the new example script.
- Word cloud and stability tests pass (18).
- The docs job runs `uv sync --locked --all-extras --dev`, and #688 put `wordcloud` in both the `wordcloud` extra and the dev group, so it is present at render time.

Quarto is not installed in this environment, so the page was not rendered locally — the substantive risk is the Python chunk, and that was executed.

---

### DIFF
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -299,11 +299,6 @@ cloud = WordCloud(
 ).generate_from_frequencies(mentions)
 
 fig, ax = plt.subplots(figsize=(8, 4))
-# A cloud has no x or y scale -- the glyph positions are packing, not data --
-# so the axis labels name what a point *holds* rather than where it sits.
-# Without them a term reads as "X: braille", which says nothing.
-ax.set_xlabel("Topic")
-ax.set_ylabel("Share of mentions")
 ax.set_title("What the accessibility reports talk about")
 ax.imshow(cloud) #<<
 ax.set_axis_off()
@@ -311,12 +306,18 @@ ax.set_axis_off()
 plt.show() #<<
 ```
 
-**The weights are relative, and the default axis label says so.** `WordCloud`
+A cloud has no x or y scale — the glyph positions are packing, not data —
+so maidr names the two axes after what a point *holds* rather than where it
+sits. No labels were set above, so the defaults fire and a term reads as
+"Term: braille, Relative frequency: 0.607". The generic "X"/"Y" the base
+class would otherwise give says nothing about a word.
+
+**The weights are relative, and that y axis name says so.** `WordCloud`
 divides every frequency by the largest one in `generate_from_frequencies` and
 keeps only the ratio — the raw counts are on no attribute of the object. So the
-counts above are announced as `1.0`, `0.728`, `0.607`, and the default y axis
-name is "Relative frequency" rather than "Occurrences". Naming it after counts
-would hand a reader "accessibility, 1.0" for a term that occurred 412 times.
+counts above are announced as `1.0`, `0.728`, `0.607`, and the axis is called
+"Relative frequency" rather than "Occurrences". Naming it after counts would
+hand a reader "accessibility, 1.0" for a term that occurred 412 times.
 
 That is also what the chart draws: glyph size is proportional to the ratio, so a
 reader hearing 1.0 and 0.728 hears what a sighted reader sees. (The R binding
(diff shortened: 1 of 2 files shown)


jooyoungseo:
The main finding was right, and it was a real defect — fixed in `1005357`.

## The mismatch

The paragraph explained the default weight axis name and why it matters, while the chunk directly above it called `ax.set_ylabel("Share of mentions")` — and an authored label always wins. So the one thing the paragraph was about was the one thing a reader running the example would never see, and the prose read as if it described the output it sat under.

Measured both ways rather than reasoning about `_extract_axes_data`:

```
no labels set  ->  x: Term,  y: Relative frequency
labels set     ->  x: Topic, y: Share of mentions
```

## What I did

Took your first option — dropped the two label calls — because it's the better example independent of the bug: it shows maidr naming the axes sensibly **on its own**, which is exactly the point the section is making, and it's less code to read.

Overriding them is still worth knowing, so it's now its own short note after the reading, with the caveat that a hand-written y name still has to stay honest about being a ratio.

Re-verified by extracting the chunk from `examples.qmd` itself and running that, so the check is against the file rather than against my earlier copy:

```
VERIFIED type : PlotType.WORD_CLOUD
VERIFIED axes : {x: 'Term', y: 'Relative frequency'}
VERIFIED first: ['accessibility', 1.0]
```

The standalone script keeps its authored labels — naming axes in the language of the data is good practice — but its docstring had the same conflation and now distinguishes the default from its own override.

## On the two precedent points

Both corrections are accurate and I'd rather have them on the record than argue them:

- The **callout-warning** is not something Hexbin/Boxen/Error Bar/Area do. It matches `stability.qmd`, not those four. I'd defend it on the merits — a prototype box is worth having on an experimental section — but you're right that my PR text implied a precedent that isn't there, and applying it to the other four is a reasonable follow-up rather than something this PR settled.
- The **`example/` script** likewise matches the *stable*-type convention (bar, box, heatmap all have one), not the four experimental sections, which have none. My commit message said "matching the per-type layout the other plot types use", which was true of the repo generally and misleading about the comparison I'd just drawn.

## On rendering

Still not rendered locally — Quarto isn't installed here. The docs job on this PR builds the site, so the callout fencing and the `#<<` markers get their check there rather than from me.

---

jooyoungseo:
Point 1 acted on in `505c96b` — extended rather than dropped.

You raised it twice, which was the right call, and the second framing made the choice clear: Word Cloud looked more provisional than four types carrying exactly the same promise. Of your two options I took extension, because dropping it would leave Hexbin, Boxen, Error Bar and Area quietly implying a stability they don't have — which is the drift `stability.qmd` exists to stop, and it would be odd to reintroduce it two PRs later.

All five now carry the same box, in the same place — directly under the heading, so a reader meets the warning before the example rather than after it. The word cloud one moved up to match and its wording is now type-neutral and shared.

I checked the set against the enum split rather than by eye: those five are exactly the experimental types that have example sections, and the remaining eighteen have no section to flag.

Word cloud chunk re-verified after the edit:

```
VERIFIED axes : {x: 'Term', y: 'Relative frequency'}
VERIFIED first: ['accessibility', 1.0]
```

On your points 2–4, agreed and no action taken:

- The prose/script duplication is the repo-wide pattern for every plot type, not something this PR introduces. Worth noting for the day `WordCloud`'s normalisation changes, and that's what your comment now records.
- No test changes and no security surface — correct on both.

Thanks for the read on the label-override mismatch in the first round. That one was a genuine doc bug: the paragraph described a default the example above it had overridden, so the one thing it was about was the one thing a reader would never see.

---

PR #690, opened 2026-08-28
docs: add gallery examples for three experimental matplotlib plot types

jooyoungseo:
## What

Adds gallery examples for three experimental plot types that py-maidr supports and that had no worked example anywhere in the docs: **contour**, **lollipop**, and **gantt**.

New sections in `docs/examples.qmd`, each placed next to the reading it relates to — Contour after Hexbin, Lollipop after Point, Gantt after Candlestick, before `## Reactive Dashboard`. Each carries the standard prototype callout the other experimental sections use.

| Section | Drawn by | Read as |
|---|---|---|
| Contour Plot | `ax.contour`, bivariate `sns.kdeplot` | `contour` |
| Lollipop Plot | `ax.stem` | `lollipop` |
| Gantt Chart | `ax.broken_barh` | `gantt` |

## Measured, not asserted

Every chunk was extracted from the committed file and executed against an installed build, and the emitted layer type, point count and axis labels recorded:

```
Contour Plot   contour    n=6                                   [x=x | y=y]
Lollipop Plot  lollipop   n=8    [x=Sample | y=Concentration (mg/L)]
Gantt Chart    gantt      lanes=['Design', 'Build', 'Launch']   [x=Week | y=Phase]
```

Two gantt claims are worth calling out because measurement corrected what I first assumed:

- **`broken_barh` is one layer, not one per call.** The chart is a single `gantt` layer whose `data` is `{"points": [...], "lanes": [...]}`. Each call contributes one *lane*; the spans in that call are that lane's spans. A phase that stops and restarts is two spans on one lane, not two rows.

- **Naming a lane takes `set_yticks()` *and* `set_yticklabels()`.** `label=` looks like the argument for it and is not — it adds a legend entry. Measured across all four combinations:

  | what the axis sets | lanes announced as |
  |---|---|
  | `set_yticks()` and `set_yticklabels()` | `Design`, `Build`, `Launch` |
  | `set_yticklabels()` alone | `13.0`, `23.0`, `33.0` |
  | `set_yticks()` alone | `13`, `23`, `33` |
  | neither | `13.0`, `23.0`, `33.0` |

For contour, `levels=` is a count of heights rather than of curves — a height crossed in two places draws one closed island per place, so a two-peaked surface gives 12 curves for `levels=6` while this section's single-peaked one gives 6. The section states the dependency rather than the coincidence.

This is a docs-only change; no package code is touched.

## Changed during review

Three review findings, all real, all fixed:

- Gantt was nested under `## Reactive Dashboard` — I anchored the insertion on `### Shiny` and missed the level-2 heading between it and Candlestick (`a6e30e5`).
- The lane table credited `set_yticklabels()` alone, and "one layer per band" for stacked area contradicted the layer/lane distinction drawn two sections down (`eaa64a2`).
- The stacked area section claimed the running total was "a drawing artefact" not part of the reading. The opposite is true and the docstrings say so — it is the stated reason `stacked_area` is distinct from `line`. I had measured the emitted payload and drawn a conclusion about the announcement that the payload does not support (`189dba8`).

**A stacked area section is no longer part of this PR** (`b7f3625`). It turned out to be redundant: the existing "Area Plot" section's example is already `ax.stackplot()` with two series, which `resolve_type()` makes a `stacked_area`, and its prose already carries the two-numbers framing. My original premise that stacked area had no example was wrong — I enumerated headings without reading them, and the heading is named for the function rather than the type. The one fact it added that the existing section lacked, what `labels=` changes about the reading, is folded into that section instead.

## Still missing

The plotly gallery (`docs/examples-plotly.qmd`) has no experimental types at all — waterfall, funnel, gauge, sankey, treemap, sunburst, icicle, parallel coordinates, alluvial, choropleth, polar area, radar, and 100% stacked bar. All measured as reachable; will follow up separately rather than grow this diff. The 100% stacked *area* is held back pending #691.

### DIFF
--- a/docs/examples.qmd
+++ b/docs/examples.qmd
@@ -635,6 +635,54 @@ count with a reduction of the values you supply, and a numeric `bins=`
 discretises the counts so the colour shows *which interval* a bin landed in.
 Pass `z_label=` to name it yourself.
 
+### Contour Plot
+
+::: {.callout-warning title="Prototype"}
+This is one of the experimental plot types. It has not been through a user
+study, and it may change without a deprecation period. See
+[Plot type stability](stability.qmd).
+:::
+
+A contour plot draws a surface as a set of level curves, each joining the
+points where the surface has one particular height. maidr reads the **levels
+themselves**: navigation walks one curve at a time, and every point on a
+curve carries the level it belongs to, so the height is never something you
+have to infer from position.
+
+That is the difference between a contour and a heatmap of the same surface.
+A heatmap gives you one number per cell on a fixed grid; a contour gives you
+the shape of a chosen height, which is what the chart was drawn to show.
+
+```{python}
+#| fig-alt: Contour plot of a two-dimensional Gaussian surface
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+import maidr #<<
+
+x = np.linspace(-3, 3, 80)
+y = np.linspace(-3, 3, 80)
+grid_x, grid_y = np.meshgrid(x, y)
+height = np.exp(-(grid_x**2 + grid_y**2) / 2)
+
+fig, ax = plt.subplots(figsize=(7, 6))
+contours = ax.contour(grid_x, grid_y, height, levels=6) #<<
+ax.clabel(contours, inline=True, fontsize=8)
+ax.set_xlabel("x")
+ax.set_ylabel("y")
+ax.set_title("A Gaussian surface, six levels")
+
+plt.show() #<<
+```
+
+`levels=` decides how many curves are drawn, and the reading follows it
+exactly — ask for six and you navigate six, because maidr reads the curves
+matplotlib actually produced rather than re-deriving its own.
+
+A bivariate `sns.kdeplot(x=..., y=...)` draws its density the same way and
+gets the same reading.
+
 ### Box Plot
 
 ```{python}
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Good catch, and the diagnosis is exactly right. I anchored the insertion on `### Shiny` without noticing that `## Reactive Dashboard` sits between it and the candlestick section, so Gantt landed inside a level-2 heading about the Shiny/Streamlit integrations.

Fixed in `a6e30e5`. The heading order is now:

```
1314:### Candlestick Chart
1356:### Gantt Chart
1412:## Reactive Dashboard
1414:### Shiny
```

Re-ran the chunk extraction against the moved file to confirm nothing broke in the move — all four sections still produce the readings the prose states:

```
Contour Plot         contour        n=6                                   [x=x | y=y]
Stacked Area Plot    stacked_area   n=3    [x=Quarter | y=Revenue (millions)]
Lollipop Plot        lollipop       n=8    [x=Sample | y=Concentration (mg/L)]
Gantt Chart          gantt          lanes=['Design', 'Build', 'Launch']   [x=Week | y=Phase]
```

Code fences are still balanced at 82.

---

jooyoungseo:
Both taken — thank you, the `set_yticks()` point in particular was a real gap in the instruction.

**Lane naming.** You're right that `set_yticklabels()` alone does nothing, and measuring all four combinations turned up a third case worth stating: `set_yticks()` alone names the lanes after the tick *positions*, as strings.

```
set_yticks + set_yticklabels    lanes=['Design', 'Build', 'Launch']
set_yticklabels only            lanes=[13.0, 23.0, 33.0]
set_yticks only                 lanes=['13', '23', '33']
neither                         lanes=[13.0, 23.0, 33.0]
```

The table now carries all four rows and the prose leads with "takes **both**", explaining why: a lane name is only trusted when the axis has fixed tick positions to attach it to.

**Layer vs. series.** Agreed, and it was worse than imprecise — the gantt section two sections down draws exactly the layer/lane distinction, so "one layer per band" read as a contradiction with its neighbour. Now "one series per band within a single layer".

Both in `eaa64a2`. Re-verified after the edit; all four sections still produce what the prose claims, fences still balanced at 82.

---

jooyoungseo:
Both correct, and the stacked area one is a straight error on my part — I asserted it instead of measuring it, in a section whose whole claim is "measured, not asserted." Thank you for catching it.

**Stacked area.** You're right and the docstrings settle it. `AreaPlot`: *"the area trace announces the value and reports the total beside it."* `PlotType.STACKED_AREA`: *"a line layer announces one number per point with nothing to say which of those two it is."* The running total is not an artefact the reading discards — it is the stated reason the type exists at all. What I measured was the emitted `{x, y, z}` payload and then wrote a conclusion about the announcement that the payload does not support.

Rewritten to say both numbers are read and why, and to add the single-band `area` case that follows from the same rule (`resolve_type`: one band has nothing stacked on it, so a total equal to its own value would be noise).

**Contour.** Also right, and measured:

```
one peak,  levels=6  ->   6 series, 6 distinct levels
two peaks, levels=6  ->  12 series, 6 distinct levels
```

A height crossed in two places draws one closed island per place. The section now states `levels=` as a count of *heights*, keeps the six-for-six result for this example as the single-peaked case it is, gives the two-peak contrast in a table, and notes that an unreached level contributes no curve.

Both in `189dba8`. Re-verified after the edit; all four sections still produce the readings claimed, fences balanced at 82.

On your point about the adjacent Area Plot section — worth noting for a follow-up that its existing example calls `ax.stackplot()` with two series, so per `resolve_type()` it already emits `STACKED_AREA` rather than `AREA` despite sitting under a heading that says otherwise. That is pre-existing and outside this diff, but it is a real mismatch and I am happy to fix it separately.

---

jooyoungseo:
You're right, and I've dropped the section. My premise was simply wrong.

I checked coverage by enumerating `###` headings and looking for one named after the type. There isn't one — the heading is named after the *function family* ("Area Plot") — so I concluded `stacked_area` was undocumented without reading the section. It is documented, and better than my version was: the existing example is `ax.stackplot()` with two series, which `resolve_type()` makes a `stacked_area`, and the prose already carries the two-numbers framing plus `fill_between`, the streamgraph baseline, and the two-curve exclusion.

Worth noting what that means for the previous round: the "correction" I made two commits ago was me deriving from docstrings a fact the doc already stated two sections above. That is a clean illustration of the cost of not reading the neighbouring section first.

Dropped in `b7f3625`. The one thing my section said that the existing one did not — what `labels=` changes about the reading — is folded into "Area Plot" instead, directly beneath the chunk that passes it:

> Pass `labels=` and each band is announced by name. Without it the bands are still read, but they can only be told apart by their order in the stack.

That is the `if label:` guard at `areaplot.py:139-140` you confirmed in your previous pass.

PR title and description updated to say three sections rather than four, and to drop the incorrect "had no worked example" claim for stacked area. Fences now 80, still balanced; the three remaining sections re-verified after the removal:

```
Contour Plot   contour    n=6                                   [x=x | y=y]
Lollipop Plot  lollipop   n=8    [x=Sample | y=Concentration (mg/L)]
Gantt Chart    gantt      lanes=['Design', 'Build', 'Launch']   [x=Week | y=Phase]
```

On your CI point — you're right that nothing renders `examples.qmd`'s chunks, so the "measured" claims here aren't independently re-verified and this page can silently drift. That is worth an issue of its own; I'll file one rather than bolt it onto this PR.

---

PR #726, opened 2026-09-03
fix(box): read a box whose caps or fliers are hidden and drop one whose statistics are NaN

### LINKED ISSUE #697: A box drawn with showfliers=False or showcaps=False is announced as an empty layer, and a box over a sample containing NaN emits NaN statistics
## Problem

`BoxPlotExtractor` (`maidr/core/plot/boxplot.py:137-197, 406-503`) zips `fliers` with `caps` in `extract_outliers`, and the final zip at line 487 stops at the shortest list. `Axes.bxp` returns an empty `caps` list for `showcaps=False` and an empty `fliers` list for `showfliers=False` (or `sym=''`), so every box is truncated: the layer is still registered with `data == []` and `selectors == []`, the figure renders, and the reader hears nothing for a box chart that is on screen. Measured: `ax.boxplot(3 samples)` default -> 3 rows; `showfliers=False`, `sym=''`, `showcaps=False`, both -> 0 rows; same for `sns.boxplot(showfliers=False)` and `orient='h'`. No comment, changelog entry or issue documents this; #593 fixed a different truncation of the same zip.

Separately, `boxplot.py:154/168/183` cast with bare `float()`. `cbook.boxplot_stats` does not drop NaN, so `ax.boxplot([[1, 2, nan, 4], [2, 3]])` (a pandas column with one missing value) draws nothing for that box and records NaN for every statistic; the emitted row is `{min: NaN, q1: NaN, ...}` and the whole figure fails `JSON.parse` (#427 class). seaborn drops NaN itself, so `sns.boxplot` is unaffected. `BoxPoint` types `min/q1/q2/q3/max` as non-nullable numbers and `Number(null)` is 0 in the core, so the honest fix is to drop the undrawn box together with its artists so data, selectors and levels stay aligned.

## Proposed fix

- `_extract_extremes` gains a `position` parameter; `extract_caps(caps, whiskers)` falls back to the whisker's far end (`position=1`) when `caps` is empty — verified equal to the cap y for every box in the default case.
- `extract_outliers` iterates over `enumerate(caps)` and reads `fliers[i]` only when it exists, else `{lowerOutliers: [], upperOutliers: []}`.
- `_extract_bxp_maidr` uses `bxp_stats['caps'] or bxp_stats['whiskers']` for min/max gid tagging, and when there are no flier artists extends the outlier element map with the box gids so `_get_selector`'s zip keeps one selector per box (the outlier selectors resolve to nothing, which the core tolerates).
- Drop any box whose q1/q3/min/max/median is non-finite, together with its artists and level, before gid tagging; drop non-finite flier values.

Default-path output verified byte-identical on `ax.boxplot` vert/horiz, `sns.boxplot`, `sns.boxplot(hue=)` vert/horiz; 327 existing box/heat/bar tests pass with the prototype.

## Acceptance criteria

- New `tests/core/plot/test_box_hidden_parts.py` parametrised over `ax.boxplot`/`sns.boxplot` x `showfliers=False` / `showcaps=False` / both / `sym=''` / `orient='h'`: `len(data) == number of boxes`, min/max equal the default chart's, outliers `[]`, `len(selectors) == len(data)`, strict JSON round trip; default chart unchanged.
- New `tests/core/plot/test_box_gaps.py`: `ax.boxplot([[1, 2, nan, 4], [2, 3, 5]])` emits one row with finite statistics and `z == '2'`, one selector, strict JSON passes; same for an empty group and an all-NaN group; NaN-free chart unchanged.

jooyoungseo:
## What

Closes #697.

**Hidden caps or fliers emptied the layer.** `BoxPlotExtractor` zipped `fliers` with `caps`, and the final zip stopped at the shortest list. `Axes.bxp` returns an empty `caps` list for `showcaps=False` and an empty `fliers` list for `showfliers=False` or `sym=''`, so every box was truncated: the layer registered with `data == []` and `selectors == []`, the figure rendered, and a reader heard nothing for a box chart that is on screen.

```
ax.boxplot(3 samples)                       rows 3  -> 3   unchanged
ax.boxplot(..., showfliers=False)           rows 0  -> 3
ax.boxplot(..., showcaps=False)             rows 0  -> 3
ax.boxplot(..., sym='')                     rows 0  -> 3
sns.boxplot(hue=..., showfliers=False)      rows 0  -> 6
```

`extract_caps` now falls back to the whiskers' far end when there are no caps (verified equal to the cap y for every box in the default case), `extract_outliers` reads `fliers[i]` only when it exists and otherwise emits two empty lists, and when no flier artist survives the outlier selectors are built on the box gid so `_get_selector`'s zip keeps one selector per box — the core already reads a selector that matches nothing as "no outliers".

**A NaN in the sample poisoned the box.** `cbook.boxplot_stats` does not drop NaN, so `ax.boxplot([[1, 2, nan, 4], [2, 3, 5]])` drew nothing for the first box and recorded NaN for every statistic; the emitted row was `{min: NaN, q1: NaN, ...}` and the whole figure failed `JSON.parse` (#427). `BoxPoint` types the statistics as non-nullable numbers and `Number(null)` is 0 in the core, so a box whose statistics are non-finite is now dropped together with its artists and its level, keeping data, selectors and levels aligned. seaborn drops missing values itself, so `sns.boxplot` was never affected by this half.

Default-path output is byte-identical after uuid normalisation across 22 cases (box default/horizontal, `sns.boxplot` plain and with hue in both orientations, plus every heat/stacked/dodged case in the audit's comparison script); only the four cases named above change. The five duplicated gid-tagging loops became one `_claim(artists, slot)` helper that preserves the original element order.

## Tests

New `tests/core/plot/test_box_hidden_parts.py` (parametrised over `ax.boxplot`/`sns.boxplot` × `showfliers=False` / `showcaps=False` / both / `sym=''` / `orient='h'`: one row per box, min/max equal to the default chart's, empty outliers when fliers are hidden, one selector per row, strict JSON round trip, default chart unchanged) and `tests/core/plot/test_box_gaps.py` (one NaN in a group, an empty group, an all-NaN group: the finite box keeps `z='2'`, one selector, strict JSON passes, NaN-free chart unchanged). Against `main`'s `boxplot.py`: 26 failed, 43 passed (the passing ones are the "must not change" guards).

## Verified

```
uv run pytest      3673 passed, 68 skipped
```

With caps hidden the min/max selectors highlight the whisker paths, where the values were read.

---

### DIFF
--- a/maidr/core/plot/boxplot.py
+++ b/maidr/core/plot/boxplot.py
@@ -536,17 +536,6 @@ class BoxPlot(
         key = MaidrKey.X if self._orientation == "vert" else MaidrKey.Y
         levels = self.extract_level(self.ax, key) or []
 
-        # A `hue` draws one box per category *per level*, and the axis still
-        # carries one tick per category -- so the `zip` below, which ends at
-        # the shortest of what it is given, stopped at the ticks and dropped
-        # every box past them. Measured on three categories and two levels:
-        # six boxes drawn, three announced, and the three that survived were
-        # the first level's paired with the tick labels in order (#593). The
-        # selector list is built from the artists rather than from this, so
-        # the same chart also emitted six selectors against three rows.
-        if len(medians) > len(levels):
-            levels = self._named_boxes(bxp_stats["boxes"], levels, key)
-
         # A box whose statistics are not finite was never on screen.
         # `cbook.boxplot_stats` does not drop a NaN, so one missing value in
         # a group -- or an empty group -- makes every statistic NaN, and
@@ -567,8 +556,28 @@ class BoxPlot(
             # artists at all when the fliers are hidden.
             return [items[index] for index in keep if index < len(items)]
 
-        # One level per box is the only shape a dropped box can be taken out
-        # of; more ticks than boxes is a chart the levels never lined up with.
+        # Two charts whose boxes and tick labels do not share an index, and
+        # so are named off the drawing -- each box by the tick nearest it --
+        # rather than paired with the labels in order:
+        #
+        # - A `hue` draws one box per category *per level*, and the axis
+        #   still carries one tick per category -- so the `zip` below, which
+        #   ends at the shortest of what it is given, stopped at the ticks and
+        #   dropped every box past them. Measured on three categories and two
+        #   levels: six boxes drawn, three announced, and the three that
+        #   survived were the first level's paired with the tick labels in
+        #   order (#593). The selector list is built from the artists rather
+        #   than from this, so the same chart also emitted six selectors
+        #   against three rows.
+        # - A chart with a box dropped. The labels are the ticks inside the
+        #   *data* limits, and a box that drew nothing contributes nothing to
+        #   those, so its own tick may or may not still be in the list --
+        #   the counts can agree while the alignment is off, and filtering
+        #   the labels by `keep` would then hand a surviving box the label
+        #   of an extra tick beside it. A chart missing nothing takes neither
+        #   branch and reads as it always has.
+        if len(medians) > len(levels) or len(keep) != len(medians):
+            levels = self._named_boxes(bxp_stats["boxes"], levels, key)
         if len(levels) == len(medians):
             levels = kept(levels)
         whiskers, caps, medians, outliers = (
--- a/maidr/core/plot/boxplot.py
+++ b/maidr/core/plot/boxplot.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 import uuid
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Follow-up on the levels-alignment note: 838e6da. It turned out wider than "more ticks than boxes" — `_ticks_in_view` keeps only ticks inside the data limits and a NaN box contributes nothing to them, so after a drop the tick and box counts can be *equal* yet misaligned. Whenever a box was dropped, every box is now named off the drawing by nearest tick through the existing `_named_boxes` path (the same branch the #593 hue case takes), and a test pins the case where an extra in-range tick sits beside a survivor (`z == ["b", "d"]`, read as `["c", "d"]` before).

---

PR #727, opened 2026-09-03
fix(schema): emit a gap rather than a NaN token for heatmap cells, stacked bars and areas

### LINKED ISSUE #696: A NaN or masked heatmap cell, a NaN stacked/dodged bar and a NaN area point are emitted as a bare NaN token that JSON.parse rejects; the heatmap loop is also 15-30x slower than needed
## Problem

Three matplotlib emitters still write non-finite numbers straight into the schema, which `json.dumps` serialises as the bare token `NaN`. `JSON.parse` in the browser rejects it inside a `catch`, so `initMaidrOnElement` never runs and the whole figure loses audio, text, braille and highlight (the #427 class; `BarPlot._magnitude`, `LinePlot`, `ScatterPlot`, `ErrorbarPlot` and `HexbinPlot` already emit `None` for exactly this reason).

- `HeatPlot._extract_scalar_mappable_data` (`maidr/core/plot/heatmap.py:278,306`) reads `sm.get_array().data`, the buffer **under** the mask, and converts every cell with `float(format(x, fmt))`. `sns.heatmap([[1, nan], [3, 4]])`, `ax.imshow` of the same, `pcolormesh(masked_invalid(...))` and `ax.pcolor(masked)` all emit `[[1.0, NaN], ...]`. `sns.heatmap(corr, mask=np.triu(...))` draws six blank cells but announces the full symmetric matrix.
- `GroupedBarPlot` (`maidr/core/plot/grouped_barplot.py:187,195`) uses bare `float(patch.get_width()/get_height())` where `BarPlot` routes the same read through `_magnitude`. `ax.bar(x, a); ax.bar(x, b, bottom=a)` with a gap in `a` (auto-typed STACKED, or via `maidr.stacked`) emits `NaN` and the figure fails to parse.
- `AreaPlot._extract_plot_data` (`maidr/core/plot/areaplot.py:137`) converts positions and magnitudes with `_scalar()` = `float()`; `ax.fill_between(x, y)` (AREA) and `ax.stackplot(x, series)` (STACKED_AREA) with a NaN, or a masked array, emit `{"x": 2.0, "y": NaN}`; a NaN in `x` emits `{"x": NaN, ...}`. The core's `area.ts` already treats a JSON `null` as a gap that does not poison column totals.

The heatmap loop is also the dominant Python-side cost of a large image: for the default `fmt == ''` on float64/int input `format()` is `repr` and `float()` round-trips exactly, so the per-cell call chain is a slow identity. Measured (min of 3): 500x500 0.65 s vs 0.018 s with `astype(float).tolist()`; 1000x1000 3.5 s vs 0.12 s; end-to-end 1000x1000 `imshow` render 5.85 s -> 3.53 s. The fast path must stay gated on dtype (numpy 1.x formats float32 at its own shortest repr).

## Proposed fix

- heatmap: read `ma.asarray(sm.get_array())`, keep the bool->float cast, `ma.filled(array.astype(float), np.nan)` so a masked cell is a gap, then a `_cell(value)` helper (`value if math.isfinite(value) else None`) mirroring `barplot._magnitude`; fast path `array.astype(float).tolist()` when `fmt == ''` and dtype is integer or float64, else the existing `format` loop.
- grouped bar: import `_magnitude` from `barplot.py` (no circular import) and use it at both reads.
- area: skip a point whose position is non-finite (same rule as `lineplot._has_position`; positions are shared so every series drops the same column and stays aligned) and emit `None` for a non-finite magnitude (same as `lineplot._reading` / `stairs._reading`).

## Acceptance criteria

- New `tests/core/plot/test_heatmap_gaps.py` mirroring `test_bar_gaps.py`: NaN cell -> `None`, masked `sns.heatmap(mask=)` -> `None` exactly where the artist's mask is True, `imshow`/`pcolormesh(masked_invalid)` the same, strict `json.loads` round-trip passes, a NaN-free grid is byte-identical; `fmt='.2f'` still emits formatted values; a random float64 grid and an int grid equal `[[float(format(x, '')) ...]]`.
- `tests/core/plot/test_bar_gaps.py` gains a stacked case (`ax.bar` + `ax.bar(bottom=)` + `maidr.stacked`) asserting the gap is `None` and the other points unchanged, plus a `barh(left=)` variant.
- `tests/core/plot/test_areaplot.py`: a NaN magnitude in stackplot and single-band fill_between is `None` and the schema passes strict `json.loads`; a NaN position drops that column from every series; a masked-array variant.

jooyoungseo:
## What

Three matplotlib emitters still wrote non-finite numbers straight into the schema. `json.dumps` serialises them as the bare token `NaN`, `JSON.parse` in the browser rejects it inside a `catch`, and `initMaidrOnElement` never runs — the whole figure loses audio, text, braille and highlight (#427). `BarPlot`, `LinePlot`, `ScatterPlot`, `ErrorbarPlot` and `HexbinPlot` already emit `None` for exactly this reason. Closes #696.

- **Heatmap.** `_extract_scalar_mappable_data` read `sm.get_array().data`, the buffer *under* the mask, and converted every cell with `float(format(x, fmt))`. `sns.heatmap([[1, nan], [3, 4]])`, `imshow` of the same, `pcolormesh(masked_invalid(...))` and `pcolor(masked)` all emitted `[[1.0, NaN], ...]`; `sns.heatmap(corr, mask=np.triu(...))` drew six blank cells and announced the full symmetric matrix. It now reads the masked array, fills masked cells with NaN, and routes every cell through a `_cell()` helper (finite → value, else `None`); the core's `HeatmapData.points` is `(number|null)[][]` and treats `null` as a gap.
- **Grouped bar.** `GroupedBarPlot` used bare `float(patch.get_width()/get_height())` where `BarPlot` routes the same read through `_magnitude`; `ax.bar(x, a); ax.bar(x, b, bottom=a)` with a gap in `a` emitted `NaN`. Both reads now go through `_magnitude`.
- **Area.** `AreaPlot` converted positions and magnitudes with `float()`. A NaN magnitude in `fill_between(x, y)` or `stackplot` is now `None`; a NaN position drops that column from every series (positions are shared, so the series stay aligned), the same two rules `lineplot` applies.

The heatmap loop was also the dominant Python-side cost of a large image: for the default `fmt == ''` on float64/int input, `format()` is `repr` and `float()` round-trips exactly, so the per-cell call chain was a slow identity. Those dtypes now take `astype(float).tolist()`; float32 keeps the loop because numpy formats it at its own shortest repr.

## Measured

| | before | after |
|---|---|---|
| 1000×1000 float64 `imshow`, extraction (min of 3) | 1.44 s | 0.25 s |
| same, end-to-end `render()` | 2.91 s | 2.58 s |

Twelve NaN-free renders (`sns.heatmap` ± `fmt`, `imshow` int/float32, `spy`, `pcolor`, gouraud `pcolormesh`, `hist2d`, stacked bar/barh, `stackplot`, `fill_between`) are byte-identical to `main` after normalising `<dc:date>` and matplotlib's random element ids.

## Output change

Confined to previously broken payloads: NaN or masked heatmap cells, NaN stacked/dodged bars and NaN area samples become `null`; a NaN area *position* drops that column. A masked `sns.heatmap` now announces the hidden half as missing rather than reading the mirrored values, which is what the artist's mask says.

## Tests

New `tests/core/plot/test_heatmap_gaps.py` (25 tests, mirroring `test_bar_gaps.py`), a `TestAStackedBarWithNoHeight` class in `test_bar_gaps.py`, and two cases in `test_areaplot.py`. Against `main`'s sources: 22 failed, 35 passed (every gap test fails on the baseline; every must-not-change guard passes on both).

## Verified

```
uv run pytest      3638 passed, 68 skipped
```

---

### DIFF
--- a/maidr/core/plot/areaplot.py
+++ b/maidr/core/plot/areaplot.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import math
 import uuid
 from typing import Any, Sequence
 
@@ -9,58 +8,8 @@ from matplotlib.axes import Axes
 
 from maidr.core.enum import MaidrKey, PlotType
 from maidr.core.plot import MaidrPlot
+from maidr.core.plot.lineplot import _has_position, _reading
 from maidr.exception import ExtractionError
-
-
-def _has_position(x: Any) -> bool:
-    """
-    Whether a sample sits anywhere a reader could be sent.
-
-    The rule ``lineplot._has_position`` applies, for the reason it gives: a
-    sample whose ``x`` is not finite has no place on the axis, and emitted as
-    it stands it is a bare ``NaN`` token that ``JSON.parse`` refuses, so one
-    stopped the chart initialising at all (#427). Such a sample is dropped
-    -- from **every** series, since the positions are shared, which is what
-    keeps the columns the consumer sums for the running total aligned.
-
-    Parameters
-    ----------
-    x : Any
-        A sample's position as :meth:`AreaPlot._scalar` returned it.
-
-    Returns
-    -------
-    bool
-        False only for a number that is NaN or infinite. A categorical
-        position arrives as a string, which is both placed and never a JSON
-        hazard, so it is kept.
-    """
-    return not (isinstance(x, float) and not math.isfinite(x))
-
-
-def _reading(y: Any) -> Any:
-    """
-    A sample's value, or ``None`` where it was positioned but never measured.
-
-    The counterpart of :func:`_has_position`, following
-    ``lineplot._reading``: a sample with a position and no value is kept,
-    since there is somewhere to send a reader, and its value emitted as
-    ``null``, which the core's area trace reads as a gap that stays out of
-    the running total. A bare ``NaN`` would stop the chart initialising
-    (#427) and a zero would claim a reading of zero.
-
-    Parameters
-    ----------
-    y : Any
-        A sample's magnitude as :meth:`AreaPlot._scalar` returned it.
-
-    Returns
-    -------
-    Any
-        The value unchanged, or ``None`` for a number that is NaN or
-        infinite.
-    """
-    return None if isinstance(y, float) and not math.isfinite(y) else y
 
 
 class AreaPlot(MaidrPlot):
@@ -184,6 +133,15 @@ class AreaPlot(MaidrPlot):
             label = self._labels[index] if index < len(self._labels) else None
             points = []
             for position, magnitude in zip(positions, magnitudes):
+                # The line layer's two rules, for the reason it gives: a
+                # sample with no *position* is nowhere a reader could be
+                # sent and is dropped -- from every series, since the
+                # positions are shared, so the columns the consumer sums
+                # stay aligned -- while one with a position and no *value*
+                # is kept and emitted as `null`, which the core's area trace
+                # reads as a gap that stays out of the running total. Either
+                # written out as a bare `NaN` stops the chart initialising
+                # (#427).
                 x = self._scalar(position)
                 if not _has_position(x):
                     continue
--- a/maidr/core/plot/areaplot.py
+++ b/maidr/core/plot/areaplot.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import math
 import uuid
 from typing import Any, Sequence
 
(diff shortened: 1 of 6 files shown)


jooyoungseo:
Follow-up for the review notes: 56a2efa. The area gap rules are named helpers (`_has_position` / `_reading`, mirroring lineplot's); the heatmap fast path hands a gap-free float64/int grid over as `tolist()` with no per-cell pass and only a grid carrying a NaN goes through the swap, the double cast is gone, and both the code comment and the test docstring now cite one measurement (2.0 s → 0.09 s for extraction of a 1000×1000 float64 `imshow`, min of 3, on a loaded box). Left as is: `_magnitude` is imported from `barplot` rather than promoted to a shared helper, to keep this fix to the three emitters.

---

Issue #701, opened 2026-09-03
Plotly histogram autobin walks every sample in a Python loop (80% of a 100k-sample build); contour tracing spends 40% of its time in per-curve numpy calls


jooyoungseo:
## Problem

`_auto_shift_bins` (`maidr/plotly/histogram.py:193-202`), the port of plotly's `autoShiftNumericBins`, iterates the sample in Python computing three counts with two closure calls per element. It runs once per 1-D histogram and once per axis for `histogram2d`/`histogram2dcontour`. Measured (min of 3): 50k samples 103-137 ms of a 133-205 ms `_flatten_maidr`; 100k 168-325 ms (~80 % of `PlotlyMaidr()`); 500k 1.0-1.4 s.

The three predicates are elementwise; `np.count_nonzero` over the same expressions is bit-identical because iterating an ndarray already yields `np.float64` scalars, so `%`, `+`, `*`, `/` are the same IEEE operations (numpy's remainder follows Python's sign-of-divisor rule). Three independent verifications (4 800 and 5 280 randomised trials across integer, gaussian, rounded, sub-milli, NaN/inf, huge/tiny samples, widths 1e-3..1e6, both caller seed formulas) found zero mismatches; vectorised: 50k 7 ms, 500k 85 ms; `_flatten_maidr` 133 -> 28 ms at 50k. `np.errstate(invalid='ignore')` keeps NaN/inf from raising a warning the loop would not.

`maidr/plotly/contour.py`: `_has_extent` (`:585-588`) calls `np.asarray` plus two `np.ptp` on strided column views per curve, and the series comprehension (`:192-199`) indexes the 2-D curve row by row producing a view and a `float()` per vertex. On a noisy 200x200 field (34k short curves) that is 0.79 s + 0.48 s of a 1.4 s `_extract_plot_data`; `vertices.min(axis=0)`/`max(axis=0)` and `curve.tolist()` give identical floats (contourpy returns float64 whatever `z`'s dtype) in 40 % less time. Negligible on smooth fields (89 curves), so low severity, but exact.

## Proposed fix

- `_auto_shift_bins`: `data = np.asarray(data, dtype=float)`; `near_edge` as an array-capable lambda; `int_count`/`edge_count`/`mid_count` via `np.count_nonzero` under `np.errstate(invalid='ignore')`; keep `n = data.size`, the scalar `near_edge(data_min)`/`near_edge(data_max)` calls and every branch verbatim; document that the counts must stay bit-identical to the loop.
[...]

PR #729, opened 2026-09-03
fix(plotly): read date axes as ISO strings and bin a histogram over its finite samples

### LINKED ISSUE #699: A plotly date axis makes render() raise or announces epoch nanoseconds, and a histogram over a sample with one NaN raises ValueError for the whole figure
## Problem

**Dates.** `PlotlyPlot._to_native` (`maidr/plotly/plotly_plot.py:82`) only unwraps numpy scalars via `.item()`, and `as_list` (`:937`) only decodes typed-array specs. `Figure.to_dict()` exports a date column as a `datetime64[us]` ndarray (px) or leaves `datetime`/`date` objects in a list (go). The schema then hits `json.dumps` with no `default=`:

- `px.line`/`px.scatter`/`px.bar` on a `pd.date_range` column, `go.Scatter(x=[datetime...])`, `x=[date...]`, `np.datetime64[D]`, `go.Bar(x=pd.date_range(...))` -> `TypeError: Object of type datetime is not JSON serializable` from `maidr.render(fig)`.
- `datetime64[ns]` renders but announces `x = 1704067200000000000`.

plotly's own `to_json` emits `np.datetime_as_string(arr).tolist()` for kind `M` arrays and `isoformat()` for `datetime`/`date`, so the accessible layer can say what the figure JSON in the same page says. Nothing in `maidr/plotly` handles dates; no test has a date axis.

**Histogram NaN.** `PlotlyHistogramPlot._extract_plot_data` (`maidr/plotly/histogram.py:646`) coerces with `np.array(values, dtype=float)` (None -> NaN) and hands it to `compute_bin_edges`; `arr.min()`/`max()` are NaN and `_plotly_default_size0` reaches `math.floor(math.log10(NaN))` -> `ValueError: cannot convert float NaN to integer`. Nothing catches it, so one blank in a column kills every layer of the figure. Reproduced for `go.Histogram(x=[1,2,None,...])`, `px.histogram` over a column with one NaN, `color=` (grouped path, `grouped_histogram.py:152`), `y=`, `nbinsx`, size-only, start-only `xbins`; only `start+end+size` survives. plotly.js draws the chart, skipping BADNUM in `autoBin` while keeping `e.length` for the size exponent and subtracting blanks in `autoShiftNumericBins`. `_bin_assignment` already maps NaN to -1 and the 2-D path already filters with `isfinite`. Also found: `px.histogram(color=)` with `xbins.size=-2` raises `IndexError` in `grouped_histogram` where the single-trace path declines.

**Histogram over dates.** `px.histogram` over a datetime column converts silently to epoch microseconds and emits bins of `1e11 us` (~1.16 days) that plotly does not draw (`autoShiftNumericBins` is skipped for `r.type === 'date'`); `go.Histogram(x=list_of_Timestamps)` emits a BAR per distinct timestamp. The 2-D path already declines a date sample (#636).

## Proposed fix

- `as_list`: `dtype.kind == 'M'` -> `np.datetime_as_string(value, unit='auto').tolist()`. `_to_native`: `np.datetime64` -> `datetime_as_string`, `datetime.date` (covers `datetime`, `pd.Timestamp`) -> `isoformat()`, both before the `.item()` branch.
- `compute_bin_edges`: `finite = arr[np.isfinite(arr)]`; empty -> `np.array([])`; min/max/range/stdev/distinct from `finite`; `_plotly_default_size0` gets `sample_size=arr.size` for the exponent so a finite sample is byte-identical; `_shared_edges` returns `None` for fewer than 2 edges (also closes the negative-width IndexError). Keep the original array for `_bin_assignment` so the value array stays aligned.
- `is_temporal_sample(values)` -> decline the 1-D and grouped histogram layers for a temporal sample (debug log), consistent with #636 and the 2-D path; porting plotly's date autobin is a separate feature to measure in a browser first.

## Acceptance criteria

- `tests/plotly/test_plotly_dates.py`: px.line/px.bar on a date column and go.Scatter with datetime, date, datetime64[D], datetime64[ns] x all render and every emitted x is a string `_looks_like_date` accepts.
- `test_plotly_histogram_bins.py`: a sample with None and NaN bins its finite values exactly as the finite sample alone (go/px, vertical/horizontal); an all-blank sample forms no layer; `test_plotly_grouped_histogram.py`: `color=` over a NaN column renders and matches the NaN-free grid; grouped negative width forms no layer.
- `tests/plotly/test_plotly_histogram_dates.py`: date column (px, go with Timestamps, `color=`, horizontal) forms no layer; numeric sample unaffected; unit test for `is_temporal_sample`.

jooyoungseo:
## What

Closes #699.

**Dates.** `PlotlyPlot._to_native` only unwrapped numpy scalars and `as_list` only decoded typed-array specs. `Figure.to_dict()` exports a date column as a `datetime64` ndarray (px) or leaves `datetime`/`date` objects in a list (go), and the schema then hit `json.dumps` with no `default=`: `px.line`/`px.scatter`/`px.bar` on a `pd.date_range` column, `go.Scatter` with `datetime`, `date`, `Timestamp`, `datetime64[D]` x, `go.Bar(x=pd.date_range(...))` — every one raised `TypeError: Object of type datetime is not JSON serializable`; `datetime64[ns]` rendered and announced `x = 1704067200000000000`. Dates are now spelled the way plotly's own `to_json` spells them — `np.datetime_as_string` for arrays, `isoformat()` for objects — so the accessible layer says what the figure JSON in the same page says. One refinement over plotly's per-element `unit='auto'`: an array uses the coarsest unit that loses nothing of any entry, so a daily series reads `2024-01-01` throughout instead of mixing `2024-01-01` and `2024-01-02T12:00`.

**Histogram blanks.** `_extract_plot_data` coerced with `np.array(values, dtype=float)` (None → NaN) and `compute_bin_edges` took NaN min/max into `math.floor(math.log10(NaN))`: `ValueError` for `go.Histogram(x=[1, 2, None, 3, 4, 5])`, for `px.histogram` over a column with one NaN (plain, `color=`, horizontal, `nbinsx`, size-only or start-only `xbins`), and nothing caught it, so one blank took every layer of the figure down. plotly.js draws the chart, skipping BADNUM in `autoBin` while keeping `e.length` for the size exponent. `compute_bin_edges` now bins the finite sample and hands `sample_size=arr.size` to `_plotly_default_size0` for the exponent, so a finite sample is byte-identical (1 500 random cases compared against the unpatched module: all identical). An all-blank sample forms no layer; `_shared_edges` returns `None` for fewer than two edges, which also closes the grouped `xbins.size=-2` `IndexError`.

**Histogram over dates.** `px.histogram` over a datetime column converted silently to epoch microseconds and emitted bins of `1e11 µs` that plotly does not draw; `go.Histogram(x=list_of_Timestamps)` emitted a BAR per instant. A temporal sample now declines the 1-D and grouped layers (debug log), consistent with #636 and the 2-D path; porting plotly's date autobin is a follow-up to measure in a browser first. A sample the author wrote as ISO *strings* still takes the categorical path exactly as before (pinned by a test; same class of bug, out of scope here).

## Output changes

Beyond the raising cases: any layer that received a `datetime64` array now carries ISO strings — the plotly candlestick's date labels change from `2024-01-01 00:00:00` to `2024-01-01`. For `NARROW`-style samples with blanks, the width can differ from the blank-free sample because plotly's `n ** 0.4` counts blanks (documented in the test).

## Tests

New `tests/plotly/test_plotly_dates.py` and `test_plotly_histogram_dates.py`; a `TestABlankIsNoObservation` class in `test_plotly_histogram_bins.py` and `test_plotly_grouped_histogram.py`. Against `main`'s sources: 29 failed, 108 passed, 1 import error (`is_temporal_sample` does not exist before the fix).

## Verified

```
uv run pytest      3655 passed, 68 skipped
```

`_auto_shift_bins` is untouched here; #701 vectorises it separately.

---

### DIFF
--- a/maidr/plotly/grouped_histogram.py
+++ b/maidr/plotly/grouped_histogram.py
@@ -21,6 +21,7 @@ from maidr.plotly.histogram import (
     as_numeric,
     binned_axis,
     compute_bin_edges,
+    is_temporal_sample,
     paired_arrays,
 )
 from maidr.plotly.plotly_plot import PlotlyPlot
@@ -153,7 +154,12 @@ class PlotlyGroupedHistogramPlot(PlotlyPlot):
         if not pooled.size:
             return None
         bins, nbins = group_bin_spec(self._traces, self._binned)
-        return compute_bin_edges(pooled, bins, nbins)
+        edges = compute_bin_edges(pooled, bins, nbins)
+        # Fewer than two edges is no grid: a negative width, a window past
+        # the data, or a pool of nothing but blanks. The single-trace path
+        # reads that as no layer, and so does this, rather than handing an
+        # empty array to `_bin_counts` and raising out of it (#699).
+        return edges if edges.size >= 2 else None
 
     def _extract_plot_data(self) -> list[list[dict]]:
         samples: list[np.ndarray] = []
@@ -164,6 +170,11 @@ class PlotlyGroupedHistogramPlot(PlotlyPlot):
                 samples.append(np.empty(0))
                 values.append(None)
                 continue
+            # A date axis is binned by rules this does not port yet -- see
+            # `is_temporal_sample`. Declining the group as a whole, since
+            # every trace of it shares the one axis.
+            if is_temporal_sample(trace.get(self._binned)):
+                return []
             try:
                 samples.append(np.array(sample, dtype=float))
             except (ValueError, TypeError):
--- a/maidr/plotly/histogram.py
+++ b/maidr/plotly/histogram.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import math
+from datetime import date
+from typing import Any
 
 import numpy as np
 
@@ -35,7 +37,9 @@ def _plotly_round_up(val: float, array: list[float], reverse: bool = False) -> f
     return array[lo]
 
 
-def _plotly_default_size0(arr: np.ndarray, *, is_2d: bool = False) -> float:
+def _plotly_default_size0(
+    arr: np.ndarray, *, is_2d: bool = False, sample_size: int | None = None
+) -> float:
     """Compute the default rough bin size when ``nbinsx`` is not given.
 
     Mirrors the logic in Plotly.js ``axes.autoBin`` (the ``else`` branch
@@ -49,7 +53,7 @@ def _plotly_default_size0(arr: np.ndarray, *, is_2d: bool = False) -> float:
     Parameters
     ----------
     arr : np.ndarray
-        The raw data values.
+        The finite data values.
     is_2d : bool, default False
         Bin an axis of a **two-dimensional** histogram, which plotly bins
         more coarsely: the sample-size exponent is ``0.25`` rather than
@@ -58,9 +62,17 @@ def _plotly_default_size0(arr: np.ndarray, *, is_2d: bool = False) -> float:
         axes across four figures -- gaussian, uniform, a six-value sample and
         a normalised one -- ``0.4`` matched none of them and ``0.25`` matched
         all eight.
+    sample_size : int, optional
+        How long the sample was *before* its blanks were dropped, for the
+        exponent alone. That is the one place plotly reads the whole array:
+        ``distinctVals`` sorts the blanks to the end and stops short of them,
+        and ``stdev`` divides by the numeric count, but the ``n`` in
+        ``n ** 0.4`` is ``data.length`` as given. Defaults to the length of
+        *arr*, which is the same number for a sample with no blanks.
     """
     sorted_arr = np.sort(arr)
     n = len(sorted_arr)
+    n_total = sample_size or n
     if n < 2:
         return 1.0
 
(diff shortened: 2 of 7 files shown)


jooyoungseo:
On the `_auto_shift_bins` denominator: I checked the bundle rather than the docstring. In plotly.min.js `autoShiftNumericBins` does

```
for(...)t[c]%1===0?s++:zh(t[c])||l++, ...
var f=t.length-l;
if(s===f&&r.type!=="date") ... else if(o<f*.1&&(a>f*.3||...
```

`l` counts the non-numeric entries and every threshold is against `f = t.length - l`, so plotly's `n` *excludes* blanks and passing the finite sample is exactly its rule; the docstring wording was the problem. f64260e rewords it to quote that form, and adds a test that discriminates the two readings (40 integers plus 30 `None` with an explicit width: the finite rule gives the integer-cluster start `-0.5`, the whole-length rule would give `-2.5`). The one term that does see the whole length is the width exponent, which `sample_size` already carries.

Same commit: a plain list of `np.datetime64` scalars now takes the same one-unit-per-series spelling as an ndarray (`Figure.to_dict()` does keep such a list as a list, so it was reachable), with a test.

---

PR #765, opened 2026-09-04
docs: describe what py-maidr does in the README and package metadata

jooyoungseo:
## Description

The README opened with **"Python binder for maidr library"** and the PyPI summary was the expanded acronym on its own. Neither said what the package does, and neither named a single plotting library it supports. Someone searching for an accessible matplotlib or seaborn chart had nothing to match on.

Prose, metadata and links only — no behaviour change.

### README

- **Badges**: PyPI version, Python versions, monthly downloads, CI, license, docs. There were none, which reads oddly for a package at 1.23.0 with 92 releases.
- **Tagline and overview**: names the four supported plotting libraries and the four output modalities, and says the integration is a single `import maidr`.
- **Quickstart**: the README had no runnable code at all. The example added here was executed against this branch before committing.
- **What is supported**: a table of libraries, environments and plot-type count that *links to* [Plot Type Stability](https://py.maidr.ai/stability.html) rather than duplicating its 38 rows, so there is no second copy to keep in step (per the note in `CLAUDE.md`).
- **Contributing / License** sections.

### pyproject.toml

- **`description`** rewritten — this is the one line PyPI shows in search results.
- **Keywords**: added the libraries and environments actually supported (`matplotlib`, `seaborn`, `plotly`, `altair`, `jupyter`, `quarto`, `shiny`, `streamlit`) plus `a11y`, `screen reader`, `data visualization`, `charts`.
- **Classifiers**: added `Topic :: Adaptive Technologies` (the classifier readers browse to find work like this), `Framework :: Matplotlib`, `Framework :: Jupyter`, `Topic :: Multimedia :: Sound/Audio`, `Operating System :: OS Independent`, `Intended Audience :: Education`, `Natural Language :: English`, and a `Development Status`.
- **Project URLs**: added Documentation, Issues and Changelog.

### Canonical docs URL

`docs/CNAME` says `py.maidr.ai`, but the README and `[project.urls].Homepage` pointed at `https://xability.github.io/py-maidr/`, which 301s to it. Every docs link now points at the canonical domain directly.

## One thing worth a second opinion

I set `Development Status :: 5 - Production/Stable`. There was no `Development Status` classifier before, so this is a new claim rather than a corrected one. It seemed right for 1.x with 92 releases and a documented stability policy, but `4 - Beta` is defensible given 23 of the 38 plot types are experimental. Happy to change it.

## Type of Change

- [x] Documentation update

## Checklist

## Verification

| check | result |
|---|---|
| `ruff check --diff` | clean |
| `uv lock --check` | passes |
| `uv build --wheel` | builds |
| `twine check` | PASSED |
| `pytest` (full suite) | 4337 passed, 69 skipped, 2 xfailed |

Three tests fail in my local Windows checkout (`test_collection_survives_missing_extras` ×2 and `test_shell_guard_matches_the_python_validator`). They fail identically on `main` at 16d1e21 with no changes applied, so they are pre-existing and environmental, not from this branch.

The README quickstart was run against this branch and produced a 26 KB accessible HTML file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,64 @@
 
 <hr style="color:transparent" />
 <br />
+
+[![PyPI](https://img.shields.io/pypi/v/maidr.svg)](https://pypi.org/project/maidr/)
+[![Python versions](https://img.shields.io/pypi/pyversions/maidr.svg)](https://pypi.org/project/maidr/)
+[![Downloads](https://img.shields.io/pypi/dm/maidr.svg)](https://pypistats.org/packages/maidr)
+[![CI](https://github.com/xability/py-maidr/actions/workflows/ci.yml/badge.svg)](https://github.com/xability/py-maidr/actions/workflows/ci.yml)
+[![License](https://img.shields.io/pypi/l/maidr.svg)](LICENSE)
+[![Docs](https://img.shields.io/badge/docs-py.maidr.ai-1f6feb.svg)](https://py.maidr.ai/)
+
 </div>
 
 # py-maidr
 
-Python binder for maidr library
+Make matplotlib, seaborn, Plotly, and Altair charts accessible — MAIDR
+(Multimodal Access and Interactive Data Representation) for Python.
+
+## Overview
+
+py-maidr makes data visualizations accessible to blind and low-vision readers.
+`import maidr` and your existing plotting code renders an interactive figure that
+can be **navigated by keyboard**, **heard as sonification**, **read on a braille
+display**, and **described as text** — the same chart in four modalities instead
+of one.
+
+Nothing about how you plot changes. py-maidr patches matplotlib and seaborn at
+import time, so `plt.show()` produces accessible HTML instead of a static image;
+Plotly and Altair figures go through the same `maidr.show()` entry point. It
+works in a plain script, a Jupyter notebook, Quarto, Shiny, and Streamlit.
+
+## Quickstart
+
+```python
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+import maidr  # the whole integration
+
+penguins = sns.load_dataset("penguins")
+
+fig, ax = plt.subplots()
+sns.barplot(x="species", y="body_mass_g", data=penguins, ax=ax)
+
+maidr.show(fig)
+```
+
+`maidr.save_html(fig, "penguins.html")` writes the same plot to a standalone HTML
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Went through the four items.

**2. Tagline vs. table (mplfinance)** — Fixed in 67768cc. The omission was deliberate but the table didn't say so, which is the same as an oversight from a reader's side. `maidr/patch/mplfinance.py` patches `PlotType.CANDLESTICK` plus the `BAR` volume panel and `LINE` moving-average overlays that mplfinance draws alongside it — an OHLC integration, not a general-purpose plotting library like the other four. The table now reads "plus mplfinance for financial charts (candlestick, volume, moving averages)", so the four-library tagline is visibly a scoping decision rather than a miscount.

**1. `Development Status :: 5 - Production/Stable`** — Keeping it, and here's the reasoning rather than just the preference.

`Development Status` is a property of the distribution, not the union of guarantees over every feature it ships. The argument that 23 experimental plot types force `4 - Beta` proves too much: matplotlib ships provisional APIs, and pandas ships whole experimental subsystems, and neither is Beta on PyPI. The classifier answers "should someone depend on this in production?", and three things here already answer yes — the version is `1.23.0`, which under SemVer is itself the declaration that the public API is stable; there have been 92 releases; and `docs/stability.qmd` says the fifteen stable types "have been exercised by real users over real charts."

The experimental types are also not an unlabeled hazard. They're enumerated in `docs/stability.qmd`, enforced by `tests/core/test_plot_type_stability.py`, and — as of this PR — the stability page is linked directly from the README's support table, which is where a new reader will land. `4 - Beta` on a 1.x package would contradict the version number and undersell fifteen plot types that are not beta in any sense.

That said, you're right that it's the one judgment call in the diff rather than a factual correction, so it's flagged for the maintainer in the PR description and it's a one-line revert if they read the maturity differently.

**3. CI badge tied to `ci.yml`** — Agreed, and no change. Worth noting the failure mode is visible rather than silent: a renamed workflow makes the badge render "CI — no status", not a green badge that lies. That's the right direction for a badge to fail in.

**4. License badge vs. SPDX** — Agreed, no action. `img.shields.io/pypi/l/maidr` reads PyPI's `License-Expression`, which the built wheel already emits as `GPL-3.0-or-later` (confirmed via `twine check` on the wheel from this branch), so the two sources are reading the same field.

PR #766, opened 2026-09-04
docs: source the downloads badge from pepy instead of pypistats

jooyoungseo:
## Description

Follow-up to #765. The downloads badge added there resolves through `img.shields.io/pypi/dm`, which reads pypistats.org — and pypistats has been answering 429, so the badge renders as **"downloads: rate limited by upstream service"** rather than a number. Checked repeatedly over ~40 minutes with the same result, so this is not a passing blip. A visibly broken badge is worse than no badge.

`img.shields.io/pepy/dt/maidr` reads pepy.tech instead and currently renders **"downloads: 84k"**. The count is lifetime rather than monthly; for a package with 92 releases that is arguably the more useful number, and it is the one that actually renders.

Link target moves from `pypistats.org/packages/maidr` to `pepy.tech/project/maidr` to match the data source.

## Type of Change

- [x] Documentation update

## Checklist

## Verification

```
curl img.shields.io/pypi/dm/maidr.svg  -> <title>downloads: rate limited by upstream service</title>
curl img.shields.io/pepy/dt/maidr      -> <title>downloads: 84k</title>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 [![PyPI](https://img.shields.io/pypi/v/maidr.svg)](https://pypi.org/project/maidr/)
 [![Python versions](https://img.shields.io/pypi/pyversions/maidr.svg)](https://pypi.org/project/maidr/)
-[![Downloads](https://img.shields.io/pypi/dm/maidr.svg)](https://pypistats.org/packages/maidr)
+[![Downloads](https://img.shields.io/pepy/dt/maidr.svg)](https://pepy.tech/project/maidr)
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Noted on pepy.tech being the smaller service — that risk is real and it is the reason the swap is worth recording rather than doing silently. Worth adding that the two fail differently: pypistats returning 429 renders the badge as the sentence "downloads: rate limited by upstream service", which is wide, wrong, and reads as neglect. A pepy outage renders "downloads: inaccessible", which is narrower and unmistakably a service problem rather than a project one. If it does regress, dropping the downloads badge entirely is a reasonable third option — the PyPI version and CI badges carry most of the trust signal on their own.

PR #767, opened 2026-09-05
ci: keep the commit body out of the changelog

jooyoungseo:
## Description

`CHANGELOG.md` here is 6,900 lines where the JS binding's, on the same conventional commits and squash merges, is one line per entry. The difference is not the release process but the renderer: python-semantic-release's default Markdown template appends the commit body to every entry ("to make sure to not hide contents of squash commits"), while `conventional-changelog` in xability/maidr only prints the subject. Every commit on `main` is a squash merge whose body is the whole pull request description, so v1.23.0's twelve entries ran to some six hundred lines, complete with "Closes #N", `Co-authored-by` trailers and Claude Code session links.

This PR gives python-semantic-release its own template directory with that one block removed.

## Related Issues

None filed; raised in conversation.

## Changes Made

- **`templates/`**: a verbatim copy of python-semantic-release 9.21.0's `data/templates/angular/md` (the set the `conventional` parser renders with), because a template directory replaces the bundled one outright and an `include` cannot reach back into the package. The one deliberate difference is in `.components/changes.md.j2`, where `commit.descriptions[1:]` is no longer appended to each entry. The top of `CHANGELOG.md.j2` records the upstream version and where the difference lives, so upgrading PSR is re-copy plus re-apply. A `BREAKING CHANGE:` paragraph is still printed in full in its own section.
- **`pyproject.toml`**: `template_dir = "templates"` written out (it is the default) with a comment, and `autoescape = false`. A template directory is rendered with the configured environment, whereas the bundled one is rendered with escaping off regardless; with escaping on, the apostrophe in "chart's" came out as `&#39;`.
- **`tests/core/test_changelog_template.py`**: renders `templates/` the way `semantic-release changelog` does (PSR's own loader, context and GitHub filters) against a synthetic squash commit and holds the body out, the breaking-change paragraph in, the apostrophe unescaped, and the render to exactly one output file. That last one matters: `recursive_render` copies any non-`.j2` file in the directory to the project root, so a `README.md` explaining the templates would have overwritten the project's README at the next release.
- **`CONTRIBUTING.md`**: says the PR title is the whole of the PR that reaches the changelog.

**Verified** by rendering the full history with PSR 9.21.0's own code, first with the bundled template (reproduces the committed `CHANGELOG.md` byte for byte, plus an Unreleased section) and then with `templates/`:

| | lines | bytes |
|---|---|---|
| bundled template | 6,965 | 374,112 |
| `templates/` | 2,321 | 103,586 |

v1.23.0 after:

```
## v1.23.0 (2026-08-31)

### Bug Fixes

- **seaborn**: Read a box or boxen chart's grouping from the legend that names it
  ([#676](https://github.com/xability/py-maidr/pull/676),
  [`50b2197`](https://github.com/xability/py-maidr/commit/50b2197539de5c8bc14443e4246fa855cc1cf953))

- **seaborn-objects**: Name a colour split from the legend wherever it was put
  ([#673](https://github.com/xability/py-maidr/pull/673),
  [`c5c76f5`](https://github.com/xability/py-maidr/commit/c5c76f50ca3d27bd6953b1d24ca90828f1ce26f7))
```

The committed `CHANGELOG.md` is not touched here: the changelog runs in `init` mode with no insertion flag, so the next release regenerates the whole file and the history shrinks then, without a 4,600-line diff in this PR.

N/A

## Checklist

- [ ] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass. (No manual browser test applies; `pytest tests/core/test_changelog_template.py tests/core/test_changelog_filter.py` passes, `ruff check --diff` is clean.)

## Additional Notes

Both the changelog and the GitHub release notes share the same component, so release notes become subject-only too, matching xability/maidr's releases. Entries still wrap at 100 columns with a hanging indent, as upstream does; a one-line-per-entry layout like the JS changelog would be a further, separate change to the same template.

https://claude.ai/code/session_01Mpt9bvWpGHew54YNysGT1o

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mpt9bvWpGHew54YNysGT1o)_

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,7 +89,10 @@ Types: `feat`, `fix`, `docs`, `perf`, `refactor`, `style`, `test`, `build`,
 `chore`, `ci`. Releases are cut automatically from `main` by semantic-release,
 so the type matters: `feat` makes a minor release and `fix` or `perf` a patch
 release. Pull requests are squash-merged and the PR title becomes the commit
-subject, so the title must follow the same format.
+subject, so the title must follow the same format. The title is also the
+whole of the pull request that reaches `CHANGELOG.md` and the GitHub release
+notes: the description becomes the commit body, which the changelog leaves
+out. Write the title as the one line a reader of the release notes sees.
 
 ## Documentation
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,6 +236,15 @@ prerelease_token = "rc"
 prerelease = false
 
 [tool.semantic_release.changelog]
+# ``templates/`` is a copy of python-semantic-release 9.21.0's own Markdown
+# templates with one change: an entry is the commit's subject line, and the
+# body is left out.  Every commit on ``main`` is a squash merge whose body is
+# the whole pull request description, and the upstream template appends that
+# body to every entry.  ``templates`` is the default, written out so it is
+# visible; ``tests/core/test_changelog_template.py`` renders the directory
+# the way a release does.
+template_dir = "templates"
+
 # ``exclude_commit_patterns`` belongs to *this* table, not to
 # ``changelog.default_templates``.  python-semantic-release parses its config
 # with pydantic, which silently drops unknown keys, so while this list lived
@@ -274,7 +283,10 @@ lstrip_blocks = false
 newline_sequence = "\n"
 keep_trailing_newline = false
 extensions = []
-autoescape = true
+# The bundled template is rendered with escaping off regardless; a template
+# directory of our own is rendered with *this* environment, and with escaping
+# on the apostrophe in a subject reaches the Markdown as ``&#39;``.
+autoescape = false
 
 [tool.semantic_release.commit_author]
 env = "GIT_COMMIT_AUTHOR"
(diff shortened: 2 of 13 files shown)


jooyoungseo:
Addressed in 430f372.

1. **Stale copy after a PSR bump** — two tests now fail loudly. `test_the_copy_is_of_the_release_tooling_in_use` holds the version named at the top of `CHANGELOG.md.j2`, the installed `semantic_release.__version__` (the `dev` pin) and the action tag in `release.yml` to one value. `test_everything_else_is_verbatim_upstream` compares every file in `templates/` other than the two changed on purpose byte for byte against the bundled `angular/md` set, so a bump whose templates changed fails on the file that changed rather than rendering an old layout.
2. **First-release path** — left untested on purpose: `first_release.md.j2` is now guarded as a verbatim upstream copy by the test above, and the project is 94 tags past the branch that would render it.
3. **`changelog_update.md.j2`** — same: verbatim upstream, unreachable in `init` mode, and covered by the byte comparison rather than by rendering.
4. CI is green across the 3.9–3.13 matrix on both heads.

---

PR #768, opened 2026-09-06
docs: keep every page under Googlebot's 2 MB fetch and add canonical, entity and citation metadata

jooyoungseo:
## Description

SEO and generative-engine (Google AI Overviews / AI Mode) optimization of https://py.maidr.ai, following Google's guide [Optimizing your website for generative AI features on Google Search](https://developers.google.com/search/docs/fundamentals/ai-optimization-guide) and the [Googlebot](https://developers.google.com/search/docs/crawling-indexing/googlebot) documentation. Google's position is that AI features need nothing beyond standard SEO: a page must be indexed and snippet-eligible, structured data must mirror visible text, and llms.txt / special AI files neither help nor hurt.

The decisive finding was page size. Googlebot fetches only the first **2 MB** of an HTML file (uncompressed) and discards the rest. py-maidr inlined the 1.9 MB `maidr.js` bundle once per rendered figure, so the live pages were:

| Page | Live size | Headings within 2 MB | After this PR |
|---|---|---|---|
| index.html | 3.96 MB | 10 of 33 | ~160 KB |
| examples.html | 61.4 MB | 5 of 40 | hub ~60 KB + 11 family pages of 55-430 KB |
| examples-plotly.html | 36.9 MB | 4 | ~650 KB |
| examples-altair.html | 2.76 MB | | ~870 KB |

Everything after the first plot on the home page (offline use, keyboard shortcuts, braille, AI feature setup) and 25 of the 26 plot types on the examples page were invisible to Google.

### Changes

- **Page size:** `docs/_environment` sets `MAIDR_USE_CDN=true` and `MAIDR_CDN_VERSION=bundled`, so figures load `maidr.js` from the pinned CDN copy (byte-identical to the bundled 4.6.0) instead of inlining it per figure. No package code changes.
- **Examples split:** `docs/examples.qmd` is now a hub at the same URL; the 26 plot types moved verbatim into 11 family pages under `docs/examples/` (bar, pie/word cloud, histogram/KDE, line/step, heatmap/hexbin/contour, box/boxen/violin, area/error bar/point/lollipop, scatter/regression, multi-layer/panel/facet, candlestick/gantt, dashboards), each with its own title, description, setup cell and a short intro stating how the family is navigated and its stability status. One rankable URL per plot family, navbar menu and sidebar updated.
- **Head hygiene:** `canonical-url: true` (no page had a canonical), `lang`, `author-meta`, `site-url` without trailing slash, locally hosted social image with `summary_large_image`, page footer with cross-links, `404.qmd`, `llms.txt` published via `project.resources`, home `<title>` carries the brand once.
- **quartodoc API pages** get a real `pagetitle` and description via a pre-render script (`docs/_scripts/api_front_matter.py`), and the reference now covers the public API (`show`, `render`, `save_html`, `stacked`, `close`, `set_use_cdn`, `init_notebook`, Shiny and Streamlit widgets).
- **Structured data:** one JSON-LD graph consistent with maidr.ai and r.maidr.ai (shared `@id`s for the (x)Ability Design Lab, the software nodes with `isBasedOn` the core, WebSite, home-only WebPage with `relatedLink`, people and the two MAIDR papers by DOI). `softwareVersion` is now stamped by python-semantic-release (`version_variables`) instead of being hand-edited (it was stale at 1.13.0).
- **Content:** definition paragraph at the top of the home page, "Related projects" and "Cite py-maidr" sections, consistent naming (`py-maidr`), emoji removed from headings, corrected the sentence that listed pie and contour plots as unsupported.
- **CI:** `docs.yml` renders on pull requests that touch `docs/**` with a page-size guard (fails above 1,900,000 bytes), and rebuilds after the release workflow so the site picks up each version (pushes made with `GITHUB_TOKEN` never trigger `on: push`).

### Notes for reviewers

- The docs job now runs on this PR and prints the five largest rendered pages.
- `softwareVersion` in the JSON-LD is not shown as visible text on the site; it is kept because it is auto-stamped and never contradicts the page. Drop it if strict visible-text mirroring is preferred.
- Verified locally with Quarto 1.10.18: zero inlined bundles on every page, all canonicals self-referential, every JSON-LD `@id` resolves on-page, `ruff` and `uv lock --check` clean, `tests/core` passing.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation update

## Checklist

https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn

---
_Generated by [Claude Code](https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn)_


jooyoungseo:
Thanks for the review. Status of each finding:

**1. Write-scoped token on the pull_request path: fixed.** The workflow now has a separate `render-pull-request` job with `permissions: contents: read` that runs quartodoc, `quarto render` and the page-size guard, and the `build-and-deploy` job (which keeps `contents: write` / `pages: write`) is gated with `github.event_name != 'pull_request'`. The top-level default is now `contents: read`. actionlint passes.

**2. `version_variables` and the quoted JSON key: verified to work, no change needed.** python-semantic-release 9.21.0 builds the pattern with an optional matching quote around the variable name (`(?P<quote1>['"])?softwareVersion(?P=quote1)?\s*(:=|[:=@])\s*...`, see `semantic_release/version/declarations/pattern.py`). Ran the real package against a copy of `docs/_quarto.yml`:

```
>>> decl = PatternVersionDeclaration.from_string_definition("docs/_quarto.yml:softwareVersion", tag_format="v{version}")
>>> decl.parse()
{Version(major=1, minor=23, patch=0, ...)}
>>> decl.update_file_w_version(new_version=Version.parse("1.24.0", tag_format="v{version}"), noop=False)
"softwareVersion": "1.24.0"
```

Exactly one match in the file, and the YAML/JSON stays valid after the rewrite.

**3. Docked sidebar: renders as intended.** Quarto attaches a named sidebar to every page listed in its `contents`, so no per-page `sidebar:` key is needed. In the local render `examples.html` and every `examples/*.html` carry `<nav id="quarto-sidebar">` with the family list; `index.html` (not listed) has none, which is the intended layout.

**Minor items:** the front-matter script now skips leading blank lines and exits non-zero if a page has no quartodoc heading, so a format change fails the build instead of silently shipping an untitled page. The size guard stays on the pull-request path only; the deploy job renders inside `quarto-actions/publish`, so guarding it would mean rendering twice.

---

jooyoungseo:
Thanks, both reviews. Everything actionable is now pushed in `966fe4b`; here is each item.

**Fixed**

1. **The size guard did not cover the publish path.** Correct, and the most useful finding here: `quarto-actions/publish` renders on its way out, so a push to `main` could still ship an oversized page. The deploy job now renders, runs the same guard, and publishes with `render: false` so what ships is what was checked.
2. **No concurrency group.** Added, keyed so publishes serialise (`cancel-in-progress: false`) while pull request renders supersede themselves per branch.
3. **Checkout ref for `workflow_run`.** Pinned to `main` for that event, so the chained run publishes the release commit rather than relying on default checkout behaviour.
4. **Malformed cell options.** Fixed in `histogram-kde.qmd` and `line-step.qmd` (`# |` is not a cell option, `#|` is). To be exact about the effect: these cells emit a maidr iframe rather than a matplotlib image, so `fig-alt` would not have applied even when spelled correctly. The accessible name comes from the chart title on the iframe, which is present (`title="Petal Lengths in Iris Dataset, accessible chart"`). The lines were dead config either way and are now valid.
5. **`_truncate` could exceed its own limit** by two characters on the ellipsis path. Fixed by cutting at `limit - 3`.
6. **`jupyter: python3` on `dashboards.qmd`.** Dropped; the page has zero executable cells, so it no longer starts a kernel.
7. **Silent skip in the front-matter script.** Addressed in the previous push: it now skips leading blank lines and raises `SystemExit` when a page has no quartodoc heading, so a format change fails the build instead of shipping an untitled page.

**Verified, no change needed**

- **`version_variables` matching the quoted JSON key.** Both reviews flagged this as unverified, so I ran the pinned package against a copy of the file:
  ```
  >>> decl = PatternVersionDeclaration.from_string_definition("docs/_quarto.yml:softwareVersion", tag_format="v{version}")
  >>> decl.parse()
  {Version(major=1, minor=23, patch=0, ...)}
  >>> decl.update_file_w_version(new_version=Version.parse("1.24.0", ...), noop=False)
  "softwareVersion": "1.24.0"
  ```
  python-semantic-release 9.21.0 builds the pattern with an optional matching quote around the key, so exactly one span matches and the YAML/JSON stays valid.
- **`website.image` not in `project.resources`.** Quarto picks it up as a referenced resource; `_site/media/logo.jpg` is present in the local build (260 KB), so `og:image` resolves.

**Not taking**

- **Inline stability wording duplicating `stability.qmd`.** Each family page names whether its plot types are stable or experimental because that is the sentence a reader (or an AI answer) needs at the top of the page. It is the same duplication the README already carries, and `test_plot_type_stability.py` remains the single enforced source.
- **`paths: docs/**` not covering `maidr/**`.** Adding it would render all 60 example plots on every package pull request. The docs build already runs on `main`, so a package change that breaks quartodoc surfaces there.
- **Redirects for the moved plot-type anchors.** `examples.html` stays a hub at the same URL and links every family page, so old deep links land on a page that leads where the reader wants. Quarto `aliases` create whole-page redirects, not fragment ones, so they cannot map `examples.html#bar-plots` to the new page.

The merge commit message will not survive a squash merge, so the final message is the PR title, which is conventional.

---

jooyoungseo:
Thanks both. Taking the one actionable nit and answering the rest.

**Duplicated size guard: fixed** in `a482571`. Fair catch, and it was a regression I introduced when I moved the guard onto the publish path: the byte limit ended up in two `run:` blocks that could drift. Both jobs now call `docs/_scripts/check-page-sizes.sh docs/_site`, so the threshold lives in one place. actionlint and shellcheck are clean on it, and it reports the same numbers locally:

```
Five largest HTML files under docs/_site:
868567 docs/_site/examples-altair.html
645743 docs/_site/examples-plotly.html
429098 docs/_site/examples/box-boxen-violin.html
400260 docs/_site/examples/heatmap-hexbin-contour.html
275789 docs/_site/examples/multi-layer-panel-facet.html
All HTML pages are at most 1900000 bytes.
```

**Weekly no-op rebuild.** Correct reading, and intended. The release workflow's cron finishing without releasing does trigger a rebuild. That is a few minutes of runner time once a week in exchange for never serving a stale `softwareVersion`, and gating it would mean comparing deployed state, which is more machinery than the problem deserves.

**Cross-site `@id` coupling.** Also correct: `py.maidr.ai` references `https://maidr.ai/#organization` and `#jooyoung-seo`, which the sibling site defines. That is deliberate, since the point is one entity across three hosts, and each page still emits a self-describing stub of any node it references so nothing dangles on-page. The companion PRs (xability/maidr#1215, xability/r-maidr#299) land the same `@id` values, and each carries the shared graph in a comment header so a future edit sees the constraint.

**Old plot-type fragments.** Agreed there is no fragment-level redirect in Quarto. The hub keeps the URL and links every family page, so an old bookmark lands somewhere that leads to the right chart rather than a 404.

Two things one review noted that I want to confirm rather than let pass as assumption, because they are the kind of thing that is easy to state and wrong:

- `MAIDR_CDN_VERSION=bundled` does resolve through `_bundled_version()` rather than pinning a literal, so the CDN copy always matches the JS in the installed wheel. That is why the docs cannot drift from the package.
- The split did fix two latent chunk-ordering bugs: `scatter-regression.qmd` and `area-errorbar-point-lollipop.qmd` now load `iris` and `penguins` in their own setup cells instead of inheriting them from an earlier chunk on the old single page.

---

jooyoungseo:
Thanks — taking the three observations in turn.

**2 (`jupyter: python3` on `examples.qmd`) — fixed in `f6d77be`.** You were right that this needed a second look, and the answer was clearer than "maybe use `eval: false`": after the split the page has *no* executable cells at all. Its two remaining `python` blocks are plain fenced code, not `{python}` cells, so the kernel started on every render and executed nothing. `examples.qmd` was the only page in `docs/` declaring an engine with zero cells. I dropped the key and re-rendered with Quarto 1.10.18 to confirm the page still builds without an engine, and that the canonical link, the description, and the `#sec-examples` anchor `index.qmd` points at all come out unchanged.

**1 (workflow duplication) — deliberately left as is.** The two jobs share four steps, but they diverge on the things that matter: the token scope, the checkout `ref`, and `render: false` on the publish. A composite action would have to take all three as inputs, which trades a visible duplication for an indirection that hides exactly the security-relevant difference between the jobs. The drift risk you're pointing at is really about the size limit, and that one is already in a single script both jobs call. I'd revisit this if a third caller appears.

**3 (`softwareVersion` in JSON-LD) — flagging for a maintainer, not changing it.** Google's guidance is that structured data must not contradict the page, and this doesn't: it's a machine-readable field with no visible counterpart rather than a claim the page disputes. It's also the one field that must stay correct without anyone remembering to touch it, which is why it goes through `version_variables`. If a maintainer would rather it be visible too, the natural home is next to the citation block on `index.qmd`, and I'm happy to add that.

---

jooyoungseo:
Both "worth confirming" items checked against the source rather than assumed, and the test gap is closed.

**`render: false` on `quarto-actions/publish@v2` — confirmed honored, not silently ignored.** I read the composite action at the `v2` tag itself (`8a96df1`), not the default branch. It declares a `render` input defaulting to `"true"` and branches on it:

```bash
if [ "${{ inputs.render }}" != "true" ]; then
  quarto publish $TARGET ${{ inputs.path }} --no-render --no-browser
else
  quarto publish $TARGET ${{ inputs.path }} --no-browser
fi
```

The comparison is against the literal string `"true"`, and a composite action stringifies its inputs, so `render: false` reaches it as `"false"` and takes the `--no-render` path. Your reasoning about why this matters was right: had the flag been ignored, the publish job would have re-rendered unchecked and thrown away the output the size guard just validated.

**`version_variables` — already verified by running the tool.** This one I checked when I added it, since it is the first non-Python target in the list. I ran python-semantic-release 9.21.0 against a copy of `docs/_quarto.yml` and it rewrote the `"softwareVersion": "1.24.0"` line correctly. Its pattern allows optional matching quotes around the key, which is what makes the quoted-JSON form match at all. So it is not a silent no-op.

**Tests for the pre-render script — added in `fb6f412`.** You were right that this is the code most likely to regress quietly on a quartodoc upgrade, and "outside the package" was a weak reason to leave it uncovered. There are now 64 cases in `tests/docs/test_api_front_matter.py`, loading the script by path since it is a Quarto pre-render step. They cover paragraph extraction past fences, headings, tables and callouts, RST role cleaning, YAML escaping, and `process` for idempotence and for failing loudly.

Writing them caught something worth stating: two of my first attempts at the sentence- and semicolon-boundary cases failed, and the code was right both times. The boundary has to fall past the halfway mark of the limit or it is rejected, otherwise a text opening with a short sentence would be truncated to just that sentence. My test strings were a character too short. The guard is now pinned by a case that asserts the boundary's position, plus a parametrized case holding the output under the limit across input lengths from 155 to 340 words and for a single unbroken 400-character token.

**On the hyphenated-anchor regex:** agreed that going red is the intent, and there is now a test asserting exactly that, so the behavior is deliberate rather than incidental.

Full suite after the change: 4414 passed, 69 skipped, 2 xfailed.

---

jooyoungseo:
All three acted on in `bb9c668`, including the two you framed as "not asking for a change" — both turned out to be worth making.

**2 (cross-site `@id` sharing) — you're right about the mechanism, and it exposed a real gap.** Structured-data consumers do evaluate each document's JSON-LD independently, so shared `@id`s across origins buy nothing on their own. The design never depended on merging: every referenced `@id` has an on-page stub node, so each page's graph is self-contained. And `sameAs` was already there on the Organization and on this site's own software node, pointing at the GitHub organization and at the repository plus the PyPI listing.

What was missing is exactly where your reasoning leads. The `maidr.ai/#software` node, which this site references through `isBasedOn`, was a bare name and URL with no external identifier of its own. A consumer reading only this page saw "based on some thing called MAIDR" with nothing to resolve it against. It now carries `sameAs` to the core's repository. I applied the same fix to r-maidr in `7e1caf6`, where both the core stub and the R package's own node were missing identifiers; the R node now points at its repository and its CRAN page. The core repository already emits its own.

**3 (the bar label) — fixed, and the nav was the wrong one.** I checked which of the two variants was accurate rather than just unifying them: the page covers count plots in four places and its own description says "bar, count, stacked and dodged", matching the hub and llms.txt. Only the nav label omitted count, so the nav changed.

**1 (NumPy-style docstrings) — done.** Being outside the package is a weak reason to skip a convention the repository documents for all functions and classes. The added sections carry the two things a summary line cannot: why a sentence boundary before the halfway mark is rejected, and why an unrecognized heading exits instead of falling back.

Verified after the change: the YAML parses, the JSON-LD extracts and parses as JSON with the expected node count on both sites, r-maidr's template still opens with exactly the whisker include line, and the pre-render script produces byte-identical front matter on a synthetic API directory and stays idempotent on a second run. Lint, format and the 64 tests pass.

---

jooyoungseo:
Chased down the one item you flagged as worth confirming before merge. It works, and I confirmed it two ways rather than by reading the regex.

**Ran the tool.** I copied `docs/_quarto.yml`, pointed python-semantic-release 9.21.0's own `PatternVersionDeclaration` at `<copy>:softwareVersion`, and asked it to stamp version 9.9.9:

```
BEFORE: '                "softwareVersion": "1.23.0",'
AFTER : '                "softwareVersion": "9.9.9",'
```

Not a no-op. It also parsed the existing value back out as `1.23.0` before rewriting, so the read path works too.

**Read the pattern it builds.** Your `=` versus `:` question is answered directly in the source, whose own comment says it: the assignment operator is `\s*(:=|[:=@])\s*`, so walrus, equals, colon and `@` are all accepted. A second group handles "optional matching quotations around variable name", which is what makes the quoted-JSON key `"softwareVersion"` match at all rather than only a bare identifier.

Two things worth noting that the run surfaced:

- The comment line in `_quarto.yml` that mentions `softwareVersion` in prose is **not** rewritten. Only the assignment line is, thanks to the negative lookbehind on `[\w.-]`. I checked this specifically, since a greedy match there would have corrupted a comment on every release.
- After the rewrite the file still parses as YAML, and the JSON-LD block still parses as JSON with `softwareVersion` reading `9.9.9`. A substitution that produced valid YAML but broken JSON would have been the nastier failure, since nothing downstream would catch it.

So the "silently becomes a no-op and goes stale again" failure mode you described is real in principle but not what happens here.

**On validating the emitted JSON-LD content:** agreed that nothing in CI checks it today, only page size. I validate it with a parser locally on every edit, but that is not the same as a gate. I'd rather not bolt a structured-data check onto this PR, since doing it properly means resolving every `@id` against the nodes actually present in the rendered HTML, and that deserves its own change rather than riding along here. Worth filing separately.

---

PR #770, opened 2026-09-06
docs(seo): give reference managers real bibliographic metadata

jooyoungseo:
## Description

Saving any page of py.maidr.ai to Zotero produced a Web Page item titled `<page> – py-maidr`, with the lab name as its only "author" and no date. The site emitted Open Graph, Twitter and a single `author` meta tag; Zotero's Embedded Metadata translator looks for Highwire Press `citation_*` first, then Dublin Core, then Open Graph, so it fell back to the `<title>`.

## Changes

**`docs/_scripts/dublin_core.py`** (new, stdlib only, idempotent) runs as a Quarto `post-render` step and adds a Dublin Core block to every page: title with the site-name suffix stripped, one creator tag per author, publisher, description, canonical identifier, type, format, language, rights, and the page's own git commit date.

It is a post-render script rather than `include-in-header` because that key injects **identical** text into every page, and the title, identifier and date differ per page.

- **Creators are surname-first** (`Seo, JooYoung`) so Zotero splits them into first and last names rather than storing one opaque string. They match `authors` in `pyproject.toml`.
- **`DC.type` is per page**: the home page is `Software` because it stands for the library; the guides and the API reference are `Text`.
- **The site name is read from `og:site_name`** and only an exact trailing `<separator><site name>` is removed, so `Sonification — A Deep Dive – py-maidr` keeps its em dash and loses only the suffix.
- The generated `api/` pages are not in git, so they take the commit date of the `maidr` package they document.
- **Two guards make failure loud rather than silent**, since Quarto swallows this script's stdout: it exits 1 when no page could be tagged at all, and it omits `DC.date` entirely rather than guessing when the checkout holds a single commit.

**`.github/workflows/docs.yml`** checks out with `fetch-depth: 0` on **both** jobs — the render and the publish — so the two paths behave identically. See the correction below.

**`CITATION.cff`** gives GitHub its "Cite this repository" button, with authors matching `pyproject.toml` and the CHI 2024 paper as the preferred citation.

## Deliberately no `citation_*` tags

Google Scholar's [inclusion guidelines](https://scholar.google.com/intl/en/scholar/inclusion.html) reserve those tags for scholarly articles, require that a site consist "primarily of scholarly articles", and say explicitly that `citation_title` must not carry the name of a repository. Putting them on software documentation misrepresents the page and risks the site being dropped from Scholar, without gaining indexing. The two MAIDR papers stay cited by DOI in the JSON-LD and the visible citation section, and reach Scholar through ACM and Eurographics. The reasoning is recorded in the module docstring so it is not quietly undone later, and a test asserts no emitted tag falls outside `DC.`.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation update

## Checklist

## Verification

- `tests/docs/`: **97 passed / 2 skipped on Python 3.9 and 3.10**, **99 passed on 3.11 and 3.12** — each interpreter the matrix runs, not one extrapolated to the rest. The two skips are the `pyproject.toml` consistency cases, which need `tomllib` (3.11+).
- `ruff check` and `ruff format` clean.
- Rendered with Quarto 1.10.18: `index.html` emits `DC.type` `Software` with the ` – py-maidr` suffix stripped and both creators as separate tags; `examples/bar.html` emits `Text` with its own canonical identifier; a second run leaves both byte-identical.

## Three defects this PR introduced and fixed, all caught in review

Recorded because the fixes are the interesting part of the diff.

**1. Per-page dates collapsed on CI.** `actions/checkout@v4` defaults to depth 1, where the single commit has no parent, so `git log -1 -- <path>` reports it for *every* tracked path. Every page would have carried the build date while looking correctly per-page. My original "verified" claim ran against this sandbox's 56-commit clone and never exercised the environment the workflow produces. Reproduced against a real depth-1 clone, then fixed in both the workflow and the script.

**2. A backslash in a title or description crashed the whole render.** `re.sub` parses a *string* replacement for escapes and group references, so `C:\1 drive` raised `invalid group reference` and `regex \d+` raised `bad escape` — and `process` runs in a loop with no per-page handler, so one page aborted the entire site. `html.escape` does not help; it guards a different context. The `api/` descriptions are extracted from docstrings automatically and this codebase already uses backslashes for math notation there, so it was a matter of time. Fixed with a function replacement; five regression cases confirmed to fail before the fix and pass after.

**3. The new tests broke Python 3.9 and 3.10.** They import `tomllib`, which is 3.11+, so those jobs failed at *collection*. Same shape as the first: verified on one interpreter, reported as verified. Now guarded and skipped, with the matrix still covering 3.11 through 3.13.

## Known limits

- The `<title>` / description / canonical / `og:site_name` extraction uses regexes tied to Quarto's current attribute order, not an HTML parser — a consequence of the stdlib-only constraint. The exit-1 guard catches a *total* regression; a partial one affecting a subset of pages would ship quietly. Worth revisiting at a Quarto upgrade.
- `fetch-depth: 0` on the render job means every docs PR clones full history for a build whose output is discarded. That is deliberate: it keeps the render and publish paths identical, so a date bug cannot surface only in production.

https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -45,6 +45,12 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v4
+        with:
+          # Full history: _scripts/dublin_core.py dates each page from the
+          # commit that last touched its source. A shallow clone has one
+          # commit with no parent, so git reports that commit for every path
+          # and every page would carry the build date instead.
+          fetch-depth: 0
 
       - name: Install uv and set the python version
         uses: astral-sh/setup-uv@v5
@@ -96,6 +102,8 @@ jobs:
           # A workflow_run event checks out the default branch by default; name
           # it so the run publishes the release commit it was chained to.
           ref: ${{ github.event_name == 'workflow_run' && 'main' || github.ref }}
+          # Full history, for the same reason as the render job above.
+          fetch-depth: 0
 
       - name: Install uv and set the python version
         uses: astral-sh/setup-uv@v5
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,82 @@
+cff-version: 1.2.0
+title: "py-maidr: Accessible Data Visualization for Python"
+message: >-
+  If you use this software, please cite both the software itself and the
+  MAIDR paper below.
+type: software
+authors:
+  - family-names: Seo
+    given-names: JooYoung
+    email: jseo1005@illinois.edu
+    affiliation: University of Illinois Urbana-Champaign
+  - family-names: Venkatesh
+    given-names: Saairam
+    email: saairam2@illinois.edu
+    affiliation: University of Illinois Urbana-Champaign
+repository-code: "https://github.com/xability/py-maidr"
+url: "https://py.maidr.ai/"
+abstract: >-
+  A Python library that makes matplotlib, seaborn, Plotly and Altair charts
+  accessible to blind and low-vision users through sonification, braille,
+  text and AI description. Distributed on PyPI as `maidr`.
+keywords:
+  - accessibility
+  - python
+  - matplotlib
+  - seaborn
+  - data visualization
+  - sonification
+  - braille
+  - screen reader
+  - blind
+  - low vision
+license: GPL-3.0-or-later
+preferred-citation:
+  type: conference-paper
+  title: >-
+    MAIDR: Making Statistical Visualizations Accessible with Multimodal Data
+    Representation
+  authors:
+    - family-names: Seo
+      given-names: JooYoung
+    - family-names: Xia
+      given-names: Yilin
+    - family-names: Lee
+      given-names: Bongshin
+    - family-names: Mccurry
+      given-names: Sean
+    - family-names: Yam
+      given-names: "Yu Jun"
+  collection-title: >-
+    Proceedings of the CHI Conference on Human Factors in Computing Systems
+    (CHI '24)
+  year: 2024
+  publisher:
+    name: Association for Computing Machinery
+  doi: 10.1145/3613904.3642730
+  isbn: "9798400703300"
+references:
+  - type: conference-paper
+    title: >-
+      Designing Born-Accessible Courses in Data Science and Visualization:
+      Challenges and Opportunities of a Remote Curriculum Taught by Blind
+      Instructors to Blind Students
+    authors:
+      - family-names: Seo
+        given-names: JooYoung
+      - family-names: O'Modhrain
+        given-names: Sile
+      - family-names: Xia
+        given-names: Yilin
+      - family-names: Kamath
+        given-names: Sanchita
+      - family-names: Lee
+        given-names: Bongshin
+      - family-names: Coughlan
+        given-names: "James M."
+    collection-title: "EuroVis 2024 - Education Papers"
+    year: 2024
+    publisher:
+      name: The Eurographics Association
+    doi: 10.2312/eved.20241053
+    isbn: "978-3-03868-257-8"
--- a/docs/_quarto.yml
+++ b/docs/_quarto.yml
@@ -8,6 +8,11 @@ project:
   # Runs before every render, after CI's `quartodoc build`: gives the
   # generated api/*.qmd pages a <title> and a meta description.
   pre-render: _scripts/api_front_matter.py
+  # Runs after every render: adds the Dublin Core block reference managers
+  # read. It is a post-render step rather than `include-in-header` because
+  # that key injects identical text into every page, and the title,
+  # identifier and date differ per page.
+  post-render: _scripts/dublin_core.py
 
 
 metadata-files:
--- a/docs/_scripts/dublin_core.py
+++ b/docs/_scripts/dublin_core.py
@@ -108,15 +108,23 @@ def _tags(
     )
 
 
-def process(page: Path, site_dir: Path, repo: Path, package_date: str) -> bool:
-    """Insert the Dublin Core block into ``page``; True when it changed."""
+def process(page: Path, site_dir: Path, repo: Path, package_date: str) -> str:
+    """Insert the Dublin Core block into ``page``.
+
+    Returns
+    -------
+    str
+        ``"added"`` when the block was written, ``"present"`` when the page
+        already had one, ``"skipped"`` when the page carries no title or no
+        canonical URL to build one from.
+    """
     text = page.read_text(encoding="utf-8")
     if _ALREADY.search(text):
-        return False
+        return "present"
 
     title_match = _TITLE.search(text)
     if title_match is None:
-        return False
+        return "skipped"
     # Quarto renders "<page title> – <site name>"; the suffix is site
     # furniture, and a reference manager should record the page's own title.
     title = html.unescape(title_match.group("title")).strip()
@@ -131,7 +139,7 @@ def process(page: Path, site_dir: Path, repo: Path, package_date: str) -> bool:
     if canonical_match is None:
         # Without a canonical URL there is no stable identifier to record, and
         # `canonical-url: true` in _quarto.yml means every page has one.
-        return False
+        return "skipped"
     identifier = html.unescape(canonical_match.group("href"))
 
     rel = page.relative_to(site_dir).as_posix()
@@ -143,7 +151,7 @@ def process(page: Path, site_dir: Path, repo: Path, package_date: str) -> bool:
         dc_type="Software" if rel == SOFTWARE_PAGE else "Text",
     )
     page.write_text(_HEAD_END.sub(block + "</head>", text, count=1), encoding="utf-8")
-    return True
+    return "added"
 
 
 def main() -> int:
@@ -159,12 +167,27 @@ def main() -> int:
     repo = Path(__file__).resolve().parents[2]
     package_date = _git_date(repo, "maidr")
 
-    changed = sum(
-        process(page, site_dir, repo, package_date)
-        for page in sorted(site_dir.rglob("*.html"))
-    )
+    counts = {"added": 0, "present": 0, "skipped": 0}
+    for page in sorted(site_dir.rglob("*.html")):
+        counts[process(page, site_dir, repo, package_date)] += 1
+
     if not os.environ.get("QUARTO_PROJECT_SCRIPT_QUIET"):
-        print(f"dublin_core: added Dublin Core to {changed} page(s)")
+        print(
+            f"dublin_core: {counts['added']} page(s) tagged, "
+            f"{counts['present']} already tagged, {counts['skipped']} skipped"
+        )
+
+    # Quarto swallows this script's stdout, so a silent no-op would ship a
+    # site with no bibliographic metadata and nothing in the log to show it.
+    # Fail loudly instead: the site always has pages, and they always have a
+    # canonical URL, so ending with none tagged means something broke.
+    if counts["added"] + counts["present"] == 0:
+        print(
+            f"dublin_core: no page under {site_dir} could be tagged "
+            f"({counts['skipped']} skipped)",
+            file=sys.stderr,
+        )
+        return 1
     return 0
 
 
--- a/docs/_scripts/dublin_core.py
+++ b/docs/_scripts/dublin_core.py
@@ -326,6 +326,10 @@ def main() -> int:
         or Path(__file__).parent.parent / "_site"
     )
     if not site_dir.is_dir():
+        # Not the fail-loud case below: that one means the render produced
+        # pages this script could not read. No directory at all means it was
+        # run outside a render -- by hand, or on a clean checkout -- where
+        # there is genuinely nothing to do.
         print(f"dublin_core: {site_dir} not found, nothing to do", file=sys.stderr)
         return 0
 
@@ -348,11 +352,10 @@ def main() -> int:
     for page in sorted(site_dir.rglob("*.html")):
         counts[process(page, site_dir, repo, package_date, dated)] += 1
 
-    if not os.environ.get("QUARTO_PROJECT_SCRIPT_QUIET"):
-        print(
-            f"dublin_core: {counts['added']} page(s) tagged, "
-            f"{counts['present']} already tagged, {counts['skipped']} skipped"
-        )
+    print(
+        f"dublin_core: {counts['added']} page(s) tagged, "
+        f"{counts['present']} already tagged, {counts['skipped']} skipped"
+    )
 
     # Quarto swallows this script's stdout, so a silent no-op would ship a
     # site with no bibliographic metadata and nothing in the log to show it.
--- a/docs/_scripts/dublin_core.py
+++ b/docs/_scripts/dublin_core.py
@@ -201,6 +201,9 @@ def _source_date(
     """
     if not dated:
         return ""
+    # Every page under docs/ is authored as .qmd today. One authored as .md
+    # or .ipynb would miss here and fall back to the package date rather than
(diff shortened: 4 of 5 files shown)


jooyoungseo:
You were right, and this was a real defect rather than a theoretical one. Fixed in `950ed40`.

**Reproduced before fixing.** I cloned this repository at `--depth 1` and ran the same lookups:

```
README.md        2026-09-06
pyproject.toml   2026-09-06
docs/index.qmd   2026-09-06
maidr/api.py     2026-09-06
```

Against full history those same paths give `2026-09-05`, `2026-09-06`, `2026-09-06` and `2026-09-03`. Your explanation of why is exactly right: the single commit has no parent, so git treats every tracked path as added by it.

**My verification claim was wrong, and worth naming.** I checked against this sandbox's clone and reported per-page dates as confirmed. That clone happens to carry 56 commits, so it does distinguish them — it never exercised the environment the workflow actually produces. The claim was true of what I ran and false of what ships, which is the worst shape for a verification claim to have.

**Both of your fix options, not one.** `fetch-depth: 0` on both jobs in `docs.yml` — the render and the publish, since both call the script — and the script now refuses to guess: it checks the commit count once and omits `DC.date` entirely when there is only one, printing why on stderr. The workflow fix alone would leave the same silent failure waiting for whoever edits that checkout step next.

One refinement to your second suggestion. I test the commit count rather than `--is-shallow-repository`, because they are not the same question: this sandbox's clone reports `is-shallow: true` **and** has 56 commits, so it dates pages correctly. Keying on shallowness would have thrown away good dates. Only the single-commit case is actually unusable.

Verified after the change: with history, `stability.html` dates from its own source and `api/show.html` from the package (`2026-08-27` vs `2026-09-03`); run from inside a depth-1 clone, the block is emitted with no `DC.date` and the warning prints. A second run still changes nothing.

**On your other two notes:**

- **Attribute-order-dependent regexes** — fair, and I've left the assumption as it is rather than reaching for an HTML parser, since the script is stdlib-only by design so it runs under whatever interpreter Quarto resolves. What I did add is a guard that makes the silent-miss case loud: if no page ends up tagged, the script exits 1 (`9b10851`). A Quarto version that reordered attributes would now fail the build instead of quietly shipping a site with no metadata.
- **No automated tests** — accepted as a real gap. The script now has a self-check in the shape of that exit-1 guard, but that is not the same as unit tests for `_truncate`-style logic, and I'd rather add them as their own change than grow this PR further.
- **Title-splitting edge case** — you and the maidr reviewer landed on the same concern independently. The sibling maidr PR now strips only an exact `separator + siteName` ending for that reason; this script still partitions from the left, which is the weaker rule. Worth aligning, and I'll note it rather than fold another change in here.

---

jooyoungseo:
Two reviews landed on the shallow-clone bug independently, which is a strong signal it was real. That one is fixed in `950ed40` (details in my reply above: reproduced against a real depth-1 clone, `fetch-depth: 0` on both jobs plus a guard in the script that omits `DC.date` rather than guessing when the checkout holds one commit).

The three remaining points are addressed in `097b4e8`.

**Tests — added, and your reasoning for expecting them was the deciding one.** `api_front_matter.py` two files over has a full suite pinning exactly this kind of string handling, so "it's a build script" was never a good reason for this one to have none. There are now 27 cases in `tests/docs/test_dublin_core.py`, loaded by path the same way its sibling is.

Two of them exist because of this PR's own history: one asserts that pages built from different sources get **different** dates, which is precisely what the shallow-clone bug removed while looking correct, and one asserts that a second run over an already-tagged site still exits 0, so the fail-loud guard cannot be mistaken for broken idempotence.

Writing them found a defect neither review caught: `_tags` emitted an **empty** `DC.description` for a page without one, instead of omitting the field as it already did for an empty date. Fixed.

They also corrected an assumption of mine. I had expected a hyphen to separate the site name; Quarto uses an en dash, and a hyphen must not be treated as a separator at all — `py-maidr` contains one and would have been truncated to `py`. That is now a parametrised case, alongside a title that legitimately contains a hyphen (`Bar - Line Comparison – py-maidr` → `Bar - Line Comparison`), which is the edge case the r-maidr reviewer raised about splitting from the left.

**Docstrings — converted.** You're right that CLAUDE.md asks for NumPy style on all functions and that the sibling script follows it; `_git_date`, `_source_date`, `_tags`, `_has_history` and `main` now have `Parameters`/`Returns` sections.

**Three copies of the metadata — now enforced rather than merely noted.** Your suggestion of a lightweight consistency test was the right shape, since deriving the values at build time would mean parsing `pyproject.toml` from a script that must run under whatever interpreter Quarto resolves. Two cases tie `CREATORS` and `RIGHTS` back to `pyproject.toml`, so a future author or licence change fails a test instead of silently leaving this script stale.

**On the `</head>` insertion — I disagree, and left the code as it is with the reasoning recorded.** The concern is that a page documenting raw HTML could contain a literal `</head>` before the real one. It cannot: such an example lives in the body, which comes after the real `</head>`, so the first match is still the right one. A literal `</head>` *inside* the head would end the head for a browser too, so that page is already broken before this script sees it. There is now a comment saying so, rather than a silent assumption.

Verified: 91 tests pass in `tests/docs/`, `ruff check` and `ruff format` clean, and a fresh render still produces the full block on `index.html`.

---

jooyoungseo:
**I broke the Python 3.9 and 3.10 jobs, and it is fixed in `ae19d67`.**

The consistency tests I added in `097b4e8` import `tomllib`, which is Python 3.11+. This package supports 3.9 and the matrix tests it, so those two jobs failed at **collection** — not the two cases failing, the whole file erroring before any test ran.

The cause is the same shape as the shallow-clone bug earlier in this PR: I verified on one interpreter and reported it as verified. The supported range is stated in `pyproject.toml` and in `CLAUDE.md`, and I did not check a new import against it.

The import is now guarded and the class skipped when `tomllib` is unavailable. There is nothing to fall back to, since reading `pyproject.toml` is the entire point of those two cases, and the matrix still covers 3.11 through 3.13, so the drift they guard against is still caught — without adding a parser dependency for two interpreters.

This time I ran each interpreter the matrix runs rather than inferring:

| Python | result |
|---|---|
| 3.9 | 91 passed, 2 skipped |
| 3.10 | 91 passed, 2 skipped |
| 3.11 | 93 passed |
| 3.12 | 93 passed |

---

On the notes in your latest review:

**Regex coupling to Quarto's attribute order** — agreed this is the main fragility, and your refinement of it is sharper than the earlier version of the same point: the fail-loud guard only catches the case where *every* page stops matching, so a partial regression where only `description` breaks would pass as a normal run. That is true. I've left it, because the alternative is either an HTML parser (the script is stdlib-only by design, so it runs under whatever interpreter Quarto resolves) or a per-field guard that would fire on pages legitimately lacking a description. Worth revisiting at a Quarto upgrade, which is exactly when it would bite.

**Empty `<title></title>`** — correct that it would write a hollow tag, and correct that real Quarto output never produces one. Left as is.

**The `_source_date` early-exit comment** — this one is already resolved, though after the review was written: `aa62a79` replaced the inference entirely. The function now takes an explicit `dated` flag rather than reading "no usable history" out of an empty `package_date`, which was conflating that with "the package directory happens to have no commit date". So there is no longer a branch for a reader to re-derive.

**The fixtures being hand-written rather than real rendered pages** — a fair limit to name. What covers it is the `render-pull-request` job, which runs the script over an actual render of the whole site on every PR touching `docs/`, and the exit-1 guard turns a total regression there into a red build rather than a quiet one.

---

jooyoungseo:
**This one was a genuine landmine and it's fixed in `8cca2d6`.** Best finding of the whole series — thank you.

I reproduced all three shapes you named before touching anything:

```
"C:\1 drive"         → error: invalid group reference 1
"regex \d+ example"  → error: bad escape \d
"C:\Users\docs"      → error: bad escape \U
```

Your analysis was exactly right on every point, including the two that make it dangerous rather than merely wrong: `html.escape` guards the HTML context and does nothing for the replacement-template context, and `process` runs in a loop with no per-page handler, so one bad page aborts the render for the whole site rather than being counted as skipped.

The `api/` path you flagged is what makes this a matter of time rather than a hypothetical. Those descriptions are extracted from docstring first paragraphs automatically, and this codebase already uses backslashes for math notation in exactly that position — `stepplot.py` and `violin_box_plot.py`, as you found. A future public API docstring doing the same would have taken the docs build down with an error naming a regex escape, which is about as far from the actual cause as an error message can get.

Fixed by passing a function replacement, which `re` uses verbatim.

**On the test gap** — you were right that the existing escaping case covers a different problem entirely (HTML-attribute breakout, not replacement-template parsing). There are now five regression cases: three title shapes, one description, and one at the `_tags` layer so the hazard is pinned on both sides of the call. I confirmed they fail against the old string replacement and pass against the new one, rather than assuming a test that passes is a test that would have caught it:

```
before: 4 failed, 1 passed
after:  5 passed
```

**On `main()` having no per-page `try`/`except`** — you read the intent correctly, it is deliberate, and I've left it that way. The script's whole posture is that a silent partial result is worse than a loud stop, which is also why it exits 1 when nothing gets tagged. Catching per page would turn a systematic breakage into a handful of quietly untagged pages. The comment at the fix now records why the substitution in particular must not be able to raise.

**For the record on the sibling repo:** I checked, and r-maidr's script is unaffected. It splices with `str.find` and slicing rather than `re.sub`, so no replacement template is ever parsed. Its plainer approach turned out to be the more robust one here.

Verified on Python 3.9 (96 passed, 2 skipped) and 3.11 (98 passed), and a fresh render still emits the full block.

---

jooyoungseo:
All three addressed in `c577d68`, including the one you flagged as worth fixing before merge.

**The live-history dependency — fixed, and you were right that it was the real problem.** Those two tests asserted that `docs/stability.qmd` and `maidr/` currently have different last-touched dates. True today, false the day one commit touches both, and at that point they fail while `_source_date` is still correct. A test that depends on incidental history tells you about the history, not the code.

A fixture now builds a repository with two files committed at fixed dates, exactly the pattern `TestHasHistory` was already using. The assertions got stronger as a side effect: they compare against exact values (`{"2026-01-02", "2026-03-04"}`) rather than "these two differ".

I also checked what still touches the real repository and confirmed none of it claims anything about history — one case asserts only that it has more than one commit, the rest pass an explicit package date, and the consistency cases read `pyproject.toml` as a file. And I added the path that was previously implicit: a page whose source is untracked falls back to the package date rather than losing its date entirely.

**`QUARTO_PROJECT_SCRIPT_QUIET` — removed.** Nothing sets it, no test covers it, and Quarto swallows this script's stdout anyway, so it was a knob with no hand on it. Worth naming where it came from: the sibling `api_front_matter.py` has the same unused check and I copied the pattern from there. That one is already merged, so I've left it rather than widen this PR into it — but it is the same dead branch if anyone wants it gone.

**Missing output directory — your read is right, and it now says so in the code.** No directory means the script ran outside a render, by hand or on a clean checkout, where there genuinely is nothing to do. The fail-loud case is narrower: a render *did* produce pages and none could be tagged. A comment at the early return distinguishes them.

---

**On the two you flagged for acknowledgement rather than change:**

**`fetch-depth: 0` on the render job.** Deliberate, and worth stating the reasoning. That job is what proves the script works on a real render before anything publishes; running it against a shallow checkout would exercise the `dated=False` path while the publish job takes a different one, so a date bug could only ever surface in production. Paying the clone cost keeps the two paths identical. Your point about revisiting as history grows is fair — the lever if it ever hurts is a bounded depth rather than dropping to depth 1.

**Where the fail-loud guarantee stops.** Correct, and it is the sharpest statement of it so far: the guard fires only when *zero* pages match, so a partial regression affecting a subset would ship quietly. Accepted as the cost of the stdlib-only constraint. Extending the guard to per-field coverage would fire on pages legitimately lacking a description, so it needs more thought than this PR should carry.

Verified on Python 3.9 (97 passed, 2 skipped) and 3.11 (99 passed).

---

jooyoungseo:
Both points addressed in `0a5fb30`, and your read of the backslash fix was correct: it was narrow.

**The per-page handler — implemented, but not quite as suggested.** You were right that patching the one call site known to raise leaves every other exception able to abort the whole step; `read_text` on a page that is not valid UTF-8 would have done exactly the same thing.

Where I differ is on counting a failing page as skipped. That would make the failure silent, which is the outcome this script's whole posture is built against — the same reason it exits 1 when nothing gets tagged. So each page is attempted inside a handler, but a page that raises is named on stderr and the run still exits 1.

The gain over the old behaviour is diagnostic rather than tolerant: the build stops either way, but now it stops having processed every other page and reported *all* the causes, instead of dying on the first one. That is the difference between one run that tells you everything and one rebuild per broken page.

Two regression cases, confirmed to fail before the change and pass after: an unreadable page alongside a good one leaves the good one tagged and names the bad one, and two unreadable pages are both named rather than just the first.

**On `test_this_repository_has_history`** — I checked, and `ci.yml`'s test job does set `fetch-depth: 0` (line 38), so it holds in CI today. But your framing is the right one: the checkout's depth is a CI setting, not a property of the code, and a test should not quietly depend on it. It now uses the same built fixture as its single-commit counterpart, so no test in the suite asserts anything about the checkout it runs in.

**On the three-way duplication** — agreed that `PUBLISHER` and the `CITATION.cff` author list are the uncovered half. Left as is: `PUBLISHER` has no machine-readable source to compare against (it is prose in the page footer), and pinning `CITATION.cff` would mean a YAML parser in the test path for a file that changes about as often as the licence. The consistency tests cover the two that have an authoritative source.

Verified on Python 3.9 (99 passed, 2 skipped) and 3.11 (101 passed), plus a fresh render.

---

jooyoungseo:
Checked the `--follow` point rather than taking it, and it does not hold. Adding `--follow` changes nothing here.

I built a repository where a file is created, left alone, then renamed, so content and path last changed on different dates:

```
last content change: 2026-01-01
rename:              2026-03-01

git log -1            -- new.qmd  ->  2026-03-01
git log -1 --follow   -- new.qmd  ->  2026-03-01
```

`--follow` lets git keep walking *past* a rename into the file's earlier history. `-1` stops at the newest commit touching the path, and the rename commit *is* that commit, so the flag never gets the chance to matter. It would matter for a "when was this created" lookup that walks to the end of history — which is exactly where the sibling maidr repo does use it, in `firstCommitDate`.

Whether the rename date is the *right* answer is a separate and more arguable question. I think it is, for this field: renaming a `.qmd` changes the page's URL, so from a reference manager's point of view the resource at that identifier did change that day. But either way, `--follow` is not the lever.

**On `QUARTO_PROJECT_OUTPUT_DIR` versus `QUARTO_PROJECT_DIR`** — right that they are different variables for different purposes, and it is confirmed empirically rather than only from the docs: the sibling r-maidr build tagged 386 pages in CI, and this repo's `render-pull-request` job runs the script over a real render and passes. If the variable were wrong the script would fall back to `docs/_site`, which is the same directory here, so the CI evidence is admittedly weak on its own — but the fail-loud guard means a wrong directory with no pages in it would exit 1 rather than pass quietly.

**On the partial-regex-break gap** — accepted, and it is already written down: the PR description has a "Known limits" section saying the exit-1 guard catches a total regression but a partial one would ship quietly, and naming the Quarto upgrade as when to revisit. Your suggestion of asserting the regexes against one real rendered page in the docs CI job is the better fix and I have not done it; it needs a fixture that is a genuine Quarto output rather than a hand-written string, which is worth its own change.

**On `except OSError`** — agreed on both counts: `git` missing from `PATH` is the realistic failure, and both call sites already treat any failure as "cannot date this", so narrowing it would not change behaviour.

---

PR #784, opened 2026-09-15
feat: carry a downloaded DotPad SDK with offline documents

### LINKED ISSUE #771: fix(offline): use_cdn=False cannot reach a DotPad without the network
## Summary

`use_cdn=False` promises a document that fetches nothing. Since the bundle moved to maidr.js 4.7.0 (a4c1096) that promise no longer holds for one path: connecting a DotPad tactile display fetches the vendor SDK from jsDelivr at runtime.

## Evidence

The bundled `maidr/static/maidr.js` carries two absolute URLs it did not carry at 4.6.0 (`grep -c 'cdn.jsdelivr.net'`: 0 at 4.6.0, 2 at 4.7.0):

```
https://cdn.jsdelivr.net/gh/dotincorp/dotpad-sdk-guide@437210b1.../Web/3.0.2
https://cdn.jsdelivr.net/gh/xability/dotpad-sdk-guide@0e8577fb.../Web/3.0.2/lib/
```

They come from `src/service/dotPadSession.ts` upstream (`VENDOR_BASE_URL`, `VENDOR_ASSET_BASE_URL`) and are fetched on the first connect attempt, not on load — so an offline document still renders, sonifies and brailles as promised, and only the DotPad path reaches the network.

## Why this is not an upstream bug

Upstream does not bundle the SDK deliberately: it ships without a licence permitting redistribution, so maidr.js references the vendor's own published copy, pinned by commit. The escape hatch is already there — a host page sets these before maidr.js loads and the CDN is never touched:

- `window.MAIDR_DOTPAD_SDK_URL`
- `window.MAIDR_DOTPAD_ASSET_BASE_URL`

py-maidr has no way to set either today, so an air-gapped install (or one behind a CSP that admits no third-party origin) has no way to use a DotPad even with the SDK sitting on its own server.

## Proposal

1. Let callers point at a locally served SDK — an option and/or `MAIDR_DOTPAD_SDK_URL` / `MAIDR_DOTPAD_ASSET_BASE_URL` environment variables — and emit the two globals ahead of the inlined bundle on every offline path: `init_notebook(use_cdn=False)`, `render()` / `save_html()`, the Shiny renderer and the Streamlit embed.
2. Document the limitation on the offline page: without that configuration, DotPad output in `use_cdn=False` needs network access to jsDelivr on first connect.

## Related test change

The CDN assertions in `tests/core/test_offline_bundle.py`, `tests/widget/test_shiny.py` and `tests/widget/test_streamlit.py` grep the emitted HTML for the bare host, so the 4.7.0 bundle broke them the moment it was inlined (3 failures, blocking [the release run](https://github.com/xability/py-maidr/actions/runs/34142625442)). They are being narrowed to the maidr loader URL (`cdn.jsdelivr.net/npm/maidr`) plus an attribute-level guard that an offline document requests no external asset — which is the part py-maidr controls. This issue is the rest of the promise.

jooyoungseo:
## Description

Closes #771.

maidr.js drives a DotPad tactile display through the vendor's SDK, which it imports from jsDelivr on first connect rather than bundling its 14 MB braille engine. That was the one path a `use_cdn=False` document still took to the network, and py-maidr had no way to point it elsewhere (the R package already had the URL options). Dot Inc. have confirmed MAIDR may redistribute the SDK, so py-maidr can now both name a copy and carry one:

- **`maidr.download_dotpad_sdk()`** fetches the commit maidr.js is pinned to (`dotincorp/dotpad-sdk-guide@781f230`, the first with an intact `liblouis.data`), verifying every file against its recorded size and SHA-256, including the LGPL-2.1 notice and liblouis wrapper sources the vendor's README asks redistributors to keep beside the engine. It writes a `manifest.json` naming the commit, is idempotent (`force=True` refetches), writes each file beside and renames over so a half-written file is never seen, and lands in a per-user cache unless `MAIDR_DOTPAD_SDK_DIR` says otherwise.
- **`save_html(..., use_cdn=False)`** then copies that directory into `lib/dotpad-sdk-3.0.2/` next to the document (following `lib_dir` and `include_version`) and declares the two `window.MAIDR_DOTPAD_*` globals with that relative path, for matplotlib and Plotly output alike. `"auto"` and `True` are untouched, and a session that never downloaded the SDK gets exactly what it did before.
- **`maidr.set_dotpad_sdk(url, asset_base_url)`** and the environment variables `MAIDR_DOTPAD_SDK_URL` / `MAIDR_DOTPAD_ASSET_BASE_URL` name a served copy instead. Every render path then declares it: as a head dependency listed ahead of the bundle for documents, as a plain tag inside iframes (where htmltools drops dependencies). Either setting, alone or together, wins over a downloaded copy, so a document never mixes a local module with a remote engine. This is what notebooks, Shiny and Flask need, since a `srcdoc` iframe cannot resolve a relative path.
- `maidr.get_dotpad_sdk()`, `maidr.dotpad_sdk_path()` and `maidr.dotpad_sdk_dir()` report the state. The pins live in `maidr/util/dotpad.py` and mirror `src/service/dotPadSdk.json` in the maidr repo.

**Also the test change #771 called for.** Three checks grep an offline document for the bare host `cdn.jsdelivr`, which the inlined maidr.js has contained since 4.7.0 because of the DotPad SDK URL: the Shiny and Streamlit `use_cdn=False` cases and `init_notebook(use_cdn=False)`. They have been failing on `main` (see the CI runs of the last few days), and the notebook one takes pytest close to two hours to report because it renders a failure diff over the 1.7 MB bundle, which is why the test matrix looked hung. They now match `cdn.jsdelivr.net/npm/maidr`, the loader URL, as the other offline checks already do; the three files run in two seconds.

Docs: a new subsection under "Offline Use and the JavaScript Bundle", three rows in the environment-variable table, and a README paragraph.

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Documentation update

## Checklist

## Tests

`tests/util/test_dotpad_sdk.py` adds 31 tests with no network access: the pins' shape and the intact `liblouis.data` size, URL configuration (setter over environment, empty means unset, a hostile URL cannot break out of the `<script>`), the declaration preceding the bundle in saved matplotlib and Plotly documents and surviving inside an iframe, the download against a fake manifest and fetch (verified, idempotent, `force`, a short file refused by its size, a same-size corruption refused by its digest, nothing written on failure), completeness detection, and the `save_html` copy (lands in `lib/`, follows `lib_dir`/`include_version`, only for `use_cdn=False`, either URL setting wins, undownloaded session unchanged). `ruff check` and `ruff format` are clean on the touched files.


jooyoungseo:
Thanks for the end-to-end run. On the two notes:

- **Formatting churn in `plotly_maidr.py`**: agreed it is noise in the diff; `ruff format` on the touched file brought the pre-existing unformatted comprehensions along. Left as is rather than spend another CI cycle on a revert, since the result is what `ruff format` wants and the DotPad change is confined to the two calls at the end.
- **Two `HTMLDependency` constructions in `dotpad_sdk_dependency()`**: left as is. The first exists only to ask htmltools for the `href` it will use, so the globals cannot drift from where the files land; setting `.head` on it afterwards would work, but the constructor is the documented surface and does no I/O.

---

jooyoungseo:
Right on the ordering; thanks for reproducing it against the real head. Fixed in `d13807b`: `save_html()` now hands the downloaded-copy dependency to `_create_html_doc()` as a leading child on both the matplotlib and Plotly paths, so its head renders above the bundle's `<script>` exactly as the URL path does, and the two local-copy tests now assert the order through the same `_declaration_precedes_the_bundle` helper. (`4a94f54` in between was a wrong turn through an attribute htmltools does not expose; it is superseded.) The two minor notes are the same as last round and stay as answered above.

---

jooyoungseo:
Thanks for the offline-Chromium run with the network blocked; that is the claim the PR makes, verified. Notes 1 and 2 are as answered above. Note 3 is a fair gotcha: `download_dotpad_sdk(directory=...)` to a custom path does need `MAIDR_DOTPAD_SDK_DIR` to point there for `save_html` to find it. The docstring's `directory` parameter and the docs both name `MAIDR_DOTPAD_SDK_DIR` as the way to move the cache, so I am leaving it at that rather than spend another CI cycle on a sentence; happy to fold a one-liner into the next change that touches the module.

---

PR #786, opened 2026-09-16
feat(dotpad): read the SDK pins from the shipped manifest and move to 3.0.3

jooyoungseo:
## Description

maidr.js now publishes the DotPad SDK it is pinned to as `dist/dotpad-sdk.json` in its npm package: version, repository, commit, base URL, module, asset directory, and the size and digests of every file. Since #784 py-maidr carried its own copy of those values as constants in `maidr/util/dotpad.py`, with a comment asking to keep the two in step by hand.

This PR ships that manifest as `maidr/static/dotpad-sdk.json`, beside `maidr.js` and `VERSION`, and makes `maidr/util/dotpad.py` read every pin from it at import. The public names and their types are unchanged (`DOTPAD_SDK_VERSION`, `DOTPAD_SDK_REPOSITORY`, `DOTPAD_SDK_COMMIT`, `DOTPAD_SDK_BASE_URL`, `DOTPAD_SDK_MODULE`, `DOTPAD_SDK_ASSET_DIR`, `DOTPAD_SDK_FILES` as `dict[str, DotPadSdkFile]`); `DOTPAD_SDK_PINS_FILENAME` is the one addition.

The pin moves from SDK **3.0.2 to 3.0.3**. Dot Inc. publish that release only as a zip archive, which a browser cannot import a module from, so the files are served from the mirror `xability/dotpad-sdk-guide` at a pinned commit, extracted from the vendor's archive and byte-verified against it. The manifest's `upstream` entry records the vendor repository, commit, archive path and archive SHA-256. The liblouis engine files are byte-identical to the 3.0.2 pin, so the "intact `liblouis.data`" guarantee still holds.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update

## Changes Made

- `maidr/static/dotpad-sdk.json`: the manifest, as maidr.js publishes it.
- `maidr/util/dotpad.py`: loads the pins from the manifest via `importlib.resources`, the same way `dependencies.py` locates `maidr.js`; the hard-coded values are gone and version-specific docstrings read `<version>`.
- `.github/scripts/fetch-maidr-bundle.sh`: copies `package/dist/dotpad-sdk.json` into `DEST_DIR` when the npm tarball ships one, checks it has the fields the module reads, and logs either outcome. A maidr release without it leaves the existing manifest in place, so the script still works for older versions.
- `tests/util/test_dotpad_sdk.py`: the pin test derives the expected jsDelivr URL from the manifest's `repository` field instead of asserting `dotincorp`; `lib/dotpad-sdk-3.0.2/` and `DotPadSDK-3.0.2.js` literals are now derived from `DOTPAD_SDK_VERSION` / `DOTPAD_SDK_MODULE`; a new test reads the JSON independently and checks every constant against it.
- `docs/index.qmd`: version-generic folder name, 3.0.3 in the concrete example, and a paragraph saying the pins come from `maidr/static/dotpad-sdk.json`, refreshed with the bundle.
- `pyproject.toml`: comment only. `packages = ["maidr"]` and the sdist's `maidr/static/**` already ship the file; verified with `uv build --wheel` (`maidr/static/dotpad-sdk.json` is in the wheel).

N/A

## Checklist

## Additional Notes

Verification run locally:

- `uvx ruff@0.3.4 check --diff` and `check --select FA102` (the two CI lint steps): pass.
- `uvx ruff@0.3.4 format --check` on the touched Python files: pass. (A repo-wide `ruff format --check` reports 144 pre-existing files; CI does not run it.)
- `uv run pytest tests/util/test_dotpad_sdk.py -vvv`: 32 passed.
- `uv run pytest -q` after `uv sync --locked --all-extras --dev`: 4482 passed, 69 skipped, 2 xfailed.
- `uv lock --check`: pass.
- `bash -n .github/scripts/fetch-maidr-bundle.sh`, plus the extraction tail run offline against two fake npm tarballs (one with `dist/dotpad-sdk.json`, one without): the manifest is copied in the first case and left as-is with a log line in the second. shellcheck is not installed here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jbm23mXb8ovppeXmTLFRBc)_


jooyoungseo:
Fixed in `219c751`. The finding was right, and the reasoning about how the manifest reaches `main` is what made it worth fixing rather than documenting: the refresh workflow commits a manifest fetched from an upstream release straight to `main`, CI does not run on that push, and this module is imported by `maidr/__init__.py`, so a renamed field upstream became `import maidr` raising for every user over a device most of them do not own.

What changed:

- `_read_pins()` catches `OSError`, `ModuleNotFoundError`, `TypeError` and `ValueError` (which covers `JSONDecodeError`), warns once through `warn_once` naming the file and what went wrong, and returns inert pins: version `0.0.0`, empty `files`. That follows the bundled-`VERSION` precedent you cite rather than raising.
- A new `_parse_pins()` validates every published field where it is read, including each file entry's `bytes` and `sha256`, so drift is reported as a broken manifest instead of a `KeyError` from further down. This also answers the redundant-`int()`/`str()` nit: the conversions are gone, replaced by checks that say what is wrong.
- The two paths that need a real pin now check for one. `download_dotpad_sdk()` raises with an actionable message, and `dotpad_sdk_path()` returns `None` when there are no pins. That second one was a latent bug your review led me to: the completeness check is a loop over the pinned files, and a loop over nothing passes, so an empty manifest would have made every directory look like a complete SDK copy and bundled it into an offline document.
- The `jq` gate in `fetch-maidr-bundle.sh` now validates each entry's shape, not just the top-level keys. I checked it against your scenario 2 manifest and three more (non-numeric `bytes`, empty `files`, empty `version`); all four are rejected and the real manifest passes.

Both of your reproductions now warn and keep the import working:

```
WARNING:maidr.util.warn:Bundled DotPad SDK manifest 'dotpad-sdk.json' will not read
([Errno 2] No such file or directory: ...). The tactile display cannot be set up
offline until it does; reinstall py-maidr or run the update-maidr-js workflow.
import ok, version = 0.0.0
```

Ten tests cover it: the rejected shapes, the warning and inert pins, no directory counting as a copy without pins, and the download naming the broken manifest. `tests/util/test_dotpad_sdk.py` is 42 passed, the full suite 4492 passed / 69 skipped / 2 xfailed, and `ruff check` and `ruff format --check` are clean on both changed files.

---

jooyoungseo:
Closed the digest-shape gap in `d84b90a`. `_parse_pins()` now requires `sha256` to be 64 hex digits, and the `jq` gate requires the same of `sha256` and 32 of `md5` when present.

Writing that second clause found a real bug in the gate I added a commit earlier. In jq, inside a `.files | ...` pipe a second `.files` resolves against the files object rather than the document, so

```jq
.files | all(has("md5") | not) or (.files | all(.md5 | test(...)))
```

iterated null and rejected every valid manifest, the real one included. It is a per-entry conditional now. I ran the gate against the shipped manifest and six malformed ones: the real manifest passes, and a bad digest, a bad md5, a missing per-file digest, a string `bytes`, an empty `files` and an empty `version` are all rejected. A manifest with no `md5` at all still passes, since nothing in Python reads it.

On your first observation, the unconditional warning is deliberate. `MAIDR_BUNDLE_STALE_WARNING` silences the staleness notice, which is about a bundle that works but is behind; a manifest that will not read is a broken install, and that is the class of thing the other unconditional `warn_once` sites in `cdn.py` report.

`tests/util/test_dotpad_sdk.py` is 43 passed, the full suite 4493 passed / 69 skipped / 2 xfailed, `ruff check` and `ruff format --check` clean, `bash -n` clean.

---

jooyoungseo:
Thanks for re-verifying the failure mode by hand rather than taking the description's word for it, and for tracing the two callers of the fetch script.

On the `_mismatch` docstring: you are right that the 7,685-byte figure is the 3.0.2-era corruption, and it is deliberately historical, so I have left it. It is also outside this diff, from `3f69287`. Rewording it here would put an unrelated line in a pin-plumbing change, so if it reads as describing the current pin to someone later, it is better fixed on its own.

The shell branch staying at the `jq`-filter level is the trade I made too. Both callers run under `set -euo pipefail`, as you found, so the only way a bad manifest survives is an invocation outside those workflows, and that path has no commit at the end of it.

Merging: 26 checks green, three reviews handled.

---

PR #787, opened 2026-09-16
fix(dotpad): refuse a manifest path that leaves its directory

jooyoungseo:
## Summary

Follow-up to #786. A review of the matching change in the R binding (xability/r-maidr#308) pointed out that the DotPad manifest's file paths were the one thing nothing checked, and they are the field that decides where bytes land on disk.

`download_dotpad_sdk()` joins each path onto the download directory:

```python
target = target_dir.joinpath(*relative.split("/"))
```

A manifest naming `../../../.ssh/authorized_keys` passed every check the parser had. The key is a non-empty string, the digests are well formed, so the download would have fetched the bytes and written them there.

## Severity

Nothing user-authored reaches this today. The manifest ships inside the wheel, and the only thing that replaces it is `fetch-maidr-bundle.sh`, which copies it out of an npm package whose checksum was already verified. So this is defence in depth rather than a live vulnerability.

What made it worth taking rather than noting is the asymmetry: #786 validates a digest to the character and a version to a non-empty string, while the field that picks a filesystem destination went unvalidated. The download is also the one place this package writes files it did not name itself.

## Changes

- `_path_is_safe()` in `maidr/util/dotpad.py`, applied to every key in `_parse_pins()`. A path is refused when it is absolute, names a home directory or a drive, carries a backslash, or has an empty, `.` or `..` segment.
- The same check in the `jq` gate in `.github/scripts/fetch-maidr-bundle.sh`, so a manifest like that fails the bundle refresh rather than landing on `main`.
- Tests over eight escaping paths and over the eight the shipped manifest really names.

## Verification

```
uv run pytest tests/util/test_dotpad_sdk.py -q   # 52 passed
uv run pytest -q                                 # 4502 passed, 69 skipped, 2 xfailed
uv run ruff check / ruff format --check          # clean on both changed files
bash -n .github/scripts/fetch-maidr-bundle.sh    # clean
```

The `jq` gate was also run directly against the shipped manifest, which passes, and against five malformed ones (traversal, absolute, drive letter, backslash, empty segment), all rejected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Jbm23mXb8ovppeXmTLFRBc)_


jooyoungseo:
Thanks for probing `_path_is_safe()` with cases beyond the checked-in ones, and for tracing both write sites back to `_parse_pins()` rather than taking the description's word for it. Those edge cases all landing on reject is the answer I wanted; a validator of this kind should fail toward refusing.

On the cross-reference comment: I agree it is the right note to leave, and I am leaving it rather than pushing for it, since the finding is explicitly not blocking and a comment-only push would restart the whole matrix. Recording it here so it is not lost. The `jq` gate in `.github/scripts/fetch-maidr-bundle.sh` and `_path_is_safe()` in `maidr/util/dotpad.py` encode the same rule in two languages on purpose, because the bundle-refresh push does not run CI and the gate therefore cannot import the Python. Whoever changes one should change the other. The matching pair in the R binding is `maidr_dotpad_path_is_safe()` and that repository's own copy of the gate, so it is three places in two repositories.

Merging: 24 checks green.

---

Issue #776, opened 2026-09-08
fix(): style consistency: spelling variant, capitalization and terminology


omark807:
# Bug Report
 
## Description
 
Inconsistent capitalization and word forms: "Braille" and "braille" alternate, sometimes within a single paragraph. Also barplot/bar plot, boxplot/box plot, lineplot, scatterplot.

## Steps to Reproduce
 
1. Open <https://py.maidr.ai/> and search for "Recognised", "centre", "serialised".
2. Search the documentation for the terms above (and their variants). 

## Actual Behavior
 
Competing spelling and terminology conventions in one document.
 
## Expected Behavior
 
Consistency in spelling and terminology conventions throughout the entire page. 
 
## Environment
 
- Operating System: N/A (documentation content)
- Browser: Any

Issue #774, opened 2026-09-08
fix(): AI model list out of date


omark807:
# Bug Report
 
## Description
 
The AI feature instructions list three specific models (OpenAI GPT-4o, Google Gemini 2.0 Flash Experimental, Anthropic Claude 3.5 Sonnet) and follow the list with "**Note**: All provided models are the latest versions." None of these are current models. The list and the note are duplicated verbatim in both the "For Alpha Testers" and "For All Other Users" subsections, so the page has two copies of the same stale claim to keep in sync.
 
## Steps to Reproduce
 
1. Open <https://py.maidr.ai/>.
2. Go to "Instruction on How to Verify and Use the AI Feature in py-maidr" → "For Alpha Testers", step 8.
3. Read the model list and the note directly beneath it. Repeat for "For All Other Users", step 3.

## Actual Behavior
 
The page names three superseded models and claims they are the latest versions.
 
## Expected Behavior
 
Either drop the note and link to a maintained model list (in the maidr.js repo or the app UI), or remove specific model names from the docs entirely and describe the categories of provider supported. Whatever is chosen, the list should exist in exactly one place on the page rather than two.
 
## Environment
 
- Operating System: N/A (documentation content)
- Browser: Any

## Additional Information
 
Either routinely update the "latest models" or don't make a claim on this.

Issue #772, opened 2026-09-08
fix(): Supported Data Visualization Libraries do not include Plotly


omark807:
# Bug Report

## Description

On py.maidr.ai, Plotly is not listed under "Supported Data Visualization Libraries". 

## Steps to Reproduce

1. Go to py.maidr.ai. 
2. Navigate to the section titled [Supported Data Visualization Libaries](https://py.maidr.ai/#supported-data-visualization-libraries).

## Actual Behavior
Plotly and its supported functions are absent. 

## Expected Behavior

Plotly should be present if it is supported. 

## Environment
- Operating System: macOS
- Browser: Chrome

jooyoungseo:
Fixed by #788 (merged as d2b2030) and closed manually — GitHub's auto-close didn't fire here even though it records #788 as a closing PR, unlike the sibling issues #773–776 which closed on merge.

`docs/index.qmd` now has a `### Plotly` section between Seaborn and Altair, written from the dispatch in `maidr/plotly/plotly_plot_factory.py` rather than from the gallery, so it lists what the adapter actually handles: bar/dodged/stacked/normalized, count, histogram, line and multi-line, step, area and stacked area, scatter, regression, box, violin, heatmap, 2-D histogram and its contour form, contour, pie, candlestick/OHLC, funnel and funnel-area, waterfall, gauge, radar and polar area, treemap/sunburst/icicle, Sankey, parallel coordinates and parallel categories, SPLOM, choropleth and geographic scatter/density maps, and subplots/facets.

It carries a callout noting most of these are experimental, linking to [Stability](https://py.maidr.ai/stability.html) and the Plotly gallery, so that promise travels with the list.

Worth recording what review caught: three bullets in the first draft named a knob *adjacent* to the real one — `barmode="relative"` described as the percentage variant when `layout.barnorm` decides that; a KDE layer the Plotly adapter has no trace type for; and `fill="tozeroy"` as the area signal when `is_area_trace()` reads `stackgroup` alone. All three are corrected in the merged version.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #789, opened 2026-09-16
docs: make the shortcut table canonical for help, chat and settings

### LINKED ISSUE #773: fix(): three conflicting shortcuts documented for help and AI chat
# Bug Report
 
## Description
 
The page gives three mutually inconsistent accounts of the help/chat shortcuts:
 
- Table 1: "Open Chat View | Question (?)" and "Open keyboard help | Ctrl + Slash (/) | CMD + Slash (/)".
- AI feature section, step 2: "press **H** to open the Help menu".
- AI feature section, final step: "press **Ctrl+Shift+/** (forward slash) or **Alt+Shift+/** (on Mac) to open the AI modal".

The Mac parenthetical is also ambiguous; it's unclear whether it modifies only the second binding, and `Alt` is the only Mac binding on the page that is not `CMD`.
 
## Steps to Reproduce
 
1. Open <https://py.maidr.ai/#keyboard-shortcuts-and-controls> and read Table 1.
2. Scroll to the AI feature section and read steps 2 and 12.
3. Attempt to determine the correct key to open AI chat on macOS.
## Actual Behavior
 
Three different key combinations are documented for what appear to be overlapping functions, with no explanation of the difference between the Help menu, the keyboard help dialog and the Chat View.
 
## Expected Behavior
 
One canonical shortcut table that includes the Help menu, keyboard help and AI chat, with correct Windows/Linux and Mac columns. The AI feature instructions reference that table rather than restating bindings.
 
## Environment
 
- Operating System: macOS, Windows, Linux
- Browser: Any

jooyoungseo:
# Pull Request

## Description

The page gave three mutually inconsistent accounts of the help/chat keys — the shortcut table, a "press **H** to open the Help menu" step, and a "**Ctrl+Shift+/** or **Alt+Shift+/** (on Mac)" step — with no explanation of how the Help menu, the keyboard help dialog and the Chat View differ. Only the table was close to correct.

I checked every affected row against the keymaps in the maidr core (`src/service/keybinding.ts`) and corrected the table to match the code. Review then turned up the root cause — the table had been **copied**, not referenced — so this also removes the duplicate and adds a test covering both the rows and the no-second-copy rule.

## Related Issues

Closes #773

### 1. The table now matches the engine

| Row | Was | Is | Source |
|---|---|---|---|
| Keyboard help | `Open keyboard help` | `Open the keyboard help`, Ctrl + / / CMD + / | `TOGGLE_HELP: key(HELP_CHORD, …)`, `HELP_CHORD = ${Platform.ctrl}+/` |
| AI chat | `Open Chat View \| Question (?)` | `Open the AI chat \| Shift + Slash (?)` | `TOGGLE_CHAT: key('shift+/', …)` — names the chord, not only the glyph |
| Settings | `Open settings` | `Open settings, including AI providers and API keys` | `TOGGLE_SETTINGS`; the AI tab in `Settings.tsx` is what a reader following the AI instructions is looking for |
| Stop Auto-play | Mac column said `Ctrl` | `CMD` | `STOP_AUTOPLAY: key(${Platform.ctrl}, …)`, and `Platform.ctrl` is `command` on Mac |
| First/last element | `Ctrl + Home` / `Ctrl + End` | replaced by `[` / `]` | Home and End are bound in **no** keymap — they appear only in braille mode's `ALLOW_DEFAULT` text-selection allow-list. `GO_TO_MIN_VALUE`/`GO_TO_MAX_VALUE` are the real jumps |
| *(new row)* | — | `Close the settings, chat or help dialog \| ESC \| ESC` | see below |

Plus a short note after the table naming the three surfaces — keyboard help is the only help dialog, nothing is bound to `H`, settings holds the API keys — and stating that `@tbl-shortcuts` is canonical so later sections cite it rather than restating bindings.

**A correction inside this PR, worth calling out.** An earlier revision reworded those three rows to "Open **or close**" and added a note saying the chat "is a toggle: the same key returns you to the plot". That is false, and it was my error, not inherited drift — I carried it from the old AI section ("The AI Chat modal is togglable…") without checking. Review caught it by pressing the keys. Each dialog runs in its own scope that does not carry the opening chord: `CHAT_KEYMAP` and `HELP_KEYMAP` rebind the same command id to `esc`, and `SETTINGS_KEYMAP` is empty because MUI's `onClose` takes Escape before hotkeys-js sees it — its comment notes the scope exists precisely to keep chart shortcuts from firing while the dialog is open, which is why a second press is inert. Hence the plain "Open …" rows, the new Esc row, and a note explaining what actually closes a dialog. The same sentence was in #790 and is fixed there too (365bd42).

### 2. There is now only one table

`docs/examples.qmd` carried a **byte-identical copy**, under the **same `{#tbl-shortcuts}` id**, and kept all three defects after the Welcome page was fixed — so a reader landing on `/examples` was still told to press `Ctrl + Home`. Two pages defining one crossref id also made `@tbl-shortcuts` ambiguous.

The copy is replaced with a link to the canonical section rather than fixed in place: a second *correct* copy is still a second copy to keep correct, and this is what that ends as. The link is plain Markdown rather than `@tbl-shortcuts`, because `_quarto.yml` is `project: type: website` and Quarto resolves cross-document crossrefs only for `book` projects; the page already links `index.qmd#bug-report` the same way. The mode-toggle steps stay — single letters that have not moved, and the section shouldn't be empty for a reader who came to it for the shortcuts.

## The drift guard

`tests/docs/test_shortcut_table_matches_bundle.py` parses the `keybinding.*` ids and their `helpKey` literals out of the vendored `maidr/static/maidr.js` and checks the six corrected rows both ways — the command still exists, and its key is unchanged — plus the invariant that actually broke: the id is defined on one page, and no other page carries a shortcut table (keyed on a row, so deleting the id from a copy doesn't hide it). Precedent is `tests/core/test_docs_quote_the_real_warning.py`.

Because it couples a test to a minified bundle, three deliberate choices keep it from becoming a false-positive generator:

- **Anchors survive minification** — only `keybinding.*` ids (i18n keys) and the `helpKey` property name. Nothing reads mangled locals; the one mangled name inside a value, `${R.ctrl}`, is normalized to `${ctrl}`.
- **A changed bundle shape fails loudly** — if the regex matches nothing, the parse asserts with "teach this test the new form rather than deleting it". An empty match set would otherwise pass every check and silently leave the table unguarded.
- **No normalizer** — expectations are three literals per row (command, helpKey, the table row verbatim), so a failure prints both sides and is readable without decoding the matcher.

Verified against the real defects rather than just run green:

| Reintroduced defect | Fails |
|---|---|
| `\| Stop Auto-play \| Ctrl \| Ctrl \|` | `test_every_row_is_present` |
| `\| Select the first element \| Ctrl + Home \| CMD + Home \|` | `test_every_row_is_present` |
| The `docs/examples.qmd` duplicate | `test_the_crossref_id_is_defined_once`, `test_no_other_page_lists_the_mode_toggles_as_a_table` |
| Simulated `maidr.js` moving chat to `ctrl + shift + c` | `test_every_documented_key_is_unchanged` |

**What it does not cover:** the toggle error above. The guard pins *keys*, not behavioral claims — nothing in it would have caught "this is a toggle". That took a reviewer driving the app, and is worth knowing before treating a green run here as the page being true.

The new Esc row is deliberately not pinned: every `esc` binding carries `showInHelp: false` and so has no `helpKey` to compare against, which would leave the check pinning the page to itself. The test says so where the rows are listed.

## Checklist

## Additional Notes

Two corrections go slightly beyond the literal report (the `Stop Auto-play` Mac cell and the fabricated Home/End rows). Shipping a PR titled "canonical shortcut table" while knowingly leaving rows in it that no keymap binds seemed worse than the small widening. Review has since confirmed both by driving a rendered chart in headless Chromium — `Ctrl+/` opens "Keyboard Shortcuts", `Shift+/` opens "Chart Assistant", `Ctrl+,` opens "Settings" with an AI tab, and `[`/`]` jump to the first/last data point.

One honest limitation, documented in the test: `test_a_ctrl_binding_says_cmd_in_the_mac_column` reads its row from the expectations rather than the page. That is deliberate — the verbatim check already pins the page, so a cell wrong in *both* places would pass it, and a row is added to both in one edit. It is exactly how the Stop Auto-play cell was wrong while its neighbours were right.

**Scope note:** this PR owns the table. The restated bindings *inside* the AI feature section are removed by #790, which rewrites that section wholesale for #774 — splitting it that way keeps the two PRs off the same lines. Either merge order works, since #790's text cites `@tbl-shortcuts`, which already exists on `main`; but until #790 lands a reader following the AI section literally hits an `H` that does nothing, so the two are best merged close together. Test-merged against #792 (which also touches `docs/examples.qmd`): clean, and `tests/docs` passes on the merged tree.

One of five PRs answering @omark807's documentation review (#772–#776).


jooyoungseo:
Thanks for driving the live app with Playwright rather than taking the table on trust — reading the in-app Help dialog's own list back is better evidence than my reading of `keybinding.ts`, and independently confirms the two corrections I flagged as going beyond the literal report.

**Took the test-coverage suggestion — added in 6870d08.**

You are right that this PR fixed the drift without preventing the next one, and that `tests/core/test_docs_quote_the_real_warning.py` is the precedent. `tests/docs/test_shortcut_table_matches_bundle.py` parses the `keybinding.*` ids and their `helpKey` literals out of `maidr/static/maidr.js` and checks the six corrected rows both ways: the command still exists (what the fabricated Home/End rows failed) and its key is unchanged (what a `maidr.js` bump would break).

Since it couples a test to a minified bundle, three things to keep it from becoming a false-positive generator on the next bump:

- **Anchors chosen to survive minification.** Only the `keybinding.*` ids (i18n keys, not locals) and the `helpKey` property name. Nothing reads `z`, `R` or `Gm`; the one mangled name inside a value, `${R.ctrl}`, is normalized to `${ctrl}`.
- **A changed bundle shape fails loudly, not silently.** If the regex matches nothing, the parse asserts with *"teach this test the new form rather than deleting it"* — otherwise an empty match set would pass every check and leave the table unguarded, which is worse than the original bug.
- **No normalizer.** Expectations are three literals per row — command, helpKey, and the table row verbatim — rather than logic mapping `${ctrl} + /` onto "Ctrl + Slash (/)". A failure prints both sides and is readable without decoding the matcher, which is what decides whether the next person fixes the drift or loosens the test.

Verified it catches the actual defects rather than just passing:

| Reintroduced defect | Fails |
|---|---|
| `\| Stop Auto-play \| Ctrl \| Ctrl \|` | `test_every_row_is_present` |
| `\| Select the first element \| Ctrl + Home \| CMD + Home \|` | `test_every_row_is_present` |
| Simulated `maidr.js` moving chat to `ctrl + shift + c` | `test_every_documented_key_is_unchanged` |

One honest limitation, documented in the test: `test_a_ctrl_binding_says_cmd_in_the_mac_column` reads its row from the expectations, not the page. That is deliberate — the verbatim check already pins the page, so a cell wrong in *both* places would pass it, and since a row is added to both in one edit that is the likely shape of the mistake. It is exactly how the Stop Auto-play cell was wrong while its neighbours were right.

`ruff check`, `ruff format --check`, and `uv run pytest tests/docs` (105 passed) are clean.

**On observation 1:** agreed, and #790 is open and green, so merging the two close together is the cleanest path.

---

jooyoungseo:
Good catch — confirmed, and it was worse than described. Fixed in 831a99e.

`docs/examples.qmd` carried a **byte-identical** copy of the table, and under the **same `{#tbl-shortcuts}` id**. So beyond the stale rows you listed, `@tbl-shortcuts` was ambiguous: two pages in the project defined one crossref id. That id is now defined exactly once.

**Took your first option — link, don't duplicate.** A second *correct* copy is still a second copy to keep correct, and this one is the demonstration of how that ends. It is also what this branch's own prose asks for, so fixing the rows in place would have left the page contradicting its own stated rule.

One adjustment to the mechanics: the pointer is a plain Markdown link to `index.qmd#keyboard-shortcuts-and-controls`, not `@tbl-shortcuts`. `docs/_quarto.yml` is `project: type: website`, and Quarto resolves cross-document crossrefs only for `book` projects — an `@tbl-shortcuts` from `examples.qmd` would have rendered as an unresolved `?tbl-shortcuts`. The page already links to `index.qmd#bug-report` that way, so this matches the established convention.

The mode-toggle steps stay on the page: single-letter bindings that have not moved, and the section should not be empty for a reader who navigated to it looking for shortcuts.

**On extending the test — I added the invariant that actually broke rather than a second copy of the row checks.** Pinning two copies of the table would have institutionalized the duplication your finding is about. `TestThereIsOnlyOneTable` asserts:

- `{#tbl-shortcuts}` is defined in `docs/index.qmd` and nowhere else under `docs/`;
- no other `.qmd` carries a shortcut table, keyed on a row rather than the id — so deleting the id from a copy does not hide it from the check.

Verified both fail against the duplicate and pass with it gone:

```
$ git stash push docs/examples.qmd      # restore the duplicate
FAILED …::TestThereIsOnlyOneTable::test_the_crossref_id_is_defined_once
FAILED …::TestThereIsOnlyOneTable::test_no_other_page_lists_the_mode_toggles_as_a_table
2 failed, 4 passed
```

`ruff check tests/docs/`, `ruff format --check`, and `uv run pytest tests/docs` (107 passed) are clean.

Worth noting what this says about the original issue: #773 was reported as "three conflicting shortcuts documented", and the root cause turns out to be that the table was *copied* rather than referenced. Fixing the rows without fixing that would have left the next drift already scheduled.

---

jooyoungseo:
**Nit taken — removed in c9bc024.** You're right that it read as a commit message that escaped into the page. A reader of the published docs doesn't need to know the page used to carry a duplicate; that's why the change was made, not something to leave behind permanently. The rationale is in the previous commit's body, where it belongs. The link to the canonical table stands on its own.

`uv run pytest tests/docs` still passes (107).

**On #2 — confirming the plan.** Yes, still deliberate: #790 rewrites that whole section for #774, so touching those lines here would guarantee a conflict between the two. #790 is open and green, and both PR bodies flag the sequencing. Your framing is the right one for a merger to act on — until #790 lands, a reader following the AI section literally hits an `H` that does nothing, so they should go in close together. Worth noting either order works, since #790's replacement text cites `@tbl-shortcuts`, which already exists on `main`.

Thanks for reimplementing the regex against the bundle independently rather than reading the docstring — that's the check that mattered, since the whole risk of this test is that it silently stops matching. Same for reintroducing the Home/End rows to confirm the failure is real.

---

jooyoungseo:
You're right, and this one is mine — not inherited drift. Fixed in 5a1d550.

The keymaps confirm your live test, and explain *why* the second press is inert:

```ts
const CHAT_KEYMAP = { TOGGLE_CHAT: key(`esc`, 'keybinding.closeChat', { showInHelp: false }) };
const HELP_KEYMAP = { TOGGLE_HELP: key(`esc`, 'keybinding.closeHelp', { showInHelp: false }) };
const SETTINGS_KEYMAP = {} as const;
```

Each dialog runs in its own scope, and that scope does not carry the opening chord. `SETTINGS_KEYMAP`'s comment gives the mechanism directly: it is empty because MUI's `onClose` takes Escape before the document-level hotkeys-js listener sees it, and *"the scope still matters — it keeps chart shortcuts from firing while the dialog is open."* That is precisely why `Ctrl+,` a second time does nothing.

**Changed:**
- The three rows are back to `Open …`.
- Added `| Close the settings, chat or help dialog | ESC | ESC |`, so the table hands the reader the keystroke that works rather than leaving the second half of the interaction undocumented.
- The note now says all three close with **Esc** and why it is not the opening key.

**Where this came from is worth naming.** The old AI section said *"The AI Chat modal is togglable. Use the same key combination to switch between the plot and the AI Chat modal."* I carried that into the note without testing it — the same failure mode as the Plotly KDE bullet earlier in this series, where Seaborn wording was carried to the Plotly section unchecked. Twice in five PRs, both caught by review rather than by me, both from trusting adjacent prose instead of the source.

**It was also in #790**, which rewrites the AI section and which I wrote separately: *"The same key closes it and returns focus to the plot."* Same claim, same error, not covered by your finding here since it is a different branch. Corrected there too (365bd42) — worth flagging so it does not merge with the bug this PR just removed.

**On the new Esc row and the test:** it is deliberately *not* pinned in `_ROWS`. Every `esc` binding here carries `showInHelp: false`, so none has a `helpKey` for the bundle parse to compare against — pinning the row would mean pinning the page to itself, which proves nothing. There is a comment saying so where the rows are listed, so the next person does not read the gap as an oversight and "fix" it into a tautology.

`uv run pytest tests/docs` passes (107), ruff clean.

---

PR #790, opened 2026-09-16
docs: replace the AI instructions with the settings flow they describe today

### LINKED ISSUE #774: fix(): AI model list out of date

jooyoungseo:
# Pull Request

## Description

The AI section named three models (GPT-4o, Gemini 2.0 Flash Experimental, Claude 3.5 Sonnet) and asserted "**Note**: All provided models are the latest versions" — twice, once per subsection, so the page carried two copies of the same stale claim to keep in sync.

Per the issue's own suggestion, this takes the "remove specific model names" option rather than the "keep them updated" one. That is the right call because the names were never load-bearing: maidr populates its model dropdowns from each provider's models API using the reader's key (`useCredentialProbe` in `Settings.tsx`), and the curated `MODEL_VERSIONS` catalog in the core is only an offline fallback, refreshed by `npm run check-models` weekly in CI. Anything this page printed was a snapshot of a list the app does not read.

Once I checked the rest of the section against the app, the model list turned out to be the least of it.

## Related Issues

Closes #774

## Changes Made

Replaced the whole section with one that matches the shipping app:

| Old text | Reality |
|---|---|
| The **alpha-tester flow** — email address, survey form, verification email, "we provide AI API keys only to alpha testers" | Gone from the app. Nothing in maidr's settings UI takes an email; the only trace left is a `clientToken` branch in `chat.ts` that the dialog cannot reach. It sent readers hunting for a text field that does not exist. **Removed entirely**, per maintainer direction that the beta-testing material can go. |
| "press **H** to open the Help menu" | `H` is bound in no keymap. API keys are in the settings dialog (`Ctrl + ,`), AI tab. |
| "**Ctrl+Shift+/** or **Alt+Shift+/** (on Mac)" | The chat binding is `shift+/`, identical on both platforms. |
| "Select up to two AI models" | No such cap; `ChatViewModel` sends the question to every enabled provider. |
| Three model names + "latest versions" | Four providers — OpenAI, Anthropic Claude, Google Gemini, and a local Ollama server. Models come from the live dropdown. |
| "These keys are not saved and are only used for the current session" | Backwards. `SettingsService` persists the whole settings object, API keys included, to `localStorage`. |

The replacement describes providers rather than models, cites `@tbl-shortcuts` instead of restating bindings, and keeps one copy of everything. The heading gains a `{#sec-ai-chat}` anchor so future links don't depend on its wording — nothing currently links to the old one, which was `#instruction-on-how-to-verify-and-use-the-ai-feature-in-py-maidr`.

## Checklist

- [ ] I have added appropriate unit tests, if applicable.

## Additional Notes

**One thing to confirm, and one thing worth a follow-up.**

The persistence note says no key reaches the Python process and none is written into `maidr.save_html()` output. That is true of the shipping code, but `Environment.initialize_llm_secrets()` in `maidr/util/environment.py` would contradict it — it reads `OPENAI_API_KEY`/`GEMINI_API_KEY` and builds a script that writes them into the iframe's `localStorage`. It is **dead code**: nothing in the package calls it, and it still speaks the maidr v2 settings schema (`openAIAuthKey`, `LLMModel`, string-valued `brailleDisplayLength`). I left it alone as out of scope here, but a documented security property resting on "the code that would break it is unreachable" is worth removing on purpose — happy to open a separate issue.

Relatedly, I pulled two absolute claims out of that callout before pushing rather than shipping them: "not sent anywhere except to the provider you enabled" ignores the `clientToken` proxy branch still present in `chat.ts`, and the exported-HTML claim is now scoped to `save_html()` specifically.

**Scope note:** this PR owns the AI section. The canonical shortcut table is #789 (for #773); the two are deliberately kept off the same lines. This PR's text cites `@tbl-shortcuts`, which already exists on `main`, so it does not depend on #789 merging first.

One of five PRs answering @omark807's documentation review (#772–#776).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Aq3ZfviNgAPRDK9D6ww7f9)_


jooyoungseo:
Thanks — the render check and the `initialize_llm_secrets` confirmation are both useful. Two responses.

**The callout-fence nit does not hold, so nothing changed.**

The claim is that "every other callout in this file consistently uses 4 colons". Counting them:

```
$ awk '/^:+[[:space:]]*\{\.callout/ {n=0; while(substr($0,n+1,1)==":") n++; print n}' docs/index.qmd | sort | uniq -c
     14 3
      1 4
```

Fourteen of the fifteen callouts use three colons. The single four-colon fence is at line 392, and it is four because it is an *outer* fence wrapping nested content — Pandoc requires the outer fence to be longer than what it contains. So three colons is the file's convention and this callout already matches it; switching to four would make it the second exception, not the first conformer.

**On the claims you could not verify from this checkout — I can give you the citations.**

You are right that they describe `xability/maidr`, not this repo, and right to flag the persistence claim specifically since it is stated to readers as flat fact. They were checked against that repo, at these locations:

| Claim | Evidence in `xability/maidr` |
|---|---|
| API keys persist to `localStorage` | `src/service/settings.ts:137` — `saveSettings()` calls `this.storage.save(SETTINGS_KEY, this.currentSettings)`; `src/service/storage.ts:36` is `localStorage.setItem`. `apiKey` is part of that object (`src/type/settings.ts:208`, defaults at 310–331) |
| `H` is unbound | No `key('h', …)` in any keymap in `src/service/keybinding.ts` |
| Chat is `shift+/` | `TOGGLE_CHAT: key('shift+/', 'keybinding.openChat', { helpKey: '?' })` |
| Settings is `Ctrl/CMD + ,`, AI tab | `TOGGLE_SETTINGS: key(`${Platform.ctrl}+,`, …)`; `SETTINGS_TABS` in `src/ui/component/Settings.tsx:140` includes `{ id: 'ai', label: 'settings.tabAi' }` |
| Question fans out to every enabled provider | `src/service/chat.ts:117` — "ChatViewModel sends one message to all enabled providers at once" |
| Four providers | `DEFAULT_SETTINGS.llm.models` — OPENAI, ANTHROPIC_CLAUDE, GOOGLE_GEMINI, OLLAMA |
| Model list is probed live | `useCredentialProbe` in `Settings.tsx`; `MODEL_VERSIONS` in `src/service/modelVersions.ts` is documented there as a fallback snapshot, refreshed by `npm run check-models` weekly in CI |

The `clientToken` branch you would have found at `chat.ts:380` is why the callout says keys are sent to the provider rather than claiming they are sent *nowhere else* — that branch exists, it is just not reachable from the settings dialog since there is no email field any more.

A maintainer double-checking the persistence sentence before merge is still worth doing, and this is a doc-only PR so nothing is blocked on it either way.

---

jooyoungseo:
Thanks — driving the real runtime was the right call, since none of these claims are reachable from pytest.

Two notes.

**Your `?` / `Esc` finding landed while this branch was being corrected for exactly that.** 365bd42 (pushed just before your review) fixes a sentence I had written here — *"The same key closes it and returns focus to the plot"* — which was the same false toggle claim, carried across from the old text I was replacing (*"The AI Chat modal is togglable"*). Your Playwright result is the confirmation: second `?` leaves the dialog open, `Escape` closes it and returns focus to the `role=application` element. `CHAT_KEYMAP` binds only `esc` in the chat scope, so the opening chord is not rebound there. The same error was in #789's table and is fixed there too.

**The duplicate `@tbl-shortcuts` label is already fixed — in #789, not here.** You're right that `docs/index.qmd:631` and `docs/examples.qmd:114` both carried that id. `examples.qmd` held a byte-identical copy of the whole table, still with the fabricated `Ctrl + Home`/`Ctrl + End` rows. #789 replaces that copy with a link to the canonical section, so the id ends up defined exactly once, and adds a test asserting that (`test_the_crossref_id_is_defined_once`, plus a row-keyed check so deleting the id from a future copy doesn't hide it). Nothing to chase in the build log once both land — though if #790 merges first, the duplicate persists until #789 does.

On scope: agreed on leaving `initialize_llm_secrets()` alone here. Worth restating that the security callout in this PR is true of shipping behavior *because* that function is unreachable — which is a fragile thing for a documented property to rest on, and is why it's flagged for follow-up rather than quietly relied on.

---

Issue #753, opened 2026-09-03
tag_elements overwrites a user-set gid on every draw and tags figures maidr never registered


jooyoungseo:
## Problem

`maidr/patch/highlight.py::tag_elements` runs inside the patched `draw` of every `Patch`, `Line2D` and collection, on every backend and every figure, and mints a fresh `maidr-<uuid>` gid unconditionally:

- A plain `fig.savefig("chart.png")` rewrites every gid the user set with `set_gid(...)` (for their own SVG post-processing, CSS, or tests) to `maidr-<uuid>`.
- A bare `Figure()` that was never registered with `FigureManager` gets tagged too, although no render is in progress and nothing will ever read the ids.

Every other gid-minting site in the package guards with `get_gid() is None`, nothing depends on gids minted outside a render (the per-render mapping in `HighlightContextManager` is the only reader), and `tests/core/plot/test_box_gaps.py` already asserts that undrawn boxes keep `None`.

## Proposed fix

In `tag_elements`: return early when no render mapping is active (`HighlightContextManager.is_rendering()`), mint a gid only when `get_gid()` is `None`, and key the per-render mapping on the existing gid otherwise. Prototyped: user gids survive a save, unregistered figures are untouched, and ordinary charts tag identically (461 tests in the gid/highlight/render subset pass with and without the change).

## Acceptance criteria

`tests/core/test_highlight_context_manager.py`: a patch with `set_gid("mine")` keeps `mine` through `savefig` and through a maidr render (the render's selector uses `mine`); a figure never handed to `FigureManager` has no `maidr-` gids after `savefig`.

Issue #695, opened 2026-09-03
Highlight tagging is O(n²) in the number of tagged artists: 60% of a 20k-bar render is two list scans per draw


jooyoungseo:
## Problem

`HighlightContextManager.set_maidr_element` (`maidr/core/context_manager.py:136-149`) runs inside the patched `draw` of every `Patch`/`Line2D`/collection during the SVG `savefig`. It does `element not in to_highlight` and then `to_highlight.index(element)` against the flat list of every tagged artist of the render. Both are linear identity scans, so a layer with n patches (bar, hist, pie: one artist per value) costs n² comparisons per render.

## Evidence (min of 3, this machine)

| bars | render today | bookkeeping alone |
|---|---|---|
| 1 000 | 0.44 s | 0.02 s |
| 5 000 | 3.27 s | 0.39 s |
| 20 000 | 15.06 s | 9.05 s (60 %) |

cProfile at 5 000 bars: `set_maidr_element` + `list.index` are the two largest entries (0.92 s + 0.57 s of ~5 s). Three independent prototypes (dict keyed by `id(element)` built once in `set_maidr_elements` with `setdefault`, so the first index still wins as `list.index` did, per #376) rendered byte-identical HTML after normalising the per-render uuids, and cut 5 000-bar renders from 3.67 s to 1.10 s and 20 000-bar renders from 17.4 s to 4.2 s.

`Artist` does not override `__eq__` (verified), so `in`/`index` are identity tests and `id()` keys are semantically identical; the elements list keeps every artist alive for the render so ids cannot be reused.

Also in the same class: `ContextManager.__new__` and `HighlightContextManager.__new__` call `super().__new__()` without `cls`, so instantiating either raises `TypeError`. Nothing instantiates them (every method is a classmethod), so the singleton scaffolding (`_instance`, `_lock`, `__new__`) is dead and misleading.

## Proposed fix

- Replace the `_elements_to_highlight` / `_selector_ids` pair with one per-render `ContextVar[dict]` mapping `id(element) -> selector id`, built with `setdefault` in `set_maidr_elements` (fresh mapping per render, per the #504 docstring); `set_maidr_element` does one `.get(id(element))` and yields immediately when absent.
- Remove the unused singleton `__new__`/`_instance`/`_lock` from both classes (or pass `cls`).

## Acceptance criteria

- A duplicated element with two selector ids resolves to the first (the #376 rule); an element not in the list is not tagged.
[...]

PR #795, opened 2026-09-17
fix(clear): drop the layers on the child axes a clear discards

### LINKED ISSUE #498: MaidrPlot holds live matplotlib artists, so no lifetime fix in FigureManager can work

jooyoungseo:
## Description

Closes #498, on the one item that thread left unmeasured rather than on the fix it proposed. The audit is in [this comment](https://github.com/xability/py-maidr/issues/498#issuecomment-5706643636); the short version is that the issue's central claim was already retracted in its own last comment, both proposals are dead, and the thing that survived is C2f point 1 — "`ax.clear()` on a live figure: artists die while the figure survives". That is real and reachable.

`Axes.inset_axes` parents an inset to the axes it sits in rather than to the figure, and `_AxesBase.__clear` empties `child_axes` — so a clear detaches every inset as surely as it removes the artists drawn directly on the axes. `Maidr.clear_axes` keys layers by their own axes, which is exactly what spares a `twinx` twin (a sibling on the figure, not a child), so the inset's layer stayed registered against an axes nothing will draw again.

## Measured

Bar chart with an inset line, `ax.clear()`, then re-plot and render:

| | before | after |
|---|---|---|
| layers registered after the clear | `[MultiLinePlot]` — the inset's | `[]` |
| the phantom layer in the schema | offered **first**, with its old data | gone |
| its `g[id='maidr-…']` selector, as an `id=` in the SVG | **0 nodes** | n/a |
| warning emitted | none | n/a |
| 5 clear-and-rebuild cycles | **6 layers** | **2** |
| detached inset `Axes` after `gc.collect()` | alive | collected |
| its `Line2D` | alive | collected |

So the reader was offered a chart that is not drawn, announced with confident values and with a highlight resolving to nothing — the failure class `clear.py`'s own docstring exists to prevent — and the detached axes and its artists were kept alive by the layer alone. Reproduced identically with `ax.clear()`, `ax.cla()` and `plt.cla()`.

The selector check is worth stating precisely, because the naive version of it lies: the gid does appear in the emitted HTML, once, inside the escaped schema JSON. As an actual `id=` attribute in the SVG markup it appears zero times.

## Changes Made

- **`maidr/patch/clear.py`** — `_discarded_child_axes` walks `child_axes` depth first and is read **before** `wrapped`, since matplotlib empties the list; `_clear_axes` then drops the layers for `(instance, *discarded)`. `clear_axes` is already idempotent, so the double firing from the `clear`/`cla` delegation stays safe.
- The discarded children also get `forget_axes_state`, for the same reason the axes itself does. This is not symmetry for its own sake — measured, the latch survives on the discarded inset and `DRAWN_SERIES` still holds the `Line2D` the clear detached. Dropping the layer while leaving the latch set is precisely the failure that helper exists to prevent: the redraw registers nothing and the chart goes from mis-described to **undescribed**.
- **`tests/patch/test_clear.py`** — six tests, plus a paragraph on the module docstring recording the child-axes route.

No change to `Maidr.clear_axes`, `_layer_axes_key`, `MaidrPlot`, or the highlight machinery.

## Verification

`uv run pytest` — **4508 passed, 69 skipped, 2 xfailed**, against **4502 passed, 69 skipped, 2 xfailed** on `main`: the same suite plus these six, no other movement. `ruff check` unchanged at the 6 pre-existing `F401` findings; both touched files `ruff format`-clean.

Falsified rather than assumed. Against the unfixed `clear.py`, **5 of the 6 fail**:

```
FAILED test_clearing_an_axes_drops_the_layers_on_the_insets_it_discards
FAILED test_clearing_an_axes_reaches_an_inset_inside_an_inset
FAILED test_repeated_clear_cycles_do_not_accumulate_inset_layers
FAILED test_a_discarded_inset_is_not_kept_alive_by_its_layer
FAILED test_a_discarded_inset_can_be_described_again_after_a_redraw
```

The sixth, `test_clearing_an_inset_leaves_the_axes_it_sits_in_registered`, passes either way **by design** — it is a neighbour guard against taking too much, the mirror of the existing twin and second-panel tests, and it is not claimed as a detector.

The recursion test was falsified separately against a one-level `list(ax.child_axes)`: it is then the *only* failure, 1 failed / 18 passed. So it earns its place rather than restating its neighbour.

| test | what it catches |
|---|---|
| `…drops_the_layers_on_the_insets_it_discards` | the defect: a layer announced for a chart matplotlib no longer draws |
| `…reaches_an_inset_inside_an_inset` | a one-level implementation |
| `…do_not_accumulate_inset_layers` | the #499 unbounded growth through this route |
| `…is_not_kept_alive_by_its_layer` | the retention half — the `Axes` and its artists pinned by the layer |
| `…can_be_described_again_after_a_redraw` | the latch regression: undescribed rather than mis-described |
| `…leaves_the_axes_it_sits_in_registered` | taking too much |

The retention test allocates and drops everything inside the function body. The same measurement at module scope, or with the candidate objects bound by a loop variable, holds the inset by that binding and reports a leak that is the harness's own — the false negative recorded on #498, which is easy to reproduce by accident and reads exactly like a confirmed bug.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Related Issues

Closes #498. Same failure class as #499, through a route that fix did not reach. Nothing here depends on #456/#508, and it needs neither of #498's proposals — see the issue comment for why both are dead (the gid rekey's target was replaced by the `id(artist)` dict for #695, and rekeying that by gid now would regress the shared-gid behaviour #753 added).

## Checklist

## Additional Notes

Deliberately excluded, all measured during the audit and each wanting its own issue:

- **`fig.delaxes(ax)` / `ax.remove()`** leave the same phantom layer, through APIs maidr does not patch. Already flagged in `tests/patch/test_clear.py`; a separate logical change with its own blast radius.
- **`set_visible(False)` / `artist.remove()`** leave N announced points against N−k highlightable elements, so point *i* outlines mark *i+k*. Real and reachable on a first render, but a pairing bug rather than a lifetime one — no keying scheme fixes it, since the artist never draws.
- **The stale comment at `maidr/core/maidr.py:505`** ("resolves an artist by its *first* index") is mechanism drift from #695; the conclusion it draws still holds via `setdefault`. Cosmetic, and a different commit.

https://claude.ai/code/session_01Qq25aruGRtqxR11SWTLrVR


jooyoungseo:
Thanks — the reformatting point was right and is fixed in e0b43fd. The diagnosis differs slightly from yours in a way worth recording: `uv run ruff` here **is** the pinned `0.3.4`, and this file is written in a *newer* ruff's style, so 0.3.4's formatter downgraded those lines rather than main having drifted. It also caught three of the new tests' own asserts.

The root cause is that I ran `ruff format` at all. Nothing in this repository asks for it — CI runs `ruff check --diff` and the pre-commit hook is `ruff --fix`, both the linter; there is no format check anywhere. Rebuilt from the base version plus the new tests: **8 hunks to 2, 0 deleted lines**, additions only. No test body changed, and the same five still fail against the unfixed `clear.py`.

Leaving `getattr(ax, "child_axes", ())` as is, as you suggested.

Your Playwright pass is the check I could not make from here — confirming the stale `X is 0, Y is 3` is what a reader actually hears on entering the chart, rather than inferring it from the schema, is the evidence that matters for this one.

---

jooyoungseo:
Correction to my last comment: **e0b43fd also reverted every line of `maidr/patch/clear.py`**, so for two commits this PR carried the tests and none of the fix. Restored in d752c54, which is `82f9974`'s production file byte for byte — `git diff 82f9974 HEAD -- maidr/patch/clear.py` is empty.

Worth recording how, since it is the same shape as the harness bug this PR's retention test is written to avoid. Falsifying the tests meant putting the base `clear.py` back, and I did that with `git checkout <base> -- maidr/patch/clear.py`, which **stages** the file. `git add tests/patch/test_clear.py` did not unstage it, and the commit carried the whole index. The full suite I ran before pushing measured the working tree, which was correct the whole time — so a green local run sat directly on top of a commit that was not.

CI would have caught it on the next run. Verification that cannot see what you are about to ship is the part that failed, not the fix.

Now on d752c54, with the working tree clean so the two agree: **4508 passed, 69 skipped, 2 xfailed**; `ruff check --diff` and `ruff check --select FA102` both clean; the same 5 of 6 still fail against the base `clear.py`, this time falsified with a file copy that never touches the index. Diff against `main` is the two files, +223/−3, additions only in the test file.

---

Issue #668, opened 2026-08-25
A `scatterpolargl` reads as nothing, so a WebGL radar chart falls back to a picture


jooyoungseo:
`go.Scatterpolar` reads as a `radar`. Its WebGL twin `go.Scatterpolargl` — the same `r` and `theta`, drawn by regl instead of by SVG — reads as nothing, so the whole figure takes the static-image path.

This is the polar half of a distinction the cartesian side already makes. `scattergl` reads exactly as `scatter` does and declines only its *selector*, because it is painted to a canvas; `renders_through_webgl()` is what draws that line, and `_WEBGL_TRACE_TYPES` lists one member.

## Measured

Every plotly trace type through `PlotlyMaidr(fig)._flatten_maidr()`, one trace per figure:

```
scatter              [('SCATTER', 3, selector=yes)]
scattergl            [('SCATTER', 3, selector=no )]
scatterpolar         [('RADAR',   1, selector=yes)]
scatterpolargl       NO LAYERS
barpolar             [('POLAR_AREA', 1, selector=no)]
```

`scatterpolargl` is the only member of the polar family that produces nothing. The reason is a literal in `_extract_plots`:

```python
polar_traces = [
    t for t in group_traces if t.get("type") in ("scatterpolar", "barpolar")
]
```

which predates it.

## It draws to a canvas, so it names nothing

Measured in Chromium against plotly's own page (`include_plotlyjs=True`, no network), counting inside the polar subplot:

| trace | `.scatterlayer .trace.scatter .point` | `… path.js-line` | `.scatterlayer > .trace` | `canvas` |
| --- | --- | --- | --- | --- |
| `scatterpolar` markers | 3 | 0 | 1 | 0 |
| `scatterpolar` lines | 0 | 1 | 1 | 0 |
| `scatterpolargl` markers | 0 | 0 | **0** | 3 |
| `scatterpolargl` lines | 0 | 0 | **0** | 3 |

So the reading is: audio, braille and text from the same `r`/`theta` a `scatterpolar` gives, and **no selector** — the outcome `scattergl` already has, and the one #145 established for a layer with nothing to point at.

## And it must not be counted in the numbering

`PlotlyPolarPlot` numbers a `scatterpolar` among *its own polar subplot's* scatter traces, because that is what one `.scatterlayer` holds. A `scatterpolargl` never enters that layer, so counting it would push its SVG siblings' `nth-child` off by one — the trap the cartesian path documents at length.

[...]

PR #798, opened 2026-09-17
feat(plotly): outline the region a choropleth reader is standing on

### LINKED ISSUE #640: A plotly choropleth ships without a highlight, for want of a map to measure
Follow-up to #627's choropleth tranche.

## The situation

`go.Choropleth` is now read as a `choropleth` layer — its regions, their
values, and the colour bar's title. What it does **not** get is a selector,
so a reader arrowing from region to region hears each one but sees nothing
outlined.

That is different from the other three plotly layers that ship without a
highlight, each of which has a reason in the chart itself:

| layer | why no selector |
|---|---|
| `barpolar` (#635) | draws one bar per spoke and no per-series path, and a radar layer resolves one selector per *series* |
| `parcoords` (#637) | renders its observations to WebGL — three `<canvas>`, no per-observation SVG |
| `parcats` (#639) | plotly lays its ribbons out in its own order (`key` = `0, 2, 1, 3` in document order), which is not computable offline |

A choropleth is none of those. It almost certainly **is** addressable —
plotly draws one `path` per region.

## Why it was left

The map could not be measured where the tranche was written. Plotly fetches
its geometry at render time, and with no network it never arrives. Measured
in Chromium:

```
topojson requests: ['https://cdn.plot.ly/un/world_110m.json']
{'hasGeo': True, 'topojsonURL': 'https://cdn.plot.ly/un/',
 'loaded': False, 'calcdataLen': 1, 'pathCount': 0}
```

`calcdata` has its one entry — plotly computed the trace fine — and
`.geolayer` holds a single empty `<g class="geo geo">`. Only the geometry is
missing.

A selector that has never resolved would be a guess, and a highlight that
lands on the wrong region is worse than none at all. So the layer ships
without one, keeping its audio, braille and text — the #145 outcome, reached
for an unusual reason.

## What finishing it needs

Somewhere the topojson can load. Then:

1. Confirm the element per region — likely `path.choroplethlocation` inside
   `.geolayer .geo .choroplethlayer .trace`.
2. Confirm whether document order matches `locations` order, or whether
   plotly sorts them the way it sorts parcats ribbons. If it sorts, the same
   conclusion as #639 applies and the layer stays unaddressed.
3. Scope per geo subplot — `layout.geo` / `layout.geo2` — the way #635 scopes
   polar. `subplot_block(trace, "geo")` in `maidr/plotly/plotly_plot.py`
   already resolves the block name; the grid placement uses it today.

The data side needs nothing: `ChoroplethTrace` reads a region list without
centroids, which is what the trace carries.

## Related, and deliberately not this

`ChoroplethPoint` also takes `lon`/`lat` and `neighbors`, which would give a
reader a spatial walk and the bordering regions instead of a flat list. A
`go.Choropleth` carries none of them — it names regions and lets plotly
resolve the geometry — so supplying them would mean shipping a gazetteer.
The grammar anticipates this: without them "the map is read as a region list
in declared order, which is a poorer reading but the one the data supports".

jooyoungseo:
## Description

A `go.Choropleth` shipped without a selector, so a reader arrowing from region to region heard each one and saw nothing outlined. The reason was never the chart — plotly fetches its geometry from `cdn.plot.ly` while it draws, so with no network the map never rendered and the elements could not be measured. Served that topojson locally, they can.

Measured in Chromium: plotly draws one `path.choroplethlocation` per **declared** `(location, z)` pair, in **declared** order — eight countries declared out of alphabetical order read back in the order they were declared. Nothing is sorted, so unlike the parcats ribbons of #639 the element a region is drawn as is computable offline. That is what #640 was waiting on.

## Type of Change

- [x] New feature

## Related Issues

Closes #640. Follows the #627 choropleth tranche; the "no highlight" siblings it was grouped with are #635 (barpolar), #637 (parcoords) and #639 (parcats).

## Changes Made

**One selector per region, keyed on the declared index.** Plotly keeps a DOM slot for a pair it can draw nothing for — a region with no value, or a name it cannot resolve like `"ZZZ"` — with the `d` attribute simply absent, while the payload drops it. A single selector matching every path would therefore resolve more elements than the layer announces on any map with a hole, and `ChoroplethTrace.mapToSvgElements` withdraws the highlight **entirely and silently** when the counts disagree. Addressed by declared index, a dropped region does not shift the ones after it, and a region plotly could not resolve highlights a zero-area element rather than the wrong country.

**Scoped by the geo subplot's exact class attribute, not its class token.** This is the one place a geo subplot differs from the polar one #635 scoped. Plotly writes `class="geo " + id`, so the first map's attribute is the doubled string `"geo geo"` whose token list is the single token `geo` — which every other geo subplot carries too. Measured: `.geolayer > g.geo.geo` matched **both** maps of a two-map figure and all **three** of a three-map one; the exact-attribute form matched exactly one on each, and on `geo2` and `geo10` alike.

**The trace position counts what plotly draws.** A `scattergeo` on the same map draws into the sibling `.scatterlayer`, so a choropleth declared after one is still `nth-of-type(1)`.

Three cases still name nothing, each for a measured reason rather than a missing measurement:

| case | why |
|---|---|
| `choroplethmap` / `choroplethmapbox` | every region is painted to a `<canvas>` via a GL base map — zero SVG paths and an empty `.geolayer` on a fully painted figure, the #668 conclusion |
| an animated figure | the payload is the base trace (plotly's *first* frame) while plotly.js advances the drawing to the **last** frame within ~0.5s of load, with no interaction. Measured on two frames naming four countries in opposite orders: the counts agree, so the highlight stays live on the wrong country at every step |
| a trace plotly draws no group for | an empty trace between two maps would otherwise number the second past its own element |

The stale reading on an animated figure is older than this and shared with every plotly layer — nothing under `maidr/plotly/` reads `frames`. What this declines is only the confident highlight laid over it.

**A `NaN` value is now dropped the way `None` already was.** This is the ordinary case rather than an exotic one: plotly's validators coerce a gap in a `pd.Series` to `NaN`, so a dataframe column with a missing value never reached the `is not None` test. The region was announced with a value the reader cannot be told anything about — in a token that stops the payload parsing as JSON at all, which is what #427 named for a coordinate.

Files: `maidr/plotly/choropleth.py` (the selector, the drop rule, the two predicates), `maidr/plotly/plotly_maidr.py` (per-geo-subplot position counter, animation flag), `maidr/plotly/plotly_plot_factory.py` and `maidr/plotly/geo.py` (comments — `geo.py` cited the choropleth's "not measured yet" reason, which is gone, so it now states its own).

## Testing

`tests/browser/test_plotly_selectors.py` resolves every emitted selector against a real Chromium, and still needs no network: the map is handed to the page before plotly.js parses rather than fetched. A `file://` page cannot read a sibling `file://` topojson (Chromium refuses it as cross-origin), so an ~800-byte generated stub is installed on `window.PlotlyGeoAssets` via `add_init_script`, which is plotly's own skip condition for the fetch. No new data file, and `test_the_map_is_drawn_without_a_network_fetch` asserts no non-`file:` request is made.

Three figures are covered: the ordinary map, one with a hole (the case the declared-index rule exists for), and a two-geo-subplot grid (the case the exact-attribute scope exists for).

Verified by mutation — each of these turns the suite red, and each was green before the tests were strengthened:

| mutation | caught by |
|---|---|
| exact attribute → `g.geo.{block}` class token | `test_a_layer_names_as_many_elements_as_it_announces[choropleth grid]` |
| declared index → announced index | `test_a_region_is_outlined_on_the_region_it_announces` |
| stub ids blanked so no region resolves | that test plus `test_the_map_is_drawn_without_a_network_fetch` |

`_selector_strings` in that file was popping from the end of a stack, so it read a selector list back to front. Invisible while every caller only counted; the new identity test is the first that pairs selectors with points by position.

- `uv run pytest` → **4533 passed, 77 skipped, 2 xfailed**
- `uv run pytest tests/browser --run-browser` → **76 passed**
- `ruff check --diff` → clean
- `uv lock --check` → clean

## Checklist

- [x] I have updated the documentation, if applicable. (`docs/stability.qmd` lists `CHOROPLETH` as experimental and makes no claim this falsifies; gaining a highlight does not promote it.)

## Additional Notes

Two things found while measuring that are deliberately **not** in this PR:

1. **The JS core's own plotly adapter has the same scoping gap.** `src/adapters/plotly/selectors.ts` builds `.geolayer > g.geo.${geoId} > ...`, which resolves one path in *every* geo subplot on a multi-map figure — the count gate then withdraws the highlight. Worth a matching fix in `xability/maidr`; it does not affect this path, since py-maidr emits its own selectors.
2. **`PlotlyPolarPlot.trace_position` is keyword-only with no default**, while `PlotlyPlotFactory` constructs it without one and comments that it is "left at its default". Pre-existing, unrelated to the choropleth, and left alone. `PlotlyChoroplethPlot` takes a default so it does not repeat it.

https://claude.ai/code/session_01Fb22qFkq8uRkgsK2WXDQiT

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fb22qFkq8uRkgsK2WXDQiT)_


jooyoungseo:
Thanks — confirmed the nit is real: `block` is bound at the top of the loop for the counter and `geo_block(choropleth_trace)` is called again for `_block_domain_start` a few lines down, where the polar block above reuses its `name`. It's an optional/cosmetic finding on an otherwise green head, so it isn't worth a CI cycle on its own; it rides this PR's next code push if one comes, and otherwise stays as is — `geo_block` is a pure lookup, and the `scattergeo` block just below already calls it inline the same way, so changing only the choropleth one would leave the geo section less internally consistent, not more.

Noting for the record that the two items in the PR description's "Additional Notes" are still open and deliberately out of scope here: the same ambiguous `g.geo.${geoId}` scoping in the JS core's own plotly adapter (`src/adapters/plotly/selectors.ts`), and `PlotlyPolarPlot.trace_position` being keyword-only with no default while the factory constructs it without one.

---

PR #800, opened 2026-09-18
fix(pie): walk matplotlib's counterclockwise slices clockwise so Right moves clockwise

jooyoungseo:
## Description

Checked the pie chart example on https://py.maidr.ai/examples/pie-wordcloud.html in Chromium (Playwright) by stepping through it with the Right arrow. The highlight and the announced data agree with each other on every step, but the walk runs the **wrong way round the dial**: Sat (drawn from 12 o'clock down the left-hand side) → Sun (bottom) → Thur (right) → Fri (top-right), i.e. counterclockwise.

The cause is a mismatch between producer and renderer. MAIDR steps Right through a pie in data order and takes that order to run clockwise from the start (that is how `PieTrace` computes the audio pan and how `p` names the clock position, see xability/maidr#780 / #784). `Axes.pie` draws **counterclockwise** unless `counterclock=False`, and py-maidr emitted the slices as drawn, so:

- Right moved the highlight counterclockwise;
- the audio panned each slice to the side it is not on (Sat, on the left, panned hard right);
- `p` announced Sat as "from 12 o'clock to 4 o'clock" while it is drawn from 12 o'clock to 8.

## Changes Made

- `maidr/patch/pieplot.py`: read `counterclock` off the call (by name or position) and hand it to the layer; `None` means matplotlib's default, `True`.
- `maidr/core/plot/pieplot.py`: a counterclockwise pie is emitted in reverse, the same wedges walked the other way round, so data order is clockwise. `counterclock=False` pies are already clockwise and are emitted as drawn. `startangle`, `explode` and `shadow` do not change the direction and leave the reading alone.
- `maidr/core/maidr.py` + `maidr/core/plot/maidr_plot.py`: the selector is positional and resolves in document order, which is draw order, so `_get_svg` now puts a layer's tagged `<g>` groups into walking order when the layer says its elements are walked out of draw order (`orders_svg_by_elements`, `False` for every other layer). Each group takes a slot another of them held, so nothing else in the SVG moves and paint order of other artists is untouched. A match that is not one to one (a user gid shared between artists, #753) leaves the document as drawn.
- `docs/examples/pie-wordcloud.qmd`: say that Right walks clockwise and why a matplotlib-default pie is read in reverse; drop the sentence that listed `counterclock` among the options that do not change the reading.
- Tests: `TestSliceOrder` rewritten around the direction (default reversed, `counterclock=False` as drawn, positional `counterclock`, elements in walking order with and without shadows, document otherwise untouched); the other pie tests updated to the clockwise reading via `CLOCKWISE_FRUIT` / `CLOCKWISE_UNITS`; new `tests/core/test_svg_order.py` for the reorder pass and its guards; new browser test `tests/browser/test_pie_clockwise_walk.py` that drives the docs pie with the shipped bundle and checks the announced slice, the outlined wedge and the `p` clock position on every stop (fails on the previous source with "Day is Sat" as the first stop).

Verified against the same `maidr@4.8.0` bundle the docs page pins, rendering the docs example with this branch:

| Step | Before (live page) | After (this branch) |
|---|---|---|
| Right ×1 | Sat, highlight on the left wedge | Fri, highlight on the top-right wedge |
| Right ×2 | Sun (bottom) | Thur (right) |
| Right ×3 | Thur (right) | Sun (bottom) |
| Right ×4 | Fri (top-right) | Sat (left) |
| `p` on Sat | "from 12 o'clock to 4 o'clock" (wrong) | "from 8 o'clock to 12 o'clock" (as drawn) |

Left walks back the other way.

## Related Issues

Renderer-side context: xability/maidr#780, xability/maidr#784.

## Checklist

## Additional Notes

- `uv run pytest`: 4562 passed, 81 skipped, 2 xfailed; `uv run pytest tests/browser --run-browser` passes with the new test. `ruff check` / `ruff format --check` clean on the changed files.
- A behaviour change for every matplotlib pie drawn with the default `counterclock=True`: the first slice announced on entering the plot is now the last one in the data (the first one clockwise from the start angle). The description's data table follows the same order.
- The renderer still assumes the walk starts at 12 o'clock, which is exact for `startangle=90` (the docs example) and off by the start angle otherwise; carrying the start angle would need a grammar addition on the maidr side, so it is left out of this PR. The comment in `src/model/pie.ts` claiming matplotlib lays a pie out clockwise is also wrong (matplotlib, base R `pie()` and Plotly all default to counterclockwise) and could be corrected there.

https://claude.ai/code/session_01UCkYH6b3mVkPdHrry7AmWZ


jooyoungseo:
Thanks, the `counterclock=None` case was real. Fixed in the latest push: `_argument` now takes the default to answer for an argument the caller left out, and the pie patch asks for matplotlib's own `True` there and reads the value for truth the way `Axes.pie` does, so an explicit `None` is a clockwise pie and is emitted as drawn. Pinned by `test_an_explicit_none_is_drawn_clockwise_and_read_that_way` (which also checks the wedges really were drawn clockwise) and a `_argument` test for the default.

The positional test now silences the 3.10 deprecation and skips if a future matplotlib refuses the parameters before `counterclock` by position: once it does, no caller can reach that path and there is nothing left to pin.

The renderer-side follow-up (carrying the start angle and drawn direction through the grammar) is xability/maidr#1268.

---

